PR #10, opened 2026-05-30
fix(viewer): add Host-header allowlist to block DNS rebinding

aaronjmars:
## Summary

The viewer's HTTP server (`StartServer` in `internal/viewer/server.go`) accepts every request regardless of the `Host` header. A web page the user visits while `ocr viewer` is running can resolve its own domain to `127.0.0.1` (DNS rebinding) and then have the browser make same-origin requests to `http://127.0.0.1:5483/` and `/r/{repo}/{sessionID}` — exfiltrating every session JSONL the viewer exposes.

## Impact

Each session JSONL (`~/.opencodereview/sessions/<repo>/<session>.jsonl`) records:

- `llm_request` messages, which include the source code being reviewed,
- `llm_response` content, which is the LLM's analysis of that code,
- `tool_call` results (file reads, code searches), and
- `session_start` metadata: `cwd`, `gitBranch`, `model`.

A successful DNS-rebinding attack against a developer running `ocr viewer` reads all of that for every repo the developer has reviewed — including private repos and the model's review findings, which may themselves describe sensitive code paths.

## Location

`internal/viewer/server.go:49-55` — `http.Server{Handler: mux}` is wired with no `Host` validation; default `--addr localhost:5483` is not by itself a defense against DNS rebinding.

## Fix

A new `internal/viewer/hostguard.go`:

- `hostOnly` parses a `Host` header value (handles `host:port`, bracketed IPv6, case-insensitive matching, and rejects ambiguous bare unbracketed IPv6).
- `buildAllowedHosts` returns a default-deny set seeded with the standard loopback names (`localhost`, `127.0.0.1`, `::1`). If the bind host is a concrete hostname or IP it is also added; wildcard binds (`0.0.0.0`, `::`, empty) are NOT auto-added.
- `OCR_VIEWER_ALLOWED_HOSTS` (comma-separated) lets an operator extend the allowlist explicitly — required for any non-loopback bind.
- `hostGuard` is an `http.Handler` middleware that returns `403 forbidden host` on any `Host` not in the allowlist (loopback names always pass via `isLoopbackHost`, which also accepts the full `127.0.0.0/8` range).

`StartServer` wraps the existing mux with `hostGuard` before passing it to `http.Server`.

## Verification

```
go vet ./...                    # exit 0
go build ./...                  # exit 0
go test ./internal/viewer/...   # 5 tests / 25 subtests, all pass
go test ./...                   # all packages pass
```

New tests in `internal/viewer/hostguard_test.go` cover:

- `hostOnly` parsing (host:port stripping, bracketed IPv6, case, ambiguous-input rejection)
- `isLoopbackHost` recognition (`localhost`, `127.0.0.0/8`, `::1`)
- `buildAllowedHosts` (loopback baseline, concrete-bind-add, wildcard-bind-skip, env-extension parsing)
- `hostGuard` end-to-end: loopback variants → 200, attacker DNS-rebinding hosts → 403 with no body leak, LAN bind, wildcard bind + env override.

Verified that without the patch the inner handler returns the session payload to an attacker `Host`; with the patch it returns 403 and no body.

## Detected by

Aeon + manual review (semgrep / trufflehog / osv-scanner all surfaced no fix-PR-worthy finding here — `semgrep p/golang` flagged 1 unrelated `math/rand` jitter use, osv reported only stdlib/x/* CVEs already covered by the Go toolchain bump cadence).

- Severity: high
- CWE-346 (Origin Validation Error)
- CWE-200 (Exposure of Sensitive Information)

## Notes

- Default behavior for `ocr viewer` (no flag) is unchanged: `localhost:5483` keeps working, since `localhost` / `127.0.0.1` are in the loopback set.
- An operator who binds the viewer on a LAN interface and needs to reach it from another host now sets `OCR_VIEWER_ALLOWED_HOSTS=review.internal,ocr.lan ocr viewer --addr :3000` — surfacing the exposure as an explicit choice. Documented in the new "Viewer security" subsection of README.md.

---
Filed by [Aeon](https://github.com/aaronjmars/aeon-aaron).

### DIFF
--- a/internal/viewer/hostguard.go
+++ b/internal/viewer/hostguard.go
@@ -1,0 +1,120 @@
+package viewer
+
+import (
+	"net"
+	"net/http"
+	"os"
+	"strings"
+)
+
+// EnvAllowedHosts is the environment variable users can set to extend the
+// default loopback allowlist with additional hostnames (comma-separated).
+const EnvAllowedHosts = "OCR_VIEWER_ALLOWED_HOSTS"
+
+// hostOnly returns the bare host portion of a Host header value, with any
+// port stripped and surrounding brackets removed from IPv6 literals. Empty
+// or malformed input returns "".
+func hostOnly(h string) string {
+	h = strings.TrimSpace(h)
+	if h == "" {
+		return ""
+	}
+	// SplitHostPort handles bracketed IPv6 ([::1]:5483) and host:port forms.
+	if host, _, err := net.SplitHostPort(h); err == nil {
+		return strings.ToLower(host)
+	}
+	// No port. Strip brackets from a bare bracketed IPv6 literal like "[::1]".
+	if strings.HasPrefix(h, "[") && strings.HasSuffix(h, "]") {
+		return strings.ToLower(h[1 : len(h)-1])
+	}
+	// Reject a bare unbracketed IPv6 with embedded colons (ambiguous with port).
+	if strings.Count(h, ":") > 1 {
+		return ""
+	}
+	return strings.ToLower(h)
+}
+
+// isLoopbackHost reports whether host is a loopback name or IP literal that
+// is always safe to accept regardless of operator configuration.
+func isLoopbackHost(host string) bool {
+	switch host {
+	case "localhost", "127.0.0.1", "::1", "0:0:0:0:0:0:0:1":
+		return true
+	}
+	// Any 127.0.0.0/8 IPv4 literal.
+	if ip := net.ParseIP(host); ip != nil && ip.IsLoopback() {
+		return true
+	}
+	return false
+}
+
+// buildAllowedHosts returns the default-deny allowlist used by the host guard.
+// The set always includes the standard loopback names. If bindHost is a
+// concrete (non-wildcard) hostname or IP, it is also included so a user who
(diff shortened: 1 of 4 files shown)


internal/viewer/hostguard.go
```diff
@@ -0,0 +1,120 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		host := hostOnly(r.Host)
+		if host == "" {
+			http.Error(w, "forbidden host", http.StatusForbidden)
+			return
+		}
+		if isLoopbackHost(host) {
+			next.ServeHTTP(w, r)
+			return
+		}
```
lizhengfeng101, line 95: Nit: the security middleware silently drops rejected requests. Consider logging the rejected `Host` value and remote address to aid attack detection during incident response:

```go
log.Printf("[viewer] rejected Host %q from %s", r.Host, r.RemoteAddr)
```

This helps operators notice DNS-rebinding probes without needing to add external request logging.

lizhengfeng101: "Sorry, the comment location from Claude Code + gh CLI isn't very accurate — this comment actually refers to `internal/viewer/hostguard.go: L89`"

PR #13, opened 2026-05-30
feat(build): add Windows platform support

betterwinsone:
## Summary

- Add `windows/amd64` and `windows/arm64` to the CI build matrix
- Add `build-windows-amd64` and `build-windows-arm64` Makefile targets
- Handle `.exe` suffix in `bin/ocr.js`, `scripts/install.js`, and `scripts/update.js`
- Skip `chmodSync` on Windows where it's unsupported
- Map `win32` platform to `windows` in `detectPlatform()`
- Update `TestReleaseYmlMatchesURLPattern` to cover Windows platform naming

## Test plan

- [x] `go test ./...` passes all tests
- [x] `go vet ./...` reports no issues
- [x] Verify binary name pattern matches URL pattern for all 6 platform/arch combinations (linux, darwin, windows x amd64, arm64)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,10 @@ jobs:
             goarch: amd64
           - goos: darwin
             goarch: arm64
+          - goos: windows
+            goarch: amd64
+          - goos: windows
+            goarch: arm64
     steps:
       - uses: actions/checkout@v4
 
@@ -38,12 +42,17 @@ jobs:
           GIT_COMMIT=$(git rev-parse --short HEAD)
           BUILD_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
           LD_FLAGS="-s -w -X main.Version=${VERSION} -X main.GitCommit=${GIT_COMMIT} -X main.BuildDate=${BUILD_DATE}"
-          go build -ldflags "${LD_FLAGS}" -o opencodereview-${{ matrix.goos }}-${{ matrix.goarch }} ./cmd/opencodereview
+          BIN_NAME=opencodereview-${{ matrix.goos }}-${{ matrix.goarch }}
+          if [ "${{ matrix.goos }}" = "windows" ]; then
+            BIN_NAME="${BIN_NAME}.exe"
+          fi
+          go build -ldflags "${LD_FLAGS}" -o "${BIN_NAME}" ./cmd/opencodereview
+          echo "bin_name=${BIN_NAME}" >> $GITHUB_ENV
 
       - uses: actions/upload-artifact@v4
         with:
           name: binary-${{ matrix.goos }}-${{ matrix.goarch }}
-          path: opencodereview-${{ matrix.goos }}-${{ matrix.goarch }}
+          path: ${{ env.bin_name }}
 
   release:
     needs: build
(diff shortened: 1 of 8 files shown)


lizhengfeng101:
#### Overall
  
  Solid PR with complete coverage across CI, Makefile, install/update scripts, binary entry point, tests, and docs. Core logic is correct. A few areas
  worth addressing:

  ---
  **Issue 1: Makefile Windows targets don't reuse the BUILD_PLATFORM macro**
  
  Linux/darwin targets use the BUILD_PLATFORM macro, but Windows targets duplicate the build command inline:

  build-windows-amd64:
        GOOS=windows GOARCH=amd64 CGO_ENABLED=0 $(GO) build -ldflags "$(LD_FLAGS)" \
                -o $(DIST_DIR)/$(BINARY_NAME)-windows-amd64.exe \
                ./cmd/opencodereview

  Suggestion: Introduce a BUILD_PLATFORM_WIN macro (or extend the existing one with an optional suffix parameter) to stay DRY:

  define BUILD_PLATFORM_WIN
        GOOS=$(1) GOARCH=$(2) CGO_ENABLED=0 $(GO) build -ldflags "$(LD_FLAGS)" \
                -o $(DIST_DIR)/$(BINARY_NAME)-$(1)-$(2).exe \
                ./cmd/opencodereview
  endef

  ---
  **Issue 2: .exe suffix logic is scattered across multiple files**
  
  The .exe append logic appears in at least 4 independent locations:
  - bin/ocr.js
  - scripts/install.js (downloadUrl += ".exe" and BINARY_NAME)
  - scripts/update.js (downloadUrl += ".exe" and BINARY_NAME)
  - .github/workflows/release.yml
  
  Since install.js already exports utility functions (detectPlatform, buildUrl, etc.), consider also exporting a binaryName() helper so update.js and
  bin/ocr.js can import it instead of duplicating the logic.

  ---
  **Issue 3: urlPattern doesn't account for .exe suffix**
  
  The urlPattern in package.json is:
  opencodereview-{os}-{arch}

  But Windows URLs actually need .exe. The current workaround is hardcoded downloadUrl += ".exe" in code, making the pattern incomplete as a source of
  truth.

  Suggestion: Consider introducing an {ext} placeholder (opencodereview-{os}-{arch}{ext}) so the pattern is self-describing. This could be a follow-up
  change given the broader scope.

  ---
  **Issue 4: fs.renameSync may fail on Windows when the binary is in use**
  
  In update.js:
  fs.renameSync(tempPath, binaryPath);
  
  On Windows, if the target binary is currently running, this will throw EBUSY or EPERM. This is a known Windows limitation.

  Suggestion: Consider a rename-then-replace strategy (rename old binary to .old first), or at minimum provide a meaningful error message guiding the user
   to manually update.

  ---
  #### Minor
  
  - README Windows install instructions lack PATH guidance. Linux/macOS shows sudo mv ocr /usr/local/bin/ocr; Windows users may not know where to place
  ocr.exe.
  - Test .exe detection relies on string matching (strings.Contains(yml, ...)), which is somewhat brittle but acceptable for a test.

  ---
  #### Summary
  
  Core logic is correct and all critical Windows paths are covered. Main suggestions: centralize .exe suffix handling, reuse macros in Makefile, and
  handle the Windows file-locking scenario during self-update. None are blockers — they can be addressed as follow-ups.

PR #51, opened 2026-06-05
fix(agent): user includes should have higher priority than system sup…

dangrie158:
## Description

This PR simply allows the user to override the exclude list of allows extensions by explicitly including files in the rules set.

Before this change it was not possible to have e.g.:

```json
{
    "include": ["*.po", "*.xlf"]
}
```
in your rules.json file and have it work because the files would be excluded before the `include` rules are tested.

This feature was documented and tested for, but I do not think this is expected an beheviour, however I am open to be convinced otherwise :)

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Tested 

## Checklist

- [x] I have signed the CLA

### DIFF
--- a/internal/agent/preview.go
+++ b/internal/agent/preview.go
@@ -53,13 +53,13 @@ func (a *Agent) whyExcluded(d model.Diff) ExcludeReason {
 		return ExcludeUserRule
 	}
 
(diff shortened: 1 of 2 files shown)


stay-foolish-forever:
I am testing this feature, will merge when no problem occur.

PR #105, opened 2026-06-12
fix(diff): detect renamed and deleted files correctly in diff parsing

### LINKED ISSUE #99: exit status 128[当使用ocr review --from master --to SRQ-XXXXX --format json]
### OpenCodeReview Version

v1.3.1

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

Anthropic (Claude)

### Bug Description

执行以下命令时， 如果在SRQ-XXXXX分支上对一个文件进行了rename, 就会出现预警提示：[ocr] WARNING: cannot read file "master分支上未重命名的文件" at ref SRQ-XXXXX: exit status 128

ocr review --from master --to SRQ-XXXXX --format json

### Steps to Reproduce

SRQ-XXXXX： rename某个文件，然后提交

ocr review --from master --to SRQ-XXXXX --format json

### Expected Behavior

能给识别到该rename文件

### Logs / Error Output

```shell

```

### Additional Context

_No response_

adlternative:
## Summary

Fixes #99 — `ocr review --from master --to BRANCH --format json` emitted `[ocr] WARNING: cannot read file <old path> at ref BRANCH: exit status 128` when a file was renamed on the target branch.

## Root Cause

Two compounding bugs:

1. **Broken `/dev/null` detection** (`internal/diff/parser.go`): git emits `--- /dev/null` / `+++ /dev/null` **without** `a/`/`b/` prefixes, but the old regexes required them, so `IsNew`/`IsDeleted` were never set. Deleted files fell through to `git show ref:<old path>` → `exit status 128`.
2. **Renames never handled**: the parser ignored `rename from` / `rename to` extended headers, and the git diff/show call sites did not pass `--find-renames`, so with `diff.renames=false` a rename degraded to delete(old)+add(new); the delete half hit bug 1 and warned about the old path.

## Changes

- Parse `rename from`/`rename to` (authoritative paths, robust for paths with spaces), `new file mode`, `deleted file mode`, and unprefixed `/dev/null` markers in `ParseDiffText`
- Pass `--find-renames` to all 4 git diff/show invocations so rename detection no longer depends on user git config
- Add `IsRenamed` to `model.Diff` (json: `is_renamed`) and prefer it in `diffStatus`
- New `parser_test.go`: rename / pure rename (100% similarity, no hunks) / deleted / new file unit tests
- New `TestRangeDiffDetectsRename` integration test: `diff.renames=false` + branch rename + small edit → exactly 1 diff, `IsRenamed=true`, content read at the new path

## Testing

- `go test -race -count=1 ./...` — all packages pass
- `go vet ./...` and `go build ./...` clean

### DIFF
--- a/internal/agent/preview.go
+++ b/internal/agent/preview.go
@@ -127,6 +127,8 @@ func diffStatus(d model.Diff) string {
 		return "added"
 	case d.IsDeleted:
 		return "deleted"
+	case d.IsRenamed:
+		return "renamed"
 	case d.OldPath != d.NewPath && d.OldPath != "" && d.OldPath != "/dev/null":
 		return "renamed"
 	default:
--- a/internal/diff/git.go
+++ b/internal/diff/git.go
@@ -115,14 +115,14 @@ func (p *Provider) GetDiff(ctx context.Context) ([]model.Diff, error) {
 		if base == "" {
 			return nil, fmt.Errorf("cannot find merge-base between %s and %s", p.from, p.to)
 		}
-		out, err := p.runGit(ctx, "diff", "--no-ext-diff", "--no-textconv", "--src-prefix=a/", "--dst-prefix=b/", "--no-color", "-U"+fmt.Sprint(DiffContextLines), "--end-of-options", base, p.to, "--")
+		out, err := p.runGit(ctx, "diff", "--no-ext-diff", "--no-textconv", "--find-renames", "--src-prefix=a/", "--dst-prefix=b/", "--no-color", "-U"+fmt.Sprint(DiffContextLines), "--end-of-options", base, p.to, "--")
 		if err != nil {
 			return nil, fmt.Errorf("git diff failed: %w", err)
 		}
 		combined.WriteString(out)
 
 	case ModeCommit:
-		out, err := p.runGit(ctx, "show", "--no-ext-diff", "--no-textconv", "--src-prefix=a/", "--dst-prefix=b/", "--no-color", "-U"+fmt.Sprint(DiffContextLines), "--end-of-options", p.commit)
+		out, err := p.runGit(ctx, "show", "--no-ext-diff", "--no-textconv", "--find-renames", "--src-prefix=a/", "--dst-prefix=b/", "--no-color", "-U"+fmt.Sprint(DiffContextLines), "--end-of-options", p.commit)
 		if err != nil {
 			return nil, fmt.Errorf("git show failed: %w", err)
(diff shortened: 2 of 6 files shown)


adlternative:
## End-to-End Verification Report

To confirm the fix actually resolves issue #99 (beyond unit tests), I reproduced the reported scenario and compared the **pre-fix** and **post-fix** binaries side by side.

### Repro setup (mirrors the issue exactly)

```bash
git init -b master
git config diff.renames false        # key: simulates an environment where git does NOT auto-detect renames
seq 1 50 | sed 's/^/code line /' > service.go
git add . && git commit -m 'init on master'

git checkout -b SRQ-12345            # the SRQ-XXXXX branch from the issue
git mv service.go service_renamed.go # rename
sed -i 's/code line 25/code line 25 EDITED/' service_renamed.go  # small edit, like the reporter did
git add -A && git commit -m 'rename service.go'
```

`diff.renames=false` is the trigger: it makes `git diff` emit the rename as delete(old) + add(new), which is exactly the path that produced the warning.

### Method

- **Pre-fix binary**: built from `main` @ 022800c via `git worktree`
- **Post-fix binary**: built from this branch
- Both run the same command with `--preview`, which skips the LLM but fully exercises the diff fetch/parse path where the warning originates (`finalizeDiff` in `internal/diff/parser.go`)

```bash
ocr review --from master --to SRQ-12345 --preview
```

### Results

**Before fix** (reproduces the exact warning from the issue):

```
[ocr] WARNING: cannot read file service.go at ref SRQ-12345: exit status 128

Preview: 2 file(s) changed  |  +50  -50
  [M]  service.go           +0   -50     <- rename misclassified as delete + add
  [M]  service_renamed.go   +50  -0
```

**After fix**:

```
Preview: 1 file(s) changed  |  +1  -1
  [R]  service_renamed.go   +1   -1      <- correctly recognized as a single rename, no warning
```

### Takeaways

- The warning is gone, and behavior is also more correct:
  - File count drops from a bogus 2 (delete + add) to 1 rename (`[R]` status)
  - Line stats drop from an inflated +50/-50 to the real +1/-1, avoiding feeding the entire file to the LLM as 'new' content (saves tokens and improves review accuracy)
- This manual repro is also codified as the `TestRangeDiffDetectsRename` integration test (same `diff.renames=false` setup) to guard against regressions in CI.
- `go test -race -count=1 ./...`, `go vet ./...`, and `go build ./...` all pass.

PR #112, opened 2026-06-13
fix(security): block git ref option injection

### LINKED ISSUE #111: Git option injection via unvalidated --from/--to/--commit refs (arbitrary file write / RCE)
### OpenCodeReview Version

v1.3.4

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

Anthropic (Claude)

## Bug description
Ref values supplied via `--from`, `--to`, and `--commit` are passed as bare
positional arguments to `git` subcommands without any `--`/`--end-of-options`
guard separating them from option flags. A ref value beginning with `-` is
therefore interpreted by git as an option rather than a revision, and several of
those options have dangerous side effects:

- `git diff --output=<file> …` → arbitrary file overwrite
- `git grep -O<cmd>` / `--open-files-in-pager=<cmd>` → arbitrary command execution

Affected call sites:
- `internal/diff/git.go:118` — `git diff … <base> <p.to> "--"` (p.to before `--`)
- `internal/diff/git.go:125` — `git show … <p.commit>` (no `--`)
- `internal/diff/git.go:260` — `git merge-base <from> <to>` (no `--`)
- `internal/tool/filereader.go:130,216` — `git show <Ref>:<path>` (Ref concatenated raw)
- `internal/tool/code_search.go:70` — `<ref>` appended into `git grep` before `--`
- `internal/tool/file_find.go:75` — `git ls-tree -r --name-only <ref>`

The values are never validated as refs before reaching git (git's own ref-naming
rules forbid a leading `-`, but these raw strings bypass that check).

## Steps to reproduce
1. In any git repo: `ocr review --commit '--output=/tmp/ocr_pwn'`
   → `/tmp/ocr_pwn` is created/overwritten with git diff output.
2. (RCE path) drive a review in range/commit mode where the ref equals
   `--open-files-in-pager=...`; the `code_search` tool's `git grep` invocation
   launches the supplied program once a match exists.

## Expected behaviour
A ref value that is not a valid git revision — in particular one that begins with
`-` — must be rejected before any git invocation, and refs must be passed to git
in a way that cannot be reinterpreted as command-line options. No `--from/--to/
--commit` value should be able to cause git to write files or execute commands.

## Actual behaviour
`ocr review --commit '--output=/path/to/file'` (or `--to`, `--from`) reaches
`git diff/show` and writes the diff to an attacker-chosen path; a ref such as
`--open-files-in-pager=<cmd>` reaching the `git grep` path in `code_search` can
execute an arbitrary command. The shipped GitHub Action avoids this only because
it happens to use `head.sha` and an `origin/` prefix; any other automation that
forwards an untrusted branch/commit name to these flags is exploitable.

### Logs / Error Output

```shell

```

## Suggested fix
- Validate every non-empty `--from/--to/--commit` value at ingestion: reject any
  value starting with `-` and confirm it resolves
  (`git rev-parse --verify --end-of-options <ref>^{commit}`).
- Defense in depth: insert `--end-of-options` immediately before ref arguments in
  the `diff`/`show`/`merge-base` calls.

MuoDoo:
## Description

Fix RCE issue.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [x] I have signed the CLA

## Related Issues

Fixes #111

### DIFF
--- a/cmd/opencodereview/git.go
+++ b/cmd/opencodereview/git.go
@@ -13,7 +13,7 @@ func runGitCmd(repoDir string, args ...string) ([]byte, error) {
 }
 
 func getCommitMessage(repoDir, commit string) (string, error) {
-	out, err := runGitCmd(repoDir, "log", "-1", "--format=%B", commit)
+	out, err := runGitCmd(repoDir, "log", "-1", "--format=%B", "--end-of-options", commit)
 	if err != nil {
 		return "", fmt.Errorf("git log failed: %w", err)
 	}
--- a/cmd/opencodereview/review_cmd.go
+++ b/cmd/opencodereview/review_cmd.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/open-code-review/open-code-review/internal/agent"
@@ -47,6 +48,9 @@ func runReview(args []string) error {
 	repoDir, err := resolveRepoDir(opts.repoDir)
 	if err != nil {
 		return fmt.Errorf("resolve repo: %w", err)
+	}
+	if err := validateReviewRefs(repoDir, opts); err != nil {
+		return err
 	}
 
 	if opts.commit != "" && opts.background == "" {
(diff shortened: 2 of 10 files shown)


Heosswitch:
您的邮件已经收到，感谢您的来信！

PR #150, opened 2026-06-16
fix(tool): correct git grep ref argument order in code_search

zhuichen:
## Summary

Fix `code_search` failures in commit/range review modes by correcting `git grep` argument ordering and preserving option-like ref safety checks.

## Problem

In commit/range mode, `code_search` constructed `git grep` args with an invalid revision placement, which caused errors like:

- `fatal: unable to resolve revision: --end-of-options`

As a result, `code_search` became unreliable for `--commit` and `--from/--to` workflows.

## Root Cause

`buildGrepArgs` assembled arguments in a way that made Git parse an invalid token as a revision.

## Changes

- Corrected `git grep` arg order in:
  - `internal/tool/code_search.go`
- Added/kept defensive guard for option-like refs (`-...`) to avoid ambiguous/unsafe execution.
- Updated tests to match the corrected command structure and behavior:
  - `internal/tool/code_search_test.go`

## Validation

Executed per contribution guide:

- `go fmt ./...`
- `go vet ./...`
- `make test` (race enabled)
- `make build`

All passed.

## Impact

- Restores `code_search` correctness in commit/range mode.
- Improves stability of AI review tool-calls for non-workspace reviews.
- Keeps security behavior for option-like refs.

### DIFF
--- a/internal/tool/code_search.go
+++ b/internal/tool/code_search.go
@@ -67,7 +67,6 @@ func (p *CodeSearchProvider) buildGrepArgs(searchText string, caseSensitive bool
 	cmdArgs = append(cmdArgs, "-e", searchText)
 
 	if ref := p.FileReader.Ref; ref != "" {
-		cmdArgs = append(cmdArgs, "--end-of-options")
 		cmdArgs = append(cmdArgs, ref)
 	}
 
@@ -104,6 +103,10 @@ func (p *CodeSearchProvider) runGitGrep(parentCtx context.Context, cmdArgs []str
 }
 
 func (p *CodeSearchProvider) gitGrep(ctx context.Context, searchText string, caseSensitive bool, usePerlRegexp bool, pathspec []string) (string, error) {
+	if ref := p.FileReader.Ref; ref != "" && strings.HasPrefix(ref, "-") {
+		return "Error: invalid ref: refs must not start with '-'", nil
+	}
+
 	cmdArgs := p.buildGrepArgs(searchText, caseSensitive, usePerlRegexp, pathspec)
 
 	outStr, errStr, err := p.runGitGrep(ctx, cmdArgs)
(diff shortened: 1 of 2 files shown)


MuoDoo:
I couldn’t reproduce the `--end-of-options` revision error with the current args. Could you share the exact failing case?

guqihao-7:
> I couldn’t reproduce the `--end-of-options` revision error with the current args. Could you share the exact failing case?我无法用当前参数复现 `--end-of-options` 修订版本错误。你能分享具体的失败案例吗？

I can reproduce
image

guqihao-7:
> I couldn’t reproduce the `--end-of-options` revision error with the current args. Could you share the exact failing case?

my git version is: git version 2.39.5 (Apple Git-154)

I suspect this is related to the git version. I couldn't reproduce it on my Ubuntu, but my Mac.

lizhengfeng101:
Hi @zhuichen, thanks for taking the time to investigate and submit this PR! We really appreciate the effort you put into diagnosing the `code_search` behavior in commit/range mode.

After reviewing the changes, we believe the current implementation on `main` is actually correct and intentional:

- The `--end-of-options` flag in `buildGrepArgs` is a Git-native safety mechanism that prevents refs starting with `-` from being misinterpreted as command-line options. This is the recommended way to handle untrusted/arbitrary ref names in Git commands.
- The current argument order (`-e <pattern> --end-of-options <ref> -- <pathspec>`) is valid for `git grep` — Git will correctly treat the first argument after `--end-of-options` as a revision.

Removing `--end-of-options` and replacing it with an application-level `-` prefix check would actually reduce security coverage, since `--end-of-options` guards against a broader class of edge cases at the Git layer itself.

If you're seeing `fatal: unable to resolve revision: --end-of-options` errors in practice, it might be caused by something else (e.g., an older Git version that doesn't support `--end-of-options` for `git grep`). If so, we'd love to hear more details about your environment so we can look into it further!

Thanks again for the contribution 🙏

PR #151, opened 2026-06-16
fix(cli): reject --to without --from in review flags

zhuichen:
## Summary

Reject `--to` when used without `--from` in `ocr review` to prevent silent fallback to workspace mode.

## Problem

The CLI previously accepted `ocr review --to <ref>`, but internal review mode detection only entered range mode when both `--from` and `--to` were provided. This caused a silent fallback to workspace mode, which did not match user intent.

## Root Cause

Flag validation only enforced one direction (`--from` requires `--to`) and missed the symmetric rule (`--to` requires `--from`).

## Changes

- Added symmetric flag validation in `parseReviewFlags`:
  - keep existing: `--from` requires `--to`
  - add new: `--to` requires `--from`
- Added regression tests for:
  - reject `--to` without `--from`
  - reject `--from` without `--to`
  - allow `--from` + `--to`

Modified files:
- `cmd/opencodereview/flags.go`
- `cmd/opencodereview/review_cmd_test.go`

## Validation

Executed per contribution guide:

- `go fmt ./...`
- `go vet ./...`
- `make test` (race enabled)
- `make build`

Behavior check:

- `go run ./cmd/opencodereview review --to HEAD --preview`
  - now returns: `--from is required when --to is specified`

## Impact

- Prevents ambiguous CLI behavior.
- Aligns flag semantics with internal range-mode logic.
- Improves predictability for CI and local scripted usage.

### DIFF
--- a/cmd/opencodereview/flags.go
+++ b/cmd/opencodereview/flags.go
@@ -155,6 +155,9 @@ func parseReviewFlags(args []string) (reviewOptions, error) {
 	if opts.from != "" && opts.to == "" {
 		return opts, fmt.Errorf("--to is required when --from is specified")
 	}
+	if opts.to != "" && opts.from == "" {
+		return opts, fmt.Errorf("--from is required when --to is specified")
+	}
 
 	switch opts.audience {
 	case "human", "agent":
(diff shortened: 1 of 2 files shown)


lizhengfeng101 (approved):
LGTM

PR #153, opened 2026-06-16
Add Claude support via Google Vertex AI

### LINKED ISSUE #56: Support Claude on Google Vertex AI
### Problem Statement

OCR cannot use Claude over Vertex AI, because the API is a bit different[0].

[0] https://platform.claude.com/docs/en/build-with-claude/claude-on-vertex-ai

### Proposed Solution

Add a OCR_USE_ANTHROPIC_VERTEX=true to use the proper API, 

or it can be autodetected using something like:
```go
isVertex := strings.Contains(cfg.URL, "aiplatform.googleapis.com")
```
(not sure about this one)

### Alternatives Considered

_No response_

### Affected Area

CLI / Commands, Review Agent / LLM interaction

### Additional Context

I can work on it, if this feature is desirable.

Devwallem:
## What

This PR adds support for using Anthropic Claude through Google Vertex AI.

Users can enable it with Application Default Credentials (ADC), without configuring an Anthropic API key or custom LLM endpoint.

## Changes

- Add Vertex AI configuration resolution for Claude
- Support `OCR_USE_ANTHROPIC_VERTEX`, `OCR_VERTEX_PROJECT_ID`, and `OCR_VERTEX_REGION`
- Support Claude Code-compatible Vertex env vars:
  - `ANTHROPIC_VERTEX_PROJECT_ID`
  - `CLOUD_ML_REGION`
- Support `GOOGLE_CLOUD_PROJECT` as a project ID fallback
- Add legacy `llm` config fields for Vertex AI
- Initialize the Anthropic SDK with Vertex AI Google auth
- Update English and Chinese docs
- Add tests for Vertex config resolution and client initialization

## Testing

- `go test ./...`

I also tested against a real GCP project. The request reached Vertex AI with `claude-sonnet-4-6`, but the project currently has zero quota for `global_online_prediction_requests_per_base_model`, so the live request was blocked by `429 RESOURCE_EXHAUSTED`.

This confirms the local ADC and Vertex routing path, while a full live completion still requires Anthropic model quota on the GCP project.

Fixes #56


lizhengfeng101 (commented):
## Code Review

Overall well-structured PR with clean design and good test coverage. A few items to consider:

### High Priority

**1. Strategy priority & error behavior in `resolver.go`**

`tryOCRVertexEnv` is placed **first** in the strategy list, ahead of the config file:

```go
{"OCR Vertex environment", tryOCRVertexEnv},
{"OCR config file", ...},
{"OCR environment", tryOCREnv},
...
```

The original priority order is config file > env. Now Vertex env jumps to the top. More critically, when `OCR_USE_ANTHROPIC_VERTEX=true` but model/project/region are missing, `tryOCRVertexEnv` **returns an error instead of falling back**, which blocks all subsequent strategies. A user who sets `OCR_USE_ANTHROPIC_VERTEX=true` but forgets `OCR_LLM_MODEL` will get an error even if they have a fully valid config file.

Suggestion: either move Vertex env after the config file (consistent with other env strategies), or return `(empty, false, nil)` instead of an error when required fields are missing, so resolution can fall through to the next strategy.

### Medium Priority

**2. Missing locale README updates**

`README.md` and `README.zh-CN.md` are updated, but `README.ru-RU.md`, `README.ja-JP.md`, and `README.ko-KR.md` are not. These should be synced for consistency.

**3. `defer/recover` for panic handling in `NewAnthropicVertexClient`**

```go
defer func() {
    if recovered := recover(); recovered != nil {
        client = nil
        err = fmt.Errorf("initialize Google Vertex AI authentication: %v", recovered)
    }
}()
```

Using `recover` for regular error paths is unusual in Go. If the Anthropic SDK's `WithGoogleAuth` can panic in specific scenarios (e.g., missing ADC), it would be helpful to add a comment explaining when/why this happens. Otherwise, if it doesn't actually panic, this can be removed.

### Low Priority

**4. No integration with the modern provider system**

Vertex AI is only supported via the legacy `llm` config block and env vars, not through the modern `providers` system (e.g., `ocr config set provider vertex-ai`). Fine for now, but worth noting as a follow-up.

**5. `UseVertex` type inconsistency**

`UseAnthropic` is `*bool` (to distinguish unset from false), while `UseVertex` is `bool`. The type choice is reasonable since the default behavior differs, but the JSON tag naming is inconsistent: `use_anthropic` vs `use_anthropic_vertex`. Consider `use_vertex` for brevity.

**6. Transitive dependency weight**

The PR pulls in `cloud.google.com/go/auth`, `google.golang.org/api`, `go.opencensus.io` (deprecated, Google is migrating to OpenTelemetry), etc. This is expected for Google Cloud integration but worth mentioning the binary size trade-off.

### Positives

- **Testability**: The `var vertexGoogleAuthOption` package-level variable for mock injection is a clean testing pattern.
- **Error messages**: Clear and actionable when configuration is incomplete.
- **`vertexBaseURL` function**: Correctly handles `global`, `us`, `eu` special regions.
- **Documentation**: English and Chinese READMEs are well-written with complete examples.
- **Test coverage**: Both resolver and client paths are covered, including the panic recovery edge case.

internal/llm/resolver.go
```diff
@@ -27,17 +34,23 @@ const (
 	envOCRLLMModel      = "OCR_LLM_MODEL"
 	envOCRLLMAuthHeader = "OCR_LLM_AUTH_HEADER"
 	envOCRUseAnthropic  = "OCR_USE_ANTHROPIC"
+	envOCRUseVertex     = "OCR_USE_ANTHROPIC_VERTEX"
```
MuoDoo: Maybe we should just use some name like `OCR_USE_VERTEX`, not just focus on ANTHROPIC models on it.

lizhengfeng101:
Hi @Devwallem, thanks for the time and effort you put into this PR — it's clear you thought through the different authentication paths and tested thoroughly against a real GCP project.

After reviewing the design, I've decided not to merge this. OCR is a short-lived, non-interactive CLI tool — a single review completes in seconds to minutes, well within the lifetime of a Google ADC token. This means users can already use Vertex AI today with the existing provider system:

```bash
export OCR_LLM_TOKEN=$(gcloud auth print-access-token)
export OCR_LLM_URL="https://REGION-aiplatform.googleapis.com/v1/projects/PROJECT/locations/REGION/publishers/anthropic/models"
export OCR_LLM_MODEL=claude-sonnet-4-6
export OCR_LLM_PROTOCOL=anthropic
```

Given that, adding a full ADC integration inside OCR would be over-engineering for a problem that a one-liner can solve. If there's interest, a documentation PR showing the Vertex AI setup example would be a great lightweight alternative.

That said, I really appreciate you identifying this use case and contributing. You're welcome to pick up other issues or ideas — looking forward to your next PR!

razor-1:
@lizhengfeng101 have you actually tried the above configuration? The Google aiplatform endpoints have a different shape as documented here: https://platform.claude.com/docs/en/build-with-claude/claude-on-vertex-ai

In my testing and analysis there is no way just changing the URL will work for Vertex.

lizhengfeng101:
@razor-1 Thanks for pointing this out! You're right — we don't actually have a Google AI Platform account, so our conclusion was based on a quick read of the docs rather than hands-on testing. That was not rigorous enough on our part.

If you have access to a GCP account and have been testing this, would you be interested in participating in further discussion or contributing a PR for proper Vertex AI support? We'd be very happy to collaborate on this and would love to have your input given your firsthand experience with the platform.

razor-1:
I'd be happy to help. Perhaps the author can clean up the review comment nits, but using the anthropic-sdk-go vertex support and allowing ADC seem like no-brainers for this integration.

Devwallem:
@razor-1 Thanks for offering to help! I think trying your approach first would be the most helpful way to move this issue forward, especially since you have hands-on experience with the Vertex AI integration.

Please feel free to take the lead on the implementation. You’re also welcome to reuse anything useful from this PR and its tests. I’d be happy to provide context if needed.

PR #174, opened 2026-06-18
feat: add HTTP(S) proxy support for install script

### LINKED ISSUE #173: feat: add HTTP(S) proxy support for install script and auto-update
### Problem Statement

Users behind corporate firewalls or in restricted network environments (e.g., China mainland accessing GitHub) cannot download the `opencodereview` binary via npm postinstall (`scripts/install.js`) because Node.js's `https.get` does not honor `HTTPS_PROXY` / `HTTP_PROXY` environment variables by default. The download simply hangs or times out.

The same issue affects the silent auto-updater (`scripts/update.js`).

### Proposed Solution

1. Add an `HttpsProxyAgent` class in `scripts/install.js` that implements an HTTPS CONNECT tunnel through an HTTP proxy, and wire it into the `download()` function.

2. Proxy source priority:
   - Environment variables: `HTTPS_PROXY` > `HTTP_PROXY` > `https_proxy` > `http_proxy`
   - Config file: `~/.opencodereview/config.json` → `proxy.url`

3. Provide a CLI helper script (`scripts/setup-proxy.js`) for users to set/unset/show the proxy configuration without manually editing JSON:
   - `node scripts/setup-proxy.js set http://127.0.0.1:7897`
   - `node scripts/setup-proxy.js unset`
   - `node scripts/setup-proxy.js show`

### Alternatives Considered

- Using a third-party npm package like `https-proxy-agent`: rejected to keep the install script dependency-free (it runs at postinstall, before any dependencies are available).
- Implementing only in the Go binary: rejected because the proxy is needed before the binary is downloaded — the install script itself must support it.

### Affected Area

Configuration, CLI / Commands

### Additional Context

This is common practice: pip (Python), curl, and Claude Code all support `HTTPS_PROXY` for the same reason.

grissomsh:
## Summary

Add HTTP(S) proxy support for the npm install script so users behind corporate firewalls or in restricted network environments can download the `opencodereview` binary.

### `scripts/install.js`
- Add `getProxyUrl()` — reads proxy config from env vars (`HTTPS_PROXY` > `HTTP_PROXY` > `https_proxy` > `http_proxy`) with fallback to `~/.opencodereview/config.json`
- Add `HttpsProxyAgent` class — implements HTTP CONNECT tunnel for HTTPS downloads through an HTTP proxy
- Support `Proxy-Authorization` Basic auth for authenticated proxies
- 30s timeout on proxy CONNECT handshake with proper cleanup
- Wire the agent into the `download()` function

### `scripts/setup-proxy.js` (new)
- CLI helper for managing proxy configuration without manually editing JSON
- Commands: `set <url>`, `unset`, `show`
- Validates proxy URL format on `set`

## How it works

```
HTTPS_PROXY/HTTP_PROXY env vars  (highest priority)
        ↓  falls back to
~/.opencodereview/config.json → proxy.url
        ↓
HttpsProxyAgent (HTTP CONNECT tunnel)
        ↓
HTTPS download through proxy
```

## Testing

```bash
# Configure proxy
node scripts/setup-proxy.js set http://127.0.0.1:7897
node scripts/setup-proxy.js show

# Test via env var override
HTTPS_PROXY=http://127.0.0.1:7897 npm install
```

Closes #173

### DIFF
--- a/scripts/config-paths.js
+++ b/scripts/config-paths.js
@@ -1,0 +1,9 @@
+"use strict";
+
+const path = require("path");
+const os = require("os");
+
+const CONFIG_DIR = path.join(os.homedir(), ".opencodereview");
+const CONFIG_PATH = path.join(CONFIG_DIR, "config.json");
+
+module.exports = { CONFIG_DIR, CONFIG_PATH };
--- a/scripts/install.js
+++ b/scripts/install.js
@@ -3,7 +3,9 @@
 
 const fs = require("fs");
 const path = require("path");
+const http = require("http");
 const https = require("https");
+const tls = require("tls");
 const crypto = require("crypto");
 
 const IS_WINDOWS = process.platform === "win32";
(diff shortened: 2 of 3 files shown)


lizhengfeng101:
Hi @grissomsh, thanks for the contribution and the detailed implementation! 🙏

A quick heads-up on the current state of the project: in recent versions we've switched to distributing the binary via **platform-specific npm packages** (`@alibaba-group/ocr-darwin-arm64`, etc.) as `optionalDependencies`. This means `npm install` downloads the binary through the npm registry itself — so the standard `npm config set https-proxy` / `npm config set registry` is all you need to get through a corporate proxy.

The postinstall download from GitHub Releases is now only a legacy fallback that triggers when the platform package is missing (e.g. `--ignore-optional`). For most users it will never execute.

**Suggestion:** Could you try upgrading to the latest version first? There's a good chance the proxy issue you encountered no longer exists with the new distribution model. If your npm registry is accessible (directly or via mirror/proxy), the install should just work.

If there's still a scenario where the fallback download is needed behind a proxy, we can revisit with a much smaller change — just reading `HTTPS_PROXY`/`HTTP_PROXY` env vars in `download()` would be sufficient, without the config file system and CLI tool. But let's first confirm whether the problem still exists for you on the latest version.

Thanks again for caring about the install experience!

grissomsh:
Thanks for the thorough review @lizhengfeng101! 🙏

You're absolutely right — with the new distribution model using platform-specific npm packages as `optionalDependencies`, the binary comes through the npm registry directly. Standard `npm config set https-proxy` handles proxy scenarios just fine, and the GitHub Releases fallback in `install.js` rarely executes in practice.

Closing this PR in favor of the existing npm proxy mechanism. Appreciate the detailed explanation!

PR #179, opened 2026-06-20
feat: persist review results and add viewer APIs

windy-bing:
## Summary

  This PR adds support for persisted review results and shared enterprise review rules.

  Changes include:

  - Add persisted review result storage for `ocr review --save-result`
  - Add configurable result storage via `--result-dir` and `OCR_REVIEWS_DIR`
  - Add `ocr viewer --reviews-dir` so viewer can read mounted/shared result storage
  - Add review result HTML pages and JSON APIs:
    - `/reviews`
    - `/reviews/{project}`
    - `/reviews/{project}/{reviewID}`
    - `/api/reviews`
    - `/api/reviews/{project}`
    - `/api/reviews/{project}/{reviewID}`
  - Add project/source/target branch filtering for saved review results
  - Add shared enterprise rule directory support via `--rules-dir` and `OCR_RULES_DIR`
  - Add enterprise project rule lookup using `OCR_PROJECT`, GitLab `CI_PROJECT_PATH`, or repo directory name
  - Update README files and CI examples for GitLab/Jenkins/containerized usage

  ## Motivation

  CI jobs often run inside short-lived containers. Previously, review history/results were stored under the
container home directory and were hard to expose after the job finished.

  This PR makes review results suitable for CI environments by allowing results to be written to a mounted
directory and viewed later by a long-running `ocr viewer` service.

  It also adds a shared rules directory mechanism so teams can maintain organization-wide and project-specific
review knowledge outside individual repositories.

  ## Rule Priority

  Review rules now resolve in this order:

  1. `--rule`
  2. `<repoDir>/.opencodereview/rule.json`
  3. `<rules-dir>/projects/<project>/rule.json`
  4. `<rules-dir>/global.json`
  5. `~/.opencodereview/rule.json`
  6. built-in system rules

  ## Testing

  ```bash
  go test ./...

  All tests pass.

### DIFF
--- a/README.ja-JP.md
+++ b/README.ja-JP.md
@@ -288,11 +288,29 @@ CI統合のコアコマンド：
 ```bash
 ocr review \
   --from "origin/main" \
-  --to "origin/feature-branch" \
+  --to "<commit_sha>" \
   --format json
 ```
 
+`--from`フラグはベースとしてブランチref（例：`origin/main`）またはコミットSHAを受け取り、`--to`はheadとしてコミットSHAまたはブランチrefを受け取ります。CI環境では、fork PR/MRのようにsource branchが`origin` remoteに存在しないケースに対応するため、`--to`にはコミットSHAを使うことを推奨します。
+
 `--format json`フラグは、CIスクリプトでのパースに適した機械可読な結果を出力します。
+
+WebUIや下流サービスで最終レビュー結果を参照する場合は、`--save-result`を追加します。デフォルトでは`~/.opencodereview/reviews`に保存されます。GitLab RunnerやJenkins agentなどのコンテナCIで実行する場合は、`--result-dir`または`OCR_REVIEWS_DIR`を使い、そのディレクトリを永続ストレージにマウントしてください。
+
+```bash
+ocr review \
+  --from "$CI_MERGE_REQUEST_TARGET_BRANCH_NAME" \
+  --to "$CI_COMMIT_SHA" \
+  --format json \
+  --save-result \
+  --result-dir /ocr-data/reviews \
+  --result-project "$CI_PROJECT_PATH" \
+  --result-source-branch "$CI_MERGE_REQUEST_SOURCE_BRANCH_NAME" \
+  --result-target-branch "$CI_MERGE_REQUEST_TARGET_BRANCH_NAME"
+```
+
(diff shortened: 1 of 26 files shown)


lizhengfeng101 (commented):
## Suggestion: Simplify project directory naming

In `internal/reviewstore/store.go`, the `ProjectKey` function uses base64 encoding for project directory names:

```go
func ProjectKey(project ProjectInfo) string {
    ...
    return base64.RawURLEncoding.EncodeToString([]byte(key))
}
```

This produces unreadable directory names like `VXNlcnMta2l0ZS1EZXNrdG9wLW15LXByb2plY3Q`, which makes it hard for users to browse or debug review results directly in the filesystem.

The existing session storage in `internal/session/persist.go` uses a simpler, human-readable approach (`encodeRepoPath`) — replacing `/` with `-` and handling Windows volume names:

```
~/.opencodereview/sessions/Users-kite-Desktop-my-project/xxx.jsonl
```

I'd suggest adopting the same pattern here for consistency and better usability. The project key typically doesn't contain characters that require base64 encoding, and a readable directory name makes it much easier to inspect stored reviews without tooling.

windy-bing:
Thanks for the suggestion. The original base64 encoding was intentional: it avoids exposing project names
  directly in the filesystem and makes review result directories less immediately enumerable by project name.

  That said, consistency with the existing session storage matters here. I agree the current behavior should be
  unified and easier to reason about across sessions and reviews. I’ve updated the review result project key
  generation to follow the same readable path-encoding style used by session persistence, while keeping path
  segment validation strict enough to prevent traversal issues.

  This makes the storage layout consistent and more debuggable:

  ~/.opencodereview/reviews/Users-kite-Desktop-my-project/<review-id>.json

  instead of opaque base64 directory names.

@lizhengfeng101

lizhengfeng101:
Hey @windy-bing, thanks a lot for putting this together — it's a genuinely useful direction, and the CI/container motivation was spot on.

Since you opened this, main has grown to cover most of what this PR set out to do:

- **Persisted results & audit** — `ocr review --format json --output <path>` writes a structured JSON report you can drop into a mounted volume and archive for auditing.
- **Browsing results in a browser** — the `ocr viewer` session pages already render the full review output ("Review Comments / findings"), with severity and category filters, plus branch/repo columns on the listing pages.

Between those two, the core needs here — persist, audit, and browse — are already served, and introducing a separate review-result store alongside the existing session persistence would give us two parallel storage paths to maintain. Combined with the current merge conflicts, I think the cleanest call is to close this one.

Really appreciate the effort and the thoughtful write-up — please don't let this discourage you from contributing again. 🙏

PR #182, opened 2026-06-21
fix(agent): recover from panics in per-file review and comment-pool goroutines (#171)

### LINKED ISSUE #171: Unrecovered panic in a per-file review goroutine crashes the whole run (missing recover, CWE-248)
### OpenCodeReview Version

V1.3.19

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

Anthropic (Claude)

### Bug Description

`dispatchSubtasks` reviews each changed file in its own goroutine and is built
around per-file fault isolation: an `error` from `executeSubtask` is caught,
counted, recorded as a `subtask_error` warning, and the other files continue
(internal/agent/agent.go:456-475). The same applies to `CommentWorkerPool.Submit`
(agent.go:206-216). But neither goroutine has a `recover()` — their deferred
calls only do `wg.Done()` and semaphore release.

In Go, an unrecovered panic in *any* goroutine terminates the entire process.
So a panic while reviewing one file (from any latent bug, an unexpected
LLM/SDK response shape, or a malformed-diff edge case) does not fail just that
file the way an error does — it aborts the whole `ocr` run and discards every
other file's results. This silently defeats the per-file isolation the code
otherwise implements.

### Steps to Reproduce

The crash is in the goroutine contract itself: `dispatchSubtasks` spawns one
goroutine per file whose only deferred work is `wg.Done()` + semaphore release,
with no `recover()`. Any panic in one of those goroutines aborts the whole
process (Go terminates on an unrecovered panic in any goroutine), discarding all
other files' results — whereas an `error` from the same file would be isolated
and counted.

### A. Minimal, self-contained reproduction of the exact pattern
Save and run this (it mirrors agent.go:456-475 — per-file goroutine, defer
wg.Done()+sem release, no recover):

```go
package main

import ("fmt"; "sync"; "sync/atomic")

func main() {
    files := []string{"a.go", "b.go", "EVIL.go", "c.go", "d.go"}
    var wg sync.WaitGroup
    sem := make(chan struct{}, 8)
    var completed int64
    for _, f := range files {
        wg.Add(1); sem <- struct{}{}
        go func(name string) {
            defer wg.Done()
            defer func() { <-sem }() // current code: NO recover
            executeSubtask(name)
            atomic.AddInt64(&completed, 1)
        }(f)
    }
    wg.Wait()
    fmt.Printf("review finished; %d/%d files completed\n", completed, len(files))
}
func executeSubtask(name string) {
    if name == "EVIL.go" { panic("index out of range while reviewing " + name) }
}

### Expected Behavior

A failure isolated to one file's review — including a panic — should be caught,
recorded as a per-file failure/warning, and leave the other concurrent reviews
unaffected, exactly as `error` returns already are.


### Logs / Error Output

```shell

```

## Actual behaviour
A panic in any subtask (or comment-pool) goroutine propagates uncaught and
crashes the entire process, losing all in-flight and completed-but-uncollected
results.

## Suggested fix
Add a `recover()` to each worker goroutine that converts a panic into the same
recorded per-file failure path used for errors.

eldar702:
## Problem (CWE-248: Uncaught Exception)

`ocr` reviews each file in its own goroutine, and the comment worker pool runs
each submitted unit of work in its own goroutine too. Both sites defer only
`wg.Done()` + semaphore release — **neither has a `recover()`**:

- `CommentWorkerPool.Submit` (`internal/agent/agent.go`)
- the per-file review goroutine in `dispatchSubtasks` (`internal/agent/agent.go`)

In Go, an **unrecovered panic in any goroutine terminates the entire process**.
So a panic while reviewing a single file (a malformed diff, a nil deref deep in
a tool/LLM path, etc.) aborts the whole `ocr` run and discards every other
file's results. This directly defeats the per-file fault isolation the code
*already* implements for `error` returns: an `error` from one file is counted,
recorded as a `subtask_error` warning, and lets the other files finish — but a
`panic` from the same code path crashes everything. The isolation is only
half-built.

## Fix

Add a `recover()` defer to both goroutine sites. The crux is **defer ordering**:
Go runs deferred calls LIFO, so the recover defer is registered **last** in each
goroutine and therefore runs **first** on unwind — it recovers *before* the
semaphore-release and `wg.Done()` defers run, and all of them still execute.
Result: the semaphore is always drained and the `WaitGroup` always decremented,
so there is no goroutine/semaphore leak and `Await()` / `wg.Wait()` never
deadlocks. The recover defer itself does not re-panic.

- **`CommentWorkerPool.Submit`** — on recover, log the panic to stdout. This
  matches the depth of the existing error handling on that path (a stderr log);
  the pool has no per-task counter to update.
- **`dispatchSubtasks`** — on recover, convert the panic into the **existing
  counted warning shape**: `atomic.AddInt64(&a.subtaskFailed, 1)` +
  `a.recordWarning("subtask_error", d.NewPath, fmt.Sprintf("panic: %v", r))` +
  a stderr line. It reuses the **existing `subtask_error` warning type** (no new
  type invented), so panics and errors surface identically downstream and the
  end-of-run all-failed rollup (`failed > 0 && failed == dispatched`) stays
  correct. The recover defer is registered before the timeout-`cancel()` defer,
  so `cancel()` still runs first on unwind (the file context is cancelled before
  the panic is contained) — no ordering hazard.

No new imports (`fmt`, `sync/atomic`, and the `stdout` package were already
imported). The happy path is unchanged; no signatures change.

## Tests — genuine RED → GREEN under `-race`

`TestCommentWorkerPool_PanicIsIsolated` (new `internal/agent/worker_pool_test.go`)
submits a panicking closure plus a healthy closure and asserts that `Await()`
does not crash and returns the healthy result (`len == 1`).

```
$ go test -race -run TestCommentWorkerPool_PanicIsIsolated ./internal/agent/...
```

**RED (before the Site A fix)** — the panic in one closure crashes the whole
test process:

```
panic: boom while reviewing EVIL.go

goroutine 6 [running]:
github.com/open-code-review/open-code-review/internal/agent.TestCommentWorkerPool_PanicIsIsolated.func1()
	/tmp/oss-r4/open-code-review/internal/agent/worker_pool_test.go:20 +0x34
github.com/open-code-review/open-code-review/internal/agent.(*CommentWorkerPool).Submit.func1()
	/tmp/oss-r4/open-code-review/internal/agent/agent.go:162 +0xd4
created by github.com/open-code-review/open-code-review/internal/agent.(*CommentWorkerPool).Submit in goroutine 5
	/tmp/oss-r4/open-code-review/internal/agent/agent.go:157 +0xd4
FAIL	github.com/open-code-review/open-code-review/internal/agent	0.376s
FAIL
```

**GREEN (after the fix)** — the panic is contained, the healthy closure's result
still arrives, race detector clean:

```
ok  	github.com/open-code-review/open-code-review/internal/agent	1.526s
```

Also verified: `go build ./...` (clean), `go vet ./internal/agent/...` (clean),
`gofmt -l internal/agent/agent.go internal/agent/worker_pool_test.go` (empty),
and the full `go test -race ./internal/agent/...` suite passes.

### Test-coverage honesty

`Submit` takes a caller-supplied closure, so Site A is **directly unit-tested**
with no new production seam. `dispatchSubtasks` calls the private
`executeSubtask`, which needs a fully-wired `Agent` + LLM client + parsed diffs;
forcing a panic there in a unit test would require adding a net-new production
seam purely for the test. Rather than reshape production code for a test hook,
**Site B is covered by the identical `recover()` idiom + code review** (it
converts the panic into the same `subtask_error` warning path that is already
exercised for `error` returns). This asymmetry is intentional and called out so
reviewers know Site B is pattern-covered, not unit-covered.

Fixes #171

---

*AI-assisted disclosure:* this change was prepared with AI assistance (analysis,
patch, and test authored with an AI coding assistant) and reviewed by the
submitter. The RED→GREEN `-race` output above is from a real local run on the
checked-out code.

---

## Updates since opening (review follow-ups)

- **`6d00066`** — capture `debug.Stack()` in both recover logs (OpenCodeReview bot suggestion). This adds one import, `runtime/debug`.
- **`b5e2ca3`** — Site B panic-recover now calls `telemetry.ErrorEvent(ctx, "subtask.panic", …)` for telemetry parity with the error path (uses the parent `ctx`, not the already-cancelled `fileCtx`); and `CommentWorkerPool.Await` now documents that a panicking unit of work is recovered+logged but not surfaced in the result.

**Correction:** the "No new imports" note above is superseded — `runtime/debug` is now imported (for `debug.Stack()`). `fmt`, `sync/atomic`, `stdout`, and `telemetry` were already imported.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"runtime/debug"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -347,6 +348,21 @@ func (a *Agent) dispatchSubtasks(ctx context.Context) ([]model.LlmComment, error
 		go func(d model.Diff) {
 			defer wg.Done()
 			defer func() { <-sem }() // release
+			// A panic while reviewing one file must be isolated exactly like an
+			// error return: counted in subtaskFailed and recorded as a
+			// subtask_error warning, so other files still complete and the
+			// all-failed rollup below stays correct. Registered before the
+			// timeout-cancel defer, so cancel() still runs first on unwind and
+			// fileCtx is already cancelled here — use the parent ctx for telemetry.
+			defer func() {
+				if r := recover(); r != nil {
+					atomic.AddInt64(&a.subtaskFailed, 1)
+					fmt.Fprintf(stdout.Writer(), "[ocr] Subtask panic for %s: %v\n%s\n", d.NewPath, r, debug.Stack())
+					telemetry.ErrorEvent(ctx, "subtask.panic", fmt.Errorf("panic: %v", r),
+						telemetry.AnyToAttr("file.path", d.NewPath))
+					a.recordWarning("subtask_error", d.NewPath, fmt.Sprintf("panic: %v", r))
+				}
+			}()
 
 			var fileCtx context.Context
 			var cancel context.CancelFunc
--- a/internal/llmloop/pool.go
+++ b/internal/llmloop/pool.go
@@ -9,6 +9,7 @@ package llmloop
 
 import (
 	"fmt"
+	"runtime/debug"
 	"sync"
 
 	"github.com/open-code-review/open-code-review/internal/model"
@@ -55,6 +56,14 @@ func (p *CommentWorkerPool) Submit(f func() ([]model.LlmComment, error)) {
 	p.wg.Go(func() {
 		p.semaphore <- struct{}{}
 		defer func() { <-p.semaphore }()
+		// Contain a panic in the submitted work so one bad unit of work cannot
+		// crash the whole process. The work that panics contributes no comments;
+		// the semaphore is still released via the defer above.
+		defer func() {
+			if r := recover(); r != nil {
+				fmt.Fprintf(stdout.Writer(), "[ocr] CommentWorkerPool panic: %v\n%s\n", r, debug.Stack())
+			}
+		}()
 
 		comments, err := f()
 		if err != nil {
@@ -68,6 +77,16 @@ func (p *CommentWorkerPool) Submit(f func() ([]model.LlmComment, error)) {
 
 // Await blocks until all submitted work has completed and returns
 // aggregated results from every Submit call so far.
+//
+// A panic in submitted work is recovered and logged inside Submit (see the
+// recover defer there) but is not surfaced here as an error or reflected in
+// the returned count — a unit that panics contributes no comments and is
+// indistinguishable from one that produced zero.
+//
+// Concurrency contract: Await must not run concurrently with Submit. Submit
+// calls wg.Go (which does wg.Add(1) synchronously), so a Submit racing Await
+// would risk sync.WaitGroup's "Add called concurrently with Wait" panic.
+// Callers must ensure every Submit has returned before calling Await.
 func (p *CommentWorkerPool) Await() []model.LlmComment {
 	p.wg.Wait()
 	return p.results
--- a/internal/llmloop/pool_test.go
+++ b/internal/llmloop/pool_test.go
@@ -97,3 +97,24 @@ func TestCommentWorkerPool_AwaitEmpty(t *testing.T) {
 		t.Errorf("expected nil for no submissions, got %v", results)
 	}
 }
+
+func TestCommentWorkerPool_PanicIsIsolated(t *testing.T) {
+	p := NewCommentWorkerPool(2)
+
+	p.Submit(func() ([]model.LlmComment, error) {
+		panic("boom in submitted work")
+	})
+	p.Submit(func() ([]model.LlmComment, error) {
+		return []model.LlmComment{{Path: "healthy.go", Content: "fine"}}, nil
+	})
+
+	// Await must not crash: the recovered panic contributes no comments, and the
+	// healthy task's result is still collected.
+	results := p.Await()
+	if len(results) != 1 {
+		t.Fatalf("expected 1 result after a panicking task, got %d", len(results))
+	}
+	if results[0].Path != "healthy.go" {
+		t.Errorf("Path = %q, want healthy.go", results[0].Path)
+	}
+}


eldar702:
Good catch — addressed in `6d00066`: both recover handlers now log `debug.Stack()` (via `runtime/debug`) alongside the panic value, so a swallowed goroutine panic carries a full stack for production debugging. `go build`/`go vet` are clean and the `internal/agent` tests pass.

lizhengfeng101 (commented):
## PR #182 Code Review

**Changes:** 2 files, +52/-0

### Overview

This PR adds `recover()` guards to two goroutine entry points in `agent.go`:
1. **Site A**: `CommentWorkerPool.Submit` — comment worker pool goroutines
2. **Site B**: `dispatchSubtasks` — per-file review goroutines

The goal is to prevent a panic in a single file review from crashing the entire process (CWE-248), converting panics into warnings consistent with the existing error handling path.

### Positives

- **Correct defer ordering**: Both `recover()` defers are registered last (run first in LIFO), ensuring panic recovery happens before semaphore release and `wg.Done()` — no deadlocks or goroutine leaks.
- **Consistent with existing error path**: Site B reuses the `subtask_error` warning type and `subtaskFailed` counter, so the `all-failed` rollup logic remains correct.
- **`recordWarning` is thread-safe**: Confirmed it uses `warningsMu` mutex protection (`agent.go:306-312`).
- **Good test quality**: The test case clearly validates panic isolation (panicking closure doesn't affect healthy closure). RED→GREEN evidence in PR description is thorough.
- **Excellent PR description**: Deep problem analysis, clear fix rationale, honest test-coverage disclosure.

### Issues to Address

#### 1. PR description claims "No new imports" — inaccurate
The PR body states *"No new imports"*, but `runtime/debug` **is** a newly added import (not previously present in `agent.go`). Minor inaccuracy, but the description should be precise.

#### 2. Site B is missing telemetry reporting
The existing error path has a `telemetry.ErrorEvent` call (`agent.go:421-422`), but the panic recovery path does not. This means panics will appear in warnings and stderr but **not in telemetry/tracing**.

Suggest adding telemetry in the recover block, using `ctx` (not `fileCtx`, since `cancel()` runs before recover in LIFO order and `fileCtx` may already be cancelled):

```go
if r := recover(); r != nil {
    atomic.AddInt64(&a.subtaskFailed, 1)
    fmt.Fprintf(stdout.Writer(), "[ocr] Subtask panic for %s: %v\n%s\n", d.NewPath, r, debug.Stack())
    a.recordWarning("subtask_error", d.NewPath, fmt.Sprintf("panic: %v", r))
    telemetry.ErrorEvent(ctx, "subtask.panic", fmt.Errorf("panic: %v", r),
        telemetry.AnyToAttr("file.path", d.NewPath))
}
```

#### 3. Site A silently swallows panics
`CommentWorkerPool.Submit`'s recover only prints a log. The caller of `Await()` cannot distinguish between "0 comments because nothing to report" and "0 comments because processing panicked". Consider whether a similar warning/error counting mechanism is needed, or at minimum document this behavior in `Await`'s docstring.

#### 4. What if operations inside recover themselves panic?
If `a.recordWarning` or `debug.Stack()` itself panics (extremely rare but theoretically possible), the second panic would not be caught by recover and would still crash the process. A conservative nested recover could guard against this, but this is a very low-priority edge case.

### Summary

This is a **high-quality defensive fix** addressing a real process crash risk. Defer ordering, concurrency safety, and consistency with the existing error path are all handled well. The main suggestion is to **add telemetry reporting in Site B** so panics and errors have consistent observability. The remaining issues are lower priority and can be addressed as follow-ups.

**Recommendation: merge after adding telemetry.**

eldar702:
Thanks for the careful review — addressed in `b5e2ca3`:

1. **"No new imports" was inaccurate** — correct, `runtime/debug` was added for `debug.Stack()`. I've fixed that line in the PR description.
2. **Site B telemetry parity** — added, mirroring the error path:
   ```go
   telemetry.ErrorEvent(ctx, "subtask.panic", fmt.Errorf("panic: %v", r),
       telemetry.AnyToAttr("file.path", d.NewPath))
   ```
   Using the parent `ctx` exactly as you noted — the per-file `cancel()` defer runs before this recover defer (LIFO), so `fileCtx` is already cancelled here. Now panics and errors have consistent telemetry/tracing.
3. **Site A silent swallow** — documented on `CommentWorkerPool.Await`'s docstring (a panicking unit of work is recovered + logged in `Submit` but not surfaced as an error or reflected in the count, so callers shouldn't rely on the count alone to detect failures). I kept the pool's behavior/signature unchanged per your "at minimum document" suggestion; happy to add an explicit panic counter if you'd prefer the stronger version.
4. **Nested recover in the handler** — agreed it's a very-low-priority edge. `recordWarning` (a mutex-guarded slice append) and `debug.Stack()` are panic-free in practice, so I left it out to keep the diff minimal; trivial to add a defensive nested recover if you'd like the belt-and-suspenders.

`go build` / `go vet` / `gofmt` clean; `go test -race ./internal/agent/...` green.

eldar702:
Thanks for flagging the `Await`/`Submit` WaitGroup interaction. In the current code this race can't occur: `dispatchSubtasks` joins all of its per-file goroutines via `wg.Wait()` **before** it calls `CommentWorkerPool.Await()`, and `Submit` does `wg.Add(1)` *synchronously* before launching its goroutine — so every `Add` has completed by the time `Await` runs, and no `Submit` overlaps `Await`. The async comment-tool path only submits from within those per-file goroutines, so it's covered by the same join.

Rather than redesign the pool (not needed for this submit-all-then-await-once usage), I pushed a commit (`0329f65`) that documents this as an explicit concurrency contract on `Await()`, so future callers preserve the invariant.

lizhengfeng101:
Hi @eldar702, this PR currently has merge conflicts with `main`. Could you please rebase or merge the latest `main` into your branch to resolve them? Thanks!

eldar702:
Done — rebased onto the latest `main` and force-pushed; conflicts resolved.

Heads-up on one thing the rebase surfaced: `main` moved `CommentWorkerPool` into `internal/llmloop` since this PR was opened, so to keep the fix in the right place I split it accordingly:
- per-file review panic recovery → `internal/agent/agent.go` (`dispatchSubtasks`)
- comment-pool panic recovery + the panic-isolation test → `internal/llmloop/pool.go` and `pool_test.go`

`go build ./...`, `go vet`, and the `internal/llmloop` tests pass locally. Ready for re-review — thanks for the ping!

lizhengfeng101 (approved):
LGTM

eldar702:
Thanks for the merge, @lizhengfeng101 — and for the nudge to rebase! Glad the panic-recovery fix landed cleanly on the refactored `llmloop` structure. 🙏

PR #189, opened 2026-06-23
feat: add ability to delete custom providers from configuration

### LINKED ISSUE #136: Add a way to delete custom providers from configuration
## Problem

Open Code Review currently supports adding and updating custom providers through `ocr config provider` and `ocr config set custom_providers.<name>.<field> <value>`, but there does not appear to be a supported way to delete a custom provider once it has been saved.

Users currently need to manually edit `~/.opencodereview/config.json` and remove the corresponding `custom_providers.<name>` entry. If the deleted custom provider is also the active `provider`, they also need to manually clear or change `provider` to avoid configuration resolution errors.

## Current Behavior

- `ocr config` supports `provider`, `model`, and `set`.
- `ocr config set custom_providers.<name>.<field> <value>` creates or updates custom provider fields.
- The Custom tab in `ocr config provider` supports selecting existing custom providers and adding new ones.
- There is no documented `delete`, `remove`, `unset`, or equivalent command for custom providers.

## Expected Behavior

Provide a supported way to delete custom providers.

A TUI-based deletion flow may be preferable, since custom providers are primarily managed through `ocr config provider`. We should also investigate how OpenCode handles provider deletion/removal in its TUI before finalizing the interaction design.

Possible approaches:

- Add a delete action in the Custom tab of `ocr config provider`, for example pressing `d` on a selected custom provider and confirming deletion.
- Optionally also add a non-interactive command, for example:
  ```bash
  ocr config delete custom_providers.<name>
  ```
  or:
  ```bash
  ocr config unset custom_providers.<name>
  ```

When deleting the currently active provider, the command should either:

- clear the active `provider` and related fallback `model`, or
- require confirmation and guide the user to select another provider.

## Acceptance Criteria

- Users can delete a saved custom provider without manually editing `~/.opencodereview/config.json`.
- Deleting the active provider leaves the config in a valid state or provides a clear error/confirmation flow.
- Tests cover deleting an inactive custom provider and deleting the currently active custom provider.
- README/docs mention the supported deletion workflow.

yingjiexu2002:
## Description

Add the ability to delete custom providers from configuration, addressing issue #136.

**CLI:** New `ocr config unset custom_providers.<name>` command that validates the key format, removes the provider from config, and clears `provider`/`model` fields if the deleted provider was active.

**TUI:** Press `d` on any custom provider in the Custom tab to trigger a `y/n` confirmation prompt. Shows a warning when targeting the active provider. Deletions persist even if the user cancels provider selection afterward. Cursor position adjusts automatically after deletion.

## Type of Change

## How Has This Been Tested?

10 new unit tests added (5 CLI + 5 TUI):

- **CLI:** `TestParseConfigArgsUnset`, `TestParseConfigArgsUnsetMissingKey`, `TestUnsetCustomProvider`, `TestUnsetActiveCustomProvider`, `TestUnsetInvalidKey`
- **TUI:** `TestProviderTUI_DeleteCustomProvider`, `TestProviderTUI_DeleteCustomProviderCancel`, `TestProviderTUI_DeleteOnAddOptionIgnored`, `TestProviderTUI_DeleteActiveCustomProvider`, `TestProviderTUI_DeleteEscCancels`

Full test suite passes with zero regressions (52 tests in `cmd/opencodereview`).

Manually verified:
- `ocr config unset custom_providers.my-gateway` deletes the provider and outputs confirmation
- `ocr config unset custom_providers.active-provider` clears active provider/model and prompts reconfiguration
- `ocr config unset llm.url` correctly rejects non-custom-provider keys
- TUI `d` → `y` flow deletes and adjusts cursor; `d` → `n`/`Esc` cancels; `d` on "Add" option is ignored
image
image

## Checklist

- [x] I have signed the CLA

## Related Issues

Closes #136

### DIFF
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -94,16 +94,40 @@ func runConfigUnset(key string) error {
 		return err
 	}
 
+	return unsetCustomProvider(configPath, name)
+}
+
+func unsetCustomProvider(configPath, name string) error {
 	cfg, err := loadOrCreateConfig(configPath)
 	if err != nil {
 		return fmt.Errorf("load config: %w", err)
 	}
 
+	wasActive, err := deleteCustomProvider(cfg, name)
+	if err != nil {
+		return err
+	}
+
+	if err := saveConfig(configPath, cfg); err != nil {
+		return err
+	}
+
+	fmt.Printf("Deleted custom provider %q.\n", name)
+	if wasActive {
+		fmt.Println("This was the active provider; 'provider' and 'model' have been cleared.")
+		fmt.Println("Run 'ocr config provider' to select a new provider.")
+	}
+	return nil
+}
+
+// deleteCustomProvider removes a custom provider from cfg in memory.
+// Returns true if the deleted provider was the active one.
+func deleteCustomProvider(cfg *Config, name string) (bool, error) {
 	if cfg.CustomProviders == nil {
-		return fmt.Errorf("custom provider %q not found", name)
+		return false, fmt.Errorf("custom provider %q not found", name)
 	}
 	if _, exists := cfg.CustomProviders[name]; !exists {
-		return fmt.Errorf("custom provider %q not found", name)
+		return false, fmt.Errorf("custom provider %q not found", name)
 	}
 
 	wasActive := cfg.Provider == name
@@ -117,16 +141,7 @@ func runConfigUnset(key string) error {
 		cfg.Model = ""
 	}
 
-	if err := saveConfig(configPath, cfg); err != nil {
-		return err
-	}
-
-	fmt.Printf("Deleted custom provider %q.\n", name)
-	if wasActive {
-		fmt.Println("This was the active provider; 'provider' and 'model' have been cleared.")
-		fmt.Println("Run 'ocr config provider' to select a new provider.")
-	}
-	return nil
+	return wasActive, nil
 }
 
 // ProviderEntry holds per-provider configuration in the providers map.
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -114,8 +114,8 @@ func unsetCustomProvider(configPath, name string) error {
 
 	fmt.Printf("Deleted custom provider %q.\n", name)
 	if wasActive {
-		fmt.Println("This was the active provider; 'provider' and 'model' have been cleared.")
-		fmt.Println("Run 'ocr config provider' to select a new provider.")
+		fmt.Fprintf(os.Stderr, "[ocr] WARNING: active provider was deleted; 'provider' and 'model' have been cleared.\n")
+		fmt.Fprintf(os.Stderr, "[ocr] Run 'ocr config provider' to select a new provider.\n")
 	}
 	return nil
 }
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -85,7 +85,7 @@ func runConfigSet(key, value string) error {
 func runConfigUnset(key string) error {
 	parts := strings.SplitN(key, ".", 2)
 	if len(parts) != 2 || parts[0] != "custom_providers" || parts[1] == "" {
-		return fmt.Errorf("unset only supports custom_providers.<name>\nUsage: ocr config unset custom_providers.<name>")
+		return fmt.Errorf("unset only supports custom_providers.<name>")
 	}
 	name := parts[1]
 
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -47,6 +47,8 @@ func runConfig(args []string) error {
 	switch action.subCmd {
 	case "set":
 		return runConfigSet(action.key, action.value)
+	case "unset":
+		return runConfigUnset(action.key)
 	default:
 		return fmt.Errorf("unknown config sub-command: %s", action.subCmd)
 	}
@@ -77,6 +79,53 @@ func runConfigSet(key, value string) error {
 		displayValue = maskKey(value)
 	}
 	fmt.Printf("Set %s = %s\n", key, displayValue)
+	return nil
+}
+
+func runConfigUnset(key string) error {
+	parts := strings.SplitN(key, ".", 2)
+	if len(parts) != 2 || parts[0] != "custom_providers" || parts[1] == "" {
+		return fmt.Errorf("unset only supports custom_providers.<name>\nUsage: ocr config unset custom_providers.<name>")
+	}
+	name := parts[1]
+
+	configPath, err := defaultConfigPath()
+	if err != nil {
+		return err
+	}
+
+	cfg, err := loadOrCreateConfig(configPath)
+	if err != nil {
+		return fmt.Errorf("load config: %w", err)
+	}
+
+	if cfg.CustomProviders == nil {
+		return fmt.Errorf("custom provider %q not found", name)
+	}
+	if _, exists := cfg.CustomProviders[name]; !exists {
+		return fmt.Errorf("custom provider %q not found", name)
+	}
+
+	wasActive := cfg.Provider == name
+	delete(cfg.CustomProviders, name)
+	if len(cfg.CustomProviders) == 0 {
+		cfg.CustomProviders = nil
+	}
+
+	if wasActive {
+		cfg.Provider = ""
+		cfg.Model = ""
+	}
+
+	if err := saveConfig(configPath, cfg); err != nil {
+		return err
+	}
+
+	fmt.Printf("Deleted custom provider %q.\n", name)
+	if wasActive {
+		fmt.Println("This was the active provider; 'provider' and 'model' have been cleared.")
+		fmt.Println("Run 'ocr config provider' to select a new provider.")
+	}
 	return nil
 }
 
--- a/cmd/opencodereview/config_cmd_test.go
+++ b/cmd/opencodereview/config_cmd_test.go
@@ -312,18 +312,6 @@ func TestUnsetActiveCustomProvider(t *testing.T) {
 }
 
 func TestUnsetInvalidKey(t *testing.T) {
-	dir := t.TempDir()
-	configPath := dir + "/config.json"
-
-	cfg := &Config{
-		CustomProviders: map[string]ProviderEntry{
-			"my-gateway": {URL: "https://gw.example.com/v1"},
-		},
-	}
-	if err := saveConfig(configPath, cfg); err != nil {
-		t.Fatalf("saveConfig: %v", err)
-	}
-
 	tests := []struct {
 		name    string
 		wantErr bool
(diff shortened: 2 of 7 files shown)


cmd/opencodereview/config_cmd_test.go
```diff
@@ -218,3 +221,117 @@ func TestSetConfigValueModelWithCustomProvider(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("loadOrCreateConfig: %v", err)
+	}
+
+	delete(cfg.CustomProviders, "my-gateway")
+	if len(cfg.CustomProviders) == 0 {
+		cfg.CustomProviders = nil
+	}
+	if err := saveConfig(configPath, cfg); err != nil {
+		t.Fatalf("saveConfig: %v", err)
```
lizhengfeng101: **[Critical] Tests do not call the actual `runConfigUnset` function**

This test manually inlines the `delete` + `saveConfig` logic instead of calling `runConfigUnset("custom_providers.my-gateway")`. This means if `runConfigUnset` has a bug (e.g., missing the `wasActive` check, `defaultConfigPath` error handling, etc.), the test would still pass.

**Suggestion:** Extract the core logic of `runConfigUnset` into an internal function that accepts a `configPath` parameter (e.g., `unsetCustomProvider(configPath, key string) error`), then call the actual function in tests using a `t.TempDir()` path. This way the tests will actually cover the validation logic, `wasActive` handling, and `CustomProviders = nil` cleanup.

The same issue applies to `TestUnsetActiveCustomProvider` and `TestUnsetInvalidKey` below.

cmd/opencodereview/config_cmd_test.go
```diff
@@ -218,3 +221,117 @@ func TestSetConfigValueModelWithCustomProvider(t *testing.T) {
...
+			"other-gateway": {URL: "https://other.example.com/v1", Protocol: "openai", Model: "other-model"},
+		},
+	}
+
+	name := "my-gateway"
+	if cfg.Provider == name {
+		cfg.Provider = ""
+		cfg.Model = ""
+	}
+	delete(cfg.CustomProviders, name)
```
lizhengfeng101: **[Critical] Same as above — manually implements deletion logic instead of calling `runConfigUnset`**

This manually executes `if cfg.Provider == name { cfg.Provider = ""; cfg.Model = "" }` + `delete` instead of calling the function under test. If the order of operations in `runConfigUnset` is accidentally changed (e.g., deleting before checking `wasActive`), this test would not catch the bug.

cmd/opencodereview/config_cmd_test.go
```diff
@@ -218,3 +221,117 @@ func TestSetConfigValueModelWithCustomProvider(t *testing.T) {
...
+	tests := []struct {
+		key     string
+		wantErr bool
+	}{
+		{"custom_providers.my-gateway", false},
+		{"llm.url", true},
+		{"providers.anthropic", true},
+		{"custom_providers.", true},
+		{"custom_providers", true},
+		{"something", true},
```
lizhengfeng101: **[Critical] Same as above — re-implements validation logic**

This manually replicates the `SplitN` + condition check validation logic instead of calling `runConfigUnset(tt.key)` and checking the returned error. Should be rewritten to call the actual function:

```go
for _, tt := range tests {
    err := unsetCustomProvider(configPath, tt.key)
    if (err != nil) != tt.wantErr {
        t.Errorf("key %q: err=%v, wantErr=%v", tt.key, err, tt.wantErr)
    }
}
```

cmd/opencodereview/provider_cmd.go
```diff
@@ -48,6 +62,28 @@ func runConfigProvider() error {
...
+	for _, name := range names {
+		if cfg.CustomProviders != nil {
+			delete(cfg.CustomProviders, name)
+		}
+		if cfg.Provider == name {
+			cfg.Provider = ""
+			cfg.Model = ""
+			clearedActive = true
+		}
+		fmt.Printf("Deleted custom provider %q.\n", name)
```
lizhengfeng101, line 77: **[Medium] Prints deletion message without checking if the provider actually existed**

`delete(cfg.CustomProviders, name)` is a silent no-op for non-existent keys, but this line prints `Deleted custom provider %q` regardless of whether the provider was actually present. While the current TUI flow only populates `deletedProviders` with names that exist, as a standalone function it should validate existence. Suggestion:

```go
if _, exists := cfg.CustomProviders[name]; !exists {
    continue
}
```

cmd/opencodereview/config_cmd.go
```diff
@@ -80,6 +82,53 @@ func runConfigSet(key, value string) error {
 	return nil
 }
 
+func runConfigUnset(key string) error {
+	parts := strings.SplitN(key, ".", 2)
```
lizhengfeng101, line 86: **[Minor] Duplicated logic with `applyProviderDeletions`**

`runConfigUnset` and `applyProviderDeletions` in `provider_cmd.go` implement the same core logic: delete custom provider → check if active → clear `Provider`/`Model` → nil out empty map. Consider extracting a shared internal function (e.g., `deleteCustomProvider(cfg *Config, name string) bool`) called by both, to avoid the risk of the two implementations diverging in future changes.

lizhengfeng101 (commented):
整体功能设计合理，TUI 交互考虑周全。主要问题在于 CLI 单元测试未调用实际函数，以及 CLI 和 TUI 之间存在删除逻辑重复。详见各行评论。

cmd/opencodereview/config_cmd_test.go
```diff
@@ -218,3 +220,122 @@ func TestSetConfigValueModelWithCustomProvider(t *testing.T) {
...
+	}
+	if err := saveConfig(configPath, cfg); err != nil {
+		t.Fatalf("saveConfig: %v", err)
+	}
+
+	tests := []struct {
+		name    string
+		wantErr bool
+	}{
+		{"my-gateway", false},
```
lizhengfeng101, line 319: **Bug: test pollution across sub-cases.** The first case `{"my-gateway", false}` successfully deletes the provider *and modifies the config file on disk*, which changes the state for the second case. While the result happens to be correct today (because `"nonexistent"` would fail regardless), this test is fragile:

1. Reversing the order would break the test.
2. Adding a case that depends on `my-gateway` still existing would silently fail.
3. The test is named `TestUnsetInvalidKey` but the first case is a *valid* deletion — the intent is unclear.

Suggestion: use `t.Run` sub-tests with independent config files per case, or split into separate test functions.

```go
for _, tt := range tests {
    t.Run(tt.name, func(t *testing.T) {
        dir := t.TempDir()
        configPath := dir + "/config.json"
        cfg := &Config{CustomProviders: map[string]ProviderEntry{
            "my-gateway": {URL: "https://gw.example.com/v1"},
        }}
        saveConfig(configPath, cfg)
        err := unsetCustomProvider(configPath, tt.name)
        // ...
    })
}
```

cmd/opencodereview/provider_cmd.go
```diff
@@ -30,7 +30,22 @@ func runConfigProvider() error {
...
+			return err
+		}
+		if clearedActive && !final.confirmed {
+			fmt.Fprintf(os.Stderr, "[ocr] WARNING: active provider was deleted; 'provider' and 'model' have been cleared.\n")
+			fmt.Fprintf(os.Stderr, "[ocr] Run 'ocr config provider' to select a new provider.\n")
+		}
+	}
+
 	if !final.confirmed {
+		if len(final.deletedProviders) > 0 {
```
lizhengfeng101, line 46: **Design: deletions persist even when the user cancels provider selection.** This is explicitly documented in the PR description, so it's clearly intentional. However, from a UX perspective it may surprise users — pressing Esc typically means "discard all changes", but here deletions have already been committed to disk.

Worth confirming this is the desired team UX convention. If so, consider adding a brief note in the TUI (e.g., "Provider deleted. Press Esc to exit without selecting a new provider.") so the user is aware the deletion was already applied.

cmd/opencodereview/provider_cmd.go
```diff
@@ -48,6 +63,24 @@ func runConfigProvider() error {
 	return applyOfficialProviderConfig(configPath, cfg, result)
 }
 
+func applyProviderDeletions(configPath string, cfg *Config, names []string) (bool, error) {
+	clearedActive := false
+	for _, name := range names {
+		wasActive, err := deleteCustomProvider(cfg, name)
+		if err != nil {
+			continue
```
lizhengfeng101, line 72: **Nit: silently swallowing errors.** If `deleteCustomProvider` returns an error (e.g., the provider was already removed from memory by a duplicate entry in `deletedProviders`), this `continue` drops the error without any signal to the user.

While unlikely in practice, consider at least logging it:

```go
if err != nil {
    fmt.Fprintf(os.Stderr, "[ocr] skip delete %q: %v\n", name, err)
    continue
}
```

cmd/opencodereview/provider_tui.go
```diff
@@ -1034,6 +1083,17 @@ func (m providerTUIModel) viewCustomTab(s *strings.Builder) {
 		s.WriteString(cursor + tuiDimStyle.Render(addLabel))
 	}
 	s.WriteString("\n")
+
+	if m.confirmingDelete {
+		s.WriteString("\n")
+		prompt := fmt.Sprintf("  Delete %q?", m.deleteTargetName)
+		if m.existingCfg != nil && m.existingCfg.Provider == m.deleteTargetName {
```
lizhengfeng101, line 1092: **Minor: active-provider warning uses initial config snapshot.** This checks `m.existingCfg` (the config at TUI startup) to decide whether to show the "active provider" warning. If the user deletes provider A (which was active), then tries to delete provider B, the warning correctly won't fire for B.

However, this won't reflect any provider *selection* made during the same TUI session. Not a bug in the current flow (selection happens after deletion), but worth a comment noting the assumption so future refactors don't break it.

cmd/opencodereview/config_cmd.go
```diff
@@ -80,6 +82,68 @@ func runConfigSet(key, value string) error {
...
+	if len(parts) != 2 || parts[0] != "custom_providers" || parts[1] == "" {
+		return fmt.Errorf("unset only supports custom_providers.<name>\nUsage: ocr config unset custom_providers.<name>")
```
lizhengfeng101: **Nit: error message style inconsistency.** This error includes a multi-line usage hint with `\n`, while `runConfigSet` and other error paths in the codebase return short single-line errors. Consider keeping it consistent:

```go
return fmt.Errorf("unset only supports custom_providers.<name>")
```

The full usage is already shown by `printConfigUsage()` when the user runs `ocr config` without args.

lizhengfeng101 (commented):
Overall this is a high-quality PR — clean separation of concerns, solid test coverage, and good UX details. Left a few inline comments on specific areas worth addressing.

lizhengfeng101 (approved):
## Re-review: all previous feedback addressed ✓

Verified against latest commit `b0a6960e` ("fix: address second round of PR review feedback"). All 5 issues from the previous round have been resolved:

| # | Issue | Status |
|---|-------|--------|
| 1 | `TestUnsetInvalidKey` test pollution — sub-cases shared config file on disk | **Fixed** — now uses `t.Run` with independent `t.TempDir()` per sub-case |
| 2 | Deletions persist on cancel (design question) | **Acknowledged** — intentional design; no code change needed |
| 3 | `applyProviderDeletions` silently swallowed errors | **Fixed** — now logs `[ocr] skip delete %q: %v` to stderr |
| 4 | Active-provider warning used `existingCfg` snapshot without explanation | **Fixed** — added clarifying comment noting this is a startup snapshot |
| 5 | Error message style inconsistency (multi-line vs single-line) | **Fixed** — simplified to single-line `"unset only supports custom_providers.<name>"` |

### Code quality summary

- **Architecture**: Clean separation — `deleteCustomProvider` (in-memory logic) is shared by both CLI (`unsetCustomProvider`) and TUI (`applyProviderDeletions`) paths. No duplication.
- **Tests**: 10 new tests with proper isolation. CLI and TUI paths, happy paths and edge cases all covered.
- **TUI UX**: Delete confirmation, active-provider warning, cursor adjustment after deletion, `d` on "Add" option ignored — all edge cases handled.
- **Documentation**: README command table and examples updated.

LGTM — ready to merge.

lizhengfeng101:
Hey @yingjiexu2002, great work on this PR!

One request: @Ackberry had previously claimed this issue (#136) on Jun 15 and was working on it as part of a coursework project. It seems the efforts overlapped since there was no draft PR to signal work-in-progress.

Would you mind adding a co-author credit when merging? You can append this to the final commit message:

```
Co-authored-by: Deep Akbari <180123260+Ackberry@users.noreply.github.com>
```

This way both contributors get recognized. Thanks!

lizhengfeng101 (approved):
LGTM

PR #193, opened 2026-06-23
fix: ensure current file path is always injected for code_comment and…

AlapinEnjoyer:
# Improve line number tracking in resolver

## Description

Two fixes for the `:0-0` line numbers issue in review comments:

1. **Wrong file attribution (`internal/agent/agent.go`):** The `code_comment` tool definition doesn't declare a `path` parameter, but some LLMs spontaneously add one. The agent's fallback (`if _, ok := args["path"]; !ok`) only filled the path if it was entirely absent, meaning the LLM's hallucinated value won. Changed this to always override with the correct current-file path.
2. **Blank lines break `resolveFromFileContent` (`internal/diff/resolver.go`):** `splitAndNormalize` strips blank lines from the target `existing_code`, but the old comparison loop iterated over raw file lines (with blanks intact), causing a misalignment whenever the snippet contained blank lines. The fix normalizes file lines the exact same way (skipping blanks and tracking actual line numbers via `fileLineNums`) so both sides stay aligned.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Validated against a real review run (Range mode, 3 files). 
* **Before:** 5/8 comments had `:0-0`. 
* **After:** Following both fixes, all 8 comments resolve to the correct lines when the path and snippet are accurate. See the discussion in the related issue for before/after logs.

## Checklist

- [x] I have signed the CLA

### DIFF
--- a/internal/diff/resolver.go
+++ b/internal/diff/resolver.go
@@ -165,7 +165,7 @@ func matchConsecutive(sideLines []indexedLine, targetLines []string) (startLine,
 }
 
 // resolveFromFileContent scans the new file content line-by-line for consecutive
-// matches of the normalized existing_code. Ported from Java's findConsecutiveLines.
+// matches of the normalized existing_code.
 func resolveFromFileContent(d *model.Diff, cm *model.LlmComment) bool {
 	if d.NewFileContent == "" {
 		return false
@@ -173,21 +173,37 @@ func resolveFromFileContent(d *model.Diff, cm *model.LlmComment) bool {
 
 	fileLines := strings.Split(d.NewFileContent, "\n")
 	targetLines := splitAndNormalize(cm.ExistingCode)
-	if len(targetLines) == 0 || len(fileLines) < len(targetLines) {
+	if len(targetLines) == 0 {
 		return false
 	}
 
-	for i := 0; i <= len(fileLines)-len(targetLines); i++ {
+	// Normalize file lines the same way as target (skip blanks) and track actual line numbers.
+	normalizedFileLines := make([]string, 0, len(fileLines))
+	fileLineNums := make([]int, 0, len(fileLines))
+	for i, line := range fileLines {
+		n := normalizeLine(strings.TrimRight(line, "\r"))
+		if n == "" {
+			continue
+		}
+		normalizedFileLines = append(normalizedFileLines, n)
+		fileLineNums = append(fileLineNums, i+1)
+	}
+
+	if len(normalizedFileLines) < len(targetLines) {
+		return false
+	}
+
+	for i := 0; i <= len(normalizedFileLines)-len(targetLines); i++ {
 		matched := true
 		for j, target := range targetLines {
-			if normalizeLine(strings.TrimRight(fileLines[i+j], "\r")) != target {
+			if normalizedFileLines[i+j] != target {
 				matched = false
 				break
 			}
 		}
 		if matched {
-			cm.StartLine = i + 1
-			cm.EndLine = i + len(targetLines)
+			cm.StartLine = fileLineNums[i]
+			cm.EndLine = fileLineNums[i+len(targetLines)-1]
 			return true
 		}
 	}
--- a/internal/diff/resolver.go
+++ b/internal/diff/resolver.go
@@ -177,7 +177,9 @@ func resolveFromFileContent(d *model.Diff, cm *model.LlmComment) bool {
 		return false
 	}
 
-	// Normalize file lines the same way as target (skip blanks) and track actual line numbers.
+	// Normalize file lines the same way as target: skip blanks so that
+	// blank lines in the source don't break the sliding-window match.
+	// "Consecutive" here means adjacent non-blank lines.
 	normalizedFileLines := make([]string, 0, len(fileLines))
 	fileLineNums := make([]int, 0, len(fileLines))
 	for i, line := range fileLines {
--- a/internal/diff/resolver_test.go
+++ b/internal/diff/resolver_test.go
@@ -117,6 +117,25 @@ import "fmt"`},
 	}
 }
 
+func TestResolveLineNumbers_FallbackToFileContent_BlankLines(t *testing.T) {
+	// Hunk match fails; fallback must match across blank lines in NewFileContent.
+	diffs := []model.Diff{{
+		NewPath:        "main.go",
+		NewFileContent: "package main\n\nfunc foo() {\n\n\treturn 1\n}",
+		Diff:           "@@ -1,2 +1,2 @@\n-old\n+new",
+	}}
+	comments := []model.LlmComment{{
+		Path:         "main.go",
+		ExistingCode: "func foo() {\n\treturn 1\n}",
+	}}
+
+	result := ResolveLineNumbers(comments, diffs)
+	cm := result[0]
+	if cm.StartLine != 3 || cm.EndLine != 6 {
+		t.Errorf("fallback with blank lines: expected 3..6, got %d..%d", cm.StartLine, cm.EndLine)
+	}
+}
+
 func TestResolveLineNumbers_NoMatchKeepsZero(t *testing.T) {
 	diffs := []model.Diff{
 		{NewPath: "test.go", Diff: testDiff},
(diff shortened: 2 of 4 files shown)


internal/diff/resolver.go
```diff
@@ -165,29 +165,45 @@ func matchConsecutive(sideLines []indexedLine, targetLines []string) (startLine,
...
+	// Normalize file lines the same way as target (skip blanks) and track actual line numbers.
+	normalizedFileLines := make([]string, 0, len(fileLines))
+	fileLineNums := make([]int, 0, len(fileLines))
+	for i, line := range fileLines {
+		n := normalizeLine(strings.TrimRight(line, "\r"))
+		if n == "" {
+			continue
+		}
+		normalizedFileLines = append(normalizedFileLines, n)
+		fileLineNums = append(fileLineNums, i+1)
```
lizhengfeng101, line 191: **Semantic change worth noting:** After this normalization, "consecutive match" means "adjacent *non-blank* lines" rather than "physically adjacent lines." For example, a target `["return nil", "}"]` will now match across intervening blank lines:

```
line 1: return nil
line 2: (blank)
line 3: }
```
→ `StartLine=1, EndLine=3`

This is reasonable for the use case (LLM-generated snippets typically omit blank lines), but it could theoretically produce a false positive when unrelated non-blank lines happen to align across large blank-separated regions. The risk is low since this is already the fallback path (hunk matching failed), and it returns the first match.

Consider adding a brief comment here noting the intentional design choice, e.g.:
```go
// Normalize file lines the same way as target: skip blanks so that
// blank lines in the source don't break the sliding-window match.
// "Consecutive" here means adjacent non-blank lines.
```

internal/diff/resolver.go
```diff
@@ -165,29 +165,45 @@ func matchConsecutive(sideLines []indexedLine, targetLines []string) (startLine,
...
-			if normalizeLine(strings.TrimRight(fileLines[i+j], "\r")) != target {
+			if normalizedFileLines[i+j] != target {
 				matched = false
 				break
 			}
 		}
 		if matched {
-			cm.StartLine = i + 1
-			cm.EndLine = i + len(targetLines)
+			cm.StartLine = fileLineNums[i]
```
lizhengfeng101, line 207: **Missing test coverage:** The PR checklist states "I have added tests that prove my fix is effective," but the diff contains no test file changes. The existing `TestResolveLineNumbers_FallbackToFileContent` doesn't exercise blank lines in `NewFileContent`.

Suggested test case to add:

```go
{
    name: "file content with blank lines between matching code",
    diffs: []*model.Diff{{
        NewPath:        "main.go",
        NewFileContent: "package main\n\nfunc foo() {\n\n\treturn 1\n}",
        Diff:           "@@ -1,2 +1,2 @@\n-old\n+new",
    }},
    comments: []*model.LlmComment{{
        ExistingCode: "func foo() {\n\treturn 1\n}",
        Path:         "main.go",
    }},
    expectedStart: 3,
    expectedEnd:   6,
}
```

This would guard against regressions on the exact scenario this PR fixes.

lizhengfeng101 (commented):
Thanks for the fix — both changes are well-motivated and the logic is correct. Two items to address before merging:

1. **Missing test coverage for the blank-line scenario** — the PR checklist says tests were added, but the diff has no test file changes. The existing integration tests (`TestResolveLineNumbers_FallbackToFileContent`, `TestResolveLineNumbers_MixedStrategies`) don't cover blank lines in `NewFileContent`.
2. **Subtle semantic change in "consecutive" matching** — see inline comment.

AlapinEnjoyer:
Hi, sorry for the missing test. I added it now together with the comments.

AlapinEnjoyer:
Sorry, had some internal stuff I needed to resolve and accidentally deleted the branch.

internal/diff/resolver_test.go
```diff
@@ -117,6 +117,25 @@ import "fmt"`},
...
+		ExistingCode: "func foo() {\n\treturn 1\n}",
+	}}
+
+	result := ResolveLineNumbers(comments, diffs)
+	cm := result[0]
+	if cm.StartLine != 3 || cm.EndLine != 6 {
+		t.Errorf("fallback with blank lines: expected 3..6, got %d..%d", cm.StartLine, cm.EndLine)
+	}
+}
+
```
lizhengfeng101, line 138: Consider adding a test case with multiple consecutive blank lines to verify `EndLine` tracking stays correct:

```go
func TestResolveLineNumbers_FallbackToFileContent_MultipleBlankLines(t *testing.T) {
	diffs := []model.Diff{{
		NewPath:        "main.go",
		NewFileContent: "a\n\n\nb\n\nc\n",
		Diff:           "@@ -1,2 +1,2 @@\n-old\n+new",
	}}
	comments := []model.LlmComment{{
		Path:         "main.go",
		ExistingCode: "a\nb\nc",
	}}

	result := ResolveLineNumbers(comments, diffs)
	cm := result[0]
	if cm.StartLine != 1 || cm.EndLine != 5 {
		t.Errorf("multiple blank lines: expected 1..5, got %d..%d", cm.StartLine, cm.EndLine)
	}
}
```

This covers the scenario where multiple consecutive blank lines separate matched code, ensuring `fileLineNums` correctly maps indices back to actual file positions.

AlapinEnjoyer: I added this test together with a leading blanks and windows style checking.

Since there is a blank line between "b" and "c" I corrected it to be 1..6 instead of 1..5

internal/agent/agent.go
```diff
@@ -980,12 +980,10 @@ func (a *Agent) executeToolCall(ctx context.Context, newPath string, call llm.To
 		return tool.Of(fmt.Sprintf("Error parsing tool arguments for %s: %v", t.Name(), err))
 	}
 
-	// Inject current file path as default for code_comment when not provided.
-	// The model already knows which file it's reviewing, so it omits path.
+	// Always inject the current file path for code_comment.
+	// The model sometimes hallucinates a path, so we override it.
 	if t == tool.CodeComment && newPath != "" {
```
lizhengfeng101: **Merge conflict:** On current `main`, `executeToolCall` has been moved to `internal/llmloop/loop.go` (receiver changed from `*Agent` to `*Runner`, signature now includes a `*session.TaskRecord` parameter). The equivalent code lives at `loop.go:283-288`.

Please rebase onto latest `main` and apply this fix there instead. The logic itself (always override `args["path"]`) is correct — `code_comment`'s tool definition does not declare a `path` parameter, so any LLM-provided value is by definition hallucinated.

Also consider adding a unit test for this behavior — e.g., verify that when `args` already contains `"path": "wrong.go"`, it gets overwritten with `newPath`.

internal/diff/resolver.go
```diff
@@ -165,29 +165,47 @@ func matchConsecutive(sideLines []indexedLine, targetLines []string) (startLine,
...
```
lizhengfeng101, line 207: Good fix. The `EndLine` semantics have changed subtly but correctly: previously it was `startIndex + len(targetLines)` (consecutive numbering), now it maps to the actual file line number of the last matched non-blank line. This means for a snippet spanning blank lines, `EndLine - StartLine + 1` can be larger than `len(targetLines)`, which is the right behavior for annotating the real line range in the file.

lizhengfeng101 (changes requested):
## Review Summary

Both fixes are logically correct and address real bugs. The resolver blank-line fix is well-tested. However, the PR has **merge conflicts** that must be resolved before merging — the `executeToolCall` method has been moved from `agent.go` to `llmloop/loop.go` on `main`.

### What's good
- Root cause analysis is clear and well-documented in the PR description
- `resolveFromFileContent` fix is sound — normalizing both sides identically and tracking original line numbers via `fileLineNums` is the right approach
- Test coverage for the resolver fix is thorough (6 new cases covering blank lines, CRLF, leading blanks, duplicates, all-blank target)
- Removing the stale "Ported from Java" comment is a nice cleanup

### What needs attention
1. **Merge conflict (blocking):** `executeToolCall` was refactored from `*Agent` in `agent.go` to `*Runner` in `internal/llmloop/loop.go`. The path-override fix needs to be rebased and applied there instead.
2. **Missing test for path override:** The always-override behavior change has no unit test. A test verifying that a hallucinated `path` value gets replaced by `newPath` would prevent regressions.

lizhengfeng101 (approved):
LGTM

PR #213, opened 2026-06-24
feat: improve config provider TUI interaction

5wxwxwxw5:
## Description

- Support wrap-around navigation in provider, custom provider, and model lists (up at first item jumps to last, down at last item jumps to first)
- Add d key to delete custom models in model selection step with confirmation prompt
- Add protocol selection (anthropic/openai) to manual configuration flow and respect UseAnthropic setting in saved config
- Preserve manual protocol selection when re-entering the form
- Fix manual form input fields losing focus after returning from next step
- Use safe slice removal in removeFromSlice to avoid backing array corruption

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

image

image

image

image

image

## Checklist

- [ ] I have signed the CLA

### DIFF
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"sort"
 	"strconv"
 	"strings"
 
@@ -381,22 +380,6 @@ func activeModelForProvider(cfg *Config, providerName string, entry ProviderEntr
 	return ""
 }
 
-func sortModelsByName(models []string) []string {
-	if len(models) < 2 {
-		return models
-	}
-	sorted := append([]string(nil), models...)
-	sort.SliceStable(sorted, func(i, j int) bool {
-		left := strings.ToLower(sorted[i])
-		right := strings.ToLower(sorted[j])
-		if left == right {
-			return sorted[i] < sorted[j]
-		}
-		return left < right
-	})
-	return sorted
-}
-
 func normalizeModelList(models []string) []string {
 	out := make([]string, 0, len(models))
 	seen := make(map[string]struct{}, len(models))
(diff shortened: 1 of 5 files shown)


lizhengfeng101:
## Code Review Findings

Hi! I spotted a few things while reviewing — please take a look and let me know if you agree they're worth addressing.

### 1. Indentation / brace alignment in `updateDeleteModelConfirm` (potential correctness risk)

The indentation in this function is inconsistent, making it hard to verify the brace matching is correct:

```go
			m.savedInSession = true
		if len(m.models()) > 0 {          // ← indent level drops unexpectedly
			if remaining := len(m.models()); m.modelIdx >= remaining {
				...
			}
		}
	}
		m.confirmingDeleteModel = false    // ← indent inconsistent with surrounding code
		return m, nil
```

Go uses braces (not indentation) for scoping, so this may compile fine — but the formatting makes it very difficult to audit correctness. Could you run `gofmt` on this function and verify the logic matches your intent?

### 2. Misplaced doc comment above `cloneProviderEntry`

The comment block describing `applyEditCustomProviderSave` sits directly above `cloneProviderEntry`:

```go
// applyEditCustomProviderSave persists the edited provider to disk. Returns
// true if a hard validation error stopped the save ...
func cloneProviderEntry(v ProviderEntry) ProviderEntry {
```

Looks like the comment should be moved to the actual `applyEditCustomProviderSave` method.

### 3. `removeFromSlice` appears to be dead code

The PR description mentions "Use safe slice removal in removeFromSlice to avoid backing array corruption," but `removeFromSlice` is defined and never called anywhere — model deletion uses `removeModels()` instead. Should this function be removed, or is there a missing call site?

### 4. Redundant provider deletion (double save)

In `updateDeleteConfirm`, a deleted provider is both saved to disk immediately (via `saveConfig`) *and* appended to `deletedProviders`. Then in the `confirmed` path of `runConfigProvider`, `applyProviderDeletions` attempts to delete and save again. Since `delete(map, key)` is idempotent this isn't a bug, but the double-write adds unnecessary complexity. Consider either skipping the `deletedProviders` append when saving in-session, or guarding the post-TUI path.

### 5. `m.models()` called multiple times in `updateDeleteModelConfirm`

`m.models()` (which includes sorting) is called 3 times in sequence:

```go
if len(m.models()) > 0 {
    if remaining := len(m.models()); m.modelIdx >= remaining {
```

Caching the result in a local variable would be cleaner and avoids redundant work:

```go
updated := m.models()
if len(updated) > 0 && m.modelIdx >= len(updated) {
    m.modelIdx = len(updated) - 1
}
```

lizhengfeng101 (approved):
LGTM

PR #218, opened 2026-06-25
fix: increase REVIEW_FILTER_TASK and RE_LOCATION_TASK timeout from 60s to 180s

### LINKED ISSUE #208: The root cause is that certain internal sub-tasks (such as REVIEW_FILTER_TASK and RE_LOCATION_TASK) have hardcoded 60-second timeouts that override the user-provided --timeout flag. For slow local models, 60 seconds is often not enough.
### OpenCodeReview Version

open-code-review v1.5.0 (7715639) linux/amd64

### Operating System

Linux (x86_64)

### Installation Method

GitHub Release binary

### LLM Provider

### Bug Description

The root cause is that certain internal sub-tasks (such as REVIEW_FILTER_TASK and RE_LOCATION_TASK) have hardcoded 60-second timeouts that override the user-provided --timeout flag. For slow local models, 60 seconds is often not enough.

### Steps to Reproduce

ocr review

### Expected Behavior

--timeout 生效

### Logs / Error Output

```shell

```

### Additional Context

_No response_

lizhengfeng101:
## Summary

- Increase `REVIEW_FILTER_TASK` and `RE_LOCATION_TASK` timeout from 60s to 180s in both `task_template.json` and `scan_template.json`
- Fixes the issue where slow local models hit the hardcoded 60-second timeout that the user-provided `--timeout` flag cannot override

Closes #208


MuoDoo (approved):
LGTM. However, this seems like not solving #208

PR #223, opened 2026-06-25
feat: enhance token usage resolution for OpenAI and Anthropic compati…

AlapinEnjoyer:
## Description

Cached token usage was not getting tracked for the OpenAI provider, the MR adds a way to recognize the common OpenAI-compatible cached-token fields for custom providers

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Tested with a custom provider run

## Checklist

- [x] I have signed the CLA

### DIFF
--- a/internal/llm/usage_resolver.go
+++ b/internal/llm/usage_resolver.go
@@ -32,9 +32,7 @@ var cacheReadTokensPaths = []string{
 	"data.usage.cache_read_input_tokens",             // wrapped Anthropic-compatible proxy
 	"usage.prompt_tokens_details.cached_tokens",      // OpenAI-compatible providers
 	"data.usage.prompt_tokens_details.cached_tokens", // wrapped OpenAI-compatible providers
-	"usage.prompt_tokens_details.cache_tokens_hit",   // some providers
-	"data.usage.prompt_tokens_details.cache_tokens_hit",
-	"usage.prompt_tokens_details.cache_tokens", // some providers
+	"usage.prompt_tokens_details.cache_tokens",        // Baidu Qianfan (ERNIE)
 	"data.usage.prompt_tokens_details.cache_tokens",
 }
 
@@ -42,8 +40,8 @@ var cacheWriteTokensPaths = []string{
 	"usage.cache_creation_input_tokens",                      // Anthropic / proxy
 	"cache_creation_input_tokens",                            // flat at root
 	"data.usage.cache_creation_input_tokens",                 // wrapped Anthropic-compatible proxy
-	"usage.prompt_tokens_details.cache_creation_tokens",      // OpenAI-compatible providers
-	"data.usage.prompt_tokens_details.cache_creation_tokens", // wrapped OpenAI-compatible providers
+	"usage.prompt_tokens_details.cache_creation_tokens",      // ApexRoute / LLM Gateway — proxy normalization of Anthropic cache_creation_input_tokens
+	"data.usage.prompt_tokens_details.cache_creation_tokens", // wrapped proxy normalization
 }
 
 // anthropicCacheReadPathCount is the number of Anthropic-style cache read paths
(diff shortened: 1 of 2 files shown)


lizhengfeng101 (commented):
Hey @AlapinEnjoyer, thanks for this PR! 🎉 Great catch on the missing `cached_tokens` field — that's the standard OpenAI field name and it's definitely needed. The tests are solid too. A few thoughts below:

---

### 1. Potential double-counting of TotalTokens (pre-existing, not introduced by this PR)

When `total_tokens` is absent from the response, the fallback calculation at [usage_resolver.go:77-79](https://github.com/nicobao/open-code-review/blob/main/internal/llm/usage_resolver.go#L77-L79) does:

```go
ui.TotalTokens = prompt + completion + cacheRead + cacheWrite
```

In OpenAI's semantics, `prompt_tokens` **already includes** `cached_tokens` (cached tokens are a subset of prompt tokens), so the correct total would be `prompt + completion` — adding cache tokens on top double-counts them. In contrast, Anthropic's `input_tokens` **excludes** cache tokens, so the addition is correct there.

The test `TestResolveUsageWrappedCachedTokens` asserts `TotalTokens == 205` (100+20+75+10), which would be a double-count under OpenAI semantics.

This is a pre-existing design choice, not something you introduced — but your PR does make more OpenAI-format responses reach this code path. Might be worth a follow-up to handle the two semantics differently, or at least a comment noting the trade-off. No action needed in this PR if you'd prefer to keep the scope focused.

### 2. Where does `cache_creation_tokens` come from?

The newly added paths:
```
usage.prompt_tokens_details.cache_creation_tokens
data.usage.prompt_tokens_details.cache_creation_tokens
```

This field name isn't part of OpenAI's standard response format (OpenAI doesn't expose cache creation), and it differs from Anthropic's `cache_creation_input_tokens` (which lives at the `usage` top level, not under `prompt_tokens_details`). Could you add a short comment noting which provider or proxy uses this field? It would help future readers understand where it comes from.

### 3. Minor: test idea for path priority

Since `probePath` uses first-match-wins, it might be worth adding a small test where both `cached_tokens` and `cache_read_input_tokens` (or `cache_tokens_hit`) are present in the same response, to verify the expected path takes priority. This would document the intended ordering.

---

Overall this is a solid improvement — the gap in OpenAI cached token tracking was real and this fixes it cleanly. Nice work on the test coverage too! 👍

lizhengfeng101 (commented):
Thanks for the update — the `probePathIndex` approach and the new tests look solid. One concern about scope:

### `cache_tokens` / `cache_tokens_hit` are not part of either standard

The two established protocols for cache token reporting are:

| Protocol | Cache read field | Location |
|----------|-----------------|----------|
| **Anthropic** | `cache_read_input_tokens` | top-level under `usage` |
| **OpenAI** | `cached_tokens` | under `usage.prompt_tokens_details` |

`cache_tokens` (attributed to Baidu Qianfan) and the previously existing `cache_tokens_hit` are neither OpenAI nor Anthropic standard — they're provider-specific non-standard field names. If we start adding paths for every provider's custom naming, this list will grow unboundedly.

I'd suggest:
1. **Remove** the `cache_tokens` and `data.usage.prompt_tokens_details.cache_tokens` paths (and the already-removed `cache_tokens_hit` should stay removed).
2. **Keep only** the two standard protocols: Anthropic-style (`cache_read_input_tokens`) and OpenAI-style (`cached_tokens`), plus their `data.usage.*` wrapped variants.
3. Providers using non-standard field names should handle the mapping on their proxy/adapter side, not in OCR.

This also simplifies the `anthropicCacheReadPathCount` / `anthropicCacheWritePathCount` bookkeeping since there would be fewer paths to track.

lizhengfeng101 (approved):
LGTM

PR #238, opened 2026-06-28
fix: add configurable HTTP timeout for LLM API calls

### LINKED ISSUE #237: bug: LLM HTTP client has no configurable request timeout, causing scan to hang indefinitely
## Bug Description

The `ocr scan` command can hang indefinitely when the LLM API does not respond. The HTTP client used for LLM API calls does not have a properly configurable request timeout, causing the process to block forever on stuck connections.

## Root Cause

In `internal/llm/client.go`, the `NewLLMClient` factory function creates a `ClientConfig` without setting the `Timeout` field:

```go
func NewLLMClient(ep ResolvedEndpoint) LLMClient {
    cfg := ClientConfig{
        URL:        ep.URL,
        APIKey:     ep.Token,
        Model:      ep.Model,
        AuthHeader: ep.AuthHeader,
        ExtraBody:  ep.ExtraBody,
        // Timeout is NOT set — defaults to 0
    }
```

While `NewOpenAIClient` and `NewAnthropicClient` both have a fallback default of 5 minutes:
```go
if cfg.Timeout <= 0 {
    cfg.Timeout = 5 * time.Minute
}
```

There are two problems:
1. **`ResolvedEndpoint` has no `Timeout` field** — users cannot configure the HTTP timeout via config file or environment variables
2. **Streaming responses may bypass the timeout** — the OpenAI/Anthropic SDKs `WithRequestTimeout` applies to the overall request lifecycle, but streaming responses that send data slowly (or stop sending data mid-stream) may not trigger the timeout

## Reproduction

1. Configure ocr with an LLM API that occasionally hangs (e.g., rate-limited or overloaded)
2. Run `ocr scan` on a repository
3. The process hangs indefinitely with 0% CPU usage, holding an ESTABLISHED TCP connection to the API server
4. The process never exits until manually killed

## Expected Behavior

The `ocr` binary should:
1. Support a configurable HTTP request timeout (e.g., via `OCR_LLM_TIMEOUT` env var or config file)
2. Have a reasonable default timeout (e.g., 120 seconds per LLM call)
3. Properly enforce the timeout on streaming responses

## Environment

- ocr version: v1.6.4 (darwin/arm64)
- OS: macOS
- LLM Provider: Xiaomi MiMo API (OpenAI-compatible)

## Suggested Fix

1. Add a `Timeout` field to `ResolvedEndpoint`
2. Add `OCR_LLM_TIMEOUT` environment variable support
3. Pass the timeout from `ResolvedEndpoint` to `ClientConfig` in `NewLLMClient`
4. Add per-request timeout enforcement that works with streaming responses

zy84338719:
## Problem

The LLM HTTP client has no user-configurable request timeout. When the LLM API becomes unresponsive (rate limiting, network issues, server overload), the `ocr scan` process hangs indefinitely with 0% CPU, holding an ESTABLISHED TCP connection that never completes.

**Reproduction:** Run `ocr scan` against an LLM API that occasionally hangs. The process blocks forever until manually killed.

Root cause in `internal/llm/client.go:195`: `NewLLMClient` creates `ClientConfig` without setting `Timeout` (defaults to 0). While `NewOpenAIClient`/`NewAnthropicClient` have a 5-minute fallback, there is no way for users to configure this.

Fixes #237

## Changes

- Add `Timeout` field to `ResolvedEndpoint` struct
- Add `OCR_LLM_TIMEOUT` environment variable (value in seconds)
- Add `timeout_sec` field to `config.json` llm and provider sections
- Pass timeout from `ResolvedEndpoint` to `ClientConfig` in `NewLLMClient`

## Usage

```bash
# Environment variable
export OCR_LLM_TIMEOUT=120  # 2 minutes per LLM call

# Or in ~/.opencodereview/config.json
{
  "llm": {
    "url": "https://api.example.com/v1",
    "timeout_sec": 120
  }
}
```

The default remains 5 minutes when not explicitly configured. All existing behavior is preserved.

### DIFF
--- a/internal/llm/resolver.go
+++ b/internal/llm/resolver.go
@@ -79,7 +79,11 @@ func ResolveEndpointWithModelOverride(configPath, modelOverride string) (Resolve
 			// OCR_LLM_TIMEOUT is a global override: applies regardless of
 			// which strategy resolved the endpoint, and takes precedence
 			// over config-file values when set.
-			if envTimeout, ok := parseTimeoutEnv(); ok {
+			envTimeout, ok, err := parseTimeoutEnv()
+			if err != nil {
+				return ResolvedEndpoint{}, fmt.Errorf("resolve %s: %w", s.name, err)
+			}
+			if ok {
 				ep.Timeout = envTimeout
 			}
 			return ep, nil
@@ -91,25 +95,22 @@ func ResolveEndpointWithModelOverride(configPath, modelOverride string) (Resolve
 
 // parseTimeoutEnv reads and validates the OCR_LLM_TIMEOUT environment variable.
 // Returns the parsed duration and true if set, or 0 and false if unset/empty.
-// Rejects negative values and values that would overflow time.Duration.
-func parseTimeoutEnv() (time.Duration, bool) {
+// Returns an error for invalid values (non-integer, negative, overflow) to give
+// the user clear feedback instead of silently falling back to the default.
+func parseTimeoutEnv() (time.Duration, bool, error) {
 	raw := strings.TrimSpace(os.Getenv(envOCRLLMTimeout))
 	if raw == "" {
-		return 0, false
+		return 0, false, nil
 	}
 	sec, err := strconv.Atoi(raw)
 	if err != nil {
-		return 0, false
+		return 0, false, fmt.Errorf("OCR_LLM_TIMEOUT must be an integer (seconds): %w", err)
 	}
-	if sec < 0 {
-		return 0, false
+	d, err := validateTimeoutSec(sec)
+	if err != nil {
+		return 0, false, fmt.Errorf("OCR_LLM_TIMEOUT: %w", err)
 	}
-	// Guard against overflow: time.Duration is int64 nanoseconds.
-	maxSec := int64(math.MaxInt64 / int64(time.Second))
-	if int64(sec) > maxSec {
-		return 0, false
-	}
-	return time.Duration(sec) * time.Second, true
+	return d, true, nil
 }
 
 // validateTimeoutSec converts a config-file timeout (in seconds) to time.Duration.
@@ -172,9 +173,6 @@ func tryOCREnv(modelOverride string) (ResolvedEndpoint, bool, error) {
 			return ResolvedEndpoint{}, false, fmt.Errorf("OCR environment: %w", err)
 		}
 	}
-
-	// Note: OCR_LLM_TIMEOUT is applied globally in ResolveEndpointWithModelOverride,
-	// not here, so it works regardless of which strategy resolves the endpoint.
 
 	return ResolvedEndpoint{URL: url, Token: token, Model: model, Protocol: protocol, AuthHeader: authHeader, Source: "OCR environment", ExtraHeaders: extraHeaders}, true, nil
 }
--- a/internal/llm/resolver.go
+++ b/internal/llm/resolver.go
@@ -3,6 +3,7 @@ package llm
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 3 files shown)


internal/llm/resolver.go
```diff
@@ -71,13 +76,59 @@ func ResolveEndpointWithModelOverride(configPath, modelOverride string) (Resolve
...
 		}
 	}
 
 	return ResolvedEndpoint{}, fmt.Errorf("no valid LLM endpoint configured; one of OCR_LLM_URL/OCR_LLM_TOKEN/OCR_LLM_MODEL, ~/.opencodereview/config.json, or ANTHROPIC_BASE_URL/ANTHROPIC_AUTH_TOKEN/ANTHROPIC_MODEL must be set")
 }
 
+// parseTimeoutEnv reads and validates the OCR_LLM_TIMEOUT environment variable.
+// Returns the parsed duration and true if set, or 0 and false if unset/empty.
+// Rejects negative values and values that would overflow time.Duration.
+func parseTimeoutEnv() (time.Duration, bool) {
```
lizhengfeng101: **Suggestion: avoid silently swallowing invalid values.**

`parseTimeoutEnv` returns `(0, false)` for invalid input (`"abc"`, `"-5"`), falling back to the default with no signal to the user. Meanwhile, `validateTimeoutSec` returns an explicit error for the same invalid inputs in config files.

This inconsistency can be confusing: a user sets `OCR_LLM_TIMEOUT=-5`, expects it to take effect, and has no way to know it was silently ignored.

Consider one of:
- Return an `error` and surface it in `ResolveEndpointWithModelOverride` (consistent with config-file behavior)
- Log a warning so the user gets feedback
- At minimum, document the silent-fallback behavior

lizhengfeng101 (commented):
## Review Summary

Clean, well-structured PR. The global override design in `ResolveEndpointWithModelOverride` is the right call — it ensures `OCR_LLM_TIMEOUT` works regardless of which resolution strategy fires. Overflow guards and boundary validation are solid.

A few items worth addressing before merge — see inline comments.

### Missing test coverage

1. **No test for provider config `timeout_sec`**: `tryProviderConfig` gained timeout parsing at line 344–348, but there's no corresponding test. The legacy `llm` path is covered by `TestResolveEndpoint_ConfigTimeoutSec`; the provider path should have parity.

2. **No test for env-over-config precedence**: `TestResolveEndpoint_EnvTimeoutGlobalOverride` tests env timeout with OCR env vars (which carry no config timeout). There's no test proving that `OCR_LLM_TIMEOUT` overrides a `timeout_sec` already set in the config file — that's the key precedence guarantee.

internal/llm/resolver.go
```diff
@@ -71,13 +76,59 @@ func ResolveEndpointWithModelOverride(configPath, modelOverride string) (Resolve
...
+	}
+	if sec < 0 {
+		return 0, false
+	}
+	// Guard against overflow: time.Duration is int64 nanoseconds.
+	maxSec := int64(math.MaxInt64 / int64(time.Second))
+	if int64(sec) > maxSec {
+		return 0, false
+	}
+	return time.Duration(sec) * time.Second, true
```
lizhengfeng101: **Suggestion: reuse `validateTimeoutSec` to reduce duplication.**

The overflow guard (lines 107–111) and the negative check (lines 104–106) duplicate the same logic in `validateTimeoutSec`. Consider:

```go
func parseTimeoutEnv() (time.Duration, bool) {
	raw := strings.TrimSpace(os.Getenv(envOCRLLMTimeout))
	if raw == "" {
		return 0, false
	}
	sec, err := strconv.Atoi(raw)
	if err != nil {
		return 0, false
	}
	d, err := validateTimeoutSec(sec)
	if err != nil {
		return 0, false
	}
	return d, true
}
```

This keeps one source of truth for validation rules and avoids the two functions drifting apart over time.

Note: `validateTimeoutSec(0)` returns `(0, nil)`, so `OCR_LLM_TIMEOUT=0` would still produce `(0, true)` — preserving current behavior.

internal/llm/resolver.go
```diff
@@ -122,6 +173,9 @@ func tryOCREnv(modelOverride string) (ResolvedEndpoint, bool, error) {
 		}
 	}
 
+	// Note: OCR_LLM_TIMEOUT is applied globally in ResolveEndpointWithModelOverride,
+	// not here, so it works regardless of which strategy resolves the endpoint.
```
lizhengfeng101: Nit: this comment restates what is already clear from reading `ResolveEndpointWithModelOverride` (lines 79–84). The comment at the override site itself (line 79–81) is the useful one; this one could be dropped to reduce noise.

css521 (approved):

Issue #275, opened 2026-07-02
Support a Codex-native mode that can use the current Codex session as the LLM runner


lihqi:
### Problem Statement

First of all, thank you for open-sourcing Open Code Review and for adding Codex skill/plugin support. The project is very useful, especially the deterministic diff handling and structured review workflow.

I would like to use Open Code Review from inside Codex via the installed skill/plugin. Today, the Codex integration invokes the local `ocr` CLI, and `ocr` still requires a separately configured LLM endpoint/API key. For users who already have a Codex subscription or token allowance, this means the Codex integration cannot use their existing Codex quota and requires another provider setup or extra API billing.

This is a bit confusing from a user perspective because the workflow is launched inside Codex, but the actual model calls are still managed outside Codex.

### Proposed Solution

It would be great to have a Codex-native mode for the Codex integration.

Possible shape:

- A documented mode such as `ocr review --audience agent --llm-runner codex`, or a dedicated Codex skill/plugin path.
- In this mode, Open Code Review could keep the deterministic parts of the pipeline, such as diff/file selection, grouping, rules, and output formatting.
- The LLM judgment/review step would be delegated to the current Codex session, so users can use their existing Codex quota instead of configuring `OCR_LLM_URL` / `OCR_LLM_TOKEN`.

If direct access to the Codex model runner is not technically possible, another helpful option might be to expose an `ocr` command that only prepares the deterministic review context as JSON/Markdown. The Codex skill could then ask Codex itself to perform the review from that context.

### Alternatives Considered

Current workarounds I found:

- Configure a separate OpenAI/Anthropic/OpenAI-compatible endpoint for `ocr`.
- Use a local model through Ollama/LM Studio.
- Use Codex's native code review without Open Code Review.

These work, but they either require extra provider setup/cost, or they lose the Open Code Review pipeline that makes this project valuable.

### Affected Area

Review Agent / LLM interaction

### Additional Context

Related docs/PRs I looked at:

- The README says the Codex plugin registers a skill that runs `ocr review --audience agent`.
- PR #69 notes that the Codex integration does not make Codex the internal LLM runner for OCR.

I completely understand there may be product/API limitations here. Even a short documentation note explaining the limitation and recommended Codex-native workaround would be very helpful. Thanks again for the great project!

lizhengfeng101:
Thanks for the detailed and thoughtful suggestion!

[...]

PR #276, opened 2026-07-02
feat: add Codex-owned review workflow

munsunouk:
## Summary

This re-submits the Codex-owned review flow from #245 and addresses the OCR bot feedback from that PR.

The change keeps OCR as the deterministic data plane for host-agent-led reviews:

- `ocr agent prepare` builds review bundles/manifests without invoking OCR's LLM backend
- `ocr agent context` exposes target-aware read/diff/search helpers
- `ocr agent validate-comments` validates agent findings against immutable bundle evidence
- `ocr agent report` renders validated findings

This is intended to address the Codex-native workflow requested in #275 without requiring OCR to call Codex CLI as a chat-completions provider.

## Maintainer follow-up

- Preserved existing `ocr review` / `ocr scan` include-filter behavior; strict include allowlist semantics are intentionally left out of this PR.
- Added coverage that symbolic refs still work for code search after ref pre-resolution.
- Landed the neutral `ocr agent ...` CLI surface in this PR and removed the public `ocr codex` command path.
- Session/report observability is now agent-neutral, and viewer support for legacy `codex_event` / `codex-owned` records has also been removed.

## OCR bot feedback addressed

- add top-level help hint for `ocr codex`
- finalize Codex report sessions after report rendering succeeds
- improve scan manifest diagnostics
- clamp context reads to the documented 500-line maximum
- use cleaned search patterns for context search
- add cancellation checks and cheaper partition sizing
- normalize bundle schema `$defs` naming
- fix Markdown code span/fence escaping and text warnings
- make scan target hashes cover reviewed files and make `partial` budget-driven
- avoid redundant scan batch parsing
- stop validation after excluded paths and cache target file reads
- preserve native include-filter behavior for diff and scan targets
- resolve range/commit refs to immutable SHAs before later diff loading
- stream untracked file hashing instead of reading whole files into memory
- add viewer fallbacks for empty Codex fields
- harden Codex session record serialization and avoid redundant chmod
- propagate code-search ref resolution timeouts
- escape codegraph script path interpolation
- remove the public `ocr codex` CLI surface in favor of `ocr agent`
- reject legacy `codex_event` / `codex-owned` viewer records

## Tests

- `GOTOOLCHAIN=auto go test ./...`
- `GOTOOLCHAIN=auto go build -o dist/opencodereview ./cmd/opencodereview`


lizhengfeng101:
Thanks a lot for picking this back up and carefully working through all the earlier bot feedback — this is a substantial, well-structured contribution, and the two-layer split (OCR as the deterministic data plane, the host agent as the reasoning layer) is exactly the right direction. Splitting `prepare` / `context` / `validate-comments` / `report` into pure, schema-bound CLI commands with no LLM dependency is a clean design, and the parity matrix + JSON schemas make the contract easy to reason about. Really nice work.

Two directions we'd love to see this evolve toward:

**1. An agent-agnostic ("host-agnostic") tooling model.**
The deterministic layer here isn't really Codex-specific — bundles, validation, and reports are identical regardless of who drives them. Any agent that can run a shell command (Claude Code, Cursor, Qoder, …) could reuse it as-is. Right now the naming and command space are bound to Codex (`ocr codex ...`, `codex-review-bundle-v1`, `controlPlane` values, session records tagged codex-owned). It would be great to generalize the surface — e.g. a neutral `ocr agent ...` namespace with Codex as the first adopter — so this becomes a reusable capability rather than a Codex-only path. The host-specific glue (plugin/skill manifests, or exposing these commands as MCP tools) can then stay a thin adapter on top of one shared core. Given we already ship an MCP server, exposing this layer over MCP could let MCP-capable hosts reuse it with zero per-host adaptation.

**2. Zero impact on the existing `ocr review` / `ocr scan` core.**
A hard requirement for us is that this new path must not change the behavior of the existing native review/scan agents. Most of the PR meets that bar (pure additions + behavior-preserving refactors), but there's one spot that does change existing behavior:

The refactor that extracts the inline file-filter logic from `internal/agent/preview.go` and `internal/scan/agent.go` into the shared `internal/reviewfilter` package also changes the `--include` semantics. The new `reviewfilter.Filter.ExcludeReason` adds:

```go
if f.FileFilter != nil && f.FileFilter.HasInclude() {
    if !f.FileFilter.IsUserIncluded(path) {
        return model.ExcludeUserRule
    }
}
```

Neither the old review path nor the old scan path had this branch. Previously, a file not matching the `--include` list was *not* short-circuited — it fell through to the extension/path checks and was often still reviewed. The new code turns `--include` into a hard allowlist (anything outside the list is excluded). The new test case `docs/readme.md + Include:["src/**"] → ExcludeUserRule` confirms the change (this used to be `ExcludeNone`).

The new semantics are arguably more intuitive, but they do change what native `ocr review` / `ocr scan` return for existing users. Our preference would be to keep this refactor strictly behavior-preserving for the native paths — i.e. have `reviewfilter` faithfully reproduce the existing behavior — and if the `--include` = strict-allowlist change is desired, land it as a separate, clearly-documented change rather than folding it into this feature.

(Minor, same theme: `internal/tool/code_search.go` now pre-resolves a ref via `rev-parse` before grepping. Equivalent for resolvable refs, but the error path is now stricter — worth a quick check that it doesn't alter existing review/scan search behavior.)

Overall: love the design and the discipline here. If we can (1) generalize it beyond Codex and (2) guarantee the native kernel behavior is untouched, this becomes a really elegant and safe foundation. Thanks again for the great work!

munsunouk:
Thanks for the detailed review — agreed on both points.

I’ll first address the hard compatibility requirement in this PR: the `reviewfilter` extraction should be behavior-preserving for the existing `ocr review` / `ocr scan` paths. I’ll remove the strict `--include` allowlist behavior from this PR and add/adjust tests so the native paths keep their previous semantics. If strict include semantics are still desirable, I’ll treat that as a separate, explicitly documented behavior-change PR.

On the host-agnostic direction, I agree as well. The deterministic bundle/context/validate-comments/report layer is not inherently Codex-specific. To avoid expanding this already-large PR too much, I’d prefer to keep this PR focused on the Codex-first adapter plus compatibility fixes, then follow up with a separate generalization PR for a neutral surface such as `ocr agent ...` and/or MCP exposure, with Codex as one thin host adapter on top.

I’ll push the behavior-preserving include fix first.

munsunouk:
Pushed the behavior-preserving include fix in `e4276ad`.

What changed:
- Removed the strict `--include` allowlist behavior from the shared review filter.
- Preserved the native diff-review include ordering: user exclude first, include match can pass, then extension/default-path checks.
- Preserved the native full-scan ordering separately: user exclude, extension check, include match, then default-path checks.
- Added/updated tests to lock both native paths so non-matching include patterns no longer exclude otherwise-reviewable files.

Verification:
- `GOTOOLCHAIN=auto go test ./internal/reviewfilter ./internal/agent ./internal/scan`
- `GOTOOLCHAIN=auto go test ./...`
- pre-push hook passed.

lizhengfeng101:
Thanks for the quick turnaround — I verified `e4276ad`, and the include fix is exactly right. Splitting the two native orderings into `ExcludeReason` (review: include short-circuits before ext) and `ExcludeScanReason` (scan: ext before include), and pointing `scan/classify.go` at the scan variant, faithfully preserves both native paths while still removing the duplication. The restored test expectations confirm it. And thanks for the extra `rev-parse` symbolic-ref test coverage too. 👍

Fully agree on keeping this PR focused and landing the host-agnostic generalization separately — that's the right scope call.

One thing I'd like to raise before that follow-up, specifically about the CLI surface we commit to here:

I'd suggest **not shipping `ocr codex` as a public top-level command**. The `prepare` / `context` / `validate-comments` / `report` layer is a deterministic data plane that's identical for every host — which host drives it (Codex, Claude Code, Cursor, …) belongs to the adapter layer (plugin/skill/MCP), not to the CLI command tree. Naming a top-level command after one host effectively promotes "who drives it" into a command dimension, and it means we'd be publishing an interface we already know we'll want to rename.

The concern is both debt and newcomer UX:
- If we later add `ocr agent`, we either keep both (top-level becomes `review / scan / codex / agent / ...`, and a newcomer immediately wonders which of `codex`/`agent` to use and how they differ), or we deprecate `codex` and migrate docs/plugins/user scripts. Neither is great.
- What actually confuses newcomers isn't an extra advanced command — it's two overlapping commands doing the same thing.

Since I don't think this needs to expand the PR, two low-cost options:
1. **Land the neutral name now** (e.g. `ocr agent ...`) with Codex as the first adapter on top of it — same code, just a host-agnostic surface from day one; or
2. **Keep these commands experimental / hidden** in this PR (not listed in top-level `-h`), so we don't commit to a public contract until the neutral design settles.

Either way, the guiding principle would be: host differences (codex/claude/cursor) live entirely in the adapter layer (plugin/skill/MCP), and the CLI top level stays limited to capability commands like `review` / `scan`. That keeps the surface clean for newcomers and avoids a future deprecation cycle.

Not a blocker for the compatibility work — just want to settle the public surface before `ocr codex` becomes something people depend on. Curious what you think.

munsunouk:
Thanks, that makes sense to me.

I agree that `prepare`, `context`, `validate-comments`, and `report` are deterministic data-plane primitives, not Codex-specific capabilities. Making Codex a top-level public command would bake the first adapter into the CLI shape, which gets awkward once Claude Code, Cursor, MCP, or other adapters use the same layer.

I’ll update this PR so `ocr agent ...` is the canonical public surface, and I’ll update the user-facing Codex plugin/skill instructions to use that path. I won’t keep `ocr codex` as a parallel public command, since that would create a confusing “which one should I use?” split.

To keep this PR scoped, I’ll limit the change to the CLI, docs, plugin, and skill instructions. I won’t rename broad internal package/type/schema identifiers here unless they are user-facing or required for tests. A deeper protocol/internal rename can be handled separately if we decide it is worth doing.

munsunouk:
Updated this PR to make `ocr agent ...` the only CLI surface for these deterministic primitives.

I removed the `ocr codex` dispatch entirely, including the hidden compatibility path, and updated the user-facing plugin/skill/docs instructions to use `ocr agent`.

I also made the user-facing observability path agent-neutral: `--session-id` now records `controlPlane: "agent"`, `model: "host-agent"`, `reviewMode: "agent"`, and `agent_event`; the viewer/report labels now say Agent rather than Codex. In `db947d1`, I also removed the remaining viewer read-side compatibility for legacy `codex_event` / `codex-owned` records, so new and supported viewer history is agent-only as well.

I intentionally kept broader internal/schema names unchanged to avoid expanding this PR beyond the public CLI/docs/plugin surface.

Validation: `GOTOOLCHAIN=auto go test ./...` and the pre-push hook pass.

munsunouk:
Thanks again for the detailed feedback here — especially the push toward a host-agnostic surface and the requirement to keep the existing `ocr review` / `ocr scan` behavior unchanged.

I reworked the landing plan to make the review scope smaller and cleaner. Instead of continuing this as one large PR, I’m splitting it into a stacked series and opening the first PR first:

1. host-agent review bundle runtime
2. host-agent surface / alias / docs
3. Cursor adapter and CI bridge

I’m starting with PR1 only so the core runtime/protocol/session/reporting layer can be reviewed independently before I layer the surface rename and the Cursor adapter and CI bridge on top.

New PR1: https://github.com/alibaba/open-code-review/pull/301

I’ll use that as the canonical review path going forward, and then follow with PR2/PR3 after PR1 direction is confirmed. This PR can then be closed as superseded once the split path is in place.

Thanks again — the discussion on this PR directly shaped the split.

lizhengfeng101:
Hi @munsunouk — first, thank you for the depth and persistence here (I know this is a careful resubmit of #245). We've read through the implementation in detail, and the core design — keeping OCR as a **deterministic data plane** that never calls its own LLM, plus the **evidence-based comment validation** (bundle-id binding, target freshness, byte-exact `existing_code` checks) — is exactly the right shape for host-agent-owned review.

Some context on our side: demand for a "run review on my coding-agent subscription, no standalone API key" mode has been growing steadily (see #331 and others), and we've now decided to make this an **officially owned feature** on our roadmap. Your PR is the most complete realization of that direction we've seen, and we'd like to build on it rather than reinvent it.

Because this feature carries a real **quality bar** — the model-output parity benchmark you yourself flagged as the remaining release gate in `CODEX_PARITY_MATRIX.md` (recall/false-positive parity on a fixed ground-truth corpus) — it involves a substantial evaluation effort before it can land on `main`. So we'd like to propose the following, and get your thoughts:

1. **Merge this into a dedicated integration branch (not `main`) first**, as the baseline for the feature. From there we'll iterate on it together and run the evaluation work, rather than blocking your contribution while that happens.
2. **We'd keep you credited as a co-author** on the resulting work — this started from your design and we want that reflected.
3. **We'd like to hold off on the `ocr agent context` capability for now.** Not a rejection — it's the piece most entangled with the evaluation (its behavior needs to be validated against the benchmark before we commit to the surface). We'd rather land the prepare/validate/report core first and revisit `context` once the eval tells us what it should look like. Does that sequencing make sense to you, or do you see `context` as load-bearing for the core flow?

Would you be open to this path — landing on an integration branch and continuing to collaborate on it from there? Happy to adjust the branch/process to whatever works best for you.

munsunouk:
Hi @lizhengfeng101,

Thank you for the thoughtful review and for recognizing the design and implementation effort here. I’m very happy that this is becoming an officially owned feature on the roadmap.

Yes, this sequencing makes sense to me.

I’m happy for the prepare/validate/report runtime to land first on a dedicated integration branch as the baseline for the evaluation work, with `main` remaining gated on the model-output parity benchmark.

I also agree that `ocr agent context` can be deferred. I consider it complementary rather than load-bearing for the core flow. The initial integration baseline can be limited to:

1. `ocr agent prepare` producing the deterministic review bundle
2. host-agent generated findings
3. `ocr agent validate-comments` validating findings against immutable bundle evidence
4. `ocr agent report` rendering validated findings

For the process, I suggest an upstream integration branch named `integration/host-agent-review`. I will open a new, clean PR against that branch containing only this narrowed core runtime scope.

I will keep follow-up work in this series limited to the host-agent runtime and its neutral surface/documentation. `ocr agent context` will remain deferred until the benchmark results guide its design.

I’d appreciate preserving the co-author attribution in the resulting merge or squashed commit metadata. I’m also glad to help define the fixed corpus, recall/false-positive metrics, and acceptance criteria in `CODEX_PARITY_MATRIX.md`.

Thanks again. I’m happy to continue collaborating on this path.

lizhengfeng101:
Hi @munsunouk,

Thank you for your understanding and for aligning on the sequencing — really appreciate the collaborative spirit throughout this discussion.

On our side, we've already started the design work for the integration branch and the evaluation framework. We'll open a PR soon and cc you on it — your input and guidance there would be very welcome.

When that PR lands, you'll be listed as a co-author on the merge commit, which reflects the fact that this work is something we've shaped together. Your design and implementation effort here are a meaningful part of the foundation.

Looking forward to continuing the collaboration.

munsunouk:
Thank you — that sounds great.

I’ll keep this PR unchanged for now and look forward to reviewing the integration and evaluation PR once you cc me. I’ll be glad to provide detailed feedback and help validate the host-agent workflow and benchmark criteria.

I appreciate the co-author attribution and look forward to continuing the collaboration.

lizhengfeng101:
Hi @munsunouk — thanks for the effort you put into this PR and the earlier iteration in #245.

We've since shipped **Delegation Mode**, which addresses the same core problem from a different angle: it delegates the actual LLM calls to the host agent (Claude Code, Codex, Cursor, or any other coding agent with an existing subscription), so OCR acts purely as a deterministic review scaffold without needing its own model endpoint.

You can see it live here: https://alibaba.github.io/open-code-review/#/docs/delegate

Given this overlap, we're going to close this PR. That said, we'd love to continue collaborating — if you have ideas for improving Delegation Mode or see gaps in how it serves the Codex workflow specifically, please open an issue or reach out. We'd welcome contributions that push the delegated review experience further.

Thanks again for driving this forward!

kaluli123123:
For readers of this closed PR who want a different integration trade-off: I published an independent experiment, [open-code-review-subscription](https://github.com/kaluli123123/open-code-review-subscription).

Unlike the host-agent-owned workflow discussed here, this fork keeps the native `ocr review` / `ocr scan` orchestration and tool loop inside OCR. Its `codex-cli` backend uses the official Codex app-server; `claude-cli` invokes the official Claude Code CLI. Authentication stays with those CLIs, with no OAuth-token extraction by OCR and no API-key fallback from these backends. This is a different option, not an implementation of this PR's host-driven design, and does not require reopening it.

Build the fork from source; the upstream npm package does not contain these additions. See [setup, verification scope, and known limitations](https://github.com/kaluli123123/open-code-review-subscription/blob/main/docs/cli-subscriptions.md). It is unofficial and experimental, not unlimited/free inference; subscription quotas and account extra-usage settings still apply. AI disclosure: development/review used OpenAI Codex/ChatGPT assistance, coordinated in Prime Agent, and Claude Code for the precommit OCR review.

---

给仍在关注这个已关闭 PR、希望选择另一种集成方式的用户：我发布了一个独立实验项目 [open-code-review-subscription](https://github.com/kaluli123123/open-code-review-subscription)。

与这里讨论的“宿主代理主导流程”不同，这个 fork 让 OCR 继续负责原生 `ocr review` / `ocr scan` 的编排与工具循环，底层通过官方 Codex app-server 或 Claude Code CLI 请求模型。认证仍由官方 CLI 管理，OCR 不提取 OAuth token，这两个后端也不回退到 API key 调用。它是另一种取舍，并非本 PR 的实现，也不请求重新开启此 PR。

需要从 fork 源码构建；上游 npm 包不包含这些新增能力。[配置、验证范围与已知限制](https://github.com/kaluli123123/open-code-review-subscription/blob/main/docs/cli-subscriptions.md)均已列明。当前是非官方实验版，不代表免费或无限量推理，仍受订阅配额与账户额外收费设置约束。AI 使用披露：开发及评审使用 OpenAI Codex/ChatGPT 辅助，由 Prime Agent 协调；提交前 OCR 审核使用了 Claude Code。

PR #280, opened 2026-07-02
fix(telemetry): support http:// scheme in otlp_endpoint for insecure gRPC

### LINKED ISSUE #268: [Bug] telemetry.otlp_endpoint does not support non-TLS connections, and http:// scheme causes "too many colons in address" error
### OpenCodeReview Version

open-code-review v1.7.0 (6adcb1e) darwin/amd64 built at: 2026-07-01T11:16:07Z

### Operating System

macOS (Intel)

### Installation Method

npm (global)

### LLM Provider

## Bug Description

When configuring telemetry to use a self-hosted OTLP collector over a 
private network (e.g. Tailscale), there is no way to establish a 
non-TLS gRPC connection. Two issues exist simultaneously:

**Issue 1:** When `otlp_endpoint` is set to `host:port` format, ocr 
defaults to TLS, with no way to disable it:

[ocr] WARNING: telemetry shutdown error: telemetry shutdown: [failed
to upload metrics: context deadline exceeded: rpc error: code =
Unavailable desc = connection error: desc = "transport: authentication
handshake failed: tls: first record does not look like a TLS handshake"]

**Issue 2:** When `otlp_endpoint` is set with `http://` scheme 
(per OTel spec, `http://` should indicate insecure connection), 
it causes a parsing error:

rpc error: code = Unavailable desc = invalid target address
http://***:24317/, error info: address
http://***:24317:443: too many colons in address

## Steps to Reproduce

```bash
ocr config set telemetry.enabled true
ocr config set telemetry.exporter otlp

# Issue 1: TLS handshake failure with plain host:port
ocr config set telemetry.otlp_endpoint 192.168.1.100:14317
ocr review  # → TLS handshake error

# Issue 2: "too many colons" with http:// scheme
ocr config set telemetry.otlp_endpoint http://192.168.1.100:14317
ocr review  # → too many colons in address
```

## Expected Behavior

Per the [OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md):

> A scheme of `http` indicates an insecure connection and takes 
> precedence over the insecure configuration setting.

The following should work for non-TLS connections:
```bash
ocr config set telemetry.otlp_endpoint http://192.168.1.100:14317
```

## Root Cause Analysis

In the telemetry initialization code, `otlp_endpoint` is likely 
passed directly to `otlptracegrpc.WithEndpoint()` without:

1. Stripping the `http://` or `https://` scheme prefix  
   (`WithEndpoint()` only accepts `host:port`, not full URLs)
2. Calling `otlptracegrpc.WithInsecure()` when `http://` scheme 
   is detected

The correct implementation should be:

```go
endpoint := config.Telemetry.OtlpEndpoint
opts := []otlptracegrpc.Option{}

if strings.HasPrefix(endpoint, "http://") {
    endpoint = strings.TrimPrefix(endpoint, "http://")
    opts = append(opts, otlptracegrpc.WithInsecure())
} else if strings.HasPrefix(endpoint, "https://") {
    endpoint = strings.TrimPrefix(endpoint, "https://")
    // TLS is default, no extra option needed
}

opts = append(opts, otlptracegrpc.WithEndpoint(endpoint))
```

## Environment

- ocr version: v1.7.0
- OS: macOS15
- OTLP backend: self-hosted OTel Collector over Tailscale private network
- Use case: sending telemetry to a private VPS without public TLS cert

## Suggested Fix

Either:
1. Parse the `http://` scheme from `otlp_endpoint` and call 
   `WithInsecure()` accordingly (as shown above)
2. Or add a new config key `telemetry.otlp_insecure` (bool) 
   that explicitly controls TLS behavior

### Expected Behavior

No Error

### Logs / Error Output

```shell

```

### Additional Context

_No response_

Qwenjin:
## Description

`initOTLPProviders` passed `cfg.OTLPEndpoint` directly to `WithEndpoint()`
without parsing the scheme, so:

- bare `host:port` forced TLS, breaking plaintext OTel Collectors
- `http://host:port` failed with "too many colons in address"

This PR parses the scheme before constructing the trace/metric exporters:
- `http://` → strip scheme, `WithInsecure()`
- `https://` → strip scheme, keep default TLS
- bare `host:port` → unchanged, keep default TLS (backward compatible)

Scheme matching is case-insensitive.

## Type of Change

## How Has This Been Tested?

Added `TestParseOTLPEndpoint` covering http/https/bare-host/uppercase-scheme/
short-string cases. Also manually verified against a real internal OTel
Collector configured with plaintext (insecure) gRPC — `ocr review` completes
without the previous "too many colons in address" / TLS handshake failure,
and a byte-level capture confirmed the connection is plaintext HTTP/2 (not a
TLS ClientHello) when `http://` is used.

## Checklist

- [ ] I have signed the CLA

## Related Issues

Fixes #268

### DIFF
--- a/internal/telemetry/exporter.go
+++ b/internal/telemetry/exporter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
 	"go.opentelemetry.io/otel/sdk/resource"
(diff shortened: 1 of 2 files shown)


lizhengfeng101 (approved):
LGTM

PR #286, opened 2026-07-03
fix: support Astro files in review filters

munsunouk:
## Summary
- Add `.astro` to the built-in supported file extension allowlist.
- Cover Astro extension matching with lowercase and uppercase test cases.

## Why
Astro files are source files and should be reviewable by `ocr review` and `ocr scan`. Previously they were filtered out as unsupported extensions.

## Test Plan
- `GOTOOLCHAIN=auto go test ./internal/config/allowlist ./internal/scan`

Note: `go test ./internal/reviewfilter` is not present on current `origin/main`. The broad pre-push hook also runs `go test ./...`, but that suite currently mutates the active worktree during git-related tests, so I pushed with `--no-verify` after the scoped tests passed.

### DIFF
--- a/internal/config/allowlist/allowed_ext_test.go
+++ b/internal/config/allowlist/allowed_ext_test.go
@@ -14,6 +14,8 @@ func TestIsAllowedExt(t *testing.T) {
 		{".java", true},
 		{".ts", true},
 		{".tsx", true},
+		{".astro", true},
+		{".ASTRO", true},
 		{".py", true},
 		{".rs", true},
 		{".ets", true},
--- a/internal/config/allowlist/supported_file_types.json
+++ b/internal/config/allowlist/supported_file_types.json
@@ -43,6 +43,7 @@
   ".less",
   ".html",
   ".htm",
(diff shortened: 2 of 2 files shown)


lizhengfeng101 (approved):
LGTM

lizhengfeng101:
Thanks a lot for the contribution, @munsunouk! 🎉 Adding `.astro` to the supported extension allowlist (with both lowercase and uppercase test coverage) is a clean and well-scoped change — much appreciated.

With Astro files now reviewable by `ocr review` and `ocr scan`, the natural next step would be dedicated, high-quality review rules tailored to Astro's specifics — e.g. the component script vs. template boundary, frontmatter fences, island hydration directives (`client:*`), and `.astro`-specific pitfalls. Generic rules will work, but purpose-built ones would make reviews far more valuable for Astro developers.

If you're up for it, we'd love a follow-up PR contributing professional `.astro` review rules — you clearly know the ecosystem well, and it would be a great addition for the community. Thanks again! 🙌

PR #291, opened 2026-07-03
fix(mcp): make setup timeout configurable

ScarletCarpet:
The MCP server setup command (e.g. codegraph init && codegraph index) previously had a hardcoded 5-minute timeout, which was too short for large repositories and caused the process to be killed.

- Add `timeout` field to mcp_servers config (default 30m), controlling the setup command's maximum execution time
- Update config CLI, tests, and README docs accordingly

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [x] I have signed the CLA


lizhengfeng101 (commented):
### 1. Field type & naming — `timeout` (`time.Duration`)

Two issues with the current `Timeout time.Duration`:

- **Serialization is opaque and fragile.** `time.Duration` has no JSON string form, so `30m` is written to `config.json` as nanoseconds: `"timeout": 1800000000000`. Worse, if a user hand-edits the file to the documented `"30m"`, the *entire* config fails to load (`cannot unmarshal string into ... time.Duration`).
- **`timeout` is too broad a name.** At the MCP-server level it reads like a connection/request timeout, but it only bounds the `setup` command.

Suggest storing a plain positive integer in **minutes** and renaming the field to tie it to `setup` — `setup_timeout`:

```go
SetupTimeout int `json:"setup_timeout,omitempty"` // minutes
```

```go
case "setup_timeout":
    n, err := strconv.Atoi(value)
    if err != nil || n <= 0 {
        return fmt.Errorf("invalid value for %s: must be a positive integer (minutes)", key)
    }
    entry.SetupTimeout = n
```

```go
setupTimeout := 30 * time.Minute
if serverCfg.SetupTimeout > 0 {
    setupTimeout = time.Duration(serverCfg.SetupTimeout) * time.Minute
}
```

This stores `"setup_timeout": 30` — readable and safe to hand-edit — and the name makes the association with `setup` explicit. Usage becomes `ocr config set mcp_servers.<name>.setup_timeout 30`. Since the value is a bare number in minutes, please state the unit clearly in the README/help text. (The `time` import in `config_cmd.go` is then no longer needed.)

### 2. Localized READMEs not synced

Per the repo convention, README changes must be mirrored to all localized copies. This PR updates `README.md` and `README.zh-CN.md`, but `README.ja-JP.md`, `README.ko-KR.md`, and `README.ru-RU.md` all contain the same MCP `setup` section — and still carry the stale "5-minute timeout" note. Please update those three as well.

### 3. New tests don't assert anything

`TestInitMCPClients_SetupOnly` and `TestInitMCPClients_TimeoutConfig` end with `_ = clients` and pass regardless of behavior — they don't actually verify the timeout feature. They also spawn a real `echo` and try to exec `/nonexistent/cmd`, so they carry environment side effects for no coverage gain. Consider asserting real behavior, e.g. a very short timeout against a `sleep` setup command should abort setup. The positive/negative cases in `config_cmd_test.go` are a good template.

### Nits / positives

- Nice consistency: the field is threaded through the unknown-key error (`config_cmd.go`), the `flags.go` help text, and the `setMCPServerValue` error. With the rename, update those same three strings and the `config_cmd_test.go` cases too.
- Raising the default from 5m → 30m and documenting it is a sensible change on its own.

ScarletCarpet:
all fixed:
1. timeout rename to setup_timeout
2. update all docs
3. use intergal to save timeout

lizhengfeng101:
### Test feedback (`cmd/opencodereview/review_cmd_test.go`)

**1. `TestInitMCPClients_SetupTimeoutAborts` is misnamed**

The name implies it verifies the timeout-abort path, but the logic (and its own comment) does the opposite: *"A 1-minute timeout should be plenty for `sleep 5` to succeed."* It never exercises an abort, and the body ends with `_ = clients` — no assertion at all.

- The name misleads readers into thinking timeout-abort is covered when it isn't.
- It runs a real `sleep 5`, making a unit test take ≥5s for zero verification.

Suggestion: either make it genuinely trigger a timeout (very short timeout + longer setup, then assert the server is skipped), or rename to e.g. `TestInitMCPClients_SetupWithinTimeoutSucceeds` and shorten the sleep with a real assertion.

**2. Tests spawn real processes — stability/speed concern (minor)**

`TestInitMCPClients_SetupOnly` and `TestInitMCPClients_SetupTimeoutAborts` drive the full `initMCPClients`, forcing real `echo`/`sleep` spawns through `mcp.NewClient`'s 30s init context. This couples the tests to `PATH`, shell-builtin availability, and platform behavior (the repo ships `shell_windows.go`, so Windows is a target — `sleep`/`echo` differ there), making the suite slow and non-deterministic.

**Recommended approach:** don't test the timeout logic end-to-end through `initMCPClients`. Extract the piece this PR actually changed into a pure function and test that directly — no processes, no platform dependency:

```go
func effectiveSetupTimeout(cfg MCPServerConfig) time.Duration {
	if cfg.SetupTimeout > 0 {
		return time.Duration(cfg.SetupTimeout) * time.Minute
	}
	return 30 * time.Minute
}
```

The new logic is then covered by a fast, deterministic table test (`0` → 30m default, negative → 30m fallback, `30` → 30m, `1` → 1m), and the misnamed `SetupTimeoutAborts` test can be deleted.

If you additionally want to cover the real execution paths (setup fails → server skipped; setup exceeds timeout → aborted), extract a runner that takes a `time.Duration` rather than int-minutes, so an abort can be tested in milliseconds instead of a full minute:

```go
func runMCPSetup(ctx context.Context, script, repoDir string, timeout time.Duration) ([]byte, error) {
	setupCtx, cancel := context.WithTimeout(ctx, timeout)
	defer cancel()
	cmd := shellCommand(setupCtx, script)
	cmd.Dir = repoDir
	configureProcessGroup(cmd)
	return cmd.CombinedOutput()
}
```

- success: `exit 0` with a generous timeout
- failure: `exit 1`, assert error returned
- abort: `sleep 5` with `10 * time.Millisecond`, assert the context/error is non-nil — millisecond-fast and actually validates the abort

For any remaining tests that must launch real processes, guard them with `if testing.Short() { t.Skip(...) }` and route commands through `shellCommand` (or skip on Windows), so `make test` stays fast and CI stays stable across platforms.

lizhengfeng101:
@ScarletCarpet Thanks for this PR! The configurable timeout is a good idea for large repos.

A couple of things before we can merge:

1. **Default timeout**: I'd prefer keeping the default at 5 minutes (the current behavior) rather than changing it to 30 minutes. Bumping the default 6x would silently change the experience for existing users — if someone's setup command hangs, they'd now wait 30 minutes instead of 5 before getting an error. Users who need longer timeouts can explicitly set `setup_timeout`.

2. **Rebase**: Could you rebase onto the latest `main`? There are some conflicts to resolve.

Let me know if you have any questions!

ScarletCarpet:
> @ScarletCarpet Thanks for this PR! The configurable timeout is a good idea for large repos.
> 
> A couple of things before we can merge:
> 
> 1. **Default timeout**: I'd prefer keeping the default at 5 minutes (the current behavior) rather than changing it to 30 minutes. Bumping the default 6x would silently change the experience for existing users — if someone's setup command hangs, they'd now wait 30 minutes instead of 5 before getting an error. Users who need longer timeouts can explicitly set `setup_timeout`.
> 2. **Rebase**: Could you rebase onto the latest `main`? There are some conflicts to resolve.
> 
> Let me know if you have any questions!

sorry for late responding. i m so busy these days for fixing conflict.
and i'll do that when i m free. thx for your suggestion about this PR.

lizhengfeng101:
Hey @ScarletCarpet, just a friendly nudge on this — two things still blocking merge:

1. **Default timeout → 5 minutes**: Please revert the default back to 5m (the current behavior). Users who need longer can explicitly set `setup_timeout`. We don't want to silently change the experience for everyone.

2. **Rebase needed**: There are conflicts with the latest `main`, so a rebase is required before we can merge.

One more suggestion: I'd recommend **dropping the README changes from this PR**. The config table is already getting dense, and `setup_timeout` is straightforward enough that CLI help text + the test coverage should be sufficient. We can document it on the website or in a follow-up PR dedicated to docs — keeps this one focused and easier to review.

No rush, just let us know when you get a chance to pick this back up!

lizhengfeng101:
Hey @ScarletCarpet, thanks again for this — the configurable setup timeout is a real need for large repos, and we appreciate the work you put in.

Since the branch has been idle for a while and now has conflicts with main, we're going to close this PR to keep the queue tidy. This isn't a rejection of the idea: the hardcoded 5-minute setup timeout is still there on main, so the feature is still wanted. We'll likely pick it up in a fresh PR (keeping the 5m default, adding `setup_timeout` as an opt-in).

If you'd like to come back to it yourself, you're very welcome to reopen or open a new PR — just rebase, keep the default at 5m, and drop the README changes as discussed. Thanks again! 🙏

PR #294, opened 2026-07-03
feat(pages): add per-chapter routes for docs

lizhengfeng101:
## What

Give each docs chapter its own URL under the hash router.

- Add a `/docs/:slug` route alongside the existing `/docs`.
- Derive `activeSlug` from the URL param in `DocsPage` (falling back to `quickstart` for `/docs` or an unknown slug) instead of tracking it in component state.
- Switch chapters via the router (`navigate('/docs/<slug>')`), so the sidebar, prev/next pagination, and ⌘K search all update the URL.
- Add a dev-time invariant that throws if two sidebar entries share a slug, since each slug now maps to exactly one URL.

## Why

Previously every chapter lived at the single `/#/docs` route driven by internal `useState`. As a result chapters had no shareable/bookmarkable link and browser back/forward did not work. Now links like `/#/docs/installation` deep-link straight to a chapter.


xyJen (approved):
LGTM

PR #297, opened 2026-07-03
fix(review-filter): apply language for review filter stage

ScarletCarpet:
## Description

the review filter stage was not respond in `language` for some model who deny the rule of `without any explanation`

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [x] I have signed the CLA

### DIFF
--- a/internal/config/template/template.go
+++ b/internal/config/template/template.go
@@ -157,7 +157,7 @@ func resolveLang(lang string) string {
 }
 
 // ApplyLanguage injects a language directive into all system-role messages
-// across MAIN_TASK, PLAN_TASK (if set), and MEMORY_COMPRESSION_TASK.
+// across MAIN_TASK, PLAN_TASK (if set), MEMORY_COMPRESSION_TASK, and REVIEW_FILTER_TASK (if set).
 func (t *Template) ApplyLanguage(lang string) {
 	instruction := "\n\nAlways respond in " + resolveLang(lang) + "."
 	applyLanguage(&t.MainTask, instruction)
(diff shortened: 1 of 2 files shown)


lizhengfeng101:
Thanks a lot for digging into the review-filter behavior, @ScarletCarpet — it's a genuinely subtle area and the change itself is clean (nice nil-guard and test coverage). 🙏

After looking closely, though, I don't think this is the right layer for the fix, so I'm going to close it. Here's the reasoning:

**The review-filter output is a machine-parsed control signal, not user-facing text.** The prompt asks the model to return a bare JSON array of comment IDs, and the consumer (`executeReviewFilter` → `parseFilterResponse`) does `StripMarkdownFences` + `json.Unmarshal` and then just calls `RemoveByPathAndIndices`. Nothing this task emits is ever shown to the user — the comments that *survive* filtering were already produced by `MAIN_TASK`, which already carries the language directive.

Because of that, injecting `Always respond in <language>.` here has no user-visible field to act on, and it can actually work against us:
- It nudges the model to emit prose/explanation, which directly contradicts the prompt's own "without any explanation" rule.
- Any surrounding prose makes `json.Unmarshal` fail, so `parseFilterResponse` returns nil and the filter silently becomes a no-op.

**Contrast with `DedupTask`:** it may look like precedent since `ScanTemplate.ApplyLanguage` applies the directive to dedup too — but dedup emits `merged_content`, a genuinely user-facing field, so localizing it there is correct. The review filter has no such field, which is exactly why it's different.

**Overlap with #295:** that PR moves the filter to function calling specifically to *stop* models from emitting natural language. This change pushes in the opposite direction, so if #295 lands the language directive becomes moot anyway.

If the real symptom is chatty models breaking the parse, the more robust fixes are the function-calling approach in #295 or more lenient JSON extraction inside `parseFilterResponse`. If you've hit a concrete case where filter output actually reaches the user, please do reopen and point me at that path — I couldn't find one, and it would change the assessment.

Thanks again for the contribution and for the careful tests — really appreciate the effort here!

PR #300, opened 2026-07-04
fix(pages): address CodeQL XSS alerts in markdown rendering

lizhengfeng101:
## What

Fixes two CodeQL XSS alerts in the docs pages markdown rendering path.

### Alert #4 — Incomplete multi-character sanitization (`utils/headingId.ts`)
The heading-id generator stripped HTML tags with a single-pass regex
(`text.replace(/<[^>]+>/g, '')`), which CodeQL flags as unreliable: nested
tags such as `<scr<script>ipt>` can reassemble a tag after one pass.

Replaced with `DOMPurify.sanitize(text, { ALLOWED_TAGS: [], ALLOWED_ATTR: [] })`.
As a side benefit this fixes a pre-existing bug: headings containing HTML
entities (e.g. `## Tips & Tricks`) produced **different** anchor ids on the TOC
side (raw markdown) vs the renderer side (marked HTML output), so their anchor
links didn't work. DOMPurify decodes entities consistently, aligning both sides.

### Alert #5 — DOM text reinterpreted as HTML (`components/MarkdownRenderer.tsx`)
Mermaid's rendered SVG was assigned directly to `innerHTML`. Mermaid already runs
with `securityLevel: 'strict'`, which sanitizes its own output (safe label HTML
like `<b>`/`<span>` is kept, scripts/handlers are stripped). Wrapping the whole
SVG in another `DOMPurify.sanitize()` **broke rendering** — it strips namespaces,
inline `<style>`, and `foreignObject` labels, causing diagrams (e.g. the
architecture guide) to disappear.

Instead we make `securityLevel: 'strict'` explicit and inject mermaid's trusted
output directly, with a `// codeql[js/xss-through-dom]` suppression comment
documenting why the source is trusted.

## Verification
- `tsc --noEmit` passes
- Production webpack build passes
- `ocr review --audience agent` returns 0 comments
- Verified in-browser that architecture-guide mermaid diagrams render correctly
  (bold/small-font/line-break labels intact)

### DIFF
--- a/pages/src/components/MarkdownRenderer.tsx
+++ b/pages/src/components/MarkdownRenderer.tsx
@@ -10,6 +10,10 @@ import { generateHeadingId } from '../utils/headingId';
 // Initialize mermaid with dark theme
 mermaid.initialize({
   startOnLoad: false,
+  // 'strict' makes mermaid sanitize its own SVG output (DOMPurify internally):
+  // safe label HTML like <b>/<span> is kept, scripts/handlers are stripped.
+  // This is why we can inject the returned SVG directly below without re-sanitizing.
+  securityLevel: 'strict',
   theme: 'dark',
   themeVariables: {
     primaryColor: '#1a1a2e',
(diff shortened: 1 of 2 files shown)


xyJen (approved):
LGTM

xyJen (approved):

PR #302, opened 2026-07-04
fix(tool): harden code_search ref and colon-path parsing

munsunouk:
## Why this change is needed
`code_search` had three rough edges in normal review usage:

- `file_patterns` accepted traversal-style inputs like `../pkg`, even though the tool is meant to stay within the repository.
- symbolic refs such as branch names were passed straight into `git grep`, which makes search behavior depend on git's ref parsing instead of resolving the target commit first.
- result parsing assumed `file:line:content` output, which breaks when a repository contains paths with `:` in the filename.

## How this change works
- reject `file_patterns` containing `..` at the tool boundary
- resolve non-empty refs to a verified commit with `git rev-parse --verify --end-of-options <ref>^{commit}` before running `git grep`
- switch `git grep` parsing to `-z` output so filenames containing `:` are parsed safely

## Tests
- add a traversal-pattern rejection test
- add a symbolic-ref search test
- add a colon-path search test
- validate with:
  - `GOTOOLCHAIN=local /Users/munseon-ug/sdk/go1.25.5/bin/go test ./internal/tool -count=1`


munsunouk:
Superseded by #303. Reopened from a clean branch because the previous branch picked up unrelated local test commits and showed an incorrect file set.

PR #303, opened 2026-07-04
fix(tool): reject traversal pathspecs in code_search

munsunouk:
## Why this change is needed
This PR is intentionally limited to the `file_patterns` guard so the existing `code_search` search behavior stays unchanged.

`git grep` already rejects real traversal outside the repository, but `code_search` currently passes traversal-style patterns through and surfaces git's error later. Rejecting those inputs at the tool boundary gives callers an earlier, clearer error message without changing the grep path itself.

## What changed
- reject `file_patterns` that contain `..` as a path component
- leave normal patterns untouched, including filenames such as `foo..bar.go`

## Tests
- reject leading, middle, and trailing `..` path components
- allow a filename containing `..`
- validate with:
  - `GOTOOLCHAIN=local /Users/munseon-ug/sdk/go1.25.5/bin/go test ./internal/tool -count=1`

### DIFF
--- a/internal/tool/code_search.go
+++ b/internal/tool/code_search.go
@@ -34,6 +34,9 @@ func (p *CodeSearchProvider) Execute(ctx context.Context, args map[string]any) (
 	var patterns []string
 	for _, item := range filePatternsIface {
 		if s, ok := item.(string); ok && s != "" {
+			if hasTraversalPathComponent(s) {
+				return "Error: file_patterns must not contain ..", nil
+			}
 			patterns = append(patterns, s)
 		}
 	}
@@ -83,6 +86,15 @@ func (p *CodeSearchProvider) buildGrepArgs(searchText string, caseSensitive bool
 	cmdArgs = append(cmdArgs, pathspec...)
 
 	return cmdArgs
+}
+
+func hasTraversalPathComponent(pathspec string) bool {
+	for _, part := range strings.Split(pathspec, "/") {
+		if part == ".." {
+			return true
+		}
+	}
+	return false
 }
 
 func (p *CodeSearchProvider) runGitGrep(parentCtx context.Context, cmdArgs []string) (string, string, error) {
--- a/internal/tool/code_search_test.go
+++ b/internal/tool/code_search_test.go
@@ -433,6 +433,54 @@ func TestCodeSearchProvider_Execute_WithFilePatterns(t *testing.T) {
 	}
 }
 
+func TestCodeSearchProvider_Execute_RejectsTraversalPattern(t *testing.T) {
+	dir := setupTestRepo(t)
+	p := NewCodeSearch(&FileReader{RepoDir: dir, Mode: ModeWorkspace})
+	tests := []struct {
+		name    string
+		pattern string
+		want    string
+	}{
+		{name: "leading parent", pattern: "../pkg", want: "Error: file_patterns must not contain .."},
+		{name: "middle parent", pattern: "pkg/../internal", want: "Error: file_patterns must not contain .."},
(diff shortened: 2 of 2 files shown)


lizhengfeng101:
Thanks for the PR, @munsunouk — this is careful work and the tests are thorough. I dug into all three changes, and I wanted to be transparent about where we've landed.

**On the `file_patterns` `..` guard (keeping this):**

This one is safe and reasonable, so I'm happy to move forward with it. One small note for context: `git grep` already blocks real traversal on its own — tested on git 2.54:

```
git grep ... -- ../secret
→ fatal: '../secret' is outside repository   (exit 128)
```

So there's no actual out-of-repo read being fixed; the check mainly turns git's error into a friendlier, earlier one, which is a fine defense-in-depth. If you'd like, it'd be great to trim the PR down to just this change so it can land cleanly on its own.

**On the ref→commit resolution and the `-z` parsing rewrite (not planning to pursue these):**

After digging in, I'd rather not take these two forward, and I want to explain why honestly rather than leave you guessing:

- **Ref resolution:** symbolic refs (branch names) already resolve correctly when passed straight to `git grep` — the test name itself (`...SymbolicRefPreservesSearchBehavior`) reflects that behavior is unchanged. It adds an extra `rev-parse` subprocess per search without a concrete bug behind it, so the cost/benefit doesn't quite land for us.

- **`-z` parsing:** the case it fixes (filenames containing `:`) is rare — `:` is illegal/reserved on macOS and Windows. More importantly, the rewrite regresses on binary matches: the `Binary file … matches` line carries no NUL and ends with `\n`, so the parser folds it into the next record's filename. The old `SplitN(":", …)` path skipped those lines safely. Given that trade-off, I'd prefer to leave the current parser as-is.

None of this takes away from the quality of the work — these were well-reasoned changes and the investigation was genuinely useful. If you're up for it, paring the PR back to just the `..` guard would be a welcome, low-risk improvement. Thanks again!

munsunouk:
@lizhengfeng101 Thanks for the detailed review. I trimmed this PR down to just the `file_patterns` `..` guard and removed the ref-resolution and `-z` parsing changes, along with their related tests.

The PR now only adds the earlier, friendlier rejection for traversal-style pathspecs while leaving the existing `code_search` grep behavior unchanged. I also kept a focused test showing that valid filenames containing `..` (for example `foo..bar.go`) are still allowed.

Revalidated with `GOTOOLCHAIN=local /Users/munseon-ug/sdk/go1.25.5/bin/go test ./internal/tool -count=1`, and the branch also passed the full pre-push `go test ./...` hook.

lizhengfeng101:
A small maintainability note: let's add a short comment on `hasTraversalPathComponent` clarifying that it's **not a security boundary**.

As established earlier, `git grep` already rejects genuine out-of-repo traversal (`fatal: '../secret' is outside repository`, exit 128), so this guard only surfaces a friendlier, earlier error (defense-in-depth / better UX). It's also intentionally narrow — it matches `..` only as a literal `/`-separated path component, so magic pathspecs (e.g. `:(top)../x`, `:!../x`) aren't caught here, which is fine since git still backstops them.

Making this explicit prevents a future maintainer from mistaking it for a hardened traversal check and relaxing real validation elsewhere. Something like:

```go
// hasTraversalPathComponent reports whether a pathspec contains ".." as a
// literal path component. This is not a security boundary: git grep already
// rejects real out-of-repo traversal. It exists only to surface a friendlier,
// earlier error, so magic pathspecs are intentionally not covered here.
```

Non-blocking — the change itself looks good to merge.

lizhengfeng101 (approved):
LGTM

PR #308, opened 2026-07-06
fix(cmd): resolve review repo dir to git top level for monorepo subdirs

lizhengfeng101:
## Problem

Fixes #287.

When `ocr review` runs from a subdirectory of a monorepo, the working directory differs from the git repository root:

```
/project        # monorepo root (.git here)
  /router       # user runs `ocr review` here
  /subproject2
```

`RepoDir` was resolved to the current working directory (via `os.Getwd()` + `git rev-parse --git-dir`), so all git commands ran with `cmd.Dir` set to the subdir. With `diff.relative=true` in the user's git config, `git diff` emits subdir-relative paths (e.g. `src/...`), but the `file_read` tool reads content via `git show <ref>:<path>`, which resolves paths **relative to the repo root**. Result:

```
✘ file_read failed: git show HEAD:src/... : exit status 128:
  fatal: path 'router/src/...' exists, but not 'src/...'
```

Every `file_read` failed on the first try and only succeeded after the LLM retried with the `router/` prefix from git's hint — wasting extra LLM calls on every file.

## Fix

Resolve the review working dir to `git rev-parse --show-toplevel` (the repository top level) so `git diff` and `git show` agree on repo-root-relative paths.

- **review** (`requireGit=true`): resolves to the repo root.
- **scan** (`requireGit=false`): intentionally keeps the working-directory scope, so scanning a subdirectory stays limited to that subdirectory (no scope expansion).
- `rules check` (`resolveRepoDir`) also resolves to the top level; it does not enumerate files, so there is no scope change.

| Mode | Run location | RepoDir |
|------|--------------|---------|
| review | repo subdir | repo root (toplevel) |
| scan | repo subdir | current subdir |
| scan | non-git dir | current dir |

## Tests

- `TestResolveWorkingDir_SubdirResolvesToToplevel` — review subdir → repo root
- `TestResolveWorkingDir_ScanSubdirKeepsScope` — scan subdir → keeps subdir scope

`make check` and `make test` pass.

PR #309, opened 2026-07-06
fix(tool): resolve file_read paths against git top-level in monorepos

### LINKED ISSUE #287: `file_read` call failures in a monorepo
### OpenCodeReview Version

v1.6.6

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

Anthropic (Claude)

### Bug Description

When I run `ocr review...` in a monorepo project.

My workspace isn't the project root.

```
/project        # monorepo root
  /.git
  /subproject1  # working dir 
  /subproject2
```

So the `/project` is git repo, but `ocr` often spawned in the `/project/subproject1`

Then, I see many errors from the review process like:

```
✘ file_read failed: file "src/models/request_meta.py" not found: git show HEAD:src/models/request_meta.py: exit status 128: fatal:
 path 'subproject1/src/models/request_meta.py' exists, but not 'src/models/request_meta.py'
```

I think it can handle with this situation, but with many extra indirect LLM calls.

### Steps to Reproduce

Run `ocr review` in any subdirectory of the repository.

### Expected Behavior

`file_read` should respect the project structure even the process spawend in a subdirectory.

### Logs / Error Output

```shell

```

### Additional Context

_No response_

chethanuk:
## Description

Running `ocr review` from a **subdirectory of a monorepo** failed with `file not found` (#287). Git reports diff paths and `git show HEAD:<path>` paths relative to the repository root, but `RepoDir` was scoped to the invocation subdirectory. Joining a root-relative path onto a subdirectory `RepoDir` produced a double prefix (e.g. `repo/subproject1/subproject1/src/...`), so both disk reads and git-show reads missed the file.

**Fix:** `resolveWorkingDir` now anchors `RepoDir` at the git top-level via `git rev-parse --show-toplevel` on the review path (`requireGit=true`). `ocr scan` (`requireGit=false`) deliberately keeps the CWD so its `git ls-files` walk stays scoped to the subdirectory.

Two robustness details:
- The top-level lookup goes through a stdout-only git helper (`runGitCmdStdout`, using `cmd.Output()`) so git's stderr notices (permissions, deprecations, config warnings) can't pollute the resolved path.
- If `--show-toplevel` errors or returns nothing — e.g. a bare repo, where `--git-dir` succeeds but there is no work tree — resolution now fails loudly instead of silently reusing the subdirectory and reproducing the original bug.

```mermaid
flowchart TD
    A[resolveWorkingDir input, requireGit] --> B[filepath.Abs + os.Stat]
    B --> C[git rev-parse --git-dir]
    C --> D{is git repo?}
    D -- no, requireGit=true --> E[error: not a git repository]
    D -- no, requireGit=false --> F[return subdir, isGit=false]
    D -- yes, requireGit=false scan --> G[return subdir, isGit=true<br/>git ls-files stays scoped]
    D -- yes, requireGit=true review --> H[git rev-parse --show-toplevel<br/>stdout only]
    H --> I{ok and non-empty?}
    I -- yes --> J[RepoDir = git top-level<br/>fix for #287]
    I -- no --> K[return error<br/>fail loudly, no silent fallback]
    J --> L[return repo root, isGit=true]
```

## Behavior Changes (beyond the `file_read` title)

Anchoring `RepoDir` at the git top-level has two user-visible effects worth calling out. Both are behavior-preserving when `ocr` runs from the repo root, and only change the monorepo-subdirectory path — in the direction of consistency:

1. **`.opencodereview/rule.json` resolution scope.** `rules.NewResolver` → `loadProjectRule` reads `<RepoDir>/.opencodereview/rule.json`. With `RepoDir` now at the git top-level, `ocr review` from a subdirectory loads the **repo-root** rule file instead of `<subdir>/.opencodereview/rule.json`. This is the consistent behavior — rule entries match against root-relative diff paths, so anchoring at root makes matching correct — but a subproject-local `rule.json` under the subdir is now intentionally not consulted. Put shared rules at the repo root, or pass `--rule`. Documented at `loadProjectRule`.

2. **`ocr rules check` (`resolveRepoDir`) now hoists to the top-level too.** It previously kept the invocation subdirectory; it now delegates to `resolveWorkingDir(input, true)` so rule resolution stays consistent with the review path when run from a monorepo subdirectory. Documented at `resolveRepoDir`.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

`go test ./cmd/opencodereview/ ./internal/tool/` — 592 tests pass; `gofmt -l` clean; `go vet` clean.

New regression tests:
- `TestResolveWorkingDir_MonorepoSubdir` — review path hoists to the git top-level; scan path keeps the subdirectory.
- `TestResolveWorkingDir_BareRepoFailsLoudly` — bare repo (`--git-dir` ok, `--show-toplevel` fails) returns an error rather than silently falling back.
- `TestFileReader_Read_CommitMode_MonorepoSubdirPath` — `git show HEAD:<root-relative-path>` and workspace disk reads both resolve, the exact command from the issue.

- [x] Manual testing (described above)

## Checklist

- [ ] I have signed the CLA

## Related Issues

closes #287

### DIFF
--- a/cmd/opencodereview/git.go
+++ b/cmd/opencodereview/git.go
@@ -12,6 +12,15 @@ func runGitCmd(repoDir string, args ...string) ([]byte, error) {
 	return cmd.CombinedOutput()
 }
 
+// runGitCmdStdout is like runGitCmd but returns stdout only. Use it when the
+// output is consumed as data (e.g. a resolved path) so git's stderr warnings
+// (permissions, deprecations, config notices) can't pollute the result.
+func runGitCmdStdout(repoDir string, args ...string) ([]byte, error) {
+	fullArgs := append([]string{"-C", repoDir}, args...)
+	cmd := exec.Command("git", fullArgs...)
+	return cmd.Output()
+}
+
 func getCommitMessage(repoDir, commit string) (string, error) {
 	out, err := runGitCmd(repoDir, "log", "-1", "--format=%B", "--end-of-options", commit)
 	if err != nil {
--- a/cmd/opencodereview/review_cmd.go
+++ b/cmd/opencodereview/review_cmd.go
@@ -115,23 +115,13 @@ func runReview(args []string) error {
 	return emitRunResult(ctx, ag, comments, startTime, opts.outputFormat, opts.audience, q)
 }
 
+// resolveRepoDir resolves the repo dir for `ocr rules check`. It delegates to
+// resolveWorkingDir(requireGit=true) so it anchors at the git top-level just
+// like the review path — keeping rule resolution consistent when run from a
+// monorepo subdirectory (#287).
 func resolveRepoDir(input string) (string, error) {
-	if input == "" {
-		var err error
-		input, err = os.Getwd()
-		if err != nil {
-			return "", fmt.Errorf("get working directory: %w", err)
-		}
-	}
-	absPath, err := filepath.Abs(input)
-	if err != nil {
-		return "", fmt.Errorf("resolve absolute path: %w", err)
-	}
-	out, err := runGitCmd(absPath, "rev-parse", "--git-dir")
-	if err != nil || len(out) == 0 {
-		return "", fmt.Errorf("%s is not a git repository", absPath)
-	}
-	return absPath, nil
+	absPath, _, err := resolveWorkingDir(input, true)
+	return absPath, err
 }
 
 // requireGitRepo validates that the given directory is part of a git repository.
--- a/cmd/opencodereview/shared.go
+++ b/cmd/opencodereview/shared.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/open-code-review/open-code-review/internal/agent"
@@ -99,6 +100,25 @@ func resolveWorkingDir(input string, requireGit bool) (string, bool, error) {
 	isGit := err == nil && len(out) > 0
 	if !isGit && requireGit {
 		return "", false, fmt.Errorf("%s is not a git repository", absPath)
+	}
+	// #287: git reports diff and `git show HEAD:<path>` paths relative to the
+	// repository root, not the current directory. When `ocr review` runs from a
+	// subdirectory of a monorepo, anchor RepoDir at the git top-level so those
+	// root-relative paths resolve for both disk reads and git-show reads.
+	// requireGit is true only for the review path; scan (requireGit=false) keeps
+	// the CWD so its `git ls-files` walk stays scoped to the subdirectory.
+	if isGit && requireGit {
+		// runGitCmdStdout captures stdout only so git stderr notices can't
+		// pollute the resolved path. --show-toplevel fails (or is empty) when
+		// there is no work tree — e.g. a bare repo, where --git-dir succeeds so
+		// isGit is true. Fail loudly there instead of silently reusing the
+		// subdir, which would reproduce the #287 root-relative-path bug.
+		top, topErr := runGitCmdStdout(absPath, "rev-parse", "--show-toplevel")
+		t := strings.TrimSpace(string(top))
+		if topErr != nil || t == "" {
+			return "", false, fmt.Errorf("%s is a git repository without a work tree (bare repo?); cannot resolve its top level for review", absPath)
+		}
+		absPath = t
 	}
 	return absPath, isGit, nil
 }
--- a/cmd/opencodereview/shared_test.go
+++ b/cmd/opencodereview/shared_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"os"
+	"os/exec"
 	"path/filepath"
 	"testing"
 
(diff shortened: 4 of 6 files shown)


lizhengfeng101:
Thanks for pushing this upstream, @chethanuk 🙏 We had independently opened #308 for the same issue with essentially the same approach (`git rev-parse --show-toplevel` on the review path, keeping CWD scope for scan). Happy to close ours in favor of this one — it's the same fix plus the stdout-only helper. Two things I noticed while reviewing:

**1. Bare-repo path silently falls back — this contradicts the PR description.**

The description and the mermaid diagram both state that when `--show-toplevel` errors or returns nothing (e.g. a bare repo, where `--git-dir` succeeds but there is no work tree), resolution *fails loudly instead of silently reusing the subdirectory*. But the code does the opposite:

```go
if isGit && requireGit {
    if top, topErr := runGitCmdStdout(absPath, "rev-parse", "--show-toplevel"); topErr == nil {
        if t := strings.TrimSpace(string(top)); t != "" {
            absPath = t
        }
    }
}
return absPath, isGit, nil // topErr != nil / empty -> silently keeps the subdir, no error
```

In a bare repo, `git rev-parse --git-dir` succeeds → `isGit=true`, but `--show-toplevel` fails → `absPath` is silently kept and returned with `isGit=true`. That matches your inline code comment ("on any rev-parse failure it falls back to the caller-provided dir") but contradicts the PR description / diagram node `K[return error, fail loudly]`. Either add the fail-loudly branch, or drop that claim from the description.

**2. `TestResolveWorkingDir_BareRepoFailsLoudly` is listed in the PR body but is not in the diff.**

Only `TestResolveWorkingDir_MonorepoSubdir` and `TestFileReader_Read_CommitMode_MonorepoSubdirPath` are present. The bare-repo regression test appears to be missing — which lines up with point 1.

Minor / optional:
- `resolveRepoDir` (used by `ocr rules check`) isn't updated, so running it from a subdir still anchors at the subdir. Harmless for #287 (rules don't hit file_read), just an inconsistency with the review path.
- The review path now does two git calls (`--git-dir` then `--show-toplevel`); a single `--show-toplevel` would double as the repo-ness check.

Neither blocks the fix — the direction is correct. Just worth reconciling the description/tests with the actual behavior. Once this lands we'll close #308. Thanks again!

chethanuk:
All three points fixed and pushed (rebased on `main`):

- **Bare-repo silent fallback** — `resolveWorkingDir` now returns an error when `--show-toplevel` fails or is empty, instead of falling back to the subdir. This is the behavior the description and diagram already claimed.
- **Missing `TestResolveWorkingDir_BareRepoFailsLoudly`** — added.
- **`resolveRepoDir` not hoisted** — now a thin wrapper over `resolveWorkingDir(input, true)`, so `review` and `rules check` resolve identically.

Bare repo now fails loudly on both entry points (`exit 1`, no `file_read` loop):

```
$ ocr review --repo bare.git        → Error: … is a git repository without a work tree (bare repo?) …
$ ocr rules check --repo bare.git f  → same error
```

On the single-`--show-toplevel` suggestion, I kept the two calls:

| | non-repo | bare repo | scan (needs `isGit`, must not hoist) |
|---|---|---|---|
| single `--show-toplevel` | ✓ | ✗ reports "not a git repository" | ✗ hoists scan to root |
| `--git-dir` + `--show-toplevel` | ✓ | ✓ accurate "bare repo" error | ✓ |

`--git-dir` answers "is this a repo" (true for bare repos and worktrees); `--show-toplevel` gives the worktree root (fails on bare). Keeping both is what buys the accurate bare-repo error and the `isGit` signal the scan path needs without hoisting. It's one extra call at startup — if you'd rather the single call, that's #308's shape and I'm fine with it.

Re #308: same fix at heart. Fine to close either in favor of the other; I'd keep the `file_read`-layer regression test whichever lands.

Tests green (`cmd/opencodereview` + `internal/tool`), gofmt/vet clean.

lizhengfeng101:
Nice fix — the root cause is well-diagnosed and the change is nicely scoped (single point in `resolveWorkingDir`, clean review/scan split, stdout-only git helper, loud bare-repo failure, and solid regression tests incl. the macOS symlink canonicalization).

One suggestion: two behavior changes ride along with this fix that aren't called out in the description. Since they're user-visible, it'd be worth mentioning them so users aren't surprised:

1. **`.opencodereview/rule.json` resolution scope.** `rules.NewResolver` → `loadProjectRule` reads `<repoDir>/.opencodereview/rule.json`. With `RepoDir` now anchored at the git top-level, `ocr review` from a subdirectory loads the **repo-root** rule file instead of `<subdir>/.opencodereview/rule.json`. This is arguably a fix in itself (rule entries match against root-relative diff paths, so anchoring at root makes matching consistent), but a subproject-local `rule.json` in the subdir would now be ignored in favor of the root one.

2. **`ocr rules check` (`resolveRepoDir`) now hoists to the top-level too.** It previously kept the invocation subdirectory; delegating to `resolveWorkingDir(input, true)` changes that. The comment explains it's for consistency, which makes sense — just worth a line in the description/changelog since it's beyond the `file_read` title.

Neither blocks merge — just documentation. Everything else LGTM.

lizhengfeng101 (approved):
LGTM

PR #314, opened 2026-07-07
Feat/telemetry http exporter

yingjiexu2002:
## Description

OCR's existing telemetry only supports OTLP gRPC exporter. Some deployment environments (e.g. behind reverse proxies or firewalls) only expose HTTP endpoints. This PR adds OTLP HTTP exporter support with protocol routing via `OTEL_EXPORTER_OTLP_PROTOCOL`.

Additionally, the current trace coverage is limited to diff parsing and event recording — LLM calls, tool executions, and plan/filter phases are only captured as metrics, not as spans. This PR adds comprehensive span coverage across the full review lifecycle, enabling end-to-end diagnosis from traces:

- **LLM calls**: `llm.request` spans with model, duration, token count, error attributes
- **Tool execution**: `tool.execute.*` spans for all three tool paths (dynamic, code_comment sync/async, standard)
- **Plan phase**: `plan.execute` span with file path
- **Review filter**: `review_filter.execute` span with before/after comment counts
- **Root span**: `review.run` / `scan.run` with repo, from/to ref (review only), model metadata
- **TraceID output**: included as a `trace_id` field in JSON output for programmatic consumption; printed to stderr in text format for quick lookup

Error paths use the OTel standard pattern (`span.SetStatus` + `span.RecordError` + `defer span.End()`) for correct error propagation without modifying function signatures. Trace ID is propagated via `context.Context` (the OTel-native path) using a `TraceIDFromContext` helper, keeping CLI output functions decoupled from context plumbing.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

All 22 test packages pass with `-race` flag. Traces verified end-to-end on an OTel-compatible backend — confirmed span hierarchy, error status propagation, and LLM/tool attribute completeness. Self-reviewed with OCR; all findings addressed.

## Checklist

- [x] I have signed the CLA


lizhengfeng101 (commented):
### Code Review — Telemetry HTTP exporter & span coverage

Overall this is a solid change: span lifecycles are correctly paired across all paths (verified the four `executeToolCall` branches and the `RunPerFile` LLM span), the async goroutine ends its span via `defer` with no data race, error paths consistently use `SetStatus + RecordError + End`, and nil spans are guarded. A few points below for your judgement — none are blocking, feel free to decide whether they're worth addressing.

**1. `http/json` silently falls back to gRPC (worth reconsidering)**
In `exporter.go`, `http/json` hits the default branch and falls back to gRPC with a warning. But users typically request an HTTP protocol precisely because only an HTTP port is reachable (the reverse-proxy / firewall scenario in the PR description). Falling back to gRPC will most likely fail to connect. Consider falling back to `http/protobuf` (still HTTP transport) instead of gRPC, or making the warning explicit about the connectivity consequence.

**2. HTTP endpoint path component is not handled**
`parseOTLPEndpoint` only strips the scheme and a trailing slash. OTLP/HTTP endpoints often carry a path (e.g. `http://host:4318/v1/traces`), but `WithEndpoint` expects a bare `host:port` and will treat the path as part of the host. This is uncommon for gRPC but common for HTTP. Consider `otlptracehttp.WithEndpointURL` (accepts a full URL incl. path) for the HTTP branch, or document that the endpoint must be `host:port` only.

**3. LLM span coverage has gaps vs. the description**
`llm.request` spans only wrap the main `RunPerFile` completion call. The LLM calls inside `executePlanPhase` / `executeReviewFilter` only have their phase spans (`plan.execute` / `review_filter.execute`) with no `llm.request` child, and `ReLocateComment` (inside `resolveAndCollect`) has no span at all. The description says "comprehensive span coverage / LLM calls: llm.request spans" — either extend coverage or note the current scope.

**4. Returned span contexts are discarded (`_, span :=`)**
All `StartToolSpan` / `StartLLMSpan` calls discard the returned context, so nested operations (e.g. re-location LLM calls under a tool span) won't nest under their parent. If the flat hierarchy is intentional that's fine, but it's slightly at odds with the "confirmed span hierarchy" claim.

**5. Documentation**
The doc checklist item is unchecked. This adds a user-visible env var (`OTEL_EXPORTER_OTLP_PROTOCOL`) and a `trace_id` output field — worth syncing to the telemetry docs / README (note the project convention that README changes must be mirrored to all localized versions).

**6. Test gap**
The new tests cover exporter routing and the span helpers, but nothing asserts that `trace_id` is actually populated (the output tests just pass `""`). A test asserting a non-empty `trace_id` and the stderr TraceID print would close the loop.

yingjiexu2002:
> ### 代码审查 — 遥测 HTTP 导出器和 span 覆盖率
> 总体而言，这是一个可靠的改动：所有路径中的 span 生命周期都正确配对（已验证四个`executeToolCall`分支和`RunPerFile`LLM span），异步 goroutine 通过 `return` 结束其 span `defer`，没有数据竞争，错误路径始终使用`SetStatus + RecordError + End``return`，并且 nil span 也得到了保护。以下几点供您判断——这些都不是阻塞性的，您可以自行决定是否需要处理。
> 
> **1.`http/json`静默回退到 gRPC（值得重新考虑）** 在当前情况下`exporter.go`，`http/json`会执行默认分支并回退到 gRPC 并发出警告。但用户通常请求 HTTP 协议正是因为只有 HTTP 端口可访问（PR 描述中的反向代理/防火墙场景）。回退到 gRPC 很可能会导致连接失败。考虑回退到`http/protobuf`HTTP 协议（​​仍然是 HTTP 传输）而不是 gRPC，或者明确警告连接失败的后果。
> 
> **2. HTTP 端点路径组件未被正确处理，** `parseOTLPEndpoint`仅移除协议和尾部斜杠。OTLP/HTTP 端点通常包含路径（例如 `/example.com/example.com` `http://host:4318/v1/traces`），但 `/ example.com``WithEndpoint`期望的是纯 URL `host:port`，并将路径视为主机的一部分。这在 gRPC 中并不常见，但在 HTTP 中却很常见。建议`otlptracehttp.WithEndpointURL`为 HTTP 分支考虑使用 `/example.com/example.com`（接受包含路径的完整 URL），或者明确指出端点必须`host:port`仅包含路径。
> 
> **3. LLM span 覆盖范围存在缺口，与描述不符，描述** `llm.request`中span 仅包裹了主完成调用。位于 ` /``RunPerFile`内的 LLM 调用只有其阶段 span（`/` ），没有子 span，而位于 `/` 内的调用则完全没有 span。描述中提到“全面的 span 覆盖范围 / LLM 调用：llm.request span”——要么扩展覆盖范围，要么说明当前范围。`executePlanPhase``executeReviewFilter``plan.execute``review_filter.execute``llm.request``ReLocateComment``resolveAndCollect`
> 
> **4. 返回的 span 上下文会被丢弃（`_, span :=`）** 所有`StartToolSpan`/`StartLLMSpan`调用都会丢弃返回的上下文，因此嵌套操作（例如工具 span 下的重定位 LLM 调用）不会嵌套在其父级下。如果扁平的层次结构是故意设计的，那没问题，但这与“已确认的 span 层次结构”的说法略有矛盾。
> 
> **5. 文档** 文档清单项未选中。这将添加一个用户可见的环境变量（`OTEL_EXPORTER_OTLP_PROTOCOL`）和一个`trace_id`输出字段——值得与遥测文档/README同步（请注意项目约定，README 的更改必须反映到所有本地化版本）。
> 
> **6. 测试缺口** 新的测试涵盖了导出器路由和 span 辅助函数，但没有断言数据是否`trace_id`实际填充（输出测试直接通过`""`）。如果添加一个断言数据非空`trace_id`且 stderr 输出 TraceID 的测试，就能完善测试。

Thanks for the thorough review! All points addressed:

1. http/json fallback — Fixed. http/json now routes to the HTTP exporter alongside http/protobuf. Verified in SDK source (otlptracehttp@v1.44.0/client.go) that the HTTP transport only uses protobuf serialization regardless, so this gives users the HTTP transport they need. Unknown protocols still fall back to gRPC with a warning.

2. HTTP endpoint path — Per the [OTLP spec](https://opentelemetry.io/docs/specs/otlp/#otlphttp-request), OTEL_EXPORTER_OTLP_ENDPOINT is a base URL and the SDK appends /v1/traces automatically. Users should not include a path. Documented this in the Telemetry section of all 5 README versions.

3. LLM span coverage — Added llm.request child spans to executePlanPhase, executeReviewFilter, and ReLocateComment, with Usage nil-safety consistent with loop.go.

4. _, span := discarding ctx — Intentional. In llmloop/loop.go, each iteration's LLM/tool spans are siblings under main.loop, not a nested chain. In executePlanPhase/executeReviewFilter, the llm.request span does nest under the phase span (we pass the phase's ctx), but the LLM span's returned ctx is discarded since nothing further needs to nest under it.

5. Documentation — Updated the Telemetry section in all 5 README versions (protocol selection table + endpoint format). Checklist item checked.

6. Test gap — Added trace_id assertions to outputJSONWithWarnings, outputJSONNoFiles, and two end-to-end emitRunResult tests using a real TracerProvider to verify ctx → trace_id propagation.

internal/llmloop/loop.go
```diff
@@ -361,8 +375,12 @@ func (r *Runner) executeToolCall(ctx context.Context, newPath string, call llm.T
 			asyncCtx := context.WithoutCancel(ctx)
 			toolName := t.Name()
 			pool.Submit(func() ([]model.LlmComment, error) {
+				defer func() {
+					telemetry.RecordToolResult(toolSpan, toolName, time.Since(startTime).Milliseconds(), nil)
```
lizhengfeng101: **Q1 — async span always records success. Is this intentional?**

In this async `code_comment` path, `RecordToolResult` is always called with a `nil` error, so the `tool.execute` span is marked `ok` regardless of what happens inside `resolveAndCollect(asyncCtx)`. In contrast, the synchronous path passes the real `err` through. If the async resolution fails, the trace will still show the tool as successful.

Was this a deliberate trade-off (async errors aren't surfaced here anyway), or should the span reflect failures from `resolveAndCollect`?

internal/llmloop/loop.go
```diff
@@ -361,8 +375,12 @@ func (r *Runner) executeToolCall(ctx context.Context, newPath string, call llm.T
...
+					toolSpan.End()
+					telemetry.PrintToolCallFinished(toolName, time.Since(startTime))
```
lizhengfeng101: **Q2 — `time.Since(startTime)` evaluated twice. Intentional?**

The defer calls `time.Since(startTime)` once for `RecordToolResult` (line 379) and again for `PrintToolCallFinished` (line 381), yielding two slightly different durations for the same tool call. Was this intentional, or would a single `dur := time.Since(startTime)` shared by both be preferable for consistency?

internal/llmloop/loop.go
yingjiexu2002: Q1 — Intentional. resolveAndCollect doesn't return an error — the sync path also passes nil here. Both paths are consistent.

internal/llmloop/loop.go
yingjiexu2002: Q2 — Fixed.

lizhengfeng101 (approved):
LGTM

PR #319, opened 2026-07-08
docs: remove retired Go Report Card badge

lizhengfeng101:
## What

Removes the Go Report Card badge from `README.md` and all localized versions (`README.zh-CN.md`, `README.ja-JP.md`, `README.ko-KR.md`, `README.ru-RU.md`).

## Why

Go Report Card has been **sunset** by its maintainers. The badge endpoint now returns `go report: retired`, and the report page redirects to a farewell/"Thank you" notice recommending golangci-lint as a successor. As a result the badge can never render a grade — it just perpetually shows a "retired" state — so keeping it adds no value and looks broken.

PR #323, opened 2026-07-08
test: fix golangci-lint errcheck/staticcheck issues in test code

lizhengfeng101:
## What

Fixes all 54 golangci-lint findings in **test code** (production code untouched):

- **errcheck** — unchecked error returns. Setup-critical calls (repo init, `os.WriteFile`, `os.Chdir` into the target dir, `Shutdown`) now assert via `t.Fatalf`; cleanup / best-effort calls (`defer Close`, `os.Chmod`, `os.Unsetenv`, stdout-capture scaffolding) use an explicit `_ =`.
- **ST1018** — two Unicode control characters (U+009B/U+009D) in `output_test.go` replaced with `` / `` escapes.
- **SA6003** — `range []rune("abc")` simplified to `range "abc"` in `provider_tui_test.go`.
- **ineffassign** — ineffectual assignment in `util_test.go` (added an assertion so the appended copy is actually checked).

The bulk is `internal/llm/resolver_test.go` (36 `os.WriteFile` setup calls).

## Why

Go Report Card has been sunset (its badge was removed in #319). This is the **first step** of replacing it with [golangci-lint](https://golangci-lint.run/) as a real static-analysis gate. Clearing the existing lint findings first — starting with zero-risk test-only changes — keeps the eventual "wire up golangci-lint + CI + badge" change decoupled and easy to review. Production-code findings will follow in a separate PR.

## Verification

- `make check` (gofmt + vet) passes
- `make test` passes (all 22 packages, no failures)
- `golangci-lint run --max-issues-per-linter 0 --max-same-issues 0 ./...` reports zero remaining test-code issues

### DIFF
--- a/cmd/opencodereview/emit_run_result_test.go
+++ b/cmd/opencodereview/emit_run_result_test.go
@@ -181,7 +181,7 @@ func TestEmitRunResult_NilQuietHandle(t *testing.T) {
 
 func TestEmitRunResult_JSONTraceIDFromContext(t *testing.T) {
 	tp := sdktrace.NewTracerProvider()
-	defer tp.Shutdown(context.Background())
+	defer func() { _ = tp.Shutdown(context.Background()) }()
 	otel.SetTracerProvider(tp)
 
 	ctx, span := tp.Tracer("test").Start(context.Background(), "test-root")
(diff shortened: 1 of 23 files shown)


css521 (approved):
LGTM

PR #327, opened 2026-07-08
fix(background-file) Manage path at git repository root directory Ref…

### LINKED ISSUE #324: Inconsistent path resolution: --background-file resolves against CWD, not the repo root
## Summary

The `--background-file` (`-B`) option added in #206 resolves a **relative path against the current working directory (CWD)**, while the `file_read` tool resolves relative paths against the **git top-level directory** (fixed in c973e58 for monorepo support). These two behaviors are inconsistent.

## Details

In `cmd/opencodereview/background_file.go`, `loadBackgroundFile` uses the raw path directly:

```go
func loadBackgroundFile(path string) (string, error) {
    info, err := os.Stat(path)      // raw path
    ...
    raw, err := os.ReadFile(path)   // raw path
```

The caller in `cmd/opencodereview/review_cmd.go` passes `opts.backgroundFile` unmodified — there is no `filepath.Join(cc.RepoDir, path)` and no `filepath.Abs`. So:

- **Absolute paths** work as expected.
- **Relative paths** (e.g. the `./docs/my_business_context.md` shown in the README examples) are resolved against the shell's CWD, **not** the `--repo` directory.

## Why this is a problem

Commit c973e58 (`fix(tool): resolve file_read paths against git top-level in monorepos`) deliberately made `file_read` resolve relative paths against the git top-level. `--background-file` does not follow the same convention, so the two path semantics diverge.

Typical failure case:

```
ocr review --repo /path/to/repo -B ./docs/context.md
```

A user naturally expects `./docs` to be relative to the repo root, but it is actually relative to the current shell directory, which can lead to a confusing `read background file ... no such file or directory` error.

## Proposed options

1. **(Preferred)** Resolve a relative `--background-file` path against `cc.RepoDir` in `loadBackgroundFile` (or its caller), aligning it with `file_read` path semantics for a consistent tool-wide behavior.
2. Alternatively, keep CWD-relative resolution but explicitly document in the README / `--help` that the path is relative to the current working directory.

## References

- PR that introduced the option: #206
- Related path-resolution fix: c973e58

Victor-D:
## Description

fix(background-file) Manage path at git repository root directory

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Quick tests such as `./dist/opencodereview review --repo $PWD -B ./../../../../misc/ocr/mycontext.md` were done

## Checklist

- [x] I have signed the CLA

## Related Issues

closes #324
https://github.com/alibaba/open-code-review/issues/324

### DIFF
--- a/cmd/opencodereview/background_file.go
+++ b/cmd/opencodereview/background_file.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"unicode"
@@ -17,6 +18,13 @@ const (
 )
 
 var multiNewline = regexp.MustCompile(`\n{3,}`)
+
+func resolveBackgroundFilePath(repoDir, path string) string {
+	if path == "" || filepath.IsAbs(path) {
+		return path
+	}
+	return filepath.Join(repoDir, path)
+}
 
 // mergeBackground combines the inline --background value (or an auto-populated
 // commit message) with the content read from --background-file, separated by a
--- a/cmd/opencodereview/background_file_test.go
+++ b/cmd/opencodereview/background_file_test.go
@@ -25,6 +25,65 @@ func TestLoadBackgroundFileNotFound(t *testing.T) {
 	}
 }
 
+func TestResolveBackgroundFilePath(t *testing.T) {
+	repo := filepath.FromSlash("/path/to/repo")
+
+	t.Run("relative anchored at repo", func(t *testing.T) {
+		got := resolveBackgroundFilePath(repo, filepath.FromSlash("./docs/context.md"))
+		want := filepath.Join(repo, "docs", "context.md")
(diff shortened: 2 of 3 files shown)


lizhengfeng101 (commented):
Nice fix! The implementation is clean and minimal — `resolveBackgroundFilePath` handles the three cases (empty, absolute, relative) correctly, and the test coverage is solid with both unit tests and an end-to-end scenario.

One suggestion: could you add a test case for when `repoDir` is an empty string? This would happen if `--repo` is not specified and repo detection fails. In that case `filepath.Join("", relativePath)` returns the relative path as-is, which is a reasonable fallback to the old behavior — but it would be good to have an explicit test documenting that expectation.

Other than that, LGTM. Thanks for the contribution!

lizhengfeng101 (approved):
LGTM

PR #329, opened 2026-07-09
docs: remove broken Star History section from all READMEs

lizhengfeng101:
## What

Removes the `## Star History` section (heading + embedded star-history.com chart) from `README.md` and all localized versions (`README.zh-CN.md`, `README.ja-JP.md`, `README.ko-KR.md`, `README.ru-RU.md`).

## Why

GitHub restricted the stargazers API in July 2026 so that only a repository's own admins and collaborators can read the stargazers endpoint. The README embeds the chart as an anonymous image request:

```
https://api.star-history.com/svg?repos=alibaba/open-code-review&type=Date
```

That request carries no token (a token cannot be committed to a public README), so star-history.com can no longer fetch star data on refresh and the chart renders broken/blank for README viewers. Rather than embed a badge that no longer works, we remove the section. It can be reintroduced later as a static committed image if desired.

The Contributing section now flows directly into License.

### DIFF
--- a/README.ja-JP.md
+++ b/README.ja-JP.md
@@ -785,10 +785,6 @@ ocr config set telemetry.otlp_endpoint localhost:4317
   <img src="https://contrib.rocks/image?repo=alibaba/open-code-review" />
 </a>
 
-## Star History
-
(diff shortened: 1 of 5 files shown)


MuoDoo (approved):
LGTM

MuoDoo (approved):
lgtm

PR #328, opened 2026-07-08
fix(viewer): fix monospace font rendering in pre/code blocks on Windows

Kakarot30:
## Summary
Fixes monospace font rendering in `pre` and `code` blocks on Windows
(Chrome & Edge). Closes #325.
## Root Cause
Two issues combined to break monospace rendering on Windows:
1. **Missing global rule** — bare `<pre>` and `<code>` elements had no
   explicit `font-family` declaration, so they inherited the `body`'s
   sans-serif stack (`-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, sans-serif`)
   instead of the monospace `--mono` variable.
2. **Wrong font priority** — the `--mono` CSS variable led with
   `ui-monospace` and `SFMono-Regular`, which are macOS-native aliases
   that don't resolve on Windows. The browser skipped all the way to
   the generic `monospace` fallback, which renders poorly.
## Changes
**`internal/viewer/static/style.css`**
- Added a global `code, pre` rule immediately after the `body` block:
  ```css
  code, pre {
      font-family: var(--mono);
  }

PR #339, opened 2026-07-09
fix: remove redundant font-family from .response-text .inline-code

### LINKED ISSUE #336: chore(viewer): remove redundant font-family: var(--mono) on .inline-code
## Background

PR #328 added a global CSS rule in `internal/viewer/static/style.css`:

```css
code, pre {
    font-family: var(--mono);
}
```

This makes bare `<code>` and `<pre>` elements use the monospace font stack. As a result, some element-specific `font-family: var(--mono)` declarations on `<code>`/`<pre>` elements are now redundant and can be cleaned up.

## Scope (please keep it narrow!)

Remove **only** the redundant declaration on `.response-text .inline-code`, which is rendered as a `<code>` element (`<code class="inline-code">`, see `internal/viewer/templates/session.html`) and therefore already inherits `var(--mono)` from the new global rule:

```css
.response-text .inline-code {
    background: var(--inline-code-bg);
    padding: 0.15em 0.35em;
    border-radius: 4px;
    font-size: 0.88em;
    font-family: var(--mono);  /* <-- remove this line */
}
```

## Do NOT touch these

These look similar but are **not** redundant, because they apply to non-`code`/`pre` elements that the global rule does not match:

- `.file-path` → rendered as `<span>` (`session.html`)
- `.tool-name` → rendered as `<strong>` (`session.html`)

Deleting `font-family: var(--mono)` from those would break monospace rendering. Please leave them as-is.

## How to verify

Open a review session in the viewer and confirm inline code (single backticks in agent responses) still renders in a monospace font — behavior should be identical before and after the change.

## Good first issue notes

This is a tiny, low-risk CSS cleanup — a good way to get familiar with the contribution workflow. See `CONTRIBUTING`/README for how to build and run the viewer locally.

abhiramvsmg:
Removed the redundant `font-family: var(--mono);` declaration from `.response-text .inline-code` in style.css, since it's rendered as a `<code>` element and now inherits the same font from the new global `code, pre` rule added in PR #328.

  Left `.file-path` and `.tool-name` untouched — they render as `<span>` and `<strong>`, not `<code>`/`<pre>`, so the global rule doesn't apply to them and removing their `font-family` would break monospace rendering.

  Fixes #336

### DIFF
--- a/internal/viewer/static/style.css
+++ b/internal/viewer/static/style.css
@@ -418,7 +418,6 @@ h3 { margin: 1.5rem 0 0.75rem; color: var(--text-strong); }
     padding: 0.15em 0.35em;
     border-radius: 4px;
     font-size: 0.88em;
-   
 }
 .response-text .md-h1 { font-size: 1.15em; font-weight: 700; margin: 0.6em 0 0.3em; color: var(--text-strong); }
 .response-text .md-h2 { font-size: 1.05em; font-weight: 600; margin: 0.5em 0 0.25em; color: var(--text-strong); }
--- a/internal/viewer/static/style.css
+++ b/internal/viewer/static/style.css
@@ -418,7 +418,7 @@ h3 { margin: 1.5rem 0 0.75rem; color: var(--text-strong); }
     padding: 0.15em 0.35em;
     border-radius: 4px;
     font-size: 0.88em;
-    font-family: var(--mono);
+   
 }
 .response-text .md-h1 { font-size: 1.15em; font-weight: 700; margin: 0.6em 0 0.3em; color: var(--text-strong); }
 .response-text .md-h2 { font-size: 1.05em; font-weight: 600; margin: 0.5em 0 0.25em; color: var(--text-strong); }


lizhengfeng101:
Thanks for the cleanup! The reasoning is sound and I verified it end-to-end:

- ✅ #328 added the global `code, pre { font-family: var(--mono); }` rule.
- ✅ `.inline-code` is rendered as a `<code>` element (`session.html:180`), so it still inherits `var(--mono)` after the removal — no visual regression.
- ✅ Correctly left `.file-path` (`<span>`) and `.tool-name` (`<strong>`) untouched, since the global `code, pre` rule does not apply to them.

One thing to fix before merge:

The declaration wasn't deleted cleanly — it was replaced with a line of **trailing whitespace** (and the indentation is off), leaving a dangling blank line inside the block:

```css
.response-text .inline-code {
    background: var(--inline-code-bg);
    padding: 0.15em 0.35em;
    border-radius: 4px;
    font-size: 0.88em;
       /* <- leftover blank line with trailing spaces */
}
```

Please remove the whole line instead so the block ends cleanly at `font-size`. It won't fail CI (no `.editorconfig` / CSS lint here), but it adds noise to future diffs. Once that's cleaned up, this LGTM. 👍

lizhengfeng101:
Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/alibaba/open-code-review?pullRequest=339) before we can accept your contribution.

lizhengfeng101:
Looking forward to your update. @abhiramvsmg

lizhengfeng101 (approved):
LGTM

PR #345, opened 2026-07-09
docs(pages): split install and version commands into separate code blocks

lizhengfeng101:
## Summary
- Split `npm install` and `ocr version` into separate code blocks in the quickstart guide so each command can be copied independently.
- Applied the same change across all three language versions (en, ja, zh).

## Test plan
- [ ] Verify the quickstart page renders correctly with two separate code blocks.


xyJen (approved):
LGTM

PR #353, opened 2026-07-10
fix(llm): make OCR_LLM_EXTRA_HEADERS apply to all resolver strategies

lizhengfeng101:
## Summary

- **Move `OCR_LLM_EXTRA_HEADERS` parsing** from `tryOCREnv` into the global resolution loop so it applies to every resolver strategy (config-file providers, OCR env, legacy env, etc.), not just OCR environment variables.
- **Merge semantics**: environment headers are merged into whatever extra headers the winning strategy already provides; on key conflict the env value wins.
- **Add test** `TestResolveEndpoint_EnvExtraHeadersMergedWithConfigFile` to verify that config-file headers are preserved while env headers are merged on top.

## Test plan

- [x] `make check` passes (fmt + vet)
- [x] `make test` passes (all unit tests including the new merge test)

### DIFF
--- a/internal/llm/resolver.go
+++ b/internal/llm/resolver.go
@@ -94,6 +94,22 @@ func ResolveEndpointWithModelOverride(configPath, modelOverride string) (Resolve
 			if ok {
 				ep.Timeout = envTimeout
 			}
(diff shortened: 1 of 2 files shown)


css521 (approved):
LGTM

PR #355, opened 2026-07-10
feat(pages): add blog section with i18n support

lizhengfeng101:
## Summary

Add a blog feature to the pages site:

- **BlogPage component** — list view with tag filtering and search modal (⌘K), detail view with TOC sidebar and scroll-spy heading highlight
- **Blog content system** — markdown-based posts with frontmatter metadata, i18n support for en/ja/zh, centralized blog registry in `pages/src/content/blog/index.ts`
- **Navbar integration** — added Blog tab, improved active state detection to use `startsWith` matching for nested routes
- **MarkdownRenderer enhancement** — custom image renderer to handle absolute paths with correct public path resolution
- **Webpack config** — added `copy-webpack-plugin` to serve static assets from `pages/public/` directory

Includes an initial blog post "Introducing Open Code Review Blog" in all three languages.

## Review notes

Code review flagged a few items worth addressing in follow-up:
- `publicPath` calculation in MarkdownRenderer may produce incorrect paths under multi-level routes
- `searchBlog` slug source inconsistency with `getAllBlogMetas`
- Deprecated `navigator.platform` usage
- Heavy inline styles that could be moved to CSS


css521 (approved):
lgtm

PR #360, opened 2026-07-13
feat(mcp): support remote MCP servers via Streamable HTTP

### LINKED ISSUE #335: Support remote MCP servers
### Problem Statement

OCR landed MCP support and describe in the documentation as

> When to use it
>
> Reach for an MCP server when the reviewer would benefit from context that lives outside the diff:
>
> - Issue / ticket lookup — let the agent fetch the linked Jira / GitHub issue to check whether the change matches the stated requirement.

However, most issue trackers, including GitHub, provide this as a remote MCP and do not use local commands.

### Proposed Solution

It would be nice if OCR directly supported a built-in auth flow with `"type": "remote"` MCP.

It would be like:

```
# Some remote MCPs are public and don't require auth
ocr config set mcp_servers.cloudflare_docs.type remote
ocr config set mcp_servers.cloudflare_docs.url "https://docs.mcp.cloudflare.com/mcp"

ocr config set mcp_servers.github.type remote
ocr config set mcp_servers.github.url "https://api.githubcopilot.com/mcp/"

# Optionally, when using PAT
ocr config set mcp_servers.github.headers "Authorization=$GITHUB_PAT"

# Or it returns 401, OCR should trigger auth flow or at least prompt
```

### Alternatives Considered

Bridge commands like `mcp-remote` can be used, this increases the user's unnecessary configuration surface and attack vectors.

### Affected Area

Configuration

### Additional Context

_No response_

lizhengfeng101:
## Summary

Add support for remote (HTTP-based) MCP servers, complementing the existing stdio transport.

- Extend `MCPServerConfig` with `type`, `url`, and `headers` fields
- Add `NewRemoteClient` using go-sdk's `StreamableClientTransport`
- Support environment variable expansion in header values (`$TOKEN` → runtime value)
- Validate URL scheme (http/https only) and emit warnings for empty expanded headers
- Authentication stays outside OCR — users provide tokens via headers, matching CLI tool conventions

## Configuration example

```json
{
  "mcp_servers": {
    "my-remote": {
      "type": "remote",
      "url": "https://mcp.example.com/sse",
      "headers": {
        "Authorization": "Bearer $MCP_TOKEN"
      }
    }
  }
}
```

## Design decisions

- Auth is the user's responsibility (provide via headers/env vars), not OCR's — consistent with how stdio MCP servers handle tool installation/setup externally
- Uses `StreamableClientTransport` from go-sdk v1.6.1 (the current MCP spec's recommended HTTP transport)
- Header values support `$ENV_VAR` / `${ENV_VAR}` expansion via `os.Expand` at connect time
- Backward-compatible: existing stdio configs work unchanged (type defaults to "stdio")

Closes #335

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -199,13 +200,17 @@ type ProviderEntry struct {
 	ExtraHeaders map[string]string `json:"extra_headers,omitempty"`
 }
 
-// MCPServerConfig holds configuration for a single MCP server (stdio transport).
+// MCPServerConfig holds configuration for a single MCP server.
+// Type "stdio" (default) uses a subprocess; type "remote" uses Streamable HTTP.
 type MCPServerConfig struct {
-	Command string   `json:"command"`
-	Args    []string `json:"args,omitempty"`
-	Env     []string `json:"env,omitempty"`
-	Tools   []string `json:"tools,omitempty"`
-	Setup   string   `json:"setup,omitempty"`
+	Type    string            `json:"type,omitempty"` // "stdio" (default) or "remote"
+	Command string            `json:"command,omitempty"`
+	Args    []string          `json:"args,omitempty"`
+	Env     []string          `json:"env,omitempty"`
+	URL     string            `json:"url,omitempty"`
+	Headers map[string]string `json:"headers,omitempty"`
+	Tools   []string          `json:"tools,omitempty"`
+	Setup   string            `json:"setup,omitempty"`
 }
 
 // Config represents the user-level configuration file (~/.opencodereview/config.json).
@@ -402,7 +407,7 @@ func setConfigValue(cfg *Config, key, value string) error {
 		}
 		cfg.Llm.ExtraBody = m
 	default:
-		return fmt.Errorf("unknown config key: %s\nSupported keys: provider, model, providers.<name>.<field>, custom_providers.<name>.<field>, mcp_servers.<name>.<field>, llm.url, llm.auth_token, llm.auth_header, llm.model, llm.protocol, llm.use_anthropic, llm.extra_body, llm.extra_headers, language, telemetry.enabled, telemetry.exporter, telemetry.otlp_endpoint, telemetry.content_logging\nProvider fields: api_key, url, protocol, model, models, auth_header, extra_body, extra_headers\nProtocol values: anthropic, openai, openai-responses\nMCP server fields: command, args, env, tools, setup", key)
+		return fmt.Errorf("unknown config key: %s\nSupported keys: provider, model, providers.<name>.<field>, custom_providers.<name>.<field>, mcp_servers.<name>.<field>, llm.url, llm.auth_token, llm.auth_header, llm.model, llm.protocol, llm.use_anthropic, llm.extra_body, llm.extra_headers, language, telemetry.enabled, telemetry.exporter, telemetry.otlp_endpoint, telemetry.content_logging\nProvider fields: api_key, url, protocol, model, models, auth_header, extra_body, extra_headers\nProtocol values: anthropic, openai, openai-responses\nMCP server fields: type, command, args, env, url, headers, tools, setup", key)
 	}
 	return nil
 }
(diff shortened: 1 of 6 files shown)


cometkim:
> Support environment variable expansion in header values (`$TOKEN` → runtime value)

When calling config via the CLI, `$VALUE` expansion can be resolved accidentally by the user's shell.

lizhengfeng101:
Good catch! This is indeed a common CLI pitfall — the user's shell may expand `$TOKEN` before OCR sees it.

In practice, this is the same issue every CLI tool faces (`docker run -e`, `git config`, etc.). The mitigations are:

1. **Single quotes** prevent shell expansion: `ocr config set mcp_servers.foo.headers '{"Authorization": "Bearer $MCP_TOKEN"}'`
2. Most users will **edit the JSON config file directly**, where `$MCP_TOKEN` is stored literally and only expanded at runtime by `os.Expand`.

We'll add a note in the documentation to remind users to use single quotes (or edit the config file directly) when header values contain environment variable references. No syntax change needed — `$VAR` is the widely understood convention.

cometkim:
Then, do you think #332 should follow that convention too?

> `$VAR` is the widely understood convention.

Others (e.g. OpenCode) are often using `{env:VAR}` btw. (additionally `{cmd:...}`)

lizhengfeng101:
Thanks for raising the syntax consistency question — it's a good design consideration.

I looked into how OpenCode actually handles this. From their source code and documentation:

- MCP `headers` are plain string values with no template expansion ([config.go](https://github.com/opencode-ai/opencode/blob/main/internal/config/config.go) — `Headers map[string]string`, passed directly via `client.WithHeaders`)
- Custom commands use `$NAME` placeholders (shell-style): `$ISSUE_NUMBER`, `$AUTHOR_NAME`, etc.
- Environment variables are resolved via standard `os.Getenv` calls, not a custom `{env:}` parser

So in practice, OpenCode also follows the conventional `$VAR` pattern rather than a `{env:VAR}` syntax.

Regarding consistency with #332's `{ocr_session_key}`: these serve different semantic purposes — `$VAR` means "substitute an OS environment variable at runtime" (same as Docker Compose, GitHub Actions env, Kubernetes manifests), while `{ocr_session_key}` is an OCR-internal computed value. Many tools maintain this distinction (e.g., GitHub Actions uses `${{ }}` for expressions vs `$VAR` in shell steps). We think keeping them separate actually makes the intent clearer rather than forcing both into one syntax.

That said, the shell-expansion footgun you raised is a real concern and we've addressed it with documentation tips (single quotes for CLI, or edit config.json directly). Happy to hear if you have further thoughts on this.

lizhengfeng101:
**Correction to my previous comment** — I referenced the wrong OpenCode repository (the archived `opencode-ai/opencode`). The actively maintained one is [`anomalyco/opencode`](https://github.com/anomalyco/opencode).

@cometkim is correct: anomalyco/opencode does use `{env:VAR}` syntax. Their [`packages/opencode/src/config/variable.ts`](https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/config/variable.ts) implements:

```typescript
/** Apply {env:VAR} and {file:path} substitutions to config text. */
export async function substitute(input: SubstituteInput) {
  let text = input.text.replace(/\{env:([^}]+)\}/g, (_, varName) => {
    return (input.env?.[varName] ?? process.env[varName]) || ""
  })
  // ... also supports {file:path} for reading secrets from files
}
```

So OpenCode supports `{env:VAR}` and `{file:path}` as a unified template system, applied to both URLs and headers in MCP config.

That said, our position remains to use `$VAR` for this PR, for these reasons:

1. **Broader industry convention** — Docker Compose, Kubernetes, GitHub Actions, and most CLI tools use `$VAR`/`${VAR}` for environment variable substitution. OpenCode is one data point, but the wider ecosystem leans toward shell-style syntax.
2. **Semantic separation** — `$VAR` (OS env var) and `{ocr_session_key}` (#332, OCR-internal computed value) are different concepts. Keeping distinct syntax makes the intent immediately clear.
3. **Simplicity** — Go's `os.Expand` handles `$VAR`/`${VAR}` natively. A custom `{env:}` parser adds code and edge cases (escaping `}` in values, etc.) for limited practical benefit.

The shell-expansion footgun is real, but well-mitigated by documentation (single quotes) and by the fact that most users edit `config.json` directly rather than using the CLI for header values.

If the project later decides to support richer substitution (like `{file:path}` for reading tokens from secret files), we can introduce that as a separate enhancement without breaking `$VAR` compatibility. Apologies for the confusion in my earlier comment.

css521 (approved):
LGTM

PR #363, opened 2026-07-14
feat: Add OpenAI Responses API support and refactor protocol handling

stay-foolish-forever:
## Description

This PR adds support for the **OpenAI Responses API** (`/v1/responses`) as a third LLM protocol, enabling code review with GPT-5.x / o-series models that use the Responses API rather than Chat Completions.

**New protocol: `openai-responses`**

Sits alongside the existing `anthropic` and `openai` (Chat Completions) protocols. Users can now select between all three via configuration.

- New `OpenAIResponsesClient` (stateless replay):
  - System messages are mapped to `instructions`; tool calls become `function_call` items keyed by `CallID` so the agent loop pairs results correctly.
  - `store=false` (privacy); `prompt_cache_key` derived from `sha256(instructions)[:32]`.
- Protocol handling refactor (backward-compatible):
  - Introduces `NormalizeProtocol` / `ValidateProtocol` helpers in `internal/llm/protocol.go`.
  - The registry uses canonical protocol constants; the resolver normalizes inputs everywhere.
  - `llm.protocol` config field + `OCR_LLM_PROTOCOL` env var take priority over the legacy `use_anthropic` / `OCR_USE_ANTHROPIC` (kept working via mirroring).
  - `anthropic-vertex` is rejected with a friendly "not yet implemented" message.
- TUI / CLI plumbing: the provider TUI exposes all three protocols (Custom & Manual modes); `provider_tui_test.go` added.
- Docs: protocol reference, config examples, env var table, and Responses API notes (store=false caching caveat, cache key derivation) updated across en/zh-CN/ko-KR/ja-JP/ru-RU READMEs.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Tested using anthropic, openai, openai-responses endpoints to ensure that all the three protocol functions correctly and previous behavior do not change

## Checklist

- [x] I have signed the CLA


internal/llm/responses_client.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	params := c.buildResponsesParams(model, req)
+
+	var opts []openaiopt.RequestOption
+	for k, v := range c.cfg.ExtraBody {
+		opts = append(opts, openaiopt.WithJSONSet(k, v))
+	}
+
+	sdkResp, err := c.sdk.Responses.New(ctx, params, opts...)
```
MuoDoo, line 95: When existing provider config contains extra_body: {"stream": true},these options are still passed to Responses.New.The API then returns SSE while New expects a regular JSON response, so every completion fails.

internal/llm/responses_client.go
```diff
@@ -0,0 +1,275 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return c.mapResponsesResponse(sdkResp), nil
```
MuoDoo, line 113: The Responses API can return HTTP 200 with status failed/cancelled, or queued/in_progress when background mode is enabled, so Responses.New can have a nil Go error here. This code converts those into a normal ChatResponse; callers do not inspect FinishReason, causing ocr llm test to report success and the review loop to retry or execute output instead of surfacing or polling the response.

lizhengfeng101 (approved):
LGTM

PR #376, opened 2026-07-15
fix(diff): add --end-of-options guard and no-commit regression test for workspace diff

### LINKED ISSUE #374: refactor(diff): normalize git argument ordering in workspaceTrackedDiff
## Description

In `internal/diff/git.go`, the `workspaceTrackedDiff` function places the `HEAD` positional argument between option flags (`--dst-prefix=b/` and `--no-color`), which is inconsistent with the argument ordering used in other git command invocations in the same file.

**Current** (line ~264):
```go
p.runGit(ctx, "-c", "core.quotepath=false", "diff", "--no-ext-diff", "--no-textconv", "--find-renames", "--src-prefix=a/", "--dst-prefix=b/", "HEAD", "--no-color", "-U"+fmt.Sprint(DiffContextLines), "--")
```

**Expected**: Move `HEAD` after all option flags (before `--`), consistent with how other diff commands in the file are structured — options first, then positional refs, then `--`.

## Scope

- File: `internal/diff/git.go`
- Function: `workspaceTrackedDiff`
- Two `runGit` calls need reordering (the unstaged and staged diff commands)

## Acceptance Criteria

- [ ] `HEAD` (and `--staged` in the second call) are placed after all `--flag` options and before `--`
- [ ] `make test` passes
- [ ] `make check` passes

## Context

Discovered during review of #365. This is a cosmetic/consistency improvement — the current code works correctly since git is flexible about argument ordering, but uniform style makes the codebase easier to maintain.

chethanuk:
## Description

Follow-up to #381, which normalized the git argument ordering in `workspaceTrackedDiff` (`internal/diff/git.go`) but stopped short of the `--end-of-options` guard. This PR adds the remaining pieces:

- **`--end-of-options` before `HEAD`** in the first `runGit` call, bringing workspace mode fully in line with the canonical range/commit/merge-base call sites in the same file, which already use the guard. After this change, every call site that takes a positional ref carries it; the `--staged` fallback and `ls-files` calls have no positional ref, so they correctly omit it. `--end-of-options` requires git ≥ 2.24 (Nov 2019), well within the project's documented Git ≥ 2.41 floor.
- **A clarifying comment** on why the `--staged` fallback is load-bearing: in a repository with no commits there is no `HEAD`, so `git diff HEAD` fails, but `git diff --staged` still surfaces staged changes by diffing the index against the empty tree.
- **`TestWorkspaceDiffNoCommitsUsesStagedFallback`**, a regression test that stages a file in a commit-less repository and asserts the workspace diff surfaces it — pinning the fallback so it can't be "simplified" away. The test sets no commit-related git config (`user.email`/`user.name`/`commit.gpgsign`), since it never commits.

## Type of Change

## How Has This Been Tested?

`go test -race ./...` — 2001 tests across 23 packages pass. The new regression test fails if the `--staged` fallback is removed and passes with the `--end-of-options` guard in place. `make check` (tidy + fmt + vet) passes.

## Checklist

- [ ] I have signed the CLA

## Related Issues

Follow-up to #381. Closes #374

### DIFF
--- a/internal/diff/git.go
+++ b/internal/diff/git.go
@@ -261,13 +261,18 @@ func (p *Provider) computeMergeBase(ctx context.Context, from, to string) string
 }
 
 func (p *Provider) workspaceTrackedDiff(ctx context.Context) (string, error) {
-	out, err := p.runGit(ctx, "-c", "core.quotepath=false", "diff", "--no-ext-diff", "--no-textconv", "--find-renames", "--src-prefix=a/", "--dst-prefix=b/", "--no-color", "-U"+fmt.Sprint(DiffContextLines), "HEAD", "--")
+	out, err := p.runGit(ctx, "-c", "core.quotepath=false", "diff", "--no-ext-diff", "--no-textconv", "--find-renames", "--src-prefix=a/", "--dst-prefix=b/", "--no-color", "-U"+fmt.Sprint(DiffContextLines), "--end-of-options", "HEAD", "--")
 	if err == nil && out != "" {
 		return out, nil
 	}
 	if ctx.Err() != nil {
 		return "", ctx.Err()
 	}
+	// Fall back to the staged diff when `git diff HEAD` errored or was empty. This is
+	// not redundant with the call above: in a repository with no commits yet there is no
+	// HEAD, so `git diff HEAD` fails with "bad revision 'HEAD'", but `git diff --staged`
+	// still surfaces staged changes by diffing the index against the empty tree — the only
+	// way to review a workspace before its first commit.
 	return p.runGit(ctx, "-c", "core.quotepath=false", "diff", "--no-ext-diff", "--no-textconv", "--find-renames", "--src-prefix=a/", "--dst-prefix=b/", "--no-color", "-U"+fmt.Sprint(DiffContextLines), "--staged", "--")
 }
 
--- a/internal/diff/git_test.go
+++ b/internal/diff/git_test.go
@@ -205,6 +205,36 @@ func TestWorkspaceDiffSurvivesExternalDiffTool(t *testing.T) {
 	}
 }
 
+// TestWorkspaceDiffNoCommitsUsesStagedFallback pins the second runGit call in
+// workspaceTrackedDiff: in a repository with no commits there is no HEAD, so
+// `git diff HEAD` fails, and the staged diff is the only way to review the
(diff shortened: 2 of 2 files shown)


lizhengfeng101:
Hi @chethanuk, thanks for the thorough work here — especially the `--end-of-options` alignment and the no-commit regression test. 🙏

Heads up on some overlap: #374 was also picked up in #381, which was assigned the issue first and has now been **merged to `main`**. That PR did the argument reordering (options → positional ref → `--`), so the core normalization from #374 is already in.

The good news is your PR still has real, unique value that #381 did **not** include:

1. **`--end-of-options` before `HEAD`** — the piece that actually brings `workspaceTrackedDiff` fully in line with the canonical calls at `git.go:119/126/256`. #381 stopped short of this.
2. **The clarifying comment** on why the `--staged` fallback is load-bearing (no-commit repos).
3. **`TestWorkspaceDiffNoCommitsUsesStagedFallback`** — genuinely valuable coverage that pins the fallback so it can't be "simplified" away.

Would you be up for **rebasing onto the latest `main` and re-framing this PR as a follow-up** — something like *"add `--end-of-options` guard + no-commit regression test for workspace diff"*? After the rebase the conflict collapses to a tiny delta:

- add `"--end-of-options"` before `"HEAD"` in the first `runGit` call,
- keep the fallback comment above the second call,
- keep the new test in `git_test.go` (that file shouldn't conflict at all).

That way your contribution — particularly the test — lands cleanly and keeps its credit. Also a reminder to sign the CLA when you get a chance (the checkbox is still unchecked). Thanks again for pushing on the details here! 🚀

chethanuk:
Hi @lizhengfeng101, thanks for the good guidance — done exactly as you suggested.

Rebased onto the latest main and re-framed the PR as a follow-up to #381: the delta is now just the --end-of-options guard before HEAD in the first runGit call, the comment explaining why the --staged fallback is load-bearing, and TestWorkspaceDiffNoCommitsUsesStagedFallback. git_test.go applied cleanly as you predicted. Title and description updated to match, and the CLA is signed.

One small thing I noticed while re-testing (pre-existing on main, so I left it out of this PR): the fallback condition err == nil && out != "" also triggers when git diff HEAD is legitimately empty — e.g. stage a change, then revert the worktree back to HEAD — in which case the review reports the staged diff even though the worktree matches HEAD.

lizhengfeng101 (approved):
LGTM

PR #378, opened 2026-07-15
fix(scan): check user include patterns before the extension allowlist

chethanuk:
## Description

In the full-scan path, user `--include` patterns were evaluated **after** the built-in extension allowlist, so a file with a non-allowlisted extension could never be force-included into a scan — even when the user explicitly asked for it. The preview path already evaluates user includes first, so `ocr preview` and the actual scan disagreed about which files a given `--include` selects.

This change moves the user-include check ahead of the extension-allowlist check in the scan path (`internal/scan/agent.go`), matching the preview path's decision order. Explicit user intent now wins over the built-in default in both places.

This also provides the "let users extend the allowlist themselves" escape hatch discussed in #371: `--include "**/*.ftl"` (or any other extension) now works for scans without a code change.

## Type of Change

## How Has This Been Tested?

Coverage tests extended: a non-allowlisted extension matched by a user `--include` is now included in the scan (this case fails on the previous ordering), and existing exclusion behavior without `--include` is unchanged. `make check` (tidy + fmt + vet) passes.

## Checklist

- [ ] I have signed the CLA

## Related Issues

Related to #371 (complements the FreeMarker rules PR; makes the allowlist user-extensible via `--include`)

### DIFF
--- a/internal/scan/agent.go
+++ b/internal/scan/agent.go
@@ -349,12 +349,15 @@ func (a *Agent) whyExcluded(it model.ScanItem) model.ExcludeReason {
 	if a.args.FileFilter != nil && a.args.FileFilter.IsUserExcluded(path) {
 		return model.ExcludeUserRule
 	}
+	// User-include is checked before the extension allowlist (matching the
+	// preview/diff path in internal/agent) so an explicit include glob can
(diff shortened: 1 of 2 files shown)


lizhengfeng101 (approved):
LGTM

PR #391, opened 2026-07-17
docs(pages): simplify claude-code prerequisites, recommend delegation mode

lizhengfeng101:
## Summary

- Remove redundant auto-install hints from the Claude Code integration page — the command transparently installs `ocr` when needed, so there's no reason to call it out as a prerequisite.
- Replace the LLM prerequisite warning with a friendlier tip recommending Delegation Mode for users who don't want to configure a model themselves.
- Applied consistently across all three locale versions (en, zh, ja).


xyJen (approved):
LGTM

PR #392, opened 2026-07-17
docs(pages): note that LLM API key is optional in delegation mode

lizhengfeng101:
## Summary

- Add a parenthetical note to the quickstart prerequisites clarifying that an LLM API key is not needed when using Delegation Mode.
- Add a tip blockquote before Step 2 telling delegation-mode users (e.g. running inside Claude Code) they can skip Steps 2–3 and go straight to Step 4.
- Applied across all three locale versions (en, zh, ja).


xyJen (approved):
LGTM

PR #407, opened 2026-07-20
feat(pages): serve landing site at custom-domain root with clean URLs

lizhengfeng101:
## What

Prepares the landing site (`pages/`) to be served at the **root of the `open-codereview.ai` custom domain** and drops the `/#/` from URLs.

- **`webpack.config.js`**: `publicPath` → `/` so assets load at the domain root (previously hard-coded to the `/open-code-review/` GitHub Pages subpath). Adds a second `HtmlWebpackPlugin` that emits `404.html` (a copy of `index.html`) as an SPA deep-link fallback.
- **`src/index.tsx`**: `HashRouter` → `BrowserRouter` for clean paths (e.g. `/docs` instead of `/#/docs`).
- **`src/components/HeroSection.tsx`**: replace the hard-coded `href="#/docs"` anchor with a router `<Link to="/docs">` (the in-page `#quickstart` anchor is intentionally kept).
- **`public/CNAME`**: add `open-codereview.ai` so GitHub Pages serves the site at the custom-domain root.

## Why

The site is moving from `alibaba.github.io/open-code-review/` to the short custom domain `open-codereview.ai`. Serving at a subpath forces the ugly `/open-code-review/` asset prefix and the `HashRouter` `/#/` URLs. This change makes the build target the domain root and produces clean, shareable URLs.

## Release note ⚠️

Merging this to `main` triggers the Pages deploy, which activates the `CNAME` and makes GitHub Pages claim `open-codereview.ai` (and redirect the old subpath). **This must be merged in coordination with the DNS cutover** — if DNS is not yet pointing the domain at GitHub Pages, the site will be unreachable until it is.

Client-side routing (`BrowserRouter`) relies on the `404.html` fallback on plain GitHub Pages; once the site moves behind a CDN, the fallback should be handled by the CDN (404 → `index.html`, HTTP 200).

## Verification

- `npm run build` succeeds; `dist/index.html` references assets at root (`/bundle.js`), `dist/404.html` matches `index.html`, and `dist/CNAME` contains `open-codereview.ai`.
- `ocr review` on the diff: 0 comments.

PR #413, opened 2026-07-21
fix(pages): scroll docs deep links on load

### LINKED ISSUE #412: fix(pages): DocsPage does not scroll to in-page anchor on initial load / deep link
## Description

`DocsPage` handles in-page anchor scrolling **only** on click (via `handleContentClick` and `scrollToHeading`). It never reads `location.hash` when the page first mounts, so **deep-linking to a heading does not scroll to it**.

Open a URL such as `.../#/docs/faq#超时` (or `.../#/docs/configuration#timeouts`) directly in a fresh tab: the correct doc renders, but the viewport stays at the top instead of scrolling to the `超时` / `Timeouts` heading. Clicking the same link from within the app works, because that path goes through the click handler.

Note the site uses `HashRouter` (`pages/src/index.tsx`), so the routed URL is `#/docs/faq#超时` — react-router parses this into `pathname: "/docs/faq"` and `hash: "#超时"`. A fix should read the fragment from `useLocation().hash`, not from `window.location.hash` directly.

Current behavior:

- Click an in-page link inside the app → scrolls ✅ (`handleContentClick`, `DocsPage.tsx:199-233`)
- Load/refresh a URL that already contains a heading fragment → no scroll ❌ (nothing reads the hash on mount)

Expected behavior:

- On initial mount (and whenever `location.hash` changes), scroll the matching heading into view, mirroring the existing click behavior.

## Scope

- File: `pages/src/pages/DocsPage.tsx`
- Area: add an effect that runs after the doc content renders, reads `useLocation().hash`, decodes it, and scrolls the target heading into view.
- Reuse the existing `decodeFragment` helper (already in `DocsPage.tsx`) — `marked` percent-encodes non-ASCII fragments while heading ids are raw text from `generateHeadingId`, so the fragment must be decoded before `getElementById`.
- Headings are rendered asynchronously from markdown, so the lookup likely needs the same `requestAnimationFrame` retry pattern already used in the cross-page anchor path (`DocsPage.tsx:222-230`).

## Acceptance Criteria

- [ ] Loading `.../#/docs/faq#超时` (zh), `.../#/docs/faq#no-tool-calls-parsed-local-models-ollama` (en), and `.../#/docs/configuration#timeouts` directly scrolls to the target heading.
- [ ] Works for CJK fragments (decoded via `decodeFragment`).
- [ ] Existing in-app click-to-scroll behavior is unchanged.
- [ ] No scroll/console error when the fragment is empty or does not match any heading.

## Context

Discovered during review of #400, which added CJK cross-links to the docs and fixed click-time anchor decoding (`decodeFragment`). The mount-time / deep-link case was explicitly noted as out of scope there and is filed here as a follow-up. The `decodeFragment` helper introduced by #400 should be reused for consistency.

fjbarrett:
## Description

Direct docs links now scroll to their fragment after the rendered markdown heading is available. The effect reads React Router's location hash, decodes percent-encoded fragments, retries briefly while content renders, and cancels stale retries when navigation changes.

Since #407 switched the site to BrowserRouter, verification used the equivalent clean URLs under `/docs/...`.

AI assistance: OpenAI Codex helped implement and verify this change. I reviewed the full diff and take responsibility for it.

## Type of Change

- [ ] Breaking change (fix or feature that causes existing functionality to change)
- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] Manual testing:
  - `make check`
  - `cd pages && npm run typecheck`
  - `cd pages && npm run build`
  - Playwright: direct English fragment
  - Playwright: encoded CJK fragment
  - Playwright: fragment change after mount
  - Playwright: unknown fragment produces no error
  - Playwright: existing cross-page click still scrolls

## Screenshots

Before, a directly loaded fragment remains at the top of the FAQ:

Before

After, the same URL scrolls to the requested FAQ section:

After

## Checklist

- [ ] I have added tests that prove my fix is effective or my feature works (the frontend currently has no automated test runner; covered with Playwright browser verification above)
- [x] I have signed the CLA

## Related Issues

Closes #412

### DIFF
--- a/pages/src/pages/DocsPage.tsx
+++ b/pages/src/pages/DocsPage.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useEffect, useMemo, useCallback, useRef } from 'react';
-import { useParams, useNavigate } from 'react-router-dom';
+import { useParams, useNavigate, useLocation } from 'react-router-dom';
 import { useTranslation } from '../i18n';
 import Navbar from '../components/Navbar';
 import Footer from '../components/Footer';
@@ -19,6 +19,28 @@ function decodeFragment(fragment: string): string {
   } catch {
     return fragment;
   }
+}
+
+// Markdown renders a frame or more after navigation, so a fragment's heading
+// may not exist yet. Retry across a few frames; the returned canceller stops a
+// stale chain when navigation moves on.
+function scrollToFragmentWhenReady(id: string): () => void {
+  let frame = 0;
+  let cancelled = false;
+  const tryScroll = (attempts: number) => {
+    if (cancelled) return;
+    const el = document.getElementById(id);
+    if (el) {
+      el.scrollIntoView({ behavior: 'smooth', block: 'start' });
+    } else if (attempts < 10) {
+      frame = requestAnimationFrame(() => tryScroll(attempts + 1));
+    }
+  };
+  frame = requestAnimationFrame(() => tryScroll(0));
+  return () => {
+    cancelled = true;
+    cancelAnimationFrame(frame);
+  };
 }
 
 /* ─── Sidebar tree data ─── */
@@ -138,6 +160,7 @@ if (process.env.NODE_ENV !== 'production' && validSlugs.size !== flatDocList.len
 const DocsPage: React.FC = () => {
   const { slug: slugParam } = useParams<{ slug?: string }>();
   const navigate = useNavigate();
+  const { hash } = useLocation();
   /* Active doc slug is derived from the URL param, falling back to quickstart */
   const activeSlug: DocSlug =
     slugParam && validSlugs.has(slugParam as DocSlug) ? (slugParam as DocSlug) : 'quickstart';
@@ -148,6 +171,8 @@ const DocsPage: React.FC = () => {
   const [searchQuery, setSearchQuery] = useState('');
   const [searchSelectedIdx, setSearchSelectedIdx] = useState(0);
   const searchInputRef = useRef<HTMLInputElement>(null);
+  /* Cancels an in-flight click-triggered scroll when a newer one starts */
+  const cancelPendingScroll = useRef<(() => void) | null>(null);
   const { t, language } = useTranslation();
   const { isMobile } = useResponsive();
   const contentRef = React.useRef<HTMLDivElement>(null);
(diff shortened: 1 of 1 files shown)


lizhengfeng101:
Thanks for the fix — the core deep-link scrolling logic is correct and the retry/cancel handling is solid. A few things to address before merging:

## 1. `HISTORY.md` should not be committed

This file is a Codex agent workflow log ("Forked the repository and claimed issue #412", etc.). It's agent-internal scratch state and unrelated to the codebase — this file didn't exist in the repo before. Please remove it from the PR.

## 2. Screenshots should live in the PR description, not the repo

The two PNGs under `screenshots/` are acceptance evidence for this UI change. That's exactly what the PR description is for — and you've already embedded them there via `raw.githubusercontent.com`. Committing the binaries into the repo is redundant and bloats history. Please drop the `screenshots/` additions and keep them as links in the PR body only.

## 3. De-duplicate the `tryScroll` retry logic

The new `useEffect` reimplements almost the same rAF retry loop that already exists in `handleContentClick` (the `tryScroll(attempts)` at ~L233-241). Please extract a single shared helper so both paths use one implementation, e.g.:

```ts
function scrollToFragmentWhenReady(id: string): () => void {
  let frame = 0;
  let cancelled = false;
  const tryScroll = (attempts: number) => {
    if (cancelled) return;
    const el = document.getElementById(id);
    if (el) {
      el.scrollIntoView({ behavior: 'smooth', block: 'start' });
    } else if (attempts < 10) {
      frame = requestAnimationFrame(() => tryScroll(attempts + 1));
    }
  };
  frame = requestAnimationFrame(() => tryScroll(0));
  return () => {
    cancelled = true;
    cancelAnimationFrame(frame);
  };
}
```

The `useEffect` can then just `return scrollToFragmentWhenReady(decodeFragment(fragment))`, and `handleContentClick` can reuse it too (which also gives that path the cancellation it currently lacks).

Once these are addressed this looks good to go. Thanks!

fjbarrett:
Thanks for the careful review. All three items are addressed in `714bb79`:

1. Removed `HISTORY.md` from the PR.
2. Removed `screenshots/` from repository history. The PR description still contains the before/after evidence, hosted from the separate immutable asset commit `aca0fd1`.
3. Extracted the retry loop into `scrollToFragmentWhenReady`. Both initial deep-link scrolling and cross-page fragment clicks now use the shared helper, including cancellation of stale retries.

The PR now changes only `pages/src/pages/DocsPage.tsx` (`+35/-10`).

Verified:

- `tsc --noEmit`
- `npm run build`

lizhengfeng101 (approved):
LGTM

PR #430, opened 2026-07-21
docs(ci): pass untrusted PR fields via env to prevent script injection

chethanuk:
## Description

The CI integration docs teach a snippet that interpolates three attacker-controlled PR fields
directly inside a shell `run:` block:

```yaml
- name: Run OCR review
  run: |
    ocr review \
      --background "${{ github.event.pull_request.title }}" \
      --from "origin/${{ github.base_ref }}" \
      --to "origin/${{ github.head_ref }}" \
      --format json --audience agent
```

`${{ }}` is textual substitution performed *before* the shell parses the line, so the quotes in
the rendered command belong to the attacker. Anyone who copies this snippet, which is what the
page tells them to do, gets arbitrary command execution on their runner from a PR title.

This moves all three fields into `env:` and references them as shell variables, plus a short
note.

This is already the repo's own rule. `internal/config/rules/rule_docs/github_workflows.md:6`:

> **Script injection**: Expressions like `${{ github.event.issue.title }}` used directly in
> `run:` blocks enable code injection. These must be passed through environment variables instead

`action.yml` complies (all ten inputs hoisted at `:228-238`), and the GitLab half of this same
page already uses `"$CI_MERGE_REQUEST_TITLE"`. The GitHub Actions section was the lone outlier.

## Verification

A live run of the attack: two jobs differing only in how the title reaches the shell, one PR
titled `demo"; echo PWNED_INJECTION_SUCCEEDED; #`.

Run [`29827744839`](https://github.com/chethanuk/open-code-review/actions/runs/29827744839). The
`vulnerable` job's rendered command, as GitHub logged it:

```
##[group]Run echo "background=demo"; echo PWNED_INJECTION_SUCCEEDED; #"
```

The payload became a second statement and executed. In the `safe` job the identical title printed
as literal text. The harness step was an `echo "background=..."`, not a real `ocr review`, and the
injected payload was itself a bare `echo`; the test PR and branch are deleted.

Branch names are injectable too, not just the title: `a";id;"b`, `a$(id)b` and `a;id;b` are all
valid per `git check-ref-format`, and a fork PR lets the attacker choose the branch name. There
is also a non-malicious case: `a$HOMEb` is a legal branch name that the old form silently
expands, building the wrong ref. So this is a correctness fix as well as a security one.

Before: 21 `${{ github.* }}` expressions inside `run:` blocks. After: 0, with all 21 surviving as
`env:` assignments. The three YAML blocks were byte-identical across locales before and still are.

## Limitations

Docs-only, so nothing in CI renders or lints these snippets and the `test` check does not
exercise this change. The reproduction ran on a GitHub-hosted runner in a fork, not this repo's
`self-hosted` pool, though the mechanism is GitHub's expression substitution and happens before
any runner-specific behaviour.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] Reproduced the injection end to end on a real runner and captured the log line
- [x] Confirmed the `env:` form treats the same payload as literal data
- [x] Asserted 21 → 0 `${{ github.* }}` expressions inside `run:` blocks
- [x] Confirmed the three locales remain byte-identical

## Checklist

- [ ] I have added tests that prove my fix is effective — n/a, documentation
- [x] I have signed the CLA

### DIFF
--- a/pages/src/content/docs/en/integrations/ci.md
+++ b/pages/src/content/docs/en/integrations/ci.md
@@ -120,13 +120,23 @@ semantic convention like `feat(auth): add OAuth2 support`):
 
 ```yaml
 - name: Run OCR review
+  env:
+    PR_TITLE: ${{ github.event.pull_request.title }}
+    BASE_REF: ${{ github.base_ref }}
+    HEAD_REF: ${{ github.head_ref }}
   run: |
     ocr review \
-      --background "${{ github.event.pull_request.title }}" \
-      --from "origin/${{ github.base_ref }}" \
-      --to "origin/${{ github.head_ref }}" \
+      --background "$PR_TITLE" \
+      --from "origin/$BASE_REF" \
+      --to "origin/$HEAD_REF" \
       --format json --audience agent
 ```
+
+Pass PR-controlled values through `env:` rather than
+interpolating `${{ }}` directly into `run:`. GitHub substitutes
+`${{ }}` textually *before* the shell parses the line, so a PR
+title or branch name containing shell metacharacters would
+execute on your runner.
 
 #### Custom rules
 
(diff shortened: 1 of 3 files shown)


lizhengfeng101 (approved):
LGTM！

PR #449, opened 2026-07-22
fix(llmloop): drain per-file comment work without racing pool submissions

hobostay:
## Description

Fixes a `sync.WaitGroup` race in the concurrent review path: `executeSubtask` calls the **pool-wide** `CommentWorkerPool.Await()` from inside each per-file goroutine (before running the review filter), while other files' tool-use loops are still calling `pool.Submit()` for their own `code_comment` work.

The pool's own doc states the contract this violates:

> Concurrency contract: Await must not run concurrently with Submit. Submit calls wg.Go (which does wg.Add(1) synchronously), so a Submit racing Await would risk sync.WaitGroup's "Add called concurrently with Wait" panic.

`Await`/`Submit` share one `sync.WaitGroup`, so when file A's goroutine is blocked in `Await()` and file B's loop calls `Submit()`, the runtime can panic with `sync: WaitGroup misuse: Add called concurrently with Wait` or `sync: WaitGroup is reused before previous Wait has returned` (a `wg.Go` → `Add(1)` landing while `Wait` is blocked, or in the wake-up window after the counter reaches 0). The panic is recovered by the subtask wrapper in `dispatchSubtasks`, so the observable symptom is a **spurious per-file failure** (`subtask_error`, counted toward the "all N file review(s) failed" rollup); depending on the interleaving the counter can also be left corrupted so the final `Await` never returns. Every multi-file `ocr review` exercises this pattern — the pool is always installed and `code_comment` always takes the async path.

## Fix

Add `SubmitFor(key, f)` / `AwaitKey(key)` so a caller can drain only the units submitted under one key:

- `internal/llmloop/pool.go` — per-key `sync.WaitGroup` map alongside the pool-wide one; `Submit` behavior unchanged.
- `internal/llmloop/loop.go` — async `code_comment` work is submitted under the reviewed file's path.
- `internal/agent/agent.go` — the pre-review-filter drain uses `AwaitKey(newPath)`.

A file's loop submits all of its async comment work before its own `AwaitKey` runs (same goroutine, program order), so the per-key Add/Wait pairing is race-free, and draining no longer waits for — or crashes into — other files' in-flight submissions. The pool-wide `Await` remains for the end-of-run call sites (`dispatchSubtasks` after `wg.Wait()`, the scan batch loop), where all submissions have finished and the contract holds.

## Type of Change

## How Has This Been Tested?

- [x] New pool unit tests: `AwaitKey` waits for its own key while other keys still have in-flight work; concurrent `SubmitFor` (other keys) + `AwaitKey` stress test under `-race`; unknown-key wait returns immediately.
- [x] Manual testing: with the pre-fix code, a standalone reproduction of the review-path pattern (one goroutine in a global `Await` loop while others `Submit`) either hangs (10-minute test timeout, counter starvation) or panics with the `sync.WaitGroup` misuse errors above; the keyed API is panic-free under the same pattern.

## Checklist

- [x] I have updated the documentation accordingly (not applicable — internal concurrency change)

## Related Issues

None filed.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -618,9 +618,12 @@ func (a *Agent) executeSubtask(ctx context.Context, d model.Diff) (bool, string,
 	if err == nil {
 		// REVIEW_FILTER_TASK runs after the main loop and decides which of the
 		// just-collected comments to drop. It needs to see comments produced by
-		// the async CommentWorkerPool, so wait for that to drain first.
+		// this file's async CommentWorkerPool units, so wait for those to drain
+		// first. This must be keyed to newPath: a pool-wide Await here would run
+		// concurrently with other files' Submit calls and misuses sync.WaitGroup
+		// ("Add called concurrently with Wait").
 		if a.args.CommentWorkerPool != nil {
-			a.args.CommentWorkerPool.Await()
+			a.args.CommentWorkerPool.AwaitKey(newPath)
 		}
 		a.executeReviewFilter(ctx, d, newPath)
 	}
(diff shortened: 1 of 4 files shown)


lizhengfeng101 (approved):
LGTM

PR #450, opened 2026-07-22
fix(diff): review merge commits against their first parent

hobostay:
## Description

`ocr review --commit <merge-commit>` silently reviews nothing and exits 0 with "No supported files changed", even when the merge commit contains conflict resolutions — precisely the content that most needs review.

Root cause: `ModeCommit` uses plain `git show <commit>` (`internal/diff/git.go`), which renders merge commits as a **combined diff** (`diff --cc` sections with `@@@` hunks and two-column prefixes). `ParseDiffText` only recognizes `diff --git a/... b/...` sections, so every combined section is dropped and the parse yields zero diffs. (Worse, a combined section following a regular one would be absorbed into the previous file's diff text.)

Reproduced with real git: on a merge commit with a conflict resolution, the provider returns `[]` (0 diffs) today.

## Fix

Pass `--diff-merges=first-parent` to the `git show` call so a merge commit is diffed against its first parent in regular unified format — the same "what did this change introduce" semantics `--commit` has for normal commits. Non-merge commits are unaffected (verified: root commits still diff against the empty tree, regular commits unchanged). The flag requires git ≥ 2.31; the project already requires git ≥ 2.41.

## Type of Change

## How Has This Been Tested?

- [x] New regression test `TestCommitDiffMergeCommitReviewsFirstParentDiff` builds a real repo with a conflicting merge, then asserts the provider returns exactly one diff containing the conflict resolution. **Fails on `main`** (`expected exactly 1 diff for the merge commit, got 0: []`), passes with the fix.
- [x] Manual testing: `git show` with the exact provider flags on merge and root commits, before/after.

## Checklist

- [x] I have updated the documentation accordingly (not applicable — behavior now matches user expectation)

## Related Issues

None filed.

### DIFF
--- a/internal/diff/git.go
+++ b/internal/diff/git.go
@@ -123,7 +123,11 @@ func (p *Provider) GetDiff(ctx context.Context) ([]model.Diff, error) {
 		combined.WriteString(out)
 
 	case ModeCommit:
-		out, err := p.runGit(ctx, "-c", "core.quotepath=false", "show", "--no-ext-diff", "--no-textconv", "--find-renames", "--src-prefix=a/", "--dst-prefix=b/", "--no-color", "-U"+fmt.Sprint(DiffContextLines), "--end-of-options", p.commit)
+		// --diff-merges=first-parent: for merge commits, plain `git show`
+		// emits a combined diff ("diff --cc"), which ParseDiffText cannot
+		// parse — the commit would silently yield zero reviewable diffs.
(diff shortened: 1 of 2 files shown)


css521 (approved):
LGTM

jassandavisoon-star (approved):

jassandavisoon-star (approved):

PR #451, opened 2026-07-22
fix(diff): anchor binary marker and count +/- lines by hunk state

hobostay:
## Description

Fixes two line-classification bugs in `ParseDiffText` (`internal/diff/parser.go`), both of which silently lose or distort review input:

1. **Unanchored binary marker.** `binaryRe` (`Binary files `) was matched *anywhere* in a file's diff section, including hunk content lines. A text file whose change merely mentions the phrase — e.g. an added doc line `+Note: Binary files are handled specially by git.` — flips `IsBinary`, and the file is then silently excluded from review (`Skipping X — binary file`).

2. **`+++`/`---` header guard eats real content lines.** Insertions/Deletions counted `+`/`-` lines while excluding anything starting with `+++`/`---` to skip the file headers — but that guard also drops real content: an added line `++i` renders in the diff as `+++i`, a deleted line `--flag` as `---flag`. Besides skewing the reported per-file/total stats, the undercounted `changeLines` (`d.Insertions + d.Deletions`) gates the plan phase in the review agent, so a file whose changes mostly start with `++`/`--` (common in `.diff`/`.patch` fixtures and unindented increments) can wrongly skip planning.

## Fix

Track hunk state: a line belongs to a hunk only after the `@@` header, and inside a hunk every content line carries a leading `+`/`-`/` ` marker, so:

- `binaryRe` is anchored to column 0 (`^Binary files `, git's fixed emission format) and only consulted before the first hunk — content lines can never match.
- `+`/`-` counting only happens inside hunks, where `+++`/`---` strings are unambiguously content and are now counted correctly.
- The `--- /dev/null` / `+++ /dev/null` file-header markers are restricted to the pre-hunk region, so an added line `++ /dev/null` (rendered exactly as `+++ /dev/null`) can no longer be misread as a deleted-file header.

## Type of Change

## How Has This Been Tested?

- [x] Three new regression tests: content mentioning `Binary files ` no longer marks a file binary (real binary sections still detected); `+++newFlag`/`---oldFlag` content lines count as insertion/deletion; `+++ /dev/null` inside a hunk is treated as content. **All three fail on `main`**, pass with the fix.

## Checklist

## Related Issues

None filed.

### DIFF
--- a/internal/diff/parser.go
+++ b/internal/diff/parser.go
@@ -16,7 +16,10 @@ import (
 
 var (
 	diffHeaderRe = regexp.MustCompile(`^diff --git a/(.+?) b/(.+)$`)
-	binaryRe     = regexp.MustCompile(`Binary files `)
+	// Anchored: git emits the marker at column 0 ("Binary files a/x and b/y
+	// differ"). Content lines inside hunks always carry a leading "+", "-"
+	// or " " prefix, so an anchored match can never misfire on file content.
+	binaryRe = regexp.MustCompile(`^Binary files `)
 )
 
 // ParseDiffText splits the unified diff text into per-file Diff structs.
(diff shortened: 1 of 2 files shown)


css521 (approved):
LGTM

PR #452, opened 2026-07-22
fix(config): preserve hand-edited timeout_sec across config round-trips

hobostay:
## Description

`ocr config set` (and every other config write: `ocr config model`, interactive provider setup, `config unset`, ...) silently **deletes** a hand-edited `timeout_sec` from `~/.opencodereview/config.json`.

The LLM resolver honors `providers.<name>.timeout_sec` / `custom_providers.<name>.timeout_sec` / `llm.timeout_sec`, and the docs explicitly tell users to hand-edit them into `config.json` (they're intentionally not settable via `ocr config set`). But the cmd-side `ProviderEntry`/`LlmConfig` structs in `cmd/opencodereview/config_cmd.go` lack the field, so when `saveConfig` rewrites the whole file after an unrelated change, the key is gone and the per-request timeout silently reverts to the 300s default.

Reproduced on `main`: a config containing `"timeout_sec": 900` for a provider loses the key after one `loadOrCreateConfig` + `saveConfig` round-trip.

## Fix

Add `TimeoutSec int \`json:"timeout_sec,omitempty"\`` to both cmd-side structs (mirroring the resolver-side structs in `internal/llm/resolver.go`), so the value survives round-trips. This intentionally does **not** add `ocr config set <...>.timeout_sec` write support, matching the documented behavior.

## Type of Change

## How Has This Been Tested?

- [x] New regression test `TestConfigRoundTripPreservesTimeoutSec`: hand-edited `timeout_sec` values in `providers`, `custom_providers`, and `llm` sections survive an unrelated `ocr config set` round-trip. Fails on `main` (field absent), passes with the fix.

## Checklist

- [x] I have updated the documentation accordingly (not applicable — the docs already describe the key; it now actually survives)

## Related Issues

None filed.

### DIFF
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -194,6 +194,7 @@ type ProviderEntry struct {
 	Model        string            `json:"model,omitempty"`
 	Models       []string          `json:"models,omitempty"`
 	AuthHeader   string            `json:"auth_header,omitempty"`
+	TimeoutSec   int               `json:"timeout_sec,omitempty"` // per-request HTTP timeout in seconds
 	ExtraBody    map[string]any    `json:"extra_body,omitempty"`
 	ExtraHeaders map[string]string `json:"extra_headers,omitempty"`
 }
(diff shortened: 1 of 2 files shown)


lizhengfeng101 (approved):
LGTM

PR #95, opened 2026-06-11
feat(vscode): add VSCode extension for AI code review

nigulasikk:
## Description

  Add a VSCode extension under `extensions/vscode/` that brings the
  [`open-code-review`](https://www.npmjs.com/package/@alibaba-group/open-code-review)
  (`ocr`) CLI into the editor. The extension drives reviews from a sidebar,
  streams CLI logs live, and renders each finding as an in-editor
  `CommentThread` that stays in sync with the sidebar (apply / ignore /
  mark-false-positive on either side updates the other).

  ### Architecture

  Adopts a **Monolithic WebView + Thin Extension Host** design:

  | Layer | Role |
  | --- | --- |
  | `src/extension/` | Extension Host (Node.js): `services/` (CLI, Git, Config,
  ReviewSession), `providers/` (Sidebar WebView, CommentThread), `commands/` |
  | `src/webview/` | WebView SPA (Preact): `views/` (Idle / Running / Done / Empty /
  Cancelled / Failed / Config), `components/`, `store`, `bridge` |
  | `src/shared/` | TypeScript types + `postMessage` protocol shared by both ends (no
  `vscode` dependency) |

  The two layers communicate via `postMessage`, with shared types in
  `src/shared/` guaranteeing type safety across the boundary.

  ### Features

  - **Three review modes**: working-tree changes, branch comparison
    (`--from` / `--to`), single commit (`--commit`).
  - **File preview**: lists changed files from current Git state before review.
  - **Custom prompt**: optionally append a `--background` hint per review.
  - **Streaming logs**: live CLI output with cancel-at-any-time support.
  - **Result + two-way sync**: comment cards in the sidebar mirror in-editor
    `CommentThread`s; apply / ignore / false-positive stay in sync.
  - **Empty / Cancelled / Failed states**: dedicated views (Failed is retryable).
  - **Config management**: view/edit LLM provider config in-extension (written
    via `ocr config set`).
  - **Model switch / connectivity test**: switch model and test LLM reachability
    from the status bar.

  ## Type of Change

  - [ ] Bug fix (non-breaking change that fixes an issue)
  - [x] New feature (non-breaking change that adds functionality)
  - [ ] Breaking change
  - [ ] Refactoring (no functional changes)
  - [ ] Documentation update
  - [ ] CI / Build / Tooling

  ## How Has This Been Tested?

  - [x] `yarn test` passes locally (Jest)
  - [x] Manual testing (described below)

  Unit tests cover:

[...]

PR #489, opened 2026-07-24
[codex] fix(vscode): force-kill unresponsive reviews

cosin2077:
## Background

The VS Code extension sends `SIGTERM` when a user cancels an active review and is intended to escalate to `SIGKILL` after three seconds if the CLI process does not exit.

## Root cause

`ChildProcess.killed` only indicates that a signal was successfully sent. It does not indicate that the child process has exited. A process that ignores `SIGTERM` therefore has `killed === true` while `exitCode` and `signalCode` remain `null`, preventing the existing timeout from ever sending `SIGKILL`.

## Fix

- Check `exitCode` and `signalCode` to determine whether the child is still running before escalating to `SIGKILL`.
- Clear the escalation timer when the child emits `close` to avoid signaling a process that already exited.
- Add regression tests for both an unresponsive child and a child that exits normally after `SIGTERM`.

## Verification

- `yarn install --frozen-lockfile` (Node.js 24.18.0)
- `yarn lint` - 0 errors (1 pre-existing warning in `IdleView.tsx`)
- `yarn test --runInBand` - 11 suites, 94/94 tests passed
- `yarn build` - production webpack build passed

## Related

- #95 introduced the VS Code extension and its cancellation flow.

### DIFF
--- a/extensions/vscode/src/extension/services/CliService.ts
+++ b/extensions/vscode/src/extension/services/CliService.ts
@@ -158,7 +158,10 @@ export class CliService {
     if (this.current && this.current.pid) {
       this.current.kill('SIGTERM');
       const proc = this.current;
-      setTimeout(() => { if (!proc.killed) proc.kill('SIGKILL'); }, 3000);
+      const forceKillTimer = setTimeout(() => {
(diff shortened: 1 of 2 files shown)


lizhengfeng101 (approved):
LGTM

PR #485, opened 2026-07-24
feat(llm): add iFlytek Spark as a built-in provider

FenjuFu:
## What

Adds **iFlytek Spark** as a built-in LLM provider.

The provider list already covers most major Chinese vendors — DashScope/Qwen, Volcano Engine, DeepSeek, Tencent Hunyuan, Kimi, Z.AI, Xiaomi MiMo, MiniMax, Baidu Qianfan — but iFlytek Spark was missing. Its official endpoint is OpenAI-compatible, so it slots in exactly like the other `openai`-protocol providers.

## Changes

- `internal/llm/providers.go` — new registry entry:
  - `BaseURL: https://spark-api-open.xf-yun.com/v1` (Spark's OpenAI-compatible HTTP endpoint; the API password is sent as a bearer token)
  - `EnvVar: SPARK_API_KEY`
  - Models are the exact `domain` values that endpoint accepts: `4.0Ultra`, `generalv3.5`, `max-32k`, `generalv3`, `pro-128k`, `lite`
- `internal/llm/providers_test.go` — added `iflytek` to the expected provider-order list (`ListProviders()` sorts by name, so it sits between `hy-tokenplan` and `kimi`)
- `pages/src/content/docs/{en,zh,ja}/configuration.md` — added the provider to the config table in all three locales

## Notes

- No new protocol or client code is needed — Spark rides the existing `ProtocolOpenAIChatCompletions` path, same as `deepseek`/`kimi`.
- I don't have a funded Spark API password, so there's no live end-to-end call in this PR. I verified statically that the sorted registry names match the updated `TestListProviders_Order` expectation exactly (18 providers, no duplicates) and that the three doc tables each gained one row. `make test` should confirm on CI.
- Endpoint and model ids are from iFlytek's official HTTP API docs (`spark-api-open.xf-yun.com/v1`); the WebSocket host `spark-api.xf-yun.com` uses HMAC-signed auth and is deliberately **not** used here.

Issue #457, opened 2026-07-23
perf(pages): enable webpack code splitting and lazy-load heavy dependencies


lizhengfeng101:
## Description

The landing page at [open-codereview.ai](https://open-codereview.ai/) currently ships a single monolithic `bundle.js` of **2.0 MB** (uncompressed), with a total site weight of **5.8 MB**. All dependencies — including Three.js (3D rendering), Mermaid (diagram rendering), and Marked (markdown parsing) — are bundled into the initial payload regardless of whether the user's viewport needs them.

This makes the site particularly slow for users in regions with high latency to GitHub Pages (e.g., mainland China), where the combination of cross-border routing and large payloads results in 10s+ load times.

### Current webpack output config (`pages/webpack.config.js`):

```js
output: {
  path: path.resolve(__dirname, 'dist'),
  filename: 'bundle.js',  // no contenthash, no splitChunks
  publicPath: '/'
}
```

## Scope

- File(s): `pages/webpack.config.js`, `pages/src/` (component-level lazy loading)
- Area: Frontend build pipeline & bundle optimization

## Tasks

1. **Enable `optimization.splitChunks`** — separate vendor chunks (react, three, mermaid) from application code for parallel loading and long-term caching
2. **Lazy-load Three.js** — use `React.lazy()` + dynamic `import()` so the 3D background only loads after first paint
3. **Lazy-load Mermaid** — dynamically import only on pages/components that render diagrams
4. **Use contenthash in output filenames** — e.g., `[name].[contenthash:8].bundle.js` to enable aggressive CDN caching
5. **Add `compression-webpack-plugin`** — generate pre-compressed `.gz` / `.br` files at build time for CDN direct delivery

## Acceptance Criteria

- [ ] `bundle.js` is split into multiple chunks; initial (critical-path) JS is under 400 KB uncompressed
- [ ] Three.js and Mermaid are in separate async chunks, loaded on demand
- [ ] Output filenames include content hash for cache-busting
- [ ] `npm run build` in `pages/` succeeds and produces valid output
- [ ] Site renders correctly with all routes functional (SPA fallback via `404.html` still works)
- [ ] Lighthouse Performance score improves (measure before/after)

## Context

This optimization is part of a broader effort to improve site accessibility from mainland China. CDN acceleration (Alibaba Cloud CDN → OSS Hong Kong) addresses network latency, but reducing payload size provides a multiplicative improvement — smaller assets benefit more from edge caching and parallel downloads.

[...]

PR #492, opened 2026-07-24
fix(ci): pin Pages CI container to node:24.18.0

lizhengfeng101:
## Summary

- Pin `pages-ci.yml` container image from the rolling `node:24` tag to `node:24.18.0`
- The rolling tag recently updated to a newer npm version that rejects `npm ci` with the existing lockfileVersion-3 lockfile, breaking all PRs that touch `pages/**` (e.g. #485, #457)
- `node:24.18.0` is the same version already used by `translation-sync.yml` and is known to work

## Context

The `pages-ci.yml` workflow was introduced in #442 with `image: node:24`. At the time, the rolling tag pointed to a compatible npm version. Docker Hub has since updated the tag, causing `npm ci` to fail with a lockfile incompatibility error.

### DIFF
--- a/.github/workflows/pages-ci.yml
+++ b/.github/workflows/pages-ci.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: self-hosted
     timeout-minutes: 15
     container:
-      image: node:24
+      image: node:24.18.0
     steps:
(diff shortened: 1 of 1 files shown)


xyJen (approved):
LGTM

PR #519, opened 2026-07-27
fix(actions): clarify ambiguous batch review log

### LINKED ISSUE #518: fix(actions): clarify ambiguous log when findExistingBatchReview matches an earlier batch
## Description

In the multi-batch comment publishing flow (`publishBatch`), all batches within the same run share a single `REVIEW_TAG`. When batch N fails with a 5xx and `findExistingBatchReview` is called to check whether the batch landed, it may find an **earlier** batch's review (e.g., batch 1's review that succeeded) rather than the current batch's.

The log message currently reads:

```
A batch review already exists on the server (review_id=<id>). X/Y of this batch's inline comments already posted. Z missing, will retry only those.
```

This is **not a correctness bug** — `getPostedCommentIds` returns a server-global set so the filtering logic works correctly regardless of which review is found. However, the `review_id` shown may belong to a different batch, which can confuse someone debugging a partial-failure scenario (they might think the current batch landed when it didn't).

## Scope

- File: `scripts/github-actions/post-review-comments.js`
- Function: `publishBatch` (line ~406)
- The existing code comment at lines 393–398 already documents this behavior; only the user-facing log message needs clarification.

## Suggested Fix

Add a note to the log message indicating the matched review may belong to a different batch:

```javascript
log(
  `A batch review with this run's tag exists on the server ` +
    `(review_id=${existingReview.review.id}, may belong to an earlier batch). ` +
    `${succeeded}/${chunk.length} of this batch's inline comments already posted. ` +
    `${toRetry.length} missing, will retry only those.`
);
```

## Acceptance Criteria

- [ ] Log message clarifies that the matched review may not belong to the current batch
- [ ] Tests pass (`npm run test:github-actions`)

## Context

Discovered during code review of #507 (feat: chunk inline comments into bounded batches). The implementation is correct — this is a debugging-UX improvement only.

LKL-ZREO:
## Description

When a later comment batch fails with a 5xx response,
`findExistingBatchReview` searches for a review using the run-level
`REVIEW_TAG`. Because this tag is shared by all batches in the same run, the
matched review may belong to an earlier successful batch rather than the
current failed batch.

The reconciliation logic remains correct because `getPostedCommentIds`
returns a server-global set of posted comment IDs. However, the existing log
message implies that the matched review necessarily belongs to the current
batch, which can be misleading while debugging partial failures.

This change:

- clarifies that the matched review may belong to an earlier batch;
- preserves the existing reconciliation and retry behavior;
- captures `core.info` output in the test mock;
- adds a regression assertion to the existing multi-batch reconciliation test.

Closes #518.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] `npm run test:github-actions`
- [x] `node --check scripts/github-actions/post-review-comments.js`
- [x] `node --check scripts/github-actions/post-review-comments.test.js`
- [x] `go vet ./...`

## Checklist

- [x] I have added tests that prove the change is effective
- [x] New and existing GitHub Actions tests pass locally
- [x] I have updated the documentation accordingly (not applicable: log-only change)
- [x] I have signed the CLA

## Related Issues

Closes #518.

### DIFF
--- a/scripts/github-actions/post-review-comments.js
+++ b/scripts/github-actions/post-review-comments.js
@@ -403,7 +403,8 @@ async function publishBatch({
       succeeded = chunk.length - toRetry.length;
       reconciled = succeeded > 0;
       log(
-        `A batch review already exists on the server (review_id=${existingReview.review.id}). ` +
+        `A batch review with this run's tag exists on the server ` +
+          `(review_id=${existingReview.review.id}, may belong to an earlier batch). ` +
           `${succeeded}/${chunk.length} of this batch's inline comments already posted. ` +
           `${toRetry.length} missing, will retry only those.`
       );
--- a/scripts/github-actions/post-review-comments.test.js
+++ b/scripts/github-actions/post-review-comments.test.js
@@ -289,10 +289,12 @@ function makeGithub(opts = {}) {
 
 function mockCore() {
   const outputs = {};
+  const logs = [];
   return {
     outputs,
+    logs,
     setOutput(name, value) { outputs[name] = value; },
-    info() {},
+    info(message) { logs.push(message); },
   };
 }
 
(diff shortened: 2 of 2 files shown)


lizhengfeng101 (approved):
LGTM

PR #530, opened 2026-07-27
fix: path traversal bypass via backslash separators on Windows in code_search

Sunil56224972:
## Summary

**Security Fix**: `hasTraversalPathComponent()` in `internal/tool/code_search.go` only splits on `'/'` to detect `..` path traversal components in `file_patterns`. On Windows (an explicitly supported platform), backslash `\` is a valid path separator, so patterns like `..\..\etc\passwd` bypass the traversal check entirely.

### Root Cause

The function splits the pathspec on `'/'` only. When a backslash-separated path like `..\..\..\etc` is provided, `strings.Split(s, "/")` produces a single component `..\..\..\etc` which does NOT equal `..`, so the traversal is not detected.

### Impact

An LLM (potentially via prompt injection in reviewed code) could craft `file_patterns` arguments to `code_search` using backslash path separators to search files outside the repository root when OCR runs on Windows.

### Fix

Normalize the pathspec via `filepath.ToSlash()` (converts `\` to `/`) and `filepath.Clean()` (collapses redundant separators and `..` segments) before splitting to check for traversal components.

### Testing

Added `TestCodeSearchProvider_Execute_RejectsBackslashTraversal` with test cases for:
- Pure backslash traversal: `..\..\\etc`
- Backslash mid-path: `pkg\..\internal`
- Mixed separators: `pkg\../..\secret`

All existing tests continue to pass.

### DIFF
--- a/internal/tool/code_search.go
+++ b/internal/tool/code_search.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os/exec"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -89,7 +90,9 @@ func (p *CodeSearchProvider) buildGrepArgs(searchText string, caseSensitive bool
 }
 
 func hasTraversalPathComponent(pathspec string) bool {
-	for _, part := range strings.Split(pathspec, "/") {
+	// Normalize backslash separators (Windows) so "..\..\.." is caught too.
+	normalized := filepath.ToSlash(filepath.Clean(pathspec))
+	for _, part := range strings.Split(normalized, "/") {
 		if part == ".." {
 			return true
 		}
--- a/internal/tool/code_search_test.go
+++ b/internal/tool/code_search_test.go
@@ -462,6 +462,34 @@ func TestCodeSearchProvider_Execute_RejectsTraversalPattern(t *testing.T) {
 	}
 }
 
+func TestCodeSearchProvider_Execute_RejectsBackslashTraversal(t *testing.T) {
+	dir := setupTestRepo(t)
+	p := NewCodeSearch(&FileReader{RepoDir: dir, Mode: ModeWorkspace})
+	tests := []struct {
+		name    string
+		pattern string
+	}{
+		{name: "backslash leading parent", pattern: `..\..\etc`},
+		{name: "backslash middle parent", pattern: `pkg\..\internal`},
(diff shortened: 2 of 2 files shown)


lizhengfeng101 (commented):
## Review

Thanks for the contribution! The identified gap is real — `hasTraversalPathComponent` only splits on `/`, so backslash-separated paths like `..\..\..\etc` bypass the check on Windows where `\` is a valid path separator.

That said, the practical exploitability is very low: `git grep` only searches tracked files (or untracked files within the worktree with `--untracked`), and modern git rejects pathspecs that resolve above the repository root. So this is more of a defense-in-depth hardening than a directly exploitable vulnerability.

### Issue with `filepath.Clean`

The current fix uses `filepath.ToSlash(filepath.Clean(pathspec))`, which introduces platform-dependent behavior:

- **On Linux**: `filepath.Clean("pkg\\..\\internal")` → `"pkg\\..\\internal"` (backslash is a literal character, not a separator) → then `ToSlash` converts it → `"pkg/../internal"` → `..` detected ✓
- **On Windows**: `filepath.Clean("pkg\\..\\internal")` → `"internal"` (resolves the `..`) → `..` gone → **not detected** ✗

This means the test case `"backslash middle parent"` would **fail on Windows**, and the behavior diverges across platforms.

### Suggested simpler approach

A plain string replacement achieves the goal without platform-dependent semantics:

```go
	normalized := strings.ReplaceAll(pathspec, "\\", "/")
	for _, part := range strings.Split(normalized, "/") {
		if part == ".." {
			return true
		}
	}
	return false
}
```

Advantages:
- Identical behavior on all platforms
- Preserves the original strict policy (any `..` component is rejected)
- No new `path/filepath` import needed
- Easier to audit — no need to reason about `filepath.Clean` semantics
- All test cases pass consistently everywhere

Sunil56224972:
Great catch on the `filepath.Clean` platform dependency! You're right — on Windows it resolves the `..` before we can detect it, making the check silently pass.

I've adopted your suggested approach:

```go
    normalized := strings.ReplaceAll(pathspec, "\\", "/")
    for _, part := range strings.Split(normalized, "/") {
        if part == ".." {
            return true
        }
    }
    return false
}
```

Advantages (as you noted):
- Identical behavior on all platforms
- Preserves the original strict policy
- No new `path/filepath` import needed
- Easier to audit
- All test cases pass consistently everywhere

Also rebuilt both files from upstream to fix encoding.

PR #533, opened 2026-07-27
fix: ensureMessagesSuffix double-paths URLs ending with /v1

Sunil56224972:
## Summary

**Bug Fix**: `ensureMessagesSuffix()` in `internal/llm/resolver.go` double-paths API URLs that end with `/v1/` or `/v1`, producing broken endpoints like `https://api.anthropic.com/v1/v1/messages`.

### Root Cause

The function uses `strings.Contains(u, "/v1/")` to detect whether a URL already has a versioned path. After `strings.TrimRight(rawURL, "/")`, a URL like `https://api.anthropic.com/v1/` becomes `https://api.anthropic.com/v1` — which does **not** contain the substring `/v1/` (no trailing slash). The function then appends `/v1/messages`, producing:

```
https://api.anthropic.com/v1/v1/messages  ← BROKEN
```

Additionally, `strings.Contains` matches `/v1/` **anywhere** in the URL, including in hostnames or unrelated path segments (e.g., `https://proxy.v1.example.com/api` would match and skip the suffix), causing silent misrouting.

### Impact

Users who configure `ANTHROPIC_BASE_URL` or `OCR_LLM_BASE_URL` with a trailing-slash `/v1/` endpoint (a common copy-paste from Anthropic docs) will get HTTP 404 errors from the API because the request goes to `/v1/v1/messages` instead of `/v1/messages`. This is a **silent failure** — no error message indicates the URL is wrong, making it extremely hard to debug.

### Fix

Replace `strings.Contains` with `strings.HasSuffix` checks:
1. If URL already ends with `/v1/messages` → return as-is
2. If URL ends with `/v1` → append only `/messages`
3. Otherwise → append full `/v1/messages`

### Testing

Added `TestEnsureMessagesSuffix` with 7 test cases covering:
- Bare base URL
- Base URL with trailing slash
- URL ending with `/v1` (the bug case)
- URL ending with `/v1/` (the bug case)
- URL already having `/v1/messages`
- URL with `/v1/messages/` trailing slash
- Proxy URL without versioned path

### DIFF
--- a/internal/llm/resolver.go
+++ b/internal/llm/resolver.go
@@ -680,5 +680,9 @@ func ensureMessagesSuffix(rawURL string) string {
 	if strings.HasSuffix(u, "/v1") {
 		return u + "/messages"
 	}
+	// URL has /v1/ in the middle (e.g. /v1/anthropic) — already versioned, leave it alone.
+	if strings.Contains(u, "/v1/") {
+		return rawURL
+	}
 	return u + "/v1/messages"
 }
--- a/internal/llm/resolver.go
+++ b/internal/llm/resolver.go
@@ -672,11 +672,11 @@ func splitHeaderPairs(raw string) ([]string, error) {
 // ensureMessagesSuffix appends /v1/messages to base URLs that lack a versioned path.
 func ensureMessagesSuffix(rawURL string) string {
 	u := strings.TrimRight(rawURL, "/")
-	// Already ends with the full messages path â€” don't modify.
+	// Already ends with the full messages path — don't modify.
 	if strings.HasSuffix(u, "/v1/messages") {
 		return rawURL
 	}
-	// Ends with /v1 (e.g. "https://api.anthropic.com/v1") â€” only append /messages.
+	// Ends with /v1 (e.g. "https://api.anthropic.com/v1") — only append /messages.
 	if strings.HasSuffix(u, "/v1") {
 		return u + "/messages"
 	}
--- a/internal/llm/resolver.go
+++ b/internal/llm/resolver.go
@@ -672,9 +672,13 @@ func splitHeaderPairs(raw string) ([]string, error) {
 // ensureMessagesSuffix appends /v1/messages to base URLs that lack a versioned path.
 func ensureMessagesSuffix(rawURL string) string {
 	u := strings.TrimRight(rawURL, "/")
-	if strings.Contains(u, "/v1/") {
-		// Already has versioned path — don't modify.
+	// Already ends with the full messages path â€” don't modify.
+	if strings.HasSuffix(u, "/v1/messages") {
 		return rawURL
+	}
+	// Ends with /v1 (e.g. "https://api.anthropic.com/v1") â€” only append /messages.
+	if strings.HasSuffix(u, "/v1") {
+		return u + "/messages"
 	}
 	return u + "/v1/messages"
 }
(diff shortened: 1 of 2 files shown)


lizhengfeng101:
Thanks for the fix! The bug is real and the fix is correct. I traced through the call sites — `ANTHROPIC_BASE_URL` env var, config.json entries, and shell RC exports all pass user-configured URLs directly into `ensureMessagesSuffix()` with no normalization, so users setting `https://api.anthropic.com/v1` or `https://api.anthropic.com/v1/` would definitely hit the double-path issue.

The logic and test coverage look good. One issue that needs fixing before merge:

**Encoding corruption**: There are 4 places where em-dash `—` (U+2014) has been replaced with mojibake `â€"`. This appears to be a UTF-8 → Latin-1 encoding issue in your editor. Two of these are in comments unrelated to your change (in `TestNewLLMClient_DefaultTimeout` and `TestResolveEndpoint_ResponsesURLNotMutated`), so they should not be part of this diff at all.

Please fix the encoding in your editor settings and revert the unrelated comment changes. After that, LGTM to merge.

lizhengfeng101:
One more thing I noticed after closer analysis:

**Potential regression for proxy URLs with `/v1/` in the middle of the path**

The old code used `strings.Contains(u, "/v1/")`, which had a side effect of preserving URLs like `https://proxy.com/v1/anthropic` as-is (it matched the `/v1/` substring and returned early).

With the new `HasSuffix`-only logic:
- `HasSuffix(u, "/v1/messages")` → no match
- `HasSuffix(u, "/v1")` → no match
- Falls through to append `/v1/messages` → produces `https://proxy.com/v1/anthropic/v1/messages` ❌

While this is an edge case, it is a behavioral regression for users who configure a custom Anthropic-protocol proxy with a sub-path after `/v1/` (e.g., `/v1/anthropic`, `/v1/proxy`).

**Suggested fix**: add a `Contains` guard at the top to preserve the old pass-through behavior for URLs that already have a versioned path segment:

```go
func ensureMessagesSuffix(rawURL string) string {
    u := strings.TrimRight(rawURL, "/")
    if strings.HasSuffix(u, "/v1/messages") {
        return rawURL
    }
    if strings.HasSuffix(u, "/v1") {
        return u + "/messages"
    }
    // URL has /v1/ in the middle (e.g. /v1/anthropic) — already versioned, leave it alone.
    if strings.Contains(u, "/v1/") {
        return rawURL
    }
    return u + "/v1/messages"
}
```

This keeps your fix for the double-path bug while avoiding a regression for proxy configurations with sub-paths.

Sunil56224972:
Thanks for the review! You're right about the encoding corruption — my editor was converting UTF-8 em-dashes to Latin-1 mojibake.

I've pushed two follow-up commits:
1. `c63ff8a` — Restores `resolver_test.go` to upstream content verbatim + only the new `TestEnsureMessagesSuffix` test
2. `92d67af` — Restores `resolver.go` to upstream encoding, keeping only the `ensureMessagesSuffix` fix

The diff should now be clean — only the intended changes, no unrelated modifications.

Sunil56224972:
Good catch on the proxy regression! I've pushed two more commits:

1. `7c0dfb8` — Added back `strings.Contains(u, '/v1/')` as a fallback guard **after** the `HasSuffix` checks, so proxy URLs like `https://proxy.com/v1/anthropic` are preserved as-is
2. `1a541d4` — Added test case for the `/v1/` mid-path proxy scenario

The evaluation order is now:
1. `HasSuffix('/v1/messages')` — exact match, return as-is
2. `HasSuffix('/v1')` — append only `/messages`  (fixes the original double-path bug)
3. `Contains('/v1/')` — mid-path versioned URL, return as-is (preserves proxy behavior)
4. Fallback — append `/v1/messages`

lizhengfeng101 (approved):
LGTM

PR #545, opened 2026-07-28
fix(vscode): pin linkify-it below v6 to fix vsce package failure

lizhengfeng101:
## Summary

- Pin `linkify-it` resolution to `>=5.0.2 <6` to avoid the breaking ESM named-export change in v6
- Fixes `ERROR LinkifyIt is not a constructor` when running `vsce package`
- Bumps extension version to 0.1.2

## Context

`linkify-it` v6 switched from a default export to named exports (`{ LinkifyIt }`). The `markdown-it` library bundled in `@vscode/vsce` still calls `new LinkifyIt()` via the default import, causing the packaging step to fail after README markdown processing.

## Test plan

- [x] `corepack yarn package` succeeds and produces a valid `.vsix`

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -2,7 +2,7 @@
   "name": "open-code-review-vscode",
   "displayName": "Open Code Review",
   "description": "%ocr.description%",
-  "version": "0.1.1",
+  "version": "0.1.2",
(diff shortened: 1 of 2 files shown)


MuoDoo (approved):
LGTM

PR #547, opened 2026-07-28
fix(pages): keep code block copy button anchored during horizontal scroll

hanhaiqingchuan:
## Description

On the docs site (e.g. /docs/architecture), the copy icon of a code block
scrolled away together with overflowing code and could even end up out of
view, because the icon was a normal flex child inside the horizontally
scrolling `<pre>`.

This change moves horizontal scrolling from the `<pre>` to the inner `<code>`
element and floats the copy icon over the top-right corner. The horizontal
padding now lives inside the scrollport, so code text still appears and
disappears exactly at the box borders, and the icon stays anchored at the
top-right corner at any scroll position.

**Before** — the copy icon scrolls away with the code:
image

**After** — the icon stays anchored at the top-right corner:
image

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Frontend-only change under `pages/`, so `make test` (Go) is not applicable.
`npm run typecheck` and `npm run build` both pass (the same steps pages-ci
runs). Manually verified in Chrome on /docs/architecture with a horizontally
overflowing code block: the icon stays at the top-right corner at scroll
positions start/middle/end; text clips exactly at both box borders; copy
still works while scrolled; single-line blocks and mermaid diagrams are
unaffected.

## Checklist

- [ ] My code follows the project's coding style (`go fmt`, `go vet`) — N/A (frontend-only)
- [ ] I have added tests that prove my fix is effective or my feature works — N/A (pages/ has no test framework)
- [ ] New and existing unit tests pass locally with my changes — N/A
- [x] I have updated the documentation accordingly (if applicable) — no docs affected
- [ ] I have signed the CLA

## Related Issues

None.

### DIFF
--- a/pages/src/components/MarkdownRenderer.tsx
+++ b/pages/src/components/MarkdownRenderer.tsx
@@ -108,7 +108,7 @@ const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content }) => {
       // Create copy button matching reference HTML
       const btn = document.createElement('div');
       btn.className = 'code-copy-btn';
-      btn.style.cssText = 'display:flex;flex-shrink:0;justify-content:flex-start;align-items:flex-start;flex-direction:column;padding-top:4px;padding-bottom:4px;cursor:pointer;';
+      btn.style.cssText = 'position:absolute;top:4px;right:12px;display:flex;padding:4px;cursor:pointer;';
       btn.innerHTML = `<img src="${copyIcon}" alt="copy" style="width:16px;height:16px;" />`;
       btn.addEventListener('click', () => {
         const text = codeEl.textContent || '';
--- a/pages/src/components/MarkdownRenderer.tsx
+++ b/pages/src/components/MarkdownRenderer.tsx
@@ -85,9 +85,7 @@ const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content }) => {
       const trimmed = text.replace(/^\n+|\n+$/g, '');
       const content = escaped ? trimmed : trimmed.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;');
       const langClass = lang ? ` class="language-${lang}"` : '';
-      const isMultiline = trimmed.includes('\n');
-      const alignItems = isMultiline ? 'flex-start' : 'center';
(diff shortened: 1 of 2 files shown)


lizhengfeng101 (commented):
## Code Review

**Overall:** The approach is sound — moving `overflow-x` from `<pre>` to `<code>` and using `position: absolute` for the copy button is the right fix for this problem.

### Potential Issues

1. **Button may overlap code text** — With the button absolutely positioned at `top:4px; right:12px`, it will sit on top of code content when lines are short (no horizontal scroll needed). The old flex layout kept the button alongside the code without overlapping. Consider adding `padding-right` to `<code>` (e.g. ~40px) to reserve space for the button, or show the button only on hover with a semi-transparent background.

2. **Residual flex properties on `<pre>`** — `justify-content: space-between` and `align-items: center` are still on `<pre>`, but with the button now absolutely positioned (out of flow), these only affect the single `<code>` child. They're harmless but dead weight — could be cleaned up.

3. **Firefox scrollbar styling** — The scrollbar selectors were updated from `pre` to `pre code`, but only `-webkit-scrollbar` variants are covered. Consider adding `scrollbar-width: thin` and `scrollbar-color` on `.docs-markdown pre code` for Firefox consistency.

### Minor

- The inline `cssText` string is getting long but matches the existing pattern, so no objection there.

Overall this is a nice fix. The main actionable suggestion is #1 (preventing button/text overlap on short lines).

hanhaiqingchuan:
Thanks for the review! Addressed all three in 78afd96:

1. **Overlap** — went with your hover-reveal suggestion: the button is hidden by
       default and fades in on hover with a translucent `rgba(0,0,0,.7)` background,
       plus a `@media (hover: none)` override so it stays visible on touch devices.
       
2. **Dead flex props** — removed, along with the now-dead `flex: 1` /
       `min-width: 0` on `<code>` and the per-block inline `align-items` emitted by
       the renderer.
3. **Firefox scrollbar** — added `scrollbar-width: thin` and `scrollbar-color`
       on `pre code`.

Also fixed a latent issue this surfaced: the generic`.docs-markdown img { margin: 8px 0 }` content-image rule was inflating the icon to 40px and making the button overflow single-line code blocks; it's now reset inside the button.

lizhengfeng101 (approved):
LGTM

PR #574, opened 2026-07-29
feat(allowlist): add matlab support

GenosseFlosse:
## Description

This PR adds matlab language support. Most of  the work addresses the disambiguation between matlab and objective-c, which share the same file ending *.m.

Quick notes:
Grouping side effect. With the sniff no longer annotating Pattern, a repo mixing MATLAB and Objective-C .m files now emits two rule groups with identical source and pattern but different rule:
```
group_id=1  source=system  pattern=**/*.m  files=[ios/ViewController.m]  rule=objc.md
group_id=2  source=system  pattern=**/*.m  files=[Models/main.m]         rule=matlab.md
```
This is consistent with GroupRules's doc comment and grouping stays correct (the key is source+pattern+text), but it's a shape the system layer couldn't produce before — pattern → rule used to be 1:1 there. Anyone deduping on source+pattern alone will run into it.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

    - ocr rules check reports the correct rules when checking manually

## Checklist

- [x] I have signed the CLA


lizhengfeng101 (commented):
Thanks for the contribution — the MATLAB rule doc itself is excellent and very thorough!

One concern before merging: **the `.m` extension is shared between MATLAB and Objective-C**. This PR maps all `**/*.m` files to `matlab.md`, which means Objective-C files (e.g. `ios/ViewController.m`, `AppDelegate.m`) will receive MATLAB-specific review feedback — clearly incorrect for those files.

The test change confirms this: `ios/ViewController.m` was removed from the "fallback to default" test because it now matches the MATLAB rule.

### Possible approaches

1. **Path-based heuristics** — put ObjC-specific glob patterns (e.g. `**/{ios,macos,tvos,watchos,Pods,Classes}/**/*.m`) before `**/*.m` in `system_rules.json`, leveraging "first match wins" order. Downside: incomplete coverage, maintenance burden.

2. **Content-based detection** — sniff the first line of the file (`function` → MATLAB, `#import`/`#include` → ObjC). Downside: requires changes to the `Resolve` interface which currently only takes a path.

3. **Accept the limitation** — document that `.m` defaults to MATLAB and ObjC projects should use a project-level `.opencodereview/rule.json` to override. Simplest, but shifts burden to ObjC users.

### Minor formatting issues in `system_rules.json`

- Trailing space on the `"**/*.bicep": "bicep.md", ` line
- The new `"**/*.m"` entry uses a tab for indentation, while the rest of the file uses 4 spaces

Would love to hear your thoughts on how you'd like to handle the extension ambiguity. Happy to discuss further!

GenosseFlosse:
Dear @lizhengfeng101,

I opted for Option Number 2, as this will ensure that it just works instead of anyone stumbling onto this in a few months. I added a mock objc.md for someone to fill at a later date (I don't know anything about objective c). The beahviour is documented in review_rules.md

GenosseFlosse:
Dear @lizhengfeng101 
I have resolved the latest merge requests an I am happy to discuss any review you might bring up. Thank you for reviewing my changes :)

lizhengfeng101:
@GenosseFlosse Thanks for adding MATLAB support — the rule doc is thorough and the test coverage looks solid!

One architectural suggestion before we move forward: the current approach threads a `content` parameter through the entire resolver pipeline (new interfaces, signature changes across 16 files) to handle what is essentially a single special case (`.m` disambiguation). This makes the change more invasive than it needs to be.

A lighter-weight alternative would be a **decorator/sniffer wrapper** around the existing resolver:

- The `Resolver` interface stays unchanged — still just `Resolve(path string) string`.
- A thin `Sniffer` struct wraps the composed resolver. It holds the repo path and a pre-loaded `objc.md` rule.
- When `Resolve` is called on a `.m` file, the Sniffer peeks the first line from disk. If it looks like Objective-C (`#import`, `@interface`, etc.), it returns the ObjC rule directly; otherwise it delegates to the inner resolver as usual.
- `NewResolver` wraps the result in a Sniffer before returning — callers (`agent.go`, `scan/agent.go`, `delegate/rulegroup.go`) never see the difference and need zero changes.

This keeps the sniffing logic in one new file (~50 lines), avoids new interfaces (`ContentAwareResolver`, `ContentAwareDetailResolver`), and leaves the composed resolver and all its callers untouched.

Would you be open to restructuring along these lines? Happy to pair on it if that would help.

Also, please rebase onto the latest `main` when you get a chance — there have been a few recent changes that may conflict.

GenosseFlosse:
@lizhengfeng101 Yes, good point, I am on it. And sure, I can rebase.

GenosseFlosse:
Resolver is unchanged again; ContentAwareResolver, ContentAwareDetailResolver and the content parameter are all gone. Logic now is in one new file, internal/config/rules/sniffer.go. 

internal/delegate/rulegroup.go (and its test) and internal/agent/coverage_test.go are back to untouched. 

Production code outside internal/config/rules is now 46 changed lines across 7 files, six of them one-liners.

Three notes on the implementation:
1. The wrapper sits on the system layer, not the outermost resolver. Wrapping the composed resolver would break rule priority: composedResolver checks user layers first and calls back into the system layer for merge_system_rule (system_rules.go:493). An outermost sniffer would short-circuit any .m file to objc.md and discard a user's own .opencodereview/rule.json entry. So composedResolver.system became a small unexported interface and the sniffer wraps that — still fully inside the package, callers unaffected. Two tests verify this.
2. It forwards ResolveDetail and CanonicalConfig. rules_cmd.go:54 type-asserts DetailResolver (without it, ocr rules check fails outright), and agent.go:855 asserts CanonicalConfig() []string — without forwarding, rule_config_sha256 would silently stop reflecting rule config. The forwarded CanonicalConfig folds in the objc rule text so editing objc.md still invalidates the hash.
3. Content is read at the reviewed ref, not the disk file from working tree. git show <ref>:<path> for range/commit mode (mirroring finalizeDiff), working tree for scan/workspace/rules check. Defaulting to matlab.md on any read failure. NewResolver takes a ResolverOptions{Ref, Runner}. This keeps the decorator's simplicity without the correctness loss a disk read would introduce on refs that aren't checked out — there's a test for exactly that case. Cost is one extra git show per .m file, since finalizeDiff already fetched that blob; happy to drop it for a plain disk read if you'd rather trade that back.

Also fixed a bug the new tests surfaced: callers passed strings.ToLower(path), which was harmless for glob matching but broke content reads for mixed-case paths. The resolver already lowercases internally, so I removed it at the call sites.

Coverage went up — rules 85.8% → 88.2%, delegate 96.4% → 100%. make check and the full suite pass.

One open question: I used a ResolverOptions struct rather than two more positional params on NewResolver. Happy to switch if you prefer.

GenosseFlosse:
@lizhengfeng101 I rebased again so that there are no conflicts with the current main.

GenosseFlosse:
Hi @lizhengfeng101! I addressed your architectural feedback (the sniffer now decorates the system rule layer directly, no more content threading through the resolver interfaces) and have kept it rebased conflict-free against main since.

Let me know if there's anything else you'd like changed, or if this is good to merge. Happy to keep it current if it sits a bit longer. I want to make sure it hasn't fallen off the radar 😄

lizhengfeng101:
> Hi @lizhengfeng101! I addressed your architectural feedback (the sniffer now decorates the system rule layer directly, no more content threading through the resolver interfaces) and have kept it rebased conflict-free against main since.
> 
> Let me know if there's anything else you'd like changed, or if this is good to merge. Happy to keep it current if it sits a bit longer. I want to make sure it hasn't fallen off the radar 😄

@GenosseFlosse Sorry, I’ve been too busy lately and let this slide; I’ll push the merger forward as soon as possible.

lizhengfeng101 (commented):
Went through this one properly — the layering call is right (decorating the *system* layer so user rules still outrank the sniff, rather than wrapping the composed resolver), and `TestSniffer_ReadsAtRefNotCheckedOut` is earning its keep. I ran build, vet and the `internal/config/rules` + `cmd/opencodereview` suites on the head and they're all green, so nothing below is a failing check — it's behavior and contract stuff.

Three things I'd want changed before this lands, then a few small ones.

### The sniff misses most real Objective-C files

`peekFirstLine` only looks at the first non-blank line, and that line is almost never the `#import`. Xcode's file template opens with a `//` header block, and most projects with a license header open with `/*`. I ran both through the resolver as it stands:

```
//
//  ViewController.m
//  MyApp
//
//  Created by ... on 1/1/25.
//

#import "ViewController.h"
```
→ resolves to `matlab.md`

```
/* Copyright 2026. */
#import <Foundation/Foundation.h>
```
→ also `matlab.md`

So in the common case an ObjC file now gets the entire MATLAB checklist thrown at it — `arguments` blocks, `snake_case`, `inv(A)*b`, integer saturation — which is a worse outcome than the pre-PR behavior of falling through to `default.md`.

Widening the peek to N lines is tempting but it's a guess: Xcode's template puts the `#import` on line 8, an Apache header pushes it past line 15, and an internal header with a change-history block goes further still. You don't need to read more lines at all. MATLAB comments start with `%`, and a `.m` file can't legally begin with `/` in MATLAB, so a C-style comment opener is itself a reliable ObjC signal:

```go
// objcSniffPrefixes are first-line signals for Objective-C. MATLAB comments
// start with "%" and a MATLAB file cannot legally begin with "/" at all, so a
// C-style comment opener is itself a reliable ObjC signal — which matters
// because the Xcode file template and most license headers put a comment, not
// a directive, on line 1.
var objcSniffPrefixes = []string{
	"#import", "#include", "#pragma", "#ifdef", "#ifndef",
	"@import", "@interface", "@implementation", "@class", "@protocol",
	"//", "/*",
}
```

Both of those headers are worth adding as test cases. "Signal isn't on line 1" isn't covered anywhere right now, which is how this got through.

### Please keep the sniff result out of `Pattern`

`detail.Pattern += " (sniffed: objc)"` flows into `delegateRuleGroupJSON.Pattern`, i.e. `json:"pattern"` in a payload that ships a `schema_version`. That field's contract is "the glob that matched", and `**/*.m (sniffed: objc)` isn't a glob — so anything downstream that dedupes by pattern, compares it, or copies it into a `.opencodereview/rule.json` to override the rule breaks on it. Worse, the moment someone works out they can grep for `(sniffed: objc)` to check whether the sniff fired, that string is an API, and rewording or removing it becomes a breaking change.

`RuleDetail` lives in `internal/`, so giving it a field of its own costs nothing:

```go
type RuleDetail struct {
	Rule      string
	Source    string
	Pattern   string // stays a plain glob
	SniffedAs string // "" | "objc"; internal only, never serialized
}
```

```go
func (s *sniffer) resolveDetail(path string) RuleDetail {
	detail := s.inner.resolveDetail(path)
	if s.sniffsAsObjC(path) {
		detail.Rule = s.objcRule
		detail.SniffedAs = "objc"
	}
	return detail
}
```

Leave `delegateRuleGroupJSON` untouched, and surface it in `rules check`, where the output isn't a contract anyone should be parsing:

```go
fmt.Printf("Pattern: %s\n", detail.Pattern)
if detail.SniffedAs != "" {
	fmt.Printf("Note:    rule selected by file content (%s), not by path alone\n", detail.SniffedAs)
}
```

`TestSniffer_ResolveDetailAnnotatesPattern` and `TestRunRulesCheck_ObjCSniffOverridesMatlab` would move to asserting on `SniffedAs` / the `Note:` line.

One side effect worth a line in the PR body: with the suffix gone, a repo mixing MATLAB and ObjC `.m` files emits two groups with identical `source` and `pattern` and different `rule`. That's still consistent with `GroupRules`'s doc comment, but it's a shape the system layer couldn't produce before — pattern → rule used to be 1:1 there — so anyone deduping on `source+pattern` will run into it.

### The other three doc locales

`review-rules.md` also exists under `zh`, `ja` and `ru`, and all three carry the same `path_rule_map` table (roughly zh:166, ja:145, ru:185). They're missing the `**/*.m` row and the content-sniffing section entirely.

### Smaller things

`reviewContentRef` is a hand-rolled copy of `tool.ParseReviewMode` + `ReviewMode.RefValue` (`internal/tool/filereader.go:34-54`), and `review_cmd.go:168` already computes `mode` a few lines under the call site. Reusing them deletes the function plus its six-case table test, and removes a pair of copies that have to stay in sync by hand.

The `specialCaseRuleDocs` comment points at `ResolveWithContent` and says the file is read "in `LoadDefault`". Neither exists — it's `sniffer` and `loadObjCRule`. Cheap to fix now, irritating for whoever greps those names in six months.

Same with the new comment in `TestResolve_FallbackToDefault`: it mentions `.swift` now matching `swift.md`, which isn't part of this change. I went looking for that diff before working out it wasn't there.

GenosseFlosse:
Thanks. That was a big line-1 bug. All six items are addressed in two commits on top of the existing one, so the delta since your review is easy to isolate.

The sniff missing real ObjC files: Took your signal set verbatim, including // and /* — the reasoning that MATLAB can't legally begin with / is the part that makes it safe, and it avoids the guess about how deep the directive sits. Both of your header shapes are now test cases (TestSniffer_RealisticObjCHeaders); I confirmed both previously resolved to matlab.md before the fix. You were right that "signal isn't on line 1" had no coverage at all.

Keeping the sniff out of Pattern: Added RuleDetail.SniffedAs exactly as you sketched. delegateRuleGroupJSON is untouched, Pattern is always a plain glob again, and rules check surfaces it as a separate 'Note:' line. The two tests now assert on SniffedAs and the 'Note:' line. I've added your grouping side effect to the PR body — verified against the real binary, two groups with matching source+pattern and different rule.

The other three locales: Added the row and section to zh, ja and ru. One thing worth flagging: generateHeadingId only keeps [a-z0-9] plus CJK ideographs, so Cyrillic and Japanese kana collapse entirely — my ja and ru section headings would both have slugged to a bare m, and the two cross-references would have pointed at nothing. I used the explicit {#id} marker that MarkdownRenderer already supports and tests for this case. That did mean adding an id to the pre-existing ## ルールファイル形式 / ## Формат файла правил headings so the links could target them — happy to drop those links instead if you'd rather I not touch them. zh needed nothing, since ideographs survive slugging.

(While checking that, I also noticed seven pre-existing broken anchors unrelated to this PR (#filtering--rules in all four locales, #content-logging in zh/ja/ru). Left them alone — mentioning in case it's worth a separate issue.)

Smaller things: reviewContentRef is gone; both call sites now use tool.ParseReviewMode(...).RefValue(...), and its table test went with it. Both stale comments fixed — specialCaseRuleDocs now names sniffer.Resolve/loadObjCRule, and the .swift comment (a leftover from resolving an earlier rebase, you were right that it wasn't part of this diff) is removed.

make check and make test pass, and both new commits are independently green.

lizhengfeng101:
Circling back on this — I've praised the rule doc twice without actually going through it line by line. Did that now, and I think a handful of rules need to come out before this lands.

The reason I'd rather not wave it through: `matlab.md` isn't documentation, it's the `{{system_rule}}` substitution in `prompts/main_task_user.md:17`. All 145 lines land in the model's "Review Checklist" verbatim on every `.m` file, so anything in there that isn't a real MATLAB defect becomes a steady source of false positives — which the doc's own preamble ("Favor precision over recall") is trying to prevent.

Concretely, this is idiomatic MATLAB and currently trips three rules:

```matlab
function y = normalizeSignal(x)
    y = x / max(abs(x));
end
```

missing `arguments (Input)`, missing `arguments (Output)`, and `normalizeSignal` not being `snake_case`.

What I'd like removed or reworded:

| Line | Rule | Why |
|---|---|---|
| 7 | German comments / identifiers | Internal styleguide — this ships to every user. |
| 11 | "scripts are not permitted" | Scripts are a first-class MATLAB file type. Fires on every script in every repo. |
| 21 | Mandatory `arguments (Input)` + `(Output)` | Needs a recent release, and requiring it on *every* function flags all existing code. Works as "for new functions, if you target R2019b+". |
| 33 | `end` in the shadowable-builtin list | `end` is a keyword — `end = 5` is a syntax error, so it can't be shadowed. The model is being told to look for something impossible. |
| 34 | `snake_case` for functions/variables | MathWorks' own code and the common community guides use `lowerCamelCase`. |
| 35 | `b_` / `l_` prefixes on logicals | Hungarian notation, not a MATLAB convention. Keep the `is` prefix. |
| 11, 12, 16, 30 | Whole-file rules | Can't be judged from changed lines alone, and can't be anchored on one — conflicts with "Review only the lines changed in this diff" up in the preamble. Either scope them to newly added files or loosen the preamble. |

Everything else stays. The technical content is genuinely good and I don't want to lose any of it — `for k = v` iterating over columns, integer division rounding instead of truncating, `'` vs `.'` on complex data, `inv(A)*b`, `struct("field", someCell)` quietly producing a struct array. I hadn't seen those collected anywhere before and they're exactly the traps worth catching.

If the removed rules matter for your own team, they'd work nicely as a `.opencodereview/rule.json` example in the docs — usable for you, without shipping in the binary.

Also, could you rebase onto latest `main` once more? There's been some movement since your last one. And sorry again for the long gaps on my end.

GenosseFlosse:
Dear @lizhengfeng101 
You're correct. There were parts in the rule doc that are specific to my team and our internal conventions. I removed them and will stick to a local rule for our repositories. 
The rebase was successful, I hope all your critique points were addressed.

lizhengfeng101:
Thanks — the trim is exactly what I was after, and the commit message laying out which rules went and why made it easy to check. Scoping the `arguments` block rule to "new function, in a file that already uses the pattern elsewhere" is better than what I asked for, and dropping "the file should show a clean checkmark" was a catch of your own that I'd missed.

Two things from the rebase, plus one leftover from my list.

**`manual_e2e_retry_test.go` no longer compiles.** The rebase pulled in a file that landed on `main` after your last one, and it calls `loadCommonContext`:

```
$ go vet -tags manual_e2e ./cmd/opencodereview/
vet: cmd/opencodereview/manual_e2e_retry_test.go:50:54: not enough arguments in call to loadCommonContext
	have (string, string, number, number, bool)
	want (string, string, string, int, int, bool)
```

One line. The harness runs in range mode — `ParseReviewMode("HEAD~1", "HEAD", "")` a few lines down, with `FileReader.Ref` hardcoded to `"HEAD"` — so it wants the same ref `executeReview` would compute for those flags:

```go
cc, err := loadCommonContext(repoDir, "", "HEAD", 0, 4, true)
```

An empty ref would compile and the harness would still pass, since the fixture repo is only `a.go`/`b.go` and nothing ever gets sniffed. But the point of that file is that it mirrors `executeReview`'s wiring, and production passes `"HEAD"` there, so let's keep the mirror honest.

To be clear, this isn't on you. The file sits behind `//go:build manual_e2e` and nothing in CI or `make check` passes that tag, so it's invisible to every automated check we have — `go list` doesn't even report it as part of the package. You updated the other seven call sites, including all six in ordinary test files, because those *do* get type-checked. I'll add the tag to the vet line in the Makefile separately so this can't bite again.

**`objcSniffPrefixes` is missing `#if` and `#define`.** My fault — I handed you that list verbatim and never went back over it for completeness. `#if TARGET_OS_IPHONE` and `#define`-first files are both common enough to matter, and since `#ifdef`/`#ifndef` are already prefixed by `#if`, collapsing them covers all three:

```go
"#import", "#include", "#pragma", "#if", "#define",
"@import", "@interface", "@implementation", "@class", "@protocol",
"//", "/*",
```

Worth *not* widening to a bare `"#"`, though — Octave `.m` files legitimately use `#` for comments.

**One rule from my list that didn't get scoped** — `matlab.md` line 39:

```
- A function without a description header, or a header that restates the function name ...
```

Same whole-file problem as the two you did fix, and your own wording works fine here:

```
- In a function added or substantially rewritten by this diff: a function without a description header, or ...
```

Minor one — the "do not request comments on self-explanatory single-purpose lines" at the end of that section already takes some of the edge off.

Finally, two answers I've been sitting on, sorry about that. Keep `ResolverOptions` as a struct — four positional params with two optional ones would read worse. And don't trade the `git show` back for a plain disk read: you're right that `finalizeDiff` has already fetched that blob, but reusing it means threading content into the resolver again, and the ref correctness that `TestSniffer_ReadsAtRefNotCheckedOut` guards is worth more than one subprocess per `.m` file. Leave it as is.

The `objc.md` placeholder is fine — I'll open a separate issue to track filling it in so it doesn't get lost.

GenosseFlosse:
Went through the full thread again end to end to make sure nothing slipped. A few worth calling out:

Ran the manual E2E harness itself this time (not just go vet -tags manual_e2e): all three subtests (clean_first_try_success, recovered_and_failed, all_files_fail) pass against the fake server.
The rebase picked up the grouped-review refactor (executeSubtask → executeGroupSubtask), which meant the whole single-file code path our commit touched no longer exists --> resolved by dropping our copy entirely in favor of the new one. In the process I found resolveGroupSystemRule had reintroduced the strings.ToLower bug. Since it's now the only rule-resolution path left, this would have hit every real review run. Fixed, with a regression test that builds a real sniffer-backed resolver against a temp repo and fails without the fix.
Double-checked all four doc locales stay in sync with the widened #if/#define prefix list and the Octave note — they do.

lizhengfeng101 (approved):
LGTM

PR #586, opened 2026-07-29
fix(config): warn when llm.* settings are silently ignored due to active provider

### LINKED ISSUE #583: fix(config): warn when `llm.*` settings are silently ignored due to active provider
## Description

When a provider is already configured (e.g. via `ocr config provider`), running `ocr config set llm.url <URL>` (or any other `llm.*` key) succeeds silently — but the value is **never used** at runtime.

This happens because the endpoint resolver (`internal/llm/resolver.go:258`) checks `cfg.Provider` first:

```go
if cfg.Provider != "" {
    return tryProviderConfig(cfg, modelOverride) // ← provider wins
}
return tryLegacyLlmConfig(cfg, modelOverride)    // ← llm.* is only a fallback
```

Users who previously ran `ocr config provider` (which sets the `provider` field) and then try to override the URL via `ocr config set llm.url` will see the old provider URL used instead, with no indication that their setting was ignored.

**Reported in #580** — user configured `llm.url` pointing to a custom endpoint but `ocr llm test` still hit the previously-configured dashscope URL.

## Scope

- File(s): `cmd/opencodereview/config_cmd.go`
- Function/area: `runConfigSet` (line 68), `runConfigUnset` (line 96)
- Secondary: no changes needed in `internal/llm/resolver.go` (resolver priority logic is correct by design)

## Acceptance Criteria

- [ ] When `ocr config set llm.*` is called while `cfg.Provider` is non-empty, print a stderr warning like:
  ```
  [ocr] WARNING: provider "dashscope" is active and takes precedence over llm.* settings.
  [ocr] Use 'ocr config set providers.dashscope.<field> <value>' to configure the active provider,
  [ocr] or run 'ocr config unset provider' to disable provider-based config.
  ```
- [ ] Support `ocr config unset provider` — clears the `provider` (and `model`) fields so the resolver falls back to `llm.*` config
- [ ] Add/update unit tests for both behaviors

## Context

Discovered via #580. The resolver's priority order is intentional (provider > llm.* > env > shell rc), but the config CLI gives no feedback when a user writes to a lower-priority layer that will be shadowed. This is a UX gap, not a resolver bug.

aalhadxx:
## Problem

When a provider is already active, writing an `llm.*` setting via the CLI succeeds without error but is discarded at runtime. The resolver checks `cfg.Provider != ""` and invokes `tryProviderConfig`, leaving `tryLegacyLlmConfig` as the fallback path. Users have no idea their setting was ignored.

### Reproduction

1. `ocr config provider` → select any provider (e.g. anthropic)
2. `ocr config set llm.url https://custom.example.com` → succeeds silently
3. At runtime, `llm.url` is ignored because the provider takes priority.

## Fix

- `runConfigSet`: issue a stderr warning when `llm.*` values are written while a provider is present.
- `runConfigUnset`: enable unsetting the `provider` and `model` fields so users can switch back to legacy llm config.

The resolver's priority ordering is intentional and remains untouched.

## Test

Added regression tests:
- `TestRunConfigSetWarnsWhenLlmShadowedByProvider`
- `TestRunConfigSetNoWarningWhenProviderNotActive`
- `TestRunConfigUnsetProvider`
- `TestRunConfigUnsetModel`
- `TestRunConfigUnsetInvalidKey`

Fixes #583

### DIFF
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -93,7 +93,7 @@ func runConfigSet(key, value string) error {
 
 	// Warn when llm.* settings are shadowed by an active provider.
 	if cfg.Provider != "" && strings.HasPrefix(key, "llm.") {
-		fmt.Fprintf(os.Stderr, "[ocr] WARNING: %q is ignored because provider %q is active. Unset the provider first if you want to use legacy llm config.\n", key, cfg.Provider)
+		fmt.Fprintf(os.Stderr, "[ocr] WARNING: %q will have no effect while provider %q is active. The value is saved but will not be used until the provider is unset.\n", key, cfg.Provider)
 	}
 	return nil
 }
@@ -112,6 +112,9 @@ func runConfigUnset(key string) error {
 		}
 		switch key {
 		case "provider":
+			if cfg.Model != "" {
+				fmt.Fprintf(os.Stderr, "[ocr] WARNING: 'model' has also been cleared because it is tied to the provider.\n")
+			}
 			cfg.Provider = ""
 			cfg.Model = ""
 		case "model":
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -90,18 +90,45 @@ func runConfigSet(key, value string) error {
 		displayValue = maskKey(value)
 	}
 	fmt.Printf("Set %s = %s\n", key, displayValue)
+
+	// Warn when llm.* settings are shadowed by an active provider.
+	if cfg.Provider != "" && strings.HasPrefix(key, "llm.") {
+		fmt.Fprintf(os.Stderr, "[ocr] WARNING: %q is ignored because provider %q is active. Unset the provider first if you want to use legacy llm config.\n", key, cfg.Provider)
+	}
 	return nil
 }
 
 func runConfigUnset(key string) error {
-	parts := strings.SplitN(key, ".", 2)
-	if len(parts) != 2 || parts[1] == "" {
-		return fmt.Errorf("unset supports custom_providers.<name> and mcp_servers.<name>")
-	}
-
 	configPath, err := defaultConfigPath()
 	if err != nil {
 		return err
+	}
+
+	// Top-level keys that do not contain a dot.
+	if !strings.Contains(key, ".") {
+		cfg, err := loadOrCreateConfig(configPath)
+		if err != nil {
+			return fmt.Errorf("load config: %w", err)
+		}
+		switch key {
+		case "provider":
+			cfg.Provider = ""
+			cfg.Model = ""
+		case "model":
+			cfg.Model = ""
+		default:
+			return fmt.Errorf("unset supports provider, model, custom_providers.<name> and mcp_servers.<name>")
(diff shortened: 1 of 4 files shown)


lizhengfeng101 (commented):
## Code Review

### Overview

This PR addresses a real UX pain point: `llm.*` settings are silently ignored when a provider is active. The fix is conservative — it adds warnings without touching the resolver's priority logic — and comes with solid test coverage.

However, the PR bundles an unrelated VS Code fix and has a subtle bug in the `unset model` path.

---

### Issues

#### 1. Unrelated change: VS Code `shell: true` fix

The `CliService.ts` changes (Windows `shell` option) have nothing to do with config warnings. Mixing unrelated fixes complicates reverts and muddies git blame.

**Suggestion**: Split into a separate PR with its own issue reference.

#### 2. `runConfigUnset("model")` does not clear top-level `cfg.Model` when a provider is active

```go
case "model":
    if cfg.Provider != "" {
        // Only clears the provider entry's Model field
        // Top-level cfg.Model is NOT cleared
    } else {
        cfg.Model = ""
    }
```

If a user previously set both a top-level `Model` and a provider entry model, the top-level value will persist as a stale ghost. The test passes because the fixture happens to have `Model: "claude-opus-4-6"` which gets serialized with `omitempty` (or similar) — but the in-memory struct still holds the old value.

**Suggestion**: Also set `cfg.Model = ""` in the provider-active branch to avoid residual state.

#### 3. `os.Stderr` redirection in tests is fragile

```go
oldStderr := os.Stderr
r, w, _ := os.Pipe()
os.Stderr = w
```

This mutates global state. If anyone adds `t.Parallel()` in the future, tests will race. Consider injecting an `io.Writer` parameter into `runConfigSet`/`runConfigUnset`, or at minimum document that these tests must not run in parallel.

#### 4. `os.Pipe()` error is discarded

```go
r, w, _ := os.Pipe()
```

Minor, but should be:
```go
r, w, err := os.Pipe()
if err != nil {
    t.Fatalf("os.Pipe: %v", err)
}
```

#### 5. `unset provider` silently clears `model` — document this

The behavior of clearing model when unsetting provider is reasonable, but it should be mentioned in CLI help text or docs so users aren't surprised.

---

### Nits

- VS Code test: `expect.not.objectContaining({ shell: true })` works but `expect.objectContaining({ shell: false })` would be more explicit about intent.

---

### Summary

| Aspect | Verdict |
|--------|---------|
| Correctness | Bug in `unset model` path (residual top-level Model) |
| Scope | VS Code fix should be a separate PR |
| Tests | Good coverage, but fragile stderr pattern |
| Style | Clean, follows project conventions |

**Recommendation**: Split the VS Code fix, address the `cfg.Model` residual bug, then this is good to merge.

lizhengfeng101:
Hi @aalhadxx, thank you for your contribution to this issue! We appreciate the effort you put into this PR.

However, since we haven't heard back from you regarding the review feedback, and the CI checks are currently failing, we've decided to move forward and merge #588 which addresses the same problem.

That said, we'd love to see you continue contributing! Feel free to pick up any other open issues or propose new improvements — contributions of all kinds are welcome. Thanks again!

PR #587, opened 2026-07-29
feat(llm): add OpenAI Responses provider with GPT-5.6 models

### LINKED ISSUE #559: OpenAI GPT-5.6 tool calls fail during ocr review because Chat Completions is used
### OpenCodeReview Version

v1.8.0 (commit: e670b3b3, built: 2026-07-28T02:27:56Z)

### Operating System

Windows

### Installation Method

npm (global)

### LLM Provider

OpenAI

### Bug Description

When the built-in OpenAI provider is configured with `gpt-5.6-terra`, `ocr review` completes the planning phase but fails when a review subtask sends function tools.

The request is sent to `/v1/chat/completions`. OpenAI then returns a 400 error stating that function tools with reasoning effort are not supported for this model through Chat Completions and that `/v1/responses` should be used instead.

This is difficult for a new user to diagnose because:

- `gpt-5.6-terra` is not included in the built-in OpenAI model presets.
- The configuration flow does not explain that GPT-5.6 models need the Responses API for the tool-calling workflow used by `ocr review`.
- The review plan can complete successfully before the tool-using subtask fails.
- I could not find a FAQ entry explaining this error or how to configure the Responses API.

The same support is needed for the current GPT-5.6 variants:

- `gpt-5.6-luna`
- `gpt-5.6-terra`
- `gpt-5.6-sol`

### Steps to Reproduce

1. Run `ocr config provider`.
2. Select the built-in OpenAI provider and configure an OpenAI API key.
3. Run `ocr config model`.
4. Manually add `gpt-5.6-terra`, because it is not present in the built-in model list.
5. Run `ocr review` on a repository containing reviewable changes.
6. Wait for the review plan to complete.
7. Observe that the following tool-using subtask fails with a 400 response from `/v1/chat/completions`.

### Expected Behavior

A user selecting a supported GPT-5.6 model should be able to run `ocr review` without manually knowing which OpenAI API protocol is required.

Ideally, the built-in OpenAI configuration should:

- Include `gpt-5.6-luna`, `gpt-5.6-terra`, and `gpt-5.6-sol`.
- Use the Responses API for these models when no explicit protocol override is configured.
- Preserve the existing behavior of other OpenAI models and configurations.

At minimum, if the selected provider and model combination cannot support the tool-calling workflow, the configuration or connectivity test should fail early with an actionable message instead of failing after the review plan has completed.

### Logs / Error Output

```shell
[ocr] Plan completed for path/to/file
[ocr] Subtask error for path/to/file: LLM completion error: POST "https://api.openai.com/v1/chat/completions": 400 Bad Request {
    "message": "Function tools with reasoning_effort are not supported for gpt-5.6-terra in /v1/chat/completions. To use function tools, use /v1/responses or set reasoning_effort to 'none'.",
    "param": "reasoning_effort",
    "code": null
}
```

### Additional Context

OpenCodeReview already supports the Responses API through the `openai-responses` protocol, introduced in:

- https://github.com/alibaba/open-code-review/pull/363

Documentation for the `openai-responses` protocol is also being tracked in:

- https://github.com/alibaba/open-code-review/issues/556

However, this capability is not discoverable through the normal built-in OpenAI provider and model-selection flow. The current FAQ also does not provide troubleshooting guidance for this error. This leaves new users without an actionable recovery path after the failure.

One possible solution would be to add the three exact GPT-5.6 model IDs to the built-in OpenAI provider and associate them with the Responses protocol through explicit per-model metadata. Existing models could continue using the provider's current default protocol, and an explicitly configured protocol could continue to take precedence.

Another option would be a separate built-in OpenAI Responses provider, although users selecting the ordinary OpenAI provider could still encounter the same configuration trap.

Setting `reasoning_effort` to `none` may bypass this particular API restriction, but it disables reasoning rather than addressing the protocol mismatch. Adding a first-class `reasoning_effort` setting is not part of this report.

I have not started an implementation yet and would be happy to submit a focused change with tests and documentation after the preferred approach is confirmed.

### LINKED ISSUE #573: feat(llm): add GPT-5.6 models to built-in OpenAI presets with automatic openai-responses protocol
## Description

When a user configures a GPT-5.6 model (`gpt-5.6-luna`, `gpt-5.6-terra`, `gpt-5.6-sol`) through the built-in OpenAI provider, `ocr review` fails at the tool-calling subtask phase with a 400 error because the request is sent to `/v1/chat/completions` instead of `/v1/responses`.

The `openai-responses` protocol is already implemented (PR #363), but it is not discoverable through the standard provider/model configuration flow. Users should not need to know which API protocol a model requires.

## Scope

- File(s):
  - `internal/llm/providers.go` — built-in provider registry (Go, authoritative)
  - `extensions/vscode/src/shared/providers.ts` — VSCode extension provider presets (must stay in sync)
  - `internal/llm/client.go` — client factory / protocol dispatch
- Area: LLM provider configuration, model-to-protocol mapping

## Requirements

1. **Add GPT-5.6 models to the built-in OpenAI preset**: `gpt-5.6-luna`, `gpt-5.6-terra`, `gpt-5.6-sol`.
2. **Automatic protocol selection**: When a GPT-5.6 model is selected under the built-in OpenAI provider, the system should automatically use the `openai-responses` protocol instead of `openai` (Chat Completions). This should not require any manual configuration from the user.
3. **Preserve existing behavior**: All other models (e.g., `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`) must continue using the Chat Completions protocol as before. An explicitly configured protocol override must still take precedence.
4. **Keep Go and TypeScript presets in sync**: Changes to `internal/llm/providers.go` must be mirrored in `extensions/vscode/src/shared/providers.ts`.

## Suggested Approach

Add per-model metadata (e.g., a `Protocol` field on each model entry, or a model-name prefix match in the client factory) so that `NewLLMClient` can select `openai-responses` for GPT-5.6 models without requiring the user to change the provider-level protocol setting.

## Acceptance Criteria

- [ ] `gpt-5.6-luna`, `gpt-5.6-terra`, `gpt-5.6-sol` appear in `ocr config model` when the OpenAI provider is selected
- [ ] `ocr review` with a GPT-5.6 model sends requests to `/v1/responses` (not `/v1/chat/completions`)
- [ ] Existing models still use Chat Completions by default
- [ ] An explicit `protocol: openai` override still forces Chat Completions (no silent override)
- [ ] VSCode extension presets are updated to match

## Context

Reported in #559 — thanks to @JH-Wen for the detailed bug report with clear reproduction steps and analysis. This issue tracks the implementation fix.

Related:
- PR #363 (introduced `openai-responses` protocol support)
- #556 (documentation for `openai-responses` protocol)

aalhadxx:
## Problem

GPT-5.6 models (gpt-5.6-luna, gpt-5.6-terra, gpt-5.6-sol) require the OpenAI Responses API (/v1/responses). When these models are used with the standard `openai` provider (which uses `/v1/chat/completions`), the platform returns:

```
Function tools with reasoning_effort are not supported for gpt-5.6-terra in /v1/chat/completions.
```

## Fix

Add a dedicated `openai-responses` preset provider that uses `ProtocolOpenAIResponses` and includes the GPT-5.6 model family. Users can now select this provider to use these models without manual protocol configuration.

## Test

- Updated `TestListProviders_Order` to include the new provider.
- Added `TestLookupProvider_OpenAIResponsesModels` to verify the provider exists, uses the correct protocol, and lists the expected models.

Fixes #573
Closes #559

### DIFF
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -93,7 +93,7 @@ func runConfigSet(key, value string) error {
 
 	// Warn when llm.* settings are shadowed by an active provider.
 	if cfg.Provider != "" && strings.HasPrefix(key, "llm.") {
-		fmt.Fprintf(os.Stderr, "[ocr] WARNING: %q is ignored because provider %q is active. Unset the provider first if you want to use legacy llm config.\n", key, cfg.Provider)
+		fmt.Fprintf(os.Stderr, "[ocr] WARNING: %q will have no effect while provider %q is active. The value is saved but will not be used until the provider is unset.\n", key, cfg.Provider)
 	}
 	return nil
 }
@@ -112,6 +112,9 @@ func runConfigUnset(key string) error {
 		}
 		switch key {
 		case "provider":
+			if cfg.Model != "" {
+				fmt.Fprintf(os.Stderr, "[ocr] WARNING: 'model' has also been cleared because it is tied to the provider.\n")
+			}
 			cfg.Provider = ""
 			cfg.Model = ""
 		case "model":
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -90,18 +90,45 @@ func runConfigSet(key, value string) error {
 		displayValue = maskKey(value)
 	}
 	fmt.Printf("Set %s = %s\n", key, displayValue)
+
+	// Warn when llm.* settings are shadowed by an active provider.
+	if cfg.Provider != "" && strings.HasPrefix(key, "llm.") {
+		fmt.Fprintf(os.Stderr, "[ocr] WARNING: %q is ignored because provider %q is active. Unset the provider first if you want to use legacy llm config.\n", key, cfg.Provider)
+	}
 	return nil
 }
 
 func runConfigUnset(key string) error {
-	parts := strings.SplitN(key, ".", 2)
-	if len(parts) != 2 || parts[1] == "" {
-		return fmt.Errorf("unset supports custom_providers.<name> and mcp_servers.<name>")
-	}
-
 	configPath, err := defaultConfigPath()
 	if err != nil {
 		return err
+	}
+
+	// Top-level keys that do not contain a dot.
+	if !strings.Contains(key, ".") {
+		cfg, err := loadOrCreateConfig(configPath)
+		if err != nil {
+			return fmt.Errorf("load config: %w", err)
+		}
+		switch key {
+		case "provider":
+			cfg.Provider = ""
+			cfg.Model = ""
+		case "model":
+			cfg.Model = ""
+		default:
+			return fmt.Errorf("unset supports provider, model, custom_providers.<name> and mcp_servers.<name>")
+		}
+		if err := saveConfig(configPath, cfg); err != nil {
+			return err
+		}
+		fmt.Printf("Unset %s\n", key)
+		return nil
+	}
+
+	parts := strings.SplitN(key, ".", 2)
+	if len(parts) != 2 || parts[1] == "" {
+		return fmt.Errorf("unset supports provider, model, custom_providers.<name> and mcp_servers.<name>")
 	}
 
 	switch parts[0] {
@@ -110,7 +137,7 @@ func runConfigUnset(key string) error {
 	case "mcp_servers":
 		return unsetMCPServer(configPath, parts[1])
 	default:
-		return fmt.Errorf("unset supports custom_providers.<name> and mcp_servers.<name>")
+		return fmt.Errorf("unset supports provider, model, custom_providers.<name> and mcp_servers.<name>")
 	}
 }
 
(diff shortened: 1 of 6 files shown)


lizhengfeng101:
Hi @aalhadxx, thanks for adding the `openai-responses` provider — the core implementation looks clean and well-tested.

However, this PR bundles three unrelated changes together:

1. **`openai-responses` provider** (the intended feature) ✅
2. **`config set/unset` improvements** — these conflict with changes already merged on `main` (e.g., `legacyLLMShadowWarning`, `unsetActiveProvider` from #588). The inline warning in this PR is a simpler duplicate of what's already landed.
3. **VSCode `CliService` Windows `shell` fix** — unrelated to the provider feature.

**Suggestion:** Could you please:

1. **Rebase onto latest `main`** to pick up the already-merged config improvements.
2. **Keep only the core `openai-responses` provider changes** (`internal/llm/providers.go`, `internal/llm/providers_test.go`).
3. **Remove the config and VSCode changes** from this PR. If those fixes are still needed after rebasing, they'd be better as separate PRs.

This will make the PR much easier to review and merge. Thanks!

aalhadxx:
Closing this PR because upstream already merged OpenAI Responses API support in #363 (f58b0f2). The feature is now available in main. Thanks for the review feedback regardless!

PR #605, opened 2026-07-30
feat(config): resolve api_key/auth_token from a command (#236)

### LINKED ISSUE #236: Support OS keyring auth for local LLM credentials
## Problem Statement

OpenCodeReview currently documents and examples LLM credentials primarily through configuration files and environment variables such as `OCR_LLM_AUTH_TOKEN`. This works well in CI, but it is not ideal for local CLI usage on developer machines.

Environment variables can leak through shell history, child process inheritance, process inspection, crash logs, and CI logs. Plain config files are also risky unless users manage permissions carefully.

## Proposed Solution

Add first-class OS keyring support for local credentials.

The CLI could provide commands such as:

    ocr auth key save <name> <token>
    ocr auth key save <name> --stdin
    ocr auth key save <name> --prompt
    ocr auth key list
    ocr auth key delete <name>

The direct form includes the token value; the stdin/prompt forms are safer because they avoid shell history and process-list exposure. Configuration could reference a named credential instead of storing the secret value. At runtime, auth resolution could use this precedence:

1. Explicit CLI option.
2. Named keyring credential.
3. Existing config value.
4. Environment variable fallback.

## Suggested design

- Use a cross-platform keyring library where possible.
- Support macOS Keychain, Windows Credential Manager, and Linux Secret Service.
- Detect unavailable, denied, or locked keyrings with actionable messages.
- Keep environment variables supported for CI.
- Consider an encrypted file fallback for headless Linux if implemented carefully with restrictive permissions and clear warnings.

## Reference implementation

LLxprt Code has a similar multi-platform design using `@napi-rs/keyring` with an encrypted fallback. It stores only a key name in profiles and resolves the secret at runtime.

Useful references at commit `a50837636f124130cf2d6973d0641e2f4eae02c2`:

- `SecureStore` adapter/options: https://github.com/vybestack/llxprt-code/blob/a50837636f124130cf2d6973d0641e2f4eae02c2/packages/core/src/storage/secure-store.ts#L56-L75
- Dynamic `@napi-rs/keyring` loader: https://github.com/vybestack/llxprt-code/blob/a50837636f124130cf2d6973d0641e2f4eae02c2/packages/core/src/storage/secure-store.ts#L156-L218
- Keyring availability probe: https://github.com/vybestack/llxprt-code/blob/a50837636f124130cf2d6973d0641e2f4eae02c2/packages/core/src/storage/secure-store.ts#L347-L399
- Keyring-first get/set with encrypted fallback: https://github.com/vybestack/llxprt-code/blob/a50837636f124130cf2d6973d0641e2f4eae02c2/packages/core/src/storage/secure-store.ts#L404-L500
- `/key save` command: https://github.com/vybestack/llxprt-code/blob/a50837636f124130cf2d6973d0641e2f4eae02c2/packages/cli/src/ui/commands/keyCommand.ts#L66-L145
- `/key load` command: https://github.com/vybestack/llxprt-code/blob/a50837636f124130cf2d6973d0641e2f4eae02c2/packages/cli/src/ui/commands/keyCommand.ts#L155-L217
- `auth-key-name` resolution before environment variables: https://github.com/vybestack/llxprt-code/blob/a50837636f124130cf2d6973d0641e2f4eae02c2/packages/core/src/auth/precedence.ts#L645-L650
- Documentation explaining why keyring is preferred over environment variables: https://github.com/vybestack/llxprt-code/blob/a50837636f124130cf2d6973d0641e2f4eae02c2/docs/cli/authentication.md#L62-L134

## Acceptance criteria

- Users can save and reference an LLM auth token by name without putting the raw token in config, shell history, or environment variables.
- Existing environment variable workflows continue to work.
- Missing, denied, or locked keyring errors are clear and actionable.
- Documentation explains local keyring usage and CI secret usage separately.

<details>
<summary>中文说明</summary>

## 问题陈述

OpenCodeReview 目前主要通过配置文件和环境变量（例如 `OCR_LLM_AUTH_TOKEN`）来记录和演示 LLM 凭据。这种方式在 CI 环境中很好用，但在开发者本地 CLI 场景中并不理想。

环境变量可能通过 shell 历史记录、子进程继承、进程检查、崩溃日志以及 CI 日志等途径泄露。如果用户没有仔细管理文件权限，明文配置文件同样存在风险。

## 建议方案

为本地凭据提供一等公民的操作系统 keyring 支持。

CLI 可以提供类似下面的命令：

    ocr auth key save <name> <token>
    ocr auth key save <name> --stdin
    ocr auth key save <name> --prompt
    ocr auth key list
    ocr auth key delete <name>

直接形式包含 token 值；stdin/prompt 形式更安全，因为可以避免写入 shell 历史记录或暴露在进程列表中。配置中可以引用一个命名凭据，而不是直接保存密钥明文。运行时的认证解析可以采用以下优先级：

1. 显式 CLI 参数。
2. 命名的 keyring 凭据。
3. 现有配置值。
4. 环境变量回退。

## 设计建议

- 尽量使用跨平台 keyring 库。
- 支持 macOS Keychain、Windows Credential Manager 和 Linux Secret Service。
- 当 keyring 不可用、被拒绝访问或处于锁定状态时，给出可操作的错误提示。
- 继续保留环境变量支持，以适配 CI 场景。
- 对于无桌面环境的 headless Linux，可以考虑加密文件回退，但需要严格的文件权限和清晰的安全提示。

## 参考实现

LLxprt Code 已经有类似的多平台设计：使用 `@napi-rs/keyring`，并提供加密文件回退。它在 profile 中只保存密钥名称，在运行时解析实际密钥。

上面的英文部分列出了具体代码和文档链接。

## 验收标准

- 用户可以按名称保存和引用 LLM auth token，不需要把原始 token 放进配置文件、shell 历史记录或环境变量。
- 现有环境变量工作流继续可用。
- 当 keyring 缺失、拒绝访问或被锁定时，错误信息清晰且可操作。
- 文档分别说明本地 keyring 使用方式和 CI secret 使用方式。

</details>

chethanuk:
## Description

Adds `api_key_cmd` (provider entries) and `auth_token_cmd` (legacy `llm` block) so the LLM credential can be fetched from a secret manager at review time instead of stored in plaintext in `config.json` — the same pattern as `git credential.helper` and AWS `credential_process`.

```bash
ocr config set providers.anthropic.api_key_cmd "op read op://dev/anthropic/api-key"
```

Today the only options are a plaintext `api_key` in `config.json` or a preset environment variable. Neither keeps the secret off disk on a shared or backed-up machine.

### Why a command rather than a keyring library

#236 proposes an `ocr auth key save/list/delete` surface over a cross-platform keyring binding, with an encrypted-file fallback for headless Linux. A command hits the same acceptance criteria without OCR owning any of that: the OS keyring is already reachable through the tool each platform ships with, so there is no binding to vendor, no fallback to implement, and no keyring-locked/denied/unavailable error taxonomy to maintain — the helper reports its own.

```bash
# macOS Keychain
ocr config set providers.anthropic.api_key_cmd "security find-generic-password -s ocr-anthropic -w"

# Linux (Secret Service: GNOME Keyring, KWallet, …)
ocr config set providers.anthropic.api_key_cmd "secret-tool lookup service ocr-anthropic"
```

It also covers the managers a keyring binding would not: 1Password, `pass`, `gopass`, Vault. Config stores a command, never the secret; the value resolves at runtime; environment variables keep working for CI. Both are documented (en/zh/ja).

What this does not add is the `ocr auth key save` half — writing the credential into the keyring. That stays with the platform's own tooling (`security add-generic-password`, `secret-tool store`). Say the word if you would rather have the wrapper commands too and I will follow up; the resolution path here does not change either way.

### Precedence

One resolution site, presets and custom providers alike:

```text
static api_key  → wins (stderr warning if a command is also set)
api_key_cmd     → runs; any failure is a hard error
preset env var  → only when neither of the above is set
                → error
```

A failing command is never a silent fallback to the environment variable. That is the point: a helper that fails because a vault is locked must not quietly send a stale `$ANTHROPIC_API_KEY` and produce a review billed to the wrong account. `auth_token_cmd` mirrors this, with one difference — an *incomplete* legacy block (no `url`, no `model`) falls through to later strategies without running the command, so an unused legacy stanza cannot trigger a credential prompt.

The resolved value is used in memory only: never written back to config, never logged. It resolves once per process, and only after the cheap config validation has passed, so `ocr review --model nonexistent` fails on the model name instead of first prompting for Touch ID.

### Backward compatibility

Both keys are new and `omitempty`, so an existing `config.json` round-trips byte-identically. Every added branch is guarded on a non-empty command string, so behavior is unchanged for anyone who does not set them. `ocr config set` accepts both and does not mask them on read-back — a command line is not itself a secret.

Since the value is executed as a shell command, `config.json` becomes trusted input. Documented alongside the `0600` mode OCR already writes it with.

## Type of Change

- [x] Documentation update
- [x] CI / Build / Tooling

## Hardening found while validating

Each of these was a way the feature did not actually work end to end:

- **The 60s timeout was not a bound.** It killed the shell, but a helper leaving a background process on the inherited stdout pipe (`gpg-agent`, `pinentry`, a first-use `op` daemon) kept `Cmd.Wait` blocked on the read. `api_key_cmd = "sleep 200 & printf tok"` hung for over 90 seconds. Fixed by buffering stdout through a writer `os/exec` copies in its own goroutine and setting `WaitDelay`, which is what lets `Wait` force the pipe closed.
- **stdin was `/dev/null`**, so a helper needing a passphrase saw EOF, or refused to prompt at all for lack of a tty. Now inherits `os.Stdin`, which is safe because no path resolves an endpoint while the bubbletea TUI is reading stdin.
- **Output was unbounded.** `cat /dev/urandom` grew the heap without limit. Capped at 64KiB, refusing the write so the child dies of `SIGPIPE`.
- **Control bytes reached the `Authorization` header**, where `net/http` rejects them as an opaque `invalid header field value`. Now rejected up front with the offending byte and offset, matching `httpguts.ValidHeaderFieldValue`. A lone interior CR survived both `TrimRight` and `TrimSpace`, so it is caught as multi-line output.
- **A whitespace-only credential out-ranked a working command** and sent `Authorization: Bearer  `, unrecoverable without hand-editing the config. Whitespace-only now normalizes to unset for the static key, the env var and the Manual TUI tab's token.
- **`ocr config provider` rejected `api_key_cmd`-only providers** in both directions — non-interactively `applyOfficialProviderConfig` demanded a static key or env var, and interactively the API-key step could not be confirmed because the field renders blank for such a provider. Both now treat a configured command as satisfying the requirement, and errors name the option that would fix them.

`cloneProviderEntry` also has to carry the new field, since it copies `ProviderEntry` member by member and would otherwise erase `api_key_cmd` whenever a provider is edited through `ocr config provider`. That is the same shape as the `timeout_sec`/`extra_headers` drop fixed upstream in the meantime; the regression test here walks the struct with reflection, so the next added field cannot be dropped silently either.

### Windows

The command line goes to `cmd.exe` through `SysProcAttr.CmdLine` with `/S`, not through `Args`. `os/exec` quotes `Args` with `syscall.EscapeArg`, which targets `CommandLineToArgvW`; [`exec.Command`'s own documentation](https://pkg.go.dev/os/exec#Command) names `cmd.exe` as an exception with a different unquoting algorithm and says to supply the full command line yourself. Without this, `op read "op://Private/My Vault/api-key"` arrives as a single literal filename.

A command string is not portable between the two arms — `%VAR%` and `^` are `cmd.exe` metacharacters, `$VAR` and `\` escaping do not apply — so an sh-authored `api_key_cmd` generally needs a Windows-specific rewrite. Documented.

## Why this adds a `windows-latest` job

I would rather not ship a Windows code path that no test has ever executed. The existing `cross-compile` job runs `go build -o /dev/null ./...` under `GOOS=windows`, which proves the arm *compiles* and nothing more — before this PR, `keycmd_windows.go` had zero executed coverage on any platform, and the five Windows-only assertions in `encodeRepoPath` had never run either.

That gap was not theoretical. On its first green run the job found a **pre-existing bug**: `TestResolveBackgroundFilePath/absolute_unchanged` was passing `filepath.FromSlash("/etc/context.md")`, which is rooted but *not* absolute on Windows (`filepath.IsAbs` wants a volume), so the subtest named "absolute" had been silently exercising the relative branch. It also confirmed every `cmd.exe` row in the new test table passes on a real runner rather than on my reasoning about `cmd /?`.

Implementation notes:

- Installs Go via `setup-go` rather than reusing the shared `golang:1.26.5` image, because GitHub does not support `container:` on Windows runners — the runner errors with *"container operations are only supported on Linux runners"* ([actions/runner#904](https://github.com/actions/runner/issues/904), [#1402](https://github.com/actions/runner/issues/1402); the PR that attempted it, [#1801](https://github.com/actions/runner/pull/1801), was never merged). Windows containers also cannot run on an `ubuntu` host, so there is no way to matrix this in from the Linux job.
- No `-race`: the detector needs a C toolchain on Windows, and races are OS-independent, so the Linux job already covers them.
- No coverage gate: the `//go:build !windows` test files legitimately put the total under the 80% the Linux job enforces.

Six existing tests needed a guard, none of them a behavior change: three assert an unreadable path is skipped, but `Chmod(0000)` on Windows only sets the read-only bit — and their existing `os.Getuid() == 0` guard cannot cover that, since `Getuid` returns `-1` there and never `0`; `TestSaveConfig` asserts the `0600` the config is written with, which Windows reports as `0666`; the symlink-safety test needs `SeCreateSymbolicLinkPrivilege`, which an unelevated CI account lacks (six sibling symlink tests already skip for this); and the background-path case above.

**One thing to flag for the maintainers:** every other job in this repo is `runs-on: self-hosted`, and this is the first GitHub-hosted one. If GitHub-hosted runners are disabled or unbilled for the org, the job will queue rather than run. I have verified it green end to end on a fork ([run log](https://github.com/chethanuk/open-code-review/actions/runs/30526695746)), so if you would prefer it on a self-hosted Windows runner, dropped, or split into its own PR, say which and I will adjust — the feature commit does not depend on it.

## How Has This Been Tested?

- [x] Manual testing (below)

Table-driven runner matrix on both arms: success, whitespace and CRLF trimming, no trailing newline, non-zero exit, command-not-found, empty, whitespace-only, multi-line, interior CR, NUL/VT/FF/DEL control bytes, interior TAB preserved, output exactly at the 64KiB cap and one byte over, timeout, `WaitDelay` bounding an orphan holding the pipe, and inherited stdin. Plus resolver precedence rows, legacy fall-through, "runs exactly once per process", and "an invalid model never runs the command".

Verified end to end against a local fake LLM server, including reproducing the >90s hang that motivated the `WaitDelay` fix and the prompt-before-validation ordering bug.

Rebased onto current `main`. `gofmt -s -l` clean · `go vet ./...` clean · `go test -race -count=1 ./...` green · coverage 91.3% (gate 80%) · `make check` clean · `GOOS=windows go vet ./...` clean. Both commits independently pass build, vet and test.

## Checklist

- [x] I have signed the CLA

## Limitations

- The VS Code extension's config parser does not read the new keys, so an `api_key_cmd`-only config still reads as unconfigured there. Out of scope here; happy to file it separately.
- No caching — the command runs once per `ocr` invocation.

## Related Issues

Closes #236


internal/llm/keycmd_windows.go
```diff
@@ -0,0 +1,38 @@
...
+	// Still built by CommandContext so ctx cancellation and WaitDelay behave
+	// exactly as on Unix; only the command-line construction differs.
+	c := exec.CommandContext(ctx, "cmd.exe")
+	// CmdLine is the whole command line including argv[0]; the executable itself
+	// still comes from c.Path. Args stays at Command's default ([]string{"cmd.exe"})
+	// rather than nil: syscall.StartProcess uses SysProcAttr.CmdLine verbatim when
+	// non-empty and never looks at argv, so the doc's "leaving Args empty" is not
+	// load-bearing here -- and a one-element Args keeps Cmd.String() from panicking
+	// on Args[1:].
+	c.SysProcAttr = &syscall.SysProcAttr{CmdLine: `cmd.exe /S /C "` + cmd + `"`}
```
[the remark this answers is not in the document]
chethanuk: The behaviour is real, but it isn't a privilege boundary being crossed.

`api_key_cmd` is a command line the config author asked us to run, so they already have arbitrary execution by design — `api_key_cmd = "whoami"` is a supported configuration, and the Unix arm hands the same string to `sh -c`. The value is read only from the user-level `~/.opencodereview/config.json`, which OCR writes `0600` and which the docs call out as trusted input for exactly this reason. It is never read from the repository under review.

Escaping the inner quotes would defeat the single case `SysProcAttr.CmdLine` exists for: `op read "op://Private/My Vault/api-key"` has to arrive as a quoted argument rather than one literal filename.

What `/S` does with each quote shape is pinned in `keycmd_windows_test.go`, running on a real `windows-latest` runner:

- `echo sk-"a b"-token` → quotes and spaces survive verbatim
- `echo A" & echo B` → a bare interior quote leaves the `&` inside a quoted region, so it stays one command
- `echo A"" & echo B` → a doubled quote closes that region and the `&` does split the line

That last row is the shape this comment describes, asserted rather than reasoned about.

chethanuk:
@lizhengfeng101 Can you merge?

chethanuk:
@lizhengfeng101 Can you merge?

lizhengfeng101:
**Duplicate SPDX license header in `internal/config/testconnection/testconnection.go`**

The file currently has the SPDX/copyright header block duplicated at the top:

```go

// Package testconnection loads the LLM test connection task configuration.
package testconnection
```

This looks like a rebase/merge artifact rather than an intentional change — the file only needs one header. Please remove the duplicate block before merging.

pages/src/content/docs/en/configuration.md
```diff
@@ -151,6 +149,54 @@ The `timeout_sec` keys are not supported by `ocr config set` — edit
 }
 ```
 
+### API key from a command
```
lizhengfeng101, line 152: @chethanuk Missing Russian (ru) localization.

chethanuk: Updated

chethanuk:

lizhengfeng101 (approved):
LGTM

Issue #579, opened 2026-07-29
Clarify ref comparison behavior


philippgille:
### Issue Type

Unclear / confusing explanation

### Document Location

README.md, https://open-codereview.ai/docs/quickstart, https://github.com/alibaba/open-code-review/blob/3017dc160bc1171e228481d838e8c67e3eaaf72f/cmd/opencodereview/flags.go

### Description

There are three conflicting info for the usage of `ocr review --from x --to y`:

#### README

https://github.com/alibaba/open-code-review/blob/3017dc160bc1171e228481d838e8c67e3eaaf72f/README.md?plain=1#L138-L139

Currently says

```
# Branch range — compare two refs
ocr review --from main --to feature-branch
```

It's not clear if it uses `x..y` (two dots, bad if x already advanced on its own) or `x...y` (three dots, merge base mode).

#### Website docs / Quickstart

https://open-codereview.ai/docs/quickstart

Currently says

```
# Branch range — reviews `main..feature-branch`
ocr review --from main --to feature-branch
```

So it says it's using `a..b` (two dots), which is bad if a advanced independently of the feature branch b.

#### CLI help `ocr review --help`

https://github.com/alibaba/open-code-review/blob/3017dc160bc1171e228481d838e8c67e3eaaf72f/cmd/opencodereview/flags.go#L206-L207

It says:

```
  # Review a branch against its base (merge-base mode)
  ocr review --from master --to dev-ref
```

Here it says it's using the merge-base mode, so `a...b` (three dots) which is the *opposite* of what the website docs say.

### Suggested Change

Decide what to use, and document it consistently in all three places: README, website docs, CLI help

stay-foolish-forever:
Hi @philippgille, thank you for the detailed and well-researched feedback — really appreciate you cross-checking all three places.

Just to clarify the actual behavior: the implementation uses **merge-base mode** (equivalent to the three-dot `a...b` semantics) — it computes `git merge-base <from> <to>` and then diffs `merge-base..<to>` (see `internal/diff/git.go`). So the CLI help (`--help`) is the accurate one, while the website docs ("reviews `main..feature-branch`", two dots) are misleading and should be corrected, and the README could state the merge-base mode explicitly to remove the ambiguity.

Since you've already mapped out exactly where the inconsistencies are, would you be interested in opening a PR to align all three places? That would be a fantastic contribution, and we'd be happy to review and merge it. Of course, no pressure at all — if you'd prefer, we can take care of it on our side as well.

[...]

PR #613, opened 2026-07-30
docs: clarify --from/--to merge-base semantics (#579)

Ritvvvik:
## Problem

Three sources documented `ocr review --from x --to y` inconsistently:

- **README.md** — didn't specify two-dot (`a..b`) vs three-dot (`a...b`)
  semantics at all.
- **Website Quickstart** (en/ja/zh) — explicitly described it as
  `main..feature-branch` (two-dot), which is wrong.
- **CLI `--help`** (`cmd/opencodereview/flags.go:206`) — correctly
  describes it as "merge-base mode" (three-dot semantics).

## Root cause confirmed

`internal/diff/git.go`'s `ModeRange` computes the range via the two
refs' common ancestor (merge-base) — i.e. three-dot semantics, matching
the CLI help text. The Quickstart pages contradicted this.

## Fix

Updated the single-line comment above `ocr review --from main --to
feature-branch` in all four affected files to describe the actual
behavior in plain language, rather than relying on git's `..`/`...`
notation (which was the source of the original ambiguity):

- `README.md`
- `pages/src/content/docs/en/quickstart.md`
- `pages/src/content/docs/ja/quickstart.md`
- `pages/src/content/docs/zh/quickstart.md`

No behavior change — documentation only. `flags.go`'s help text was
already correct and untouched.

## Note on translations

I'm not a native Japanese or Chinese speaker; I'd appreciate a quick
sanity check on the ja/zh wording from a maintainer or native speaker
before merge.

### DIFF
--- a/README.ja-JP.md
+++ b/README.ja-JP.md
@@ -135,7 +135,7 @@ cd your-project
 # ワークスペースモード — ステージ済み・未ステージ・未追跡のすべての変更をレビュー
 ocr review
 
-# ブランチ範囲 — 2つのrefを比較
+# ブランチ範囲 — main から分岐した後の feature-branch の変更をレビュー（マージベースモード）
 ocr review --from main --to feature-branch
 
 # 単一コミット
--- a/README.ko-KR.md
+++ b/README.ko-KR.md
@@ -135,7 +135,7 @@ cd your-project
 # Workspace mode: staged, unstaged, untracked 변경을 모두 리뷰
 ocr review
 
-# Branch range: 두 ref 비교
+# 브랜치 범위 — main에서 분기된 이후 feature-branch의 변경 사항을 리뷰합니다 (머지베이스 모드)
 ocr review --from main --to feature-branch
 
 # 단일 commit
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ cd your-project
 # Workspace mode — review all staged, unstaged, and untracked changes
 ocr review
 
-# Branch range — compare two refs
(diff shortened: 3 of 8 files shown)


lizhengfeng101:
## Review

The change itself is correct and valuable — I verified that `internal/diff/git.go` does indeed use `git merge-base` (line 260) to compute the range, so the new wording accurately reflects the actual behavior.

### Issue: Missing localized README updates

The following localized README files contain the same outdated description but were not updated in this PR:

| File | Current (stale) line |
|------|------|
| `README.zh-CN.md:138` | `# 分支范围 —— 比较两个引用` |
| `README.ja-JP.md:138` | `# ブランチ範囲 — 2つのrefを比較` |
| `README.ko-KR.md:138` | `# Branch range: 두 ref 비교` |
| `README.ru-RU.md:138` | `# Диапазон веток — сравнение двух ref'ов` |

These should be updated to match the new merge-base description for consistency.

### Suggestion: PR title

The current title "Fixes #579" is quite terse for commit history. Consider something like `docs: clarify --from/--to merge-base semantics (#579)` so the change is self-describing in `git log`.

Ritvvvik:
Good catch on the locale files pushed a fix for zh-CN, ja-JP, ko-KR and ru-RU with the same merge-base wording. I'm not a native speaker in any of the five non-English locales touched across this PR, so I'd really appreciate a native speaker sanity-checking the phrasing before merge.

Also updated the PR title per your suggestion.

lizhengfeng101 (approved):
LGTM

PR #563, opened 2026-07-28
docs(configuration): document openai responses protocol

FenjuFu:
## Summary

- document `openai-responses` as the third supported custom-provider protocol
- add an OpenAI Responses API configuration example
- keep the English, Chinese, and Japanese configuration docs aligned

## Why

The Responses API has been supported since v1.7.10, but the custom-provider docs still listed only `anthropic` and `openai`. This closes the documentation gap reported in #556.

## Validation

- `git diff --check`
- verified the example against `internal/llm/responses_client.go` URL normalization
- local site build not run because `pages/node_modules` is not installed; the repository build script also uses Unix-style `NODE_ENV=...`

Closes #556.

Contributed as a member of the iFLYTEK Astron community.

PR #562, opened 2026-07-28
docs(configuration): document openai-responses protocol

zhouzhih:
## Summary

This PR updates the custom provider configuration docs to document the `openai-responses` protocol supported by OCR.

Fixes #556.

## Changes

- Lists `anthropic`, `openai`, and `openai-responses` as valid custom provider protocol values.
- Adds an OpenAI Responses API (`/v1/responses`) custom provider example.
- Keeps the English, Chinese, and Japanese configuration docs consistent.

## Non-goals

- This does not change provider resolution or LLM client behavior.
- This does not add or modify built-in providers.

## Validation

- `git diff --check origin/main...HEAD`
- `cd pages && npm run typecheck`
- `cd pages && npm run build`

Issue #556, opened 2026-07-28
docs(configuration): document openai-responses protocol in Custom Providers section


lizhengfeng101:
## Description

The Custom Providers section of the configuration docs currently states that `protocol` accepts only `anthropic` or `openai`. However, since v1.7.10 (#363), a third value `openai-responses` (for the OpenAI Responses API, `/v1/responses`) is also supported.

The docs should be updated to reflect all three valid protocol values.

**Current (inaccurate):**
> supply at least `url` and `protocol` (`protocol` is either `anthropic` or `openai`)

**Expected:**
> supply at least `url` and `protocol` (`protocol` is `anthropic`, `openai`, or `openai-responses`)

An example using the Responses API should also be added alongside the existing examples.

## Scope

- File(s):
  - `pages/src/content/docs/en/configuration.md`
  - `pages/src/content/docs/zh/configuration.md`
  - `pages/src/content/docs/ja/configuration.md`
- Area: Custom providers section (around line 62–85 in the English version)
- Add `openai-responses` to the protocol description and provide a usage example

## Acceptance Criteria

- [ ] All three protocol values (`anthropic`, `openai`, `openai-responses`) are listed in the Custom Providers section
- [ ] A brief example showing `openai-responses` configuration is included
- [ ] All localized docs (en, zh, ja) are updated consistently
- [ ] Site builds without errors

## Context

Discovered via #532 — a user requested `openai-responses` support, which was already implemented in #363 (shipped in v1.7.10). The user then suggested updating the docs. This issue tracks that documentation update specifically.

Closes #532.

PR #623, opened 2026-07-31
docs(configuration): note url accepts base URL or full /responses endpoint

FenjuFu:
## What

Follow-up to #562 as suggested by @lizhengfeng101 in #563.

#562 already documented the `openai-responses` protocol across all three locales. This adds the one incremental clarification #563 had that #562 lacked: for `openai-responses`, the `url` field accepts **either** the API base URL **or** the full `/responses` endpoint.

Added one sentence right after the `openai-responses` example in `en`, `zh`, and `ja`.

## Why it's accurate

`internal/llm/responses_client.go` → `ensureResponsesEndpoint` normalizes both forms. Its own contract comment states:

```
https://api.openai.com/v1             -> https://api.openai.com/v1/responses
https://api.openai.com/v1/            -> https://api.openai.com/v1/responses
https://api.openai.com/v1/responses   -> https://api.openai.com/v1/responses
```

So a user can supply either shape and it resolves the same way — worth stating so people don't guess.

## Scope

Docs-only, +2 lines per locale (the sentence + a blank line), no other changes. Rebased on current `main` so it sits cleanly on top of #562.

Refs #556, #563

### DIFF
--- a/pages/src/content/docs/en/configuration.md
+++ b/pages/src/content/docs/en/configuration.md
@@ -84,6 +84,8 @@ ocr config set custom_providers.openai-responses-gateway.model        gpt-5
 ocr config set custom_providers.openai-responses-gateway.api_key      "$OPENAI_API_KEY"
 ```
 
+The `url` can be either the API base URL or the full `/responses` endpoint — OCR normalizes it either way.
+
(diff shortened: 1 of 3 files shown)


lizhengfeng101 (approved):
LGTM

PR #629, opened 2026-07-31
test(rules): add integrity checks for system_rules ↔ rule_docs

### LINKED ISSUE #620: ci: add integrity test for system_rules.json and rule_docs consistency
## Problem

`system_rules.json` is embedded into the binary via `//go:embed` and maps glob patterns to rule doc files under `rule_docs/`. Currently there is no automated check to ensure the JSON entries and the actual `rule_docs/` files stay in sync. A contributor can:

- Add a new `.md` file to `rule_docs/` but forget to register it in the JSON → orphan file, never used
- Add a JSON entry pointing to a non-existent `.md` file → runtime error in `LoadDefault()`
- Write an invalid glob pattern (e.g. unclosed `{`) → silent match failure at runtime
- Introduce a duplicate pattern key → shadowed entry, confusing behavior

These issues only surface at runtime or not at all.

## Proposed Solution

Add a `TestSystemRulesIntegrity` test in `internal/config/rules/system_rules_test.go` that validates:

1. **File existence** — `default_rule` and every `path_rule_map` value must correspond to a file in `rule_docs/` (via the embedded `rulesFS`)
2. **Pattern validity** — every `path_rule_map` key must be a valid doublestar glob (call `doublestar.Match(pattern, "")` and assert no error)
3. **No orphan files** — every `.md` file in `rule_docs/` must be referenced by at least one entry (either `default_rule` or a `path_rule_map` value)
4. **No duplicate patterns** — detect duplicate keys in `path_rule_map`

### Why a Go test (not a JSON Schema tool)

- Zero external dependency — uses the already-embedded `rulesFS` and `doublestar` library
- Runs in existing `make test` / CI pipeline with no setup
- Can verify filesystem presence (orphan detection) which JSON Schema cannot
- Matches the project's existing pattern of testing embedded resources in Go

## Acceptance Criteria

- [ ] `make test` fails if a `rule_docs/*.md` file is added without a JSON entry
- [ ] `make test` fails if a JSON entry references a missing `.md` file
- [ ] `make test` fails if a glob pattern is syntactically invalid
- [ ] `make test` fails if duplicate pattern keys exist

nolanchic:
Adds `TestSystemRulesIntegrity` to `internal/config/rules`, validating that `system_rules.json` and `rule_docs/` stay in sync. Covers the four cases from #620:

1. **File existence** — `default_rule` and every `path_rule_map` value resolve to an embedded `rule_docs/` file.
2. **Pattern validity** — each `path_rule_map` key is a valid doublestar glob. Patterns are brace-expanded before validation, matching what `Resolve()` actually matches at runtime (raw `{go,py}` arms are not valid doublestar on their own).
3. **No orphan files** — every embedded `rule_docs/*.md` is referenced by at least one entry.
4. **No duplicate patterns** — duplicate `path_rule_map` keys are flagged via the ordered streaming decode in `SystemRule.UnmarshalJSON`, which preserves declaration order instead of collapsing duplicates.

The test uses the already-embedded `rulesFS` and `doublestar`, so it runs in `make test` with no extra setup. Each branch was verified to fail on purpose (dangling ref, unclosed brace, orphan file, duplicate key) before restoring the clean state.

Closes #620.

### DIFF
--- a/internal/config/rules/system_rules_test.go
+++ b/internal/config/rules/system_rules_test.go
@@ -1,10 +1,13 @@
 package rules
 
 import (
+	"encoding/json"
 	"os"
 	"path/filepath"
 	"strings"
 	"testing"
+
+	"github.com/bmatcuk/doublestar/v4"
 )
 
 func TestExpandBraces_NoBraces(t *testing.T) {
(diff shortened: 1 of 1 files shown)


lizhengfeng101 (approved):
LGTM

PR #630, opened 2026-07-31
fix(cli): refocus previous input on esc in manual provider form

### LINKED ISSUE #628: config provider TUI: Manual mode — ESC to go back to previous field makes it uneditable (cursor disappears)
### OpenCodeReview Version

v1.82.1

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

### Bug Description

In the config provider TUI, when using the Manual configuration tab, pressing ESC to navigate back to a previously filled field causes the input cursor to disappear. The field becomes completely uneditable — no cursor is visible and all keyboard input is silently ignored.

### Steps to Reproduce

1. Run ocr config provider (or equivalent command that opens the provider TUI)
2. Select the Manual tab and press Enter to enter the form
3. Fill in the URL field, press Enter to advance to Protocol
4. Select a protocol (e.g. openai), press Enter to advance to Model
5. Type a model name (e.g. step-3.5-flash), press Enter to advance to Auth Token
6. Press ESC to go back to Model
7. Observe: the cursor is gone, the field cannot be edited

### Expected Behavior

Pressing ESC should move focus back to the previous field with a visible, blinking cursor, allowing the user to edit the existing value.

### Logs / Error Output

```shell
The target field appears focused (highlighted label) but the text cursor is invisible and the field does not accept any keyboard input. The only way to recover is to press ESC all the way back to exit the form and re-enter it.
```

### Additional Context

Image

Image

Image

Image

hanhaiqingchuan:
## Description

In `ocr config provider`, the Manual tab form loses input focus when the user presses <kbd>Esc</kbd> to go back to a previous field. For example: fill in the URL, press <kbd>Enter</kbd> to reach the Protocol step, then press <kbd>Esc</kbd> to go back and edit the URL — the URL field is rendered as active, but it has no cursor and ignores all typing. The same back-navigation works correctly in the Custom provider form.

Root cause: `focusManualStep` is declared with a value receiver, while `textinput.Model.Focus()` uses a pointer receiver. In the <kbd>Esc</kbd> branch of `updateManualForm` (`return m, m.focusManualStep()`), the method receives a copy of the model, so `Focus()` marks the input as focused only on that discarded copy — the model actually returned from `Update` still holds a blurred input, and bubbles' `textinput` ignores key messages while blurred.

The fix changes `focusManualStep` to a pointer receiver, matching its Custom-form counterpart `focusCPStep` (which is why the Custom form was unaffected). All back-navigation paths with a text input are covered: Protocol→URL, Auth Token→Model, Auth Header→Auth Token. (Model→Protocol was unaffected: the protocol step has no text input.)

## Type of Change

- [X] Bug fix (non-breaking change that fixes an issue)

## How Has This Been Tested?

- [X] `make test` passes locally
- [X] Manual testing (describe below)

- Added `TestProviderTUI_ManualFormEscRefocusesPreviousInput`, a table-driven regression test covering the three back-navigation paths that land on a text input. Each case asserts the previous step's input is focused after <kbd>Esc</kbd> and accepts subsequent typing. Verified the test fails on the unfixed code (all three cases) and passes with the fix.
- Full suite: `make check` and `make test` (with `-race`) pass locally.
- Manual testing with a locally built binary (`ocr config provider`, isolated `$HOME` sandbox): Manual tab → fill URL → Enter → Esc — the URL field regains the cursor and is editable again; same verified for the Auth Token and Auth Header back-navigation paths. Reproduced the broken behavior with an unfixed main binary as an A/B control.

## Checklist

- [X] My code follows the project's coding style (`go fmt`, `go vet`)
- [X] I have performed a self-review of my code
- [X] I have added tests that prove my fix is effective or my feature works
- [X] New and existing unit tests pass locally with my changes
- [ ] I have updated the documentation accordingly (if applicable) — not applicable, no user-facing docs describe this interaction
- [X] I have signed the CLA

## Related Issues

Closes #628

### DIFF
--- a/cmd/opencodereview/provider_tui.go
+++ b/cmd/opencodereview/provider_tui.go
@@ -1642,7 +1642,7 @@ func (m *providerTUIModel) blurManualStep() {
 	}
 }
 
-func (m providerTUIModel) focusManualStep() tea.Cmd {
+func (m *providerTUIModel) focusManualStep() tea.Cmd {
 	switch m.manualStep {
 	case manualStepURL:
 		return m.manualURLInput.Focus()
--- a/cmd/opencodereview/provider_tui_test.go
+++ b/cmd/opencodereview/provider_tui_test.go
@@ -265,6 +265,62 @@ func TestProviderTUI_ManualFormEscFromURLExitsForm(t *testing.T) {
 	}
 }
 
+func TestProviderTUI_ManualFormEscRefocusesPreviousInput(t *testing.T) {
+	tests := []struct {
+		name     string
+		fromStep manualStep
+		wantStep manualStep
+		focused  func(m providerTUIModel) bool
+		value    func(m providerTUIModel) string
+	}{
+		{
+			name:     "protocol back to url",
+			fromStep: manualStepProtocol,
+			wantStep: manualStepURL,
+			focused:  func(m providerTUIModel) bool { return m.manualURLInput.Focused() },
+			value:    func(m providerTUIModel) string { return m.manualURLInput.Value() },
+		},
+		{
+			name:     "auth token back to model",
+			fromStep: manualStepAuthToken,
+			wantStep: manualStepModel,
+			focused:  func(m providerTUIModel) bool { return m.manualModelInput.Focused() },
+			value:    func(m providerTUIModel) string { return m.manualModelInput.Value() },
+		},
+		{
+			name:     "auth header back to auth token",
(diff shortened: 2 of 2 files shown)


5wxwxwxw5:
Thanks for the PR! I've gone through the changes and they look good to me.

lizhengfeng101 (approved):
LGTM

Issue #639, opened 2026-07-31
docs: add shell completion setup guide to CLI Reference and Installation pages


lizhengfeng101:
## Description

The `ocr completion` command (supporting bash, zsh, fish, and powershell) is fully implemented but has **zero documentation** — it is not mentioned in the CLI Reference, Installation page, README, or any other doc page.

Shell completion requires manual setup by the user (e.g., `source <(ocr completion zsh)`), so it needs documentation to be discoverable.

### 1. CLI Reference — add `ocr completion` section

- Files: `pages/src/content/docs/{en,zh,ja}/cli-reference.md`
- Add an `## ocr completion` section alongside the existing `ocr review`, `ocr config`, etc. sections
- Document the command format: `ocr completion [bash|zsh|fish|powershell]`
- Include per-shell one-shot and persistent setup instructions (reference the help text in `cmd/opencodereview/completion.go:32-61` for accurate content)

### 2. Installation — add optional post-install step

- Files: `pages/src/content/docs/{en,zh,ja}/installation.md` and `pages/src/content/docs/ru/installation.md`
- Add a brief "Enable Shell Completion" section after the "Verifying the install" section
- One sentence explaining what it does, a quick bash/zsh example, and a link to the CLI Reference for full details

### 3. README — no changes needed

Shell completion is an optional enhancement, not part of the core usage flow. The README already links to the docs site where users can find it.

## Acceptance Criteria

- [ ] `ocr completion` section added to CLI Reference in en, zh, ja
- [ ] Brief "Enable Shell Completion" section added to Installation in en, zh, ja, ru
- [ ] Content matches the actual supported shells and setup steps in `cmd/opencodereview/completion.go`
- [ ] Docs site renders correctly locally (`cd pages && npm run dev`)
- [ ] All three language versions (en/zh/ja) have equivalent content; ru only needs the installation section

## Context

Discovered during code review of the Cobra CLI migration PR — the `completion` command was added but documentation was not updated. The CLI Reference page is also missing `scan` and `delegate` commands, but those are separate issues.

Reference: the help text in [`cmd/opencodereview/completion.go`](https://github.com/alibaba/open-code-review/blob/main/cmd/opencodereview/completion.go) contains the per-shell instructions that should be reflected in the docs.

Ritvvvik:
Hi @lizhengfeng101 can i take up

lizhengfeng101:
> Hi @lizhengfeng101 can i take up

yes assigned

[...]

PR #652, opened 2026-07-31
docs: add shell completion setup guide for CLI Reference and Installation (#639)

Ritvvvik:
Documents the `ocr completion` command (bash/zsh/fish/powershell), which is implemented but has zero documentation.

## CLI Reference (en, zh, ja)

- New `## ocr completion` section alongside the existing command sections
- Command format: `ocr completion [bash|zsh|fish|powershell]`
- Per-shell one-shot and persistent setup instructions, matching the help text in `cmd/opencodereview/completion.go`

## Installation (en, zh, ja, ru)

- New "Enable Shell Completion" section after "Verifying the install"
- One-line explanation, quick shell example, and a link to the CLI Reference for full details

Content verified against the shells and setup steps in `cmd/opencodereview/completion.go`. Docs site renders correctly locally via `cd pages && npm run dev`.


lizhengfeng101 (approved):
LGTM

PR #655, opened 2026-07-31
refactor(config): extract supported config keys into a single source of truth

### LINKED ISSUE #637: refactor(config): extract supported config keys into a single source of truth
## Description

In `config_cmd.go`, the list of supported configuration keys is maintained in two places that must be kept in sync:

1. The `switch/case` branches in `setConfigValue()` (line 361–482) define the actual key handling logic.
2. The error message in the `default` case (line 480) hardcodes a human-readable list of all supported keys.

When a new config key is added, both locations must be updated. If the error message is forgotten, users see an incomplete list of supported keys on typos.

**Current code (line 480):**
```go
default:
    return fmt.Errorf("unknown config key: %s\nSupported keys: provider, model, providers.<name>.<field>, custom_providers.<name>.<field>, mcp_servers.<name>.<field>, llm.url, llm.auth_token, ...", key)
```

## Scope

- File: `cmd/opencodereview/config_cmd.go`
- Function: `setConfigValue`
- Extract a `var supportedConfigKeys` (or similar) as the single source of truth, and derive the error message from it.

## Acceptance Criteria

- [ ] Supported config keys are defined in one place
- [ ] The error message for unknown keys is generated from that definition
- [ ] No change in user-facing behavior (same error message content)

## Context

Discovered during code review of the Cobra CLI migration PR. The switch/case logic itself is correct; this is a maintainability improvement to prevent future drift between the key handling and the help text.

DoTuanAnh2k1:
`setConfigValue` kept the list of supported top-level config keys in two places — the `switch`/`case` branches and, hardcoded again, the "Supported keys:" line of the unknown-key error. Adding a key meant updating both, and it was easy to forget the error text.

This moves the keys into a `supportedConfigKeys` slice and builds the "Supported keys:" part of the message from it with `strings.Join`, so the two can't drift. The message content is byte-for-byte unchanged.

Added `TestSetConfigValueUnknownKeyMessage`, which pins the full message and checks it is generated from `supportedConfigKeys`. `go test ./...` and `make check` pass.

Closes #637

---
<sub>Disclosure: prepared with AI assistance.</sub>

### DIFF
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -346,6 +346,30 @@ func LoadAppConfig(path string) (*Config, error) {
 	return &cfg, nil
 }
 
+// supportedConfigKeys is the single source of truth for the top-level config
+// keys accepted by setConfigValue. The unknown-key error message is generated
+// from this list so the two cannot drift apart when a new key is added.
+var supportedConfigKeys = []string{
+	"provider",
+	"model",
+	"providers.<name>.<field>",
+	"custom_providers.<name>.<field>",
+	"mcp_servers.<name>.<field>",
+	"llm.url",
+	"llm.auth_token",
+	"llm.auth_header",
+	"llm.model",
+	"llm.protocol",
+	"llm.use_anthropic",
+	"llm.extra_body",
+	"llm.extra_headers",
+	"language",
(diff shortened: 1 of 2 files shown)


lizhengfeng101 (approved):
LGTM

PR #660, opened 2026-07-31
fix(cli): return error on unknown subcommands for parent commands

### LINKED ISSUE #641: Parent commands silently return exit code 0 for unknown subcommands
## OpenCodeReview Version

`open-code-review dev darwin/arm64` (built from source, latest main)

## Operating System

macOS (Apple Silicon)

## Installation Method

Built from source

## LLM Provider

Anthropic (Claude)

## Bug Description

Parent commands (`session`, `config`, `delegate`, `llm`, `rules`) silently display help and return exit code 0 when given an unknown subcommand. This is inconsistent with the root command and leaf commands, which correctly return exit code 1 with an error message.

| Command level | Input | Exit code | Behavior |
|---|---|---|---|
| Root | `ocr nonexistent` | 1 | `Error: unknown command "nonexistent" for "ocr"` |
| Leaf | `ocr review unknown` | 1 | `Error: unknown command "unknown" for "ocr review"` |
| Parent | `ocr session unknown` | 0 | Silently prints help, no error |
| Parent | `ocr config unknown` | 0 | Silently prints help, no error |
| Parent | `ocr delegate unknown` | 0 | Silently prints help, no error |
| Parent | `ocr llm unknown` | 0 | Silently prints help, no error |
| Parent | `ocr rules unknown` | 0 | Silently prints help, no error |

## Steps to Reproduce

```bash
# 1. Run a parent command with an unknown subcommand
ocr session unknown
echo $?   # prints 0

# 2. Compare with root command
ocr nonexistent
echo $?   # prints 1

# 3. Compare with leaf command
ocr review unknown
echo $?   # prints 1
```

## Expected Behavior

All commands should consistently return exit code 1 and print an error message when given an unrecognized subcommand. For example, `ocr session unknown` should output:

```
Error: unknown command "unknown" for "ocr session"
```

and exit with code 1, matching the behavior of the root command and leaf commands.

## Logs / Error Output

```shell
$ ocr session unknown
List and inspect saved review sessions

Usage:
  ocr session [command]

Aliases:
  session, sessions

Available Commands:
  list        List recent review sessions for the current repo
  show        Show one session's metadata and per-file items

Flags:
  -h, --help   help for session

Use "ocr session [command] --help" for more information about a command.

$ echo $?
0
```

## Additional Context

The root cause is that parent commands (`sessionCmd`, `configCmd`, `delegateCmd`, `llmCmd`, `rulesCmd`) are defined without a `RunE`/`Run` function. In Cobra, when a parent command has no `RunE` and receives an unrecognized subcommand, it falls back to displaying help and returns `nil` (exit 0). Adding a `RunE` that returns `cmd.Help()` (or any non-nil function) would cause Cobra to properly report "unknown command" errors for these parent commands, consistent with the root command's behavior.

aalhadxx:
## Problem

Parent commands (`session`, `config`, `delegate`, `llm`, `rules`) were defined without a `RunE` function. In Cobra, when a parent command has no `RunE` and receives an unrecognized subcommand, it falls back to displaying help and returns `nil` (exit 0). This is inconsistent with the root command and leaf commands, which correctly return exit code 1 with an error message.

| Command level | Input | Exit code | Behavior |
|---|---|---|---|
| Root | `ocr nonexistent` | 1 | Error: unknown command |
| Leaf | `ocr review unknown` | 1 | Error: unknown command |
| Parent | `ocr session unknown` | 0 | Silently prints help, no error |

## Fix

Add `RunE: func(cmd *cobra.Command, args []string) error { return cmd.Help() }` to all five parent commands so Cobra properly reports "unknown command" errors instead of silently succeeding.

## Changes

- `sessionCmd` — add `RunE`
- `configCmd` — add `RunE`
- `delegateCmd` — add `RunE`
- `llmCmd` — add `RunE`
- `rulesCmd` — add `RunE`
- New `parent_cmd_test.go` with 13 tests covering:
  - Unknown subcommands produce a non-nil error containing "unknown command"
  - Known subcommands still route correctly (no regression)
  - Bare parent commands (no args) still print help and return nil

## Verification

All new tests pass locally. Existing tests remain unchanged.

Closes #641

### DIFF
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -32,6 +32,12 @@ Examples:
   ocr config set provider my-gateway
   ocr config set custom_providers.my-gateway.url https://gateway.internal.com/v1
   ocr config set custom_providers.my-gateway.protocol openai`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) > 0 {
+			return fmt.Errorf("unknown command %q for %q", args[0], cmd.CommandPath())
+		}
+		return cmd.Help()
+	},
 }
 
 var configSetCmd = &cobra.Command{
--- a/cmd/opencodereview/delegate_cmd.go
+++ b/cmd/opencodereview/delegate_cmd.go
@@ -41,6 +41,12 @@ Output review spec for host-agent delegation (no LLM required).`,
 
   # Get rules for multiple files (grouped by content)
   ocr delegate rule internal/agent/agent.go internal/llm/client.go`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) > 0 {
+			return fmt.Errorf("unknown command %q for %q", args[0], cmd.CommandPath())
+		}
+		return cmd.Help()
+	},
 }
 
 var delegatePreviewCmd = &cobra.Command{
--- a/cmd/opencodereview/llm_cmd.go
+++ b/cmd/opencodereview/llm_cmd.go
@@ -18,6 +18,12 @@ var llmCmd = &cobra.Command{
 	Long:  "LLM utility commands.",
 	Example: `  ocr llm test                   Verify LLM connectivity and configuration
   ocr llm providers              List available built-in providers`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) > 0 {
+			return fmt.Errorf("unknown command %q for %q", args[0], cmd.CommandPath())
+		}
+		return cmd.Help()
+	},
 }
 
 var llmTestCmd = &cobra.Command{
--- a/cmd/opencodereview/parent_cmd_test.go
+++ b/cmd/opencodereview/parent_cmd_test.go
@@ -1,0 +1,139 @@
+package main
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/spf13/cobra"
+)
+
+// TestParentCommands_UnknownSubcommand verifies that parent commands which
+// previously silently printed help and exited 0 now return a non-zero exit
+// code with a clear "unknown command" error when given an unrecognized
+// subcommand. This is consistent with the root command and leaf commands.
+//
+// See: https://github.com/alibaba/open-code-review/issues/641
+func TestParentCommands_UnknownSubcommand(t *testing.T) {
+	tests := []struct {
+		name string
+		args []string
+		want string // substring expected in error message
+	}{
+		{"session", []string{"session", "bogus"}, "unknown command"},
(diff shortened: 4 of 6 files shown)


lizhengfeng101:
Hi @aalhadxx, CI is failing on the `TestParentCommands_UnknownSubcommand` tests. The issue is that when a parent command has `RunE` defined, Cobra passes unrecognized subcommands as `args` to `RunE` instead of returning an "unknown command" error automatically.

Since `cmd.Help()` always returns `nil`, the current `RunE` implementation silently succeeds even with unknown subcommands like `bogus`.

**Fix**: check `args` before falling through to help:

```go
RunE: func(cmd *cobra.Command, args []string) error {
    if len(args) > 0 {
        return fmt.Errorf("unknown command %q for %q", args[0], cmd.CommandPath())
    }
    return cmd.Help()
},
```

This needs to be applied to all parent commands (`session`, `config`, `delegate`, `llm`, `rules`).

aalhadxx:
@lizhengfeng101 thanks for the review — you are absolutely right. When a parent command has `RunE`, Cobra no longer auto-rejects unknown subcommands; it passes them as `args` to `RunE` instead.\n\nFixed in all five parent commands (session, config, delegate, llm, rules) by checking `len(args) > 0` before falling through to `cmd.Help()`:\n\n\n\nForce-pushed the amended commit. PTAL.

lizhengfeng101 (commented):
Nice fix — the silent exit-0 on unknown subcommands was definitely confusing.

A couple of thoughts:

**You're reimplementing `cobra.NoArgs`**

The 5-line RunE you added to each parent is doing exactly what `cobra.NoArgs` already does (same error format and everything). You could simplify each command to:

```go
Args: cobra.NoArgs,
RunE: func(cmd *cobra.Command, args []string) error {
    return cmd.Help()
},
```

Cobra validates `Args` before calling `RunE`, so the RunE only fires when there are zero args. Less code, same behavior, and you don't have to maintain the format string yourself.

**`cloneRootCmd` mutates global state**

`AddCommand` in Cobra doesn't remove the command from its old parent — it just reassigns the `.parent` pointer. So after your tests run, every subcommand's `.parent` points to the last clone, not the real `rootCmd`. It happens to work today because both use `Use: "ocr"`, but it's fragile. Something like a `t.Cleanup` that re-adds everything back to `rootCmd`, or just using `rootCmd` directly (it already has `SilenceUsage`/`SilenceErrors`), would be more robust.

Neither of these is a blocker — the fix works as-is. But the `cobra.NoArgs` change especially would clean things up nicely.

lizhengfeng101 (approved):

binyaaminofficial (approved):

PR #663, opened 2026-08-01
fix(tool/code_search): remove --end-of-options from git grep arguments

### LINKED ISSUE #645: code_search fails in range/commit reviews because git grep treats --end-of-options as a revision
### OpenCodeReview Version

1.8.3

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

OpenAI

## Description

  ```bash

  ## Environment

  - OpenCodeReview: v1.8.3
  - Commit: 80a579466
  - Git: 2.43.0
  - OS: Linux amd64

  ## Reproduction

  Run:

    -e RepositoryFactory \

  Actual result:

    -e RepositoryFactory \
    main -- 'src/**/*.py'

  ## Impact

  In one range review:

  ## Root Cause

  }

  ## Proposed Fix

      ctx context.Context,
      searchText string,
      caseSensitive bool,
      usePerlRegexp bool,
      pathspec []string,
  ) (string, error) {
      }

      cmdArgs := p.buildGrepArgs(
          searchText,
          caseSensitive,
          usePerlRegexp,
          false,
          pathspec,
      )
      // ...
  }

  }

  ## Suggested Tests

  -e myFunc abc1234 -- pkg/

          RepoDir: "/tmp",
          Ref:     "-O./pwn.sh",
      })

      result, err := p.gitGrep(
          context.Background(),
          "myFunc",
          false,
          false,
          nil,
      )

      if err != nil {
          t.Fatal(err)
      }
      }
  }

  ## Verification

### Steps to Reproduce

  ## Reproduction

  Run:

    -e RepositoryFactory \

  Actual result:

    -e RepositoryFactory \
    main -- 'src/**/*.py'

  ## Impact

  In one range review:

  ## Root Cause

  }

  ## Proposed Fix

      ctx context.Context,
      searchText string,
      caseSensitive bool,
      usePerlRegexp bool,
      pathspec []string,
  ) (string, error) {
      }

      cmdArgs := p.buildGrepArgs(
          searchText,
          caseSensitive,
          usePerlRegexp,
          false,
          pathspec,
      )
      // ...
  }

  }


### Expected Behavior

  ## Reproduction

  Run:

    -e RepositoryFactory \

  Actual result:

    -e RepositoryFactory \
    main -- 'src/**/*.py'

  ## Impact

  In one range review:

  ## Root Cause

  }

  ## Proposed Fix

      ctx context.Context,
      searchText string,
      caseSensitive bool,
      usePerlRegexp bool,
      pathspec []string,
  ) (string, error) {
      }

      cmdArgs := p.buildGrepArgs(
          searchText,
          caseSensitive,
          usePerlRegexp,
          false,
          pathspec,
      )
      // ...
  }

  }


### Logs / Error Output

```shell

```

### Additional Context

_No response_

aalhadxx:
## Problem

`git grep` on Git 2.43.0 (and likely other versions) does not accept `--end-of-options` between the pattern and revision arguments. It treats `--end-of-options` as a revision and fails with:

```
```

This caused every `code_search` call to fail during range or commit reviews, rendering the tool unusable in those modes. In one reported case, 375 consecutive searches all returned the same error, exhausting `MAX_TOOL_REQUEST_TIMES`.

## Root Cause

`internal/tool/code_search.go` added `--end-of-options` before the ref:

```go
}
```

Unlike `git show`, `git diff`, `git log`, etc., `git grep` does not accept `--end-of-options` in this position.

## Fix

1. Remove `--end-of-options` from `buildGrepArgs` before the ref argument.
2. Add explicit validation in `gitGrep` to reject option-like refs (starting with `-`) before invoking git, preserving the original security intent.

## Test Updates

- `TestBuildGrepArgs_CommitMode` — updated to assert `--end-of-options` is absent.
- `TestBuildGrepArgs_RefUsesEndOfOptions` — replaced with `TestGitGrep_RejectsOptionLikeRef` that validates the new guard rejects refs starting with `-` before git is invoked.

## Verification

After this change, commit/range-mode `code_search` returns matches successfully on Git 2.43.0, while option-like refs are still blocked.

Closes #645


aalhadxx:
@github-actions Thanks for the security review, but this suggestion would re-introduce the exact bug this PR fixes.

`git grep` (unlike `git show`, `git diff`, `git log`, `git rev-parse`, etc.) does **not** accept `--end-of-options` between the `-e <pattern>` arguments and the revision. On Git 2.43.0:

```bash
$ git grep -F -n --no-color --max-count 100 \
    -e RepositoryFactory \
```

Removing `--end-of-options` makes the command work:

```bash
$ git grep -F -n --no-color --max-count 100 \
    -e RepositoryFunction \
    main -- 'src/**/*.py'
# (returns matches successfully)
```

The defense-in-depth is preserved by the new guard added at the top of `gitGrep`:

```go
}
```

This rejects option-like refs **before** git is ever invoked. Combined with the existing `validateReviewRefs` (which already rejects refs starting with `-` and verifies them via `git rev-parse --verify`), the layered defense remains intact without the broken `--end-of-options` flag.

stay-foolish-forever:
@aalhadxx

Thank you very much for opening this PR and for the clear write-up — the root-cause analysis and the updated tests made it very easy to review. We appreciate you taking the time to dig into this.

We're aligned with you on the core problem: `--end-of-options` is genuinely broken for `git grep` on Git 2.43.0, and the current behavior is unacceptable. Where we'd like to push back a little is on the trade-off this PR makes.

Some background: `--end-of-options` was added specifically to close the command-line injection / arbitrary file-write surface reported in #111 (option-like `--from`/`--to`/`--commit` values such as `--output=` or `--open-files-in-pager=` being reinterpreted by git as flags). The complication is that `CodeSearchProvider` can be constructed outside the CLI path, so `validateReviewRefs` doesn't always run before git is invoked — which is exactly why the in-call `--end-of-options` guard was added as defense-in-depth.

This PR replaces that guard with a `strings.HasPrefix(ref, "-")` check. That's a real improvement over nothing, but it's a narrower gate than what `--end-of-options` provided: it only catches a leading `-`, so option-like content embedded elsewhere, or future call sites that construct refs differently, wouldn't be covered. We fully acknowledge your point that `--end-of-options` can't stay in its current position — we agree it has to go. The tension is that simply dropping it, with only a prefix check in its place, leans the trade-off further toward compatibility than we're comfortable with for a security-sensitive path.

For now, rather than merge a partial fix, the maintainer team is designing a more robust architecture that doesn't depend on `--end-of-options` at all: full ref validation and a clean option/revision separation at the source, applied uniformly across every subcommand and resilient across git versions. You can see more on our thinking in the reply on #645.

Once that design is finalized, we would genuinely welcome your participation and contribution — your familiarity with this code path and your test updates here would make you a great collaborator on the proper fix. Would you be open to holding this PR open for a little while so we can align it with the upcoming design? We'll follow up here as soon as the plan is ready, and in the meantime we're happy to share details as the design takes shape.

Thanks again for the high-quality contribution and for engaging with the security review so thoughtfully.

aalhadxx:
Thanks for the thoughtful review, @stay-foolish-forever. I am happy to keep this PR open and align it with the upcoming design you mentioned. The distinction between compatibility and defense-in-depth makes sense, and I would welcome the chance to collaborate once the architecture is finalized. Let me know whenever you have a direction or need any changes here.

lizhengfeng101:
Hey @aalhadxx, thanks for digging into this — the root cause is spot-on and the reproduction in #645 clearly shows that `git grep` doesn't support `--end-of-options` on Git < 2.45. Appreciate the effort here.

That said, we'd like to suggest a slightly different structure for the fix before we can merge this. Here's the context:

## Existing two-layer defense

The codebase already has upstream validation in `validateReviewRefs` (`review_cmd.go:340-365`) that runs **before** `FileReader.Ref` is ever assigned. It:
1. Rejects any ref starting with `-`
2. Validates the ref resolves to a real commit via `git rev-parse --verify`

So by the time `buildGrepArgs` sees the ref, it's already guaranteed to be a legitimate git object. The `--end-of-options` in each git invocation is a second layer (defense-in-depth), not the primary guard.

## What we'd suggest

Since `git grep` can't use `--end-of-options` on older Git versions, we're fine removing it — but we'd like the defense-in-depth check to live **inside `buildGrepArgs`** rather than in `gitGrep()`. This keeps the guard co-located with the argument construction (same pattern the other tools follow):

```go
    if strings.HasPrefix(ref, "-") {
        return nil // or handle the error appropriately
    }
    // NOTE: git grep < 2.45 does not support --end-of-options here.
    // Source validation in validateReviewRefs ensures only verified
    // commit refs reach this point.
}
```

This way:
- The security intent is visible right where args are built (not separated across functions)
- If anyone ever calls `buildGrepArgs` from a new path, the guard still holds
- A brief comment explains why this is the one place we don't use `--end-of-options`

The test changes look good — `TestGitGrep_RejectsOptionLikeRef` is better coverage than the old unit test on `buildGrepArgs` alone. You'd just need to adjust which function does the rejection.

Would you be up for making that adjustment? Happy to discuss further if anything's unclear!

aalhadxx:
@lizhengfeng101 Thanks for the detailed review! I've addressed the feedback:

1. Moved the defense-in-depth check from gitGrep() into uildGrepArgs() — the guard is now co-located with argument construction.
2. Added a comment explaining why --end-of-options can't be used here: git grep < 2.45 does not support --end-of-options before the revision.
3. Added TestBuildGrepArgs_RejectsOptionLikeRef to cover the new 
il-return path in uildGrepArgs.

The PR was auto-closed and I can't reopen it. Could you please reopen so the updated commit (d8fe610) can be reviewed? The changes remain on branch ix/code-search-end-of-options.

lizhengfeng101:
reopen

lizhengfeng101:
> @lizhengfeng101 Thanks for the detailed review! I've addressed the feedback:
> 
> 1. Moved the defense-in-depth check from gitGrep() into �uildGrepArgs() — the guard is now co-located with argument construction.
> 2. Added a comment explaining why --end-of-options can't be used here: git grep < 2.45 does not support --end-of-options before the revision.
> 3. Added TestBuildGrepArgs_RejectsOptionLikeRef to cover the new
>    il-return path in �uildGrepArgs.
> 
> The PR was auto-closed and I can't reopen it. Could you please reopen so the updated commit ([d8fe610](https://github.com/alibaba/open-code-review/commit/d8fe610f0fe7e942b69f552436f1812da0ee3c86)) can be reviewed? The changes remain on branch ix/code-search-end-of-options.

@aalhadxx The GitHub interface currently indicates that I cannot reopen it. Could you please create a new PR?

PR #667, opened 2026-08-01
fix(telemetry): honour a base path in the OTLP HTTP endpoint

### LINKED ISSUE #668: OTLP HTTP exporter discards a base path in OTEL_EXPORTER_OTLP_ENDPOINT
### OpenCodeReview Version

`open-code-review v1.8.0 (e670b3b3) linux/arm64`

Also reproduced against `main` at `80a5794`.

### Operating System

Linux (ARM64) — a container image, `linux/arm64`.

### Installation Method

GitHub Release binary (pinned into a container image); also built from source to confirm.

### LLM Provider

Other OpenAI-compatible endpoint (LiteLLM proxy). Not relevant to the bug — this is in the telemetry exporter, not the LLM path.

### Bug Description

When `OTEL_EXPORTER_OTLP_PROTOCOL` selects an HTTP protocol, a **base path** in `OTEL_EXPORTER_OTLP_ENDPOINT` is silently discarded. Every request goes to `/v1/traces` (and `/v1/metrics`) at the **root** of the host instead of under the configured path.

`internal/telemetry/exporter.go` configures the HTTP exporters with `WithEndpoint`, which takes a bare `host:port` and cannot express a path — `parseOTLPEndpoint` strips the scheme and the exporter never calls `WithURLPath`, which does not appear anywhere in the repository.

This excludes any backend addressed through a path prefix. Langfuse, for example, receives OTLP at `/api/public/otel`.

The failure is silent, which is what makes it costly to diagnose:

- Exporter creation is lazy, so telemetry initialises without error.
- Nothing is written to stderr.
- The backend answers `404` and the spans are simply gone.

From the outside it looks like telemetry is off, so the natural response is to check `OCR_ENABLE_TELEMETRY`, then the endpoint, then the protocol — none of which is wrong. I only found it by reading `internal/telemetry/exporter.go`.

### Steps to Reproduce

1. Start any OTLP/HTTP collector that serves under a path prefix. A one-line stand-in:

   ```bash
   python3 -c "
   import http.server
   class H(http.server.BaseHTTPRequestHandler):
       def do_POST(self):
           print('received POST', self.path, flush=True)
           self.send_response(200); self.send_header('Content-Length','0'); self.end_headers()
       def log_message(self,*a): pass
   http.server.HTTPServer(('127.0.0.1',4318),H).serve_forever()"
   ```

2. Run a review with telemetry pointed at a **base path** on that collector:

   ```bash
   export OCR_ENABLE_TELEMETRY=1
   export OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
   export OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:4318/api/public/otel
   ocr review --from origin/main --to HEAD --format json
   ```

3. Observe the path the collector receives.

### Expected Behavior

`received POST /api/public/otel/v1/traces`

The OpenTelemetry specification defines `OTEL_EXPORTER_OTLP_ENDPOINT` for the HTTP protocols as a **base URL** to which the per-signal path is appended — an endpoint of `https://backend.example.com` sends traces to `https://backend.example.com/v1/traces`.

### Logs / Error Output

`received POST /v1/traces`

The configured `/api/public/otel` prefix is gone. Against a real backend this is a `404` and the spans are dropped; `ocr` itself prints nothing.

### Additional Context

The gRPC path is correct as written and should not change — gRPC has no URL path, and `WithEndpoint` there genuinely does take a bare `host:port`.

I have opened #667 with a fix: the HTTP path switches to `WithEndpointURL` (which accepts a scheme and a path, and derives TLS from the scheme) with the signal path appended by a small helper, since `WithEndpointURL` assigns the URL's path straight to `URLPath` and so expects the full signal URL while the environment variable is specified as a base.

Existing behaviour is preserved for every endpoint form that works today, including a bare `host:port` still meaning TLS, and an endpoint that already names `/v1/traces` is left alone rather than doubled — so a configuration written to work around this does not break.

The PR includes a regression test that asserts on the request path the collector receives, and which fails against the current implementation.

fafg:
## Description

The OTLP **HTTP** exporters were configured with `WithEndpoint`, which takes a bare `host:port` and cannot express a path, and `WithURLPath` is never called. Any path in `OTEL_EXPORTER_OTLP_ENDPOINT` was therefore discarded and every request went to `/v1/traces` (or `/v1/metrics`) at the **root** of the host.

That silently excludes any backend addressed through a path prefix. Langfuse, for example, receives OTLP at `/api/public/otel`, so:

```
OTEL_EXPORTER_OTLP_ENDPOINT=http://langfuse:3000/api/public/otel
OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
```

produced requests to `http://langfuse:3000/v1/traces`. The backend answered `404` and every span was dropped. Nothing surfaced the problem: exporter creation is lazy, so telemetry appears to initialise correctly and the failure only shows up as missing data.

### The change

The HTTP path now uses `WithEndpointURL`, which accepts a scheme and a path and derives TLS from the scheme — replacing the explicit `WithInsecure` call so the decision is made from one source rather than two.

The signal path is appended by a new `otlpSignalURL` helper rather than left to the SDK, because `WithEndpointURL` assigns the URL's path straight to `URLPath` and so expects the *full* signal URL, while `OTEL_EXPORTER_OTLP_ENDPOINT` is specified as a *base*.

Existing behaviour is preserved in both directions:

| Endpoint | Before | After |
|---|---|---|
| `http://host:4318` | plaintext, `/v1/traces` | plaintext, `/v1/traces` |
| `https://host:4318` | TLS, `/v1/traces` | TLS, `/v1/traces` |
| `host:4318` (no scheme) | TLS, `/v1/traces` | TLS, `/v1/traces` |
| `http://host:3000/base` | plaintext, `/v1/traces` ❌ | plaintext, `/base/v1/traces` ✅ |
| `http://host:4318/v1/traces` | plaintext, `/v1/traces` | plaintext, `/v1/traces` (not doubled) |

The last row matters: anyone who worked around this by writing the signal path out in full keeps working.

`parseOTLPEndpoint` is unchanged and still used by the **gRPC** path, where `WithEndpoint` genuinely does take a bare `host:port` — gRPC has no URL path.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Two tests were added:

- **`TestInitOTLPHTTPProviders_HonoursBasePath`** stands up an `httptest` collector under a `/api/public/otel` prefix, exports a span, and asserts on the request path the collector actually receives. This is the regression test: it **fails against the previous implementation** (receives `/v1/traces`, wants `/api/public/otel/v1/traces`), which I verified by reverting the fix with the test in place. It also covers the TLS decision that moved into `WithEndpointURL` — `httptest.NewServer` speaks plaintext, so the request only arrives if `Insecure` was resolved to `true` from the `http://` scheme.
- **`TestOTLPSignalURL`** covers the normalisation table above, including the trailing-slash and already-has-the-signal-path cases.

Existing tests are unchanged and pass, including `TestParseOTLPEndpoint` (that function keeps its behaviour for gRPC).

Verified with the project's own targets: `gofmt -s` clean, `go vet` clean, `go test -race -count=1 ./internal/telemetry/` green.

Manual verification of the real-world case that prompted this: with `OTEL_ENABLE_TELEMETRY=1` and an endpoint carrying a base path, requests now reach `<base>/v1/traces` instead of `/v1/traces`.

## Documentation

`pages/src/content/docs/*/telemetry.md` claimed in two places that the HTTP protocols were "accepted but not yet wired up". They are — `initOTLPProviders` routes both to `initOTLPHTTPProviders`, and `TestInitOTLPProviders_ProtocolRouting` covers it.

The troubleshooting row was the more costly of the two, because it surfaces exactly when someone is debugging this and points the wrong way ("switching it won't help ... verify the collector is listening for gRPC"). Replaced with the check that does help, plus a row for the failure mode that produces no error at all: exporter creation is lazy, so a wrong endpoint fails silently at export time.

Also added an **Endpoint format** section covering three things that were undocumented and are each easy to get wrong:

- the scheme decides transport security (`http://` plaintext, `https://` and bare `host:port` TLS);
- the default port differs by protocol — `4317` for gRPC, `4318` for HTTP — while every example used `4317`;
- for the HTTP protocols the endpoint is a base URL to which the signal path is appended, and a base path is preserved.

**Translations:** the `ja` and `zh` code blocks carry that note in English, so they are corrected here. Their **troubleshooting rows are translated prose** and I have deliberately left them alone rather than guess at Japanese and Chinese I cannot review — they still carry the outdated claim and would need a translator:

- `pages/src/content/docs/ja/telemetry.md:249`
- `pages/src/content/docs/zh/telemetry.md:249`

Happy to take a suggested wording in either language and apply it here.

## Checklist

- [x] I have signed the CLA

## Related Issues

closes #668

### DIFF
--- a/internal/telemetry/exporter.go
+++ b/internal/telemetry/exporter.go
@@ -46,6 +46,39 @@ func parseOTLPEndpoint(endpoint string) (addr string, insecure bool) {
 	}
 }
 
+// otlpSignalURL builds the URL for one signal from the configured endpoint.
+//
+// OTEL_EXPORTER_OTLP_ENDPOINT is defined by the OpenTelemetry specification as a BASE
+// URL for the HTTP protocols, to which the per-signal path is appended: an endpoint of
+// "https://backend.example.com" sends traces to "https://backend.example.com/v1/traces".
+// WithEndpointURL, by contrast, takes the full signal URL — it assigns u.Path straight
+// to URLPath — so the signal path has to be appended here rather than left to the SDK.
+//
+// Two details this has to get right:
+//
+//   - A bare host:port keeps the meaning parseOTLPEndpoint gave it, TLS, by gaining an
+//     https:// scheme. WithEndpointURL derives Insecure from the scheme
+//     (`u.Scheme != "https"`), which is what the explicit WithInsecure call used to do,
+//     so an existing configuration behaves identically after this change.
+//   - An endpoint that already names the signal path is left alone. Someone who worked
+//     around the missing path support by writing it out in full should not end up with
+//     "/v1/traces/v1/traces".
+func otlpSignalURL(endpoint, signalPath string) string {
+	base := endpoint
+	switch {
+	case len(base) >= 7 && strings.EqualFold(base[:7], "http://"):
+	case len(base) >= 8 && strings.EqualFold(base[:8], "https://"):
+	default:
+		base = "https://" + base
+	}
+
+	base = strings.TrimRight(base, "/")
+	if strings.HasSuffix(base, signalPath) {
+		return base
+	}
+	return base + signalPath
+}
+
 // initOTLPProviders dispatches to the gRPC or HTTP exporter based on cfg.OTLPProtocol.
 func initOTLPProviders(ctx context.Context, res *resource.Resource, cfg Config) {
 	switch cfg.OTLPProtocol {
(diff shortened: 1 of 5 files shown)


lizhengfeng101 (approved):
LGTM

PR #670, opened 2026-08-01
fix(llm): retry once when a chat completion is truncated

### LINKED ISSUE #664: error reading response body: unexpected EOF
### OpenCodeReview Version

v1.8.1 (be5e79c1a) darwin/arm64

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

OpenAI

### Bug Description

When reviewing my code repository, an error occurred when reading files

GitHub repo: https://github.com/shawicx/scx-api-tool

Image

Image

### Steps to Reproduce

1. ocr config provider - deepseek api
2. ocr config model - deepseek-v4-pro
3. ocr view 

### Expected Behavior

It can correctly read files and review code

### Logs / Error Output

```shell

```

### Additional Context

_No response_

mvanhorn:
## Description

A truncated LLM response body surfaces from the SDK as `io.ErrUnexpectedEOF`, and the client returned it straight to the caller. A single truncated stream therefore failed the whole review run even though a retry would usually succeed.

This retries the completion exactly once when the error is `io.ErrUnexpectedEOF`, and otherwise leaves behavior unchanged.

Closes #664

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Added four table-free unit tests in `internal/llm/client_test.go` that drive the client against a stub server, then verified they actually pin the new behavior by reverting the implementation and re-running them.

Retry semantics covered by the tests:

- `TestOpenAIClient_RetriesTruncatedResponse` - one truncated response then a good one succeeds.
- `TestOpenAIClient_StopsAfterSecondTruncatedResponse` - two consecutive truncations give up and return the error (exactly 2 requests, no unbounded retry loop).
- `TestOpenAIClient_DoesNotRetryNonRetryableError` - a non-EOF error is returned immediately without a second request.
- `TestOpenAIClient_DoesNotRetryTruncatedResponseAfterCancellation` - a cancelled context returns the context error instead of retrying.

Regression proof, so these are not vacuous. With `internal/llm/client.go` reverted to `main` and only the tests applied:

```
--- FAIL: TestOpenAIClient_RetriesTruncatedResponse (0.00s)
    client_test.go:544: CompletionsWithCtx: error reading response body: unexpected EOF
--- FAIL: TestOpenAIClient_StopsAfterSecondTruncatedResponse (0.00s)
    client_test.go:587: requests = 1, want 2
FAIL
```

With the fix applied, `go test -count=1 ./internal/llm/...` is `ok`.

  Not run: I ran the package directly (`go test -count=1 ./internal/llm/...`, ok) rather than the full `make test` target, which reaches beyond this package.
  The revert-and-rerun regression proof above was done by hand.

## Checklist

  `gofmt -l internal/llm/` is empty and `go vet ./internal/llm/...` is clean.
  See the regression proof above; the tests fail against `main` and pass with the change.

## Related Issues

Closes #664. No other open issue or PR touches `internal/llm/client.go` retry behavior as far as I can see.

## Implementation notes

The retry is deliberately narrow:

- It triggers only on `errors.Is(err, io.ErrUnexpectedEOF)`, so ordinary API errors, rate limits, and auth failures are untouched.
- It retries at most once. A second truncation returns the error rather than looping.
- It checks `ctx.Err()` before retrying and again after a failed retry, so a cancelled or timed-out context returns the context error and never issues an extra request.
- On a failed retry that is itself an unexpected EOF, the original error is preserved rather than replaced, so the caller sees the same error class as before.

## Notes for reviewers

Two things worth your judgment:

1. Retry count is fixed at one, with no backoff. For a truncated stream an immediate retry is usually right, but if you would prefer this to go through a shared retry/backoff helper, point me at it and I will rework it.
2. I scoped this strictly to `internal/llm`. While testing I noticed the Python example suites under `examples/*_ci/` do not collect cleanly without `__init__.py` files, which is unrelated to this issue; I left it alone rather than widen the PR. Happy to file that separately if it is news to you.

### DIFF
--- a/internal/llm/client.go
+++ b/internal/llm/client.go
@@ -8,7 +8,9 @@ package llm
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
 	"strings"
 	"sync"
 	"time"
@@ -362,6 +364,23 @@ func (c *OpenAIClient) CompletionsWithCtx(ctx context.Context, req ChatRequest) 
 	}
 
 	sdkResp, err := c.sdk.Chat.Completions.New(ctx, params, opts...)
+	if errors.Is(err, io.ErrUnexpectedEOF) {
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			return nil, ctxErr
+		}
+		retryResp, retryErr := c.sdk.Chat.Completions.New(ctx, params, opts...)
+		if retryErr == nil {
+			sdkResp = retryResp
+			err = nil
+		} else {
+			if ctxErr := ctx.Err(); ctxErr != nil {
+				return nil, ctxErr
+			}
+			if !errors.Is(retryErr, io.ErrUnexpectedEOF) {
+				err = retryErr
+			}
+		}
+	}
 	if err != nil {
 		return nil, err
 	}
--- a/internal/llm/client_test.go
+++ b/internal/llm/client_test.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


lizhengfeng101 (approved):
LGTM

mvanhorn:
Appreciate you landing the truncation retry, @lizhengfeng101. Should make long completions much less flaky.

PR #676, opened 2026-08-02
feat(llm): add per-model reasoning effort

allenxu09:
## Summary

- store reasoning effort per model in backward-compatible `model_settings`
- send the protocol-specific request fields:
  - OpenAI Chat Completions: `reasoning_effort`
  - OpenAI Responses: `reasoning.effort`
  - Anthropic Messages: `output_config.effort`
- add reasoning-effort selection to the provider/model TUIs and VS Code configuration UI
- add CLI, environment, and GitHub Action overrides plus English/Chinese documentation

## Why

Reasoning effort is model-specific, and each supported protocol uses a different request shape. Previously OCR had no first-class way to select it, so users had to rely on global `extra_body` overrides that did not follow model switches cleanly.

## User impact

Users can select an effort while adding or switching models, set it with `ocr config set reasoning_effort <value>`, override it per run with `--reasoning-effort` or `OCR_LLM_REASONING_EFFORT`, and reset to the provider default without removing other model settings.

## Validation

- `go test ./internal/llm ./internal/agent`
- targeted CLI/config/TUI tests in `./cmd/opencodereview`
- `go vet ./internal/llm ./internal/agent ./cmd/opencodereview`
- VS Code Jest suite: 13 suites / 103 tests
- VS Code webpack compile and ESLint (one pre-existing warning in `IdleView.tsx`)
- `npm run test:github-actions`
- README and docs translation-sync checks


allenxu09:
@lizhengfeng101 ping

zhaozhenxiang:
这是一个刚性需求，需要实现。
我没有review pr的代码。

allenxu09:
@MuoDoo @lizhengfeng101 ptal

stay-foolish-forever:
Thanks for the detailed work here — the protocol mapping and the test coverage are clearly well thought through.

That said, I have reservations about landing `reasoning_effort` as a first-class configuration field in OCR.

**1. The option is neither universal nor stable across providers and models.** Support varies by provider, and even within one provider it varies by model — many models don't accept the field at all, and the valid value sets differ (this PR already has to special-case OpenAI vs. Anthropic, and notes that model-level support is left to the provider to reject). Promoting it to a dedicated config key implies a level of portability that doesn't really exist, and it puts OCR in the position of tracking each vendor's evolving effort vocabulary. Users would hit provider-side errors that OCR's own validation passed.

**2. `extra_body` already covers this use case.** It's supported both in the `llm` section and per provider entry, so anyone who needs a specific reasoning effort can set it today without new surface area:

```json
{
  "providers": {
    "anthropic": {
      "extra_body": { "output_config": { "effort": "high" } }
    }
  }
}
```

This keeps vendor-specific request fields in the one place that's explicitly designed to be vendor-specific — which also means new fields from any provider work immediately, with no OCR release needed. (Happy to see documentation updates about this.)

**3. If you still think first-class support is warranted, let's discuss it before the implementation.** Please open an Issue or a Discussion so we can settle the design questions first — particularly how per-model settings should be scoped and validated, and whether the same mechanism should generalize to other model-specific parameters rather than being built for `reasoning_effort` alone. A generic per-model settings mechanism may well be the right thing; I'd just rather agree on that shape before merging a single-field version of it.

lizhengfeng101:
Thanks for the work here, and sorry for the slow follow-up.

For the action/CI use case this is now covered by #1154, which sets `reasoning_effort` through `extra_body` — the same approach suggested above. Since the first-class config design still needs to be settled and this has drifted out of sync with main, I'll close it for now. If you'd like to pursue the generic per-model settings idea, please open a Discussion so we can agree on the shape first. Appreciate the contribution!

PR #683, opened 2026-08-02
feat(allowlist): exclude snapshots, testdata, fixtures, and generated code from review

lizhengfeng101:
## Summary

- Add 9 new patterns to `default_exclude_patterns.json` to skip files that pass existing filters but have low review value:
  - **Snapshot files**: `__snapshots__/**`, `*.snap` (Jest/Vitest auto-generated)
  - **Test data directories**: `testdata/**`, `fixtures/**` (fixture data, not manually authored logic)
  - **Generated code**: `*.generated.*`, `*.gen.go`, `*.pb.go`, `*.pb.cc`, `*.pb.h` (protobuf/codegen output)
- Add 16 test cases covering both positive matches and negative (non-match) assertions to prevent false positives

## Rationale

These files are tracked by Git and have allowed extensions, so they reach the review stage. However, reviewing auto-generated protobuf stubs or golden-file test fixtures yields no actionable feedback. This change reduces noise without affecting user-authored code.

## Test plan

- [x] `make test` — all tests pass including new `TestIsExcludedPath` cases
- [x] `make check` — formatting and vet clean
- [x] `ocr review` — no comments on the change itself

### DIFF
--- a/internal/config/allowlist/allowed_ext_test.go
+++ b/internal/config/allowlist/allowed_ext_test.go
@@ -130,6 +130,33 @@ func TestIsExcludedPath(t *testing.T) {
 		{"julia test nested", "MyPkg/test/unit/foo.jl", true},
 		{"julia non-test", "src/model.jl", false},
 
+		// Snapshot files
+		{"jest snapshot dir", "src/__snapshots__/App.test.js.snap", true},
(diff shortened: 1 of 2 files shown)


css521 (approved):
LGTM

PR #529, opened 2026-07-27
feat(action): add fail-open category/severity publication controls (#478)

nitishagar:
## Summary

First increment toward #478 (the full issue is a roadmap). Adds category/severity-aware, fail-open publication controls to the reusable GitHub Action:

1. **Render a CLI-consistent `[category · severity]` badge** on every comment that carries metadata (inline and summary), matching `buildBadge()` in `cmd/opencodereview/output.go`.
2. **One opt-in routing destination**: move low-severity and/or selected-category findings from inline comments to the PR summary, via two new `action.yml` inputs (`route_severity_below`, `route_categories`).
3. **Fail-open semantics**: no finding is ever silently dropped — unknown/malformed metadata on a finding never matches the policy, and a malformed policy itself degrades to no-routing.
4. **Accounting reconciliation**: a new `routed` bucket is disjoint from `summary`/`skipped`/`failed`, so destination counts still sum to the raw input `total` (`inline + summary + skipped + failed + routed == total`).

With **no routing input set**, behavior is byte-equivalent to today except for the additive badge prefix on comments that carry `category`/`severity` metadata. No Go changes — `category`/`severity` already serialize into the Action's input JSON via `model.LlmComment`.

## Key design decisions (invariant → mechanism)

- **I1 (fail-open):** `routeComment` requires the finding's own category/severity to be a known enum member before it can match; unknown metadata falls through to the normal inline path (visible). `buildPolicy` returns a `NO_ROUTING` sentinel on any malformed input.
- **I2 (accounting):** new `routed` bucket + `comments_routed` output; the partition loop routes each finding into exactly one bucket via mutually-exclusive `continue` branches.
- **I3 (defaults):** empty policy = no-op; the badge is an additive prefix that renders `""` (no line) when metadata is absent, so metadata-less comments are byte-identical to today.
- **I4 (idempotency):** routing is a placement decision in the partition loop — routed findings `continue` before `reviewComments.push`, so they never enter `toSend`/`toRetry`/either `createReview` call, and carry no idempotency id. No double-post surface on retry.
- **I6 (badge consistency):** shared `buildBadge` byte-matches the CLI degeneration. Control-char sanitization is intentionally stricter than the CLI's `sanitizeTerminal` (strips `\t`/`\n` too) to defend the Markdown comment-body layout — a documented divergence that only manifests for malformed model output and is safer (clean enum values match exactly).

## At-or-below severity semantics

[...]

Issue #478, opened 2026-07-23
Add fail-open, non-destructive finding publication controls to the reusable Action


acoliver:
## Problem statement

OpenCodeReview emits structured category and severity for each finding, but the reusable GitHub Action does not render or use those fields when publishing review comments. Every line-addressable finding is treated as an inline candidate regardless of category/severity, while downstream users must either post everything or write a custom publication pipeline.

Released v1.7.15 evidence:

- Category/severity schema: https://github.com/alibaba/open-code-review/blob/v1.7.15/internal/config/toolsconfig/tools.json#L55-L85
- Action formatter uses content/location/suggestion but not category/severity: https://github.com/alibaba/open-code-review/blob/v1.7.15/scripts/github-actions/post-review-comments.js#L1013-L1044
- Structured fields originated in #16 / PR #311.
- Closed PR #299 explored filtering but did not land.

Production evidence from https://github.com/vybestack/llxprt-code/actions/runs/29946076923:

- 75 raw findings;
- 4 high, 37 medium, 33 low, and 1 undocumented info severity;
- 13 bug, 15 undocumented correctness, 31 maintainability, 12 test, 3 security, and 1 other category.

After exact deduplication, a conservative presentation policy that moved only low maintainability/test/style/documentation and info-like findings to a collapsed summary would have reduced inline candidates from 74 to 46 while preserving every finding. This is evidence of the need for publication routing, not a proposal to make that exact policy the upstream default.

The key requirement is non-destructive, fail-open control. Category and severity are model-generated metadata, not calibrated truth, and unknown values already occur in real output. Filtering must never silently discard malformed or lower-tier findings.

## Proposed solution

Add neutral, opt-in publication controls to the reusable GitHub Action.

### Rendering

- Render category and severity labels for inline and summary findings.
- Make missing/unknown values explicit instead of omitting them.

### Routing

Allow callers to configure deterministic routing by category/severity to destinations such as:

- inline review comment;
- sticky summary;
- artifact/raw output only when explicitly allowed;
- overflow summary.

Defaults should preserve existing behavior and fail open. Missing, unknown, or malformed metadata must remain visible unless a caller explicitly selects a different policy.

### Preservation and accounting

- Preserve every raw finding and original metadata in machine-readable output.
- Record the destination and reason for every finding.
- Expose mutually exclusive counts for raw, exact duplicate, overlap-suppressed, policy-routed-inline, policy-routed-summary, overflow, failed, and unknown-schema findings.
[...]

PR #685, opened 2026-08-03
feat(examples/gitlab): add fail-open category/severity publication controls

SeonggwanAhn:
## Summary
- Port GitHub Action fail-open category/severity publication controls (#478 / #529) to the GitLab CI example (`examples/gitlab_ci/post_review.py`).
- Add `[category · severity]` badges on inline and summary findings, plus optional `OCR_ROUTE_SEVERITY_BELOW` / `OCR_ROUTE_CATEGORIES` routing that moves matching findings to summary notes without dropping them.
- Split destination accounting into inline / no-line fallback / routed / failed, and document the new CI variables in the GitLab example README and workflow comments.

## Test plan
- [x] `python3 examples/gitlab_ci/post_review_test.py` (78 tests)
- [x] Smoke on a GitLab MR with `OCR_ROUTE_SEVERITY_BELOW=medium` and `OCR_ROUTE_CATEGORIES=style,documentation,test`
- [x] Confirm `security`/`high` findings still land as inline discussions with `suggestion:-0+0`
- [x] Confirm routed findings appear in a summary note with `Routed to summary…` reason and Before/After details

Made with [Cursor](https://cursor.com)

### DIFF
--- a/examples/gitlab_ci/.gitlab-ci.yml
+++ b/examples/gitlab_ci/.gitlab-ci.yml
@@ -22,6 +22,10 @@
 #   OCR_FAILURE_DELAY       - Delay (ms) after a non-rate-limit failure to pace subsequent requests (default: 1000)
 #   OCR_RATE_LIMIT_THRESHOLD - Proactively slow down when GitLab RateLimit-Remaining is at/below this value (default: 10, set 0 to disable)
 #
+# Optional CI/CD Variables (for category/severity publication routing):
+#   OCR_ROUTE_SEVERITY_BELOW  - Route findings at-or-below this severity (critical/high/medium/low) to summary notes instead of inline (fail-open; empty disables)
+#   OCR_ROUTE_CATEGORIES      - Comma-separated categories to route to summary (bug, security, performance, maintainability, test, style, documentation, other)
+#
 # Note: The pipeline also configures llm.extra_body to '{"thinking": {"type": "disabled"}}'
 #   to disable thinking mode for compatibility with various LLM providers.
 #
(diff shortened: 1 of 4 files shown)


lizhengfeng101 (commented):
## Review

Nice port — the fail-open design is solid and tests are thorough (78 cases covers the surface well).

A few things I noticed:

### `sanitize_metadata` gap in `route_comment`

`build_badge` runs values through `sanitize_metadata` (strips control chars), but `route_comment` just does `.strip().lower()` on the raw value. So if an LLM spits out `"bu\ng"` as a category, the badge renders `[bug · high]` but routing won't match it against `CATEGORIES` — it stays inline silently.

Fail-open means nothing gets dropped, so it's safe, but it's confusing: the badge says "bug" while routing doesn't recognize it as one. Probably worth running `sanitize_metadata` in `route_comment` too:

```python
cat_raw = sanitize_metadata(comment.get("category") if comment else None).strip().lower()
sev_raw = sanitize_metadata(comment.get("severity") if comment else None).strip().lower()
```

### Multiple notes per MR could get noisy

When a review has all three flavors (no-line, routed, failed), you'll post 4 separate notes plus inline comments. On a busy MR that's a lot of bot noise. Might be worth collapsing no-line + routed + failed into one note with section headers. Not a blocker, just something to think about for UX.

### Minor

- Removing the unused `start_line` assignment from the loop — good cleanup, confirmed it was never passed to the discussion payload.
- Two `test_badge_prefix` methods exist across different test classes. Not a conflict in Python but easy to confuse when grepping test names.

Overall LGTM with the sanitize fix as the one actionable item.

lizhengfeng101 (approved):
LGTM

SeonggwanAhn:
Addressed the `sanitize_metadata` feedback in `route_comment` (87f1727) — PTAL.

PR #687, opened 2026-08-03
feat(cli): support per-run LLM provider and model overrides

### LINKED ISSUE #680: Allow selecting provider/model per invocation (e.g. `ocr review --provider X --model Y`)
### Problem Statement

Provider and model are global state in `~/.opencodereview/config.json` (`ocr config provider` / `ocr config set model`). 

To review the same diff with a different model I have to mutate that file, which affects every other run — and concurrent runs race over it.

My flow: I orchestrate reviews from a coding agent. I want to run the same diff through several models (Claude, Codex/GPT, others) via OCR, collect the JSON outputs, deduplicate findings across runs, and post the merged result as comments on a merge request. 

Agreement between models is a strong signal for filtering out false positives — but today there is no clean way to vary the model per run.

### Proposed Solution

Add per-run flags to `ocr review` (and `ocr scan`):

| Flag | Description |
| --- | --- |
| `--provider <name>` | A provider already present in the `providers` map |
| `--model <name>` | Model for that provider |

Both override config for that invocation only — nothing is written back.

```bash
ocr review --from main --to feat --provider anthropic --model claude-opus-4-6 --format json
ocr review --from main --to feat --provider openai    --model gpt-5-codex     --format json
```

Suggested precedence: **flag > env vars > config file**.

It would also help if `--format json` reported which provider/model produced the result, so findings can be attributed during deduplication:

```json
{ "llm": { "provider": "anthropic", "model": "claude-opus-4-6" }, "comments": [] }
```

### Alternatives Considered

- **Env vars** (`OCR_LLM_URL` / `OCR_LLM_TOKEN` / `OCR_LLM_MODEL` / `OCR_USE_ANTHROPIC`) work per process, but bypass the configured `providers` map - every call has to re-supply URL and token, which is clumsy inside an agent and error-prone in CI.
- **Rewriting `config.json` between runs** - racy under concurrency and destroys the user's own settings.
- **Delegate mode** - solves a different problem: one agent, one model, no independent second opinion.

### Affected Area

CLI / Commands

### Additional Context

**Additional Areas**: `Configuration`, `Output / Formatting`
**Related**: per-run overrides would also make it easy to benchmark models against each other on a fixed set of diffs, which is useful when picking a default for a team.

Makishima:
## Description

Implements #680 by adding per-run LLM provider and model selection for both `review` and `scan`.

- Adds the `--provider` flag alongside the existing `--model` flag.
- Supports providers configured under both `providers` and `custom_providers`.
- Applies strict configuration precedence: CLI flags > environment variables > config file.
- Prevents partial configuration strategies from being mixed.
- Preserves environment credential fallback for the selected built-in provider.
- Prevents a model configured for one provider from leaking into another provider selection.
- Adds the resolved provider and model to JSON output through the `llm` field.
- Documents the new flags, precedence rules, and JSON output in all localized READMEs.

Compatibility note: a complete `OCR_LLM_*` environment configuration now takes precedence over file-based provider configuration.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] Manual testing:
  - Verified `review --help` exposes `--provider` and `--model`.
  - Verified `scan --help` exposes `--provider` and `--model`.

Additional verification:

- `go vet ./...`
- `make build`
- Focused resolver regression tests for cross-provider model selection
- Git diff formatting and scope checks

## Checklist

- [x] I have signed the CLA

## Related Issues

Closes #680


lizhengfeng101 (commented):
It is not recommended to add too much content to the README; instead, it is suggested to add the descriptions of these commands to the official documentation section.

ref：pages/src/pages/DocsPage.tsx

Makishima:
Addressed in 1afd06e: the per-run provider/model documentation was removed from the localized READMEs and moved to the official localized CLI reference pages rendered by DocsPage.

One clarification regarding the added command-summary row: `ocr scan` and its `ocr s` alias are not introduced by this PR. They already exist in `cmd/opencodereview/scan_cmd.go`:

- `Use: "scan [flags]"`
- `Aliases: []string{"s"}`

The row only documents this existing command because the PR adds `--provider` and `--model` support to both `review` and `scan`. If you prefer keeping the documentation diff strictly limited to the new flags, I can remove the command-summary row.

> It is not recommended to add too much content to the README; instead, it is suggested to add the descriptions of these commands to the official documentation section.
> 
> ref：pages/src/pages/DocsPage.tsx

lizhengfeng101 (commented):
Thanks for this PR! The feature design is solid, with thorough tests and documentation.

One change needed: **please keep the strategy resolution order unchanged for the non-`--provider` path** (i.e., maintain config → env → Claude Code env → Shell RC).

Reasons:
- The `--provider` short-circuit logic is independent of the strategy list order — the feature works either way
- The precedence change is a separate breaking change that would silently alter behavior for existing users
- There is no need to take on this compatibility risk

Specifically, the `strategies` slice in `ResolveEndpointWithOptions` should be restored to:

```go
strategies := []struct {
    name string
    fn   func() (ResolvedEndpoint, bool, error)
}{
    {"OCR config file", func() (ResolvedEndpoint, bool, error) { return tryOCRConfig(configPath, opts) }},
    {"OCR environment", func() (ResolvedEndpoint, bool, error) { return tryOCREnv(opts.Model) }},
    {"Claude Code environment", func() (ResolvedEndpoint, bool, error) { return tryCCEnv(opts.Model) }},
    {"Shell rc file", func() (ResolvedEndpoint, bool, error) { return tryShellRC(opts.Model) }},
}
```

The corresponding tests (e.g., `TestResolveEndpoint_OCREnvironmentPrecedesConfig`) should be adjusted accordingly. Everything else LGTM.

lizhengfeng101 (changes requested):
@Makishima please keep the strategy resolution order unchanged for the non---provider path

Makishima:
Addressed in 67a0b14. Restored the legacy non-`--provider` resolution order:

1. OCR config file
2. OCR environment
3. Claude Code environment
4. Shell rc file

The explicit `--provider` path still short-circuits before the strategy list, so per-run provider selection is unaffected. I also updated the precedence/fallback tests and all three localized CLI reference pages to match the restored behavior.

> @Makishima please keep the strategy resolution order unchanged for the non---provider path

lizhengfeng101 (changes requested):
merge conflicts

lizhengfeng101 (approved):
LGTM. Strategy order restored, tests updated to reflect config-first resolution, docs are consistent. No blocking issues remain.

lizhengfeng101 (approved):
LGTM

PR #689, opened 2026-08-03
feat(cli): select provider per run

### LINKED ISSUE #680: Allow selecting provider/model per invocation (e.g. `ocr review --provider X --model Y`)

xianjianlf2:
## Summary
- add `--provider` to `ocr review` and `ocr scan`
- resolve provider/model overrides from configured provider maps without writing config.json
- preserve provider provenance in the resolved endpoint and review manifest

## Validation
- `go test ./internal/llm ./cmd/opencodereview`
- `go test ./...`

Closes #680

### DIFF
--- a/cmd/opencodereview/flags_test.go
+++ b/cmd/opencodereview/flags_test.go
@@ -35,6 +35,16 @@ func TestParseReviewFlagsModelOverride(t *testing.T) {
 	}
 }
 
+func TestParseReviewFlagsProviderOverride(t *testing.T) {
+	opts, err := parseReviewFlags([]string{"--provider", "openai"})
+	if err != nil {
+		t.Fatalf("parseReviewFlags: %v", err)
+	}
+	if opts.provider != "openai" {
+		t.Errorf("provider = %q, want openai", opts.provider)
+	}
+}
+
 func TestParseReviewFlagsResume(t *testing.T) {
 	opts, err := parseReviewFlags([]string{"--from", "main", "--to", "feature", "--resume", "session-123"})
 	if err != nil {
--- a/cmd/opencodereview/review_cmd.go
+++ b/cmd/opencodereview/review_cmd.go
@@ -33,6 +33,7 @@ type reviewOptions struct {
 	audience        string
 	background      string
 	backgroundFile  string
+	provider        string
 	model           string
 	concurrency     int
 	perFileTimeout  int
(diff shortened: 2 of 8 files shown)


lizhengfeng101:
Thank you for working on this! #687 was merged first and covers the same scope (closes #680). Appreciate the effort — the approaches were quite similar in design.

PR #704, opened 2026-08-03
fix(npm): prevent binary unavailability during auto-upgrade

### LINKED ISSUE #703: ocr command becomes unavailable during npm auto-upgrade (binary gap window + Windows file lock)
### OpenCodeReview Version

`v1.8.5` (reproduced on current main, commit `3b9c5e7`)

### Operating System

- macOS (Apple Silicon)
- Windows
- Linux (x86_64)

(All three platforms are affected; Windows is the worst case.)

### Installation Method

npm (global)

### LLM Provider

### Bug Description

When `ocr` is installed globally via npm, the `ocr` command becomes unavailable while the background auto-upgrade runs (`scripts/update.js` executing `npm i -g`), failing with:

```
[ERROR] OpenCodeReview binary not found. Run: npm install -g @alibaba-group/open-code-review
```

The root cause is that the auto-upgrade uses a "delete-then-install" approach, leaving a window where the native binary is missing on disk. Any `ocr` invocation during that window fails because `resolveNativeBinary()` returns `null`.

Upgrade flow (`bin/ocr.js` + `scripts/update.js`):

1. `bin/ocr.js:43-51` — on every `ocr` run, `scripts/update.js` is spawned as a detached background process.
2. `scripts/update.js:164` — when a newer version is found, it runs `spawnSync("npm", ["i", "-g", `${pkgName}@${latestVersion}`])` to reinstall globally.
3. npm reinstalls by first removing the old platform package `@alibaba-group/ocr-<platform>` (which contains `bin/opencodereview`), then downloading and installing the new one.
4. During the "remove old → install new" gap in step 3, `resolveNativeBinary()` in `scripts/platform.js:36` finds neither the platform-package binary nor the legacy `bin/opencodereview` path, and returns `null`.
5. This hits the `if (!resolved)` branch at `bin/ocr.js:12-17`, which prints the error and `exit(1)`.

The window length depends on network download speed (the platform package ships a multi-MB binary), and can last from a few seconds to tens of seconds.

**Additional Windows-only problem**: `ocr.js:54` immediately `spawnSync`s the native binary right after spawning the detached upgrader. This means `opencodereview.exe` is still running when the background `npm i -g` tries to delete/replace it. Windows refuses to remove a running `.exe` (EBUSY/EPERM), so the upgrade fails and can leave the install in a broken "old binary deleted, new binary not yet written" state, making `ocr` **permanently unavailable** until a manual `npm i -g` repairs it.

### Steps to Reproduce

```console
# 1. Install an older version globally
$ npm i -g @alibaba-group/open-code-review@<old-version>

# 2. Wait for / trigger the background upgrade (run any ocr command, 18-min cooldown elapsed)
$ ocr version

# 3. While the background npm i -g is still running, keep invoking ocr
#    (during the gap between old binary removal and new binary install)
$ ocr version
[ERROR] OpenCodeReview binary not found. Run: npm install -g @alibaba-group/open-code-review
```

On Windows it's easiest to reproduce: keep an `ocr review` running for a while (binary in execution) while the background upgrade triggers — `npm i -g` then fails because the file is locked.

### Expected Behavior

Auto-upgrade should never make `ocr` unavailable at any point in time. The upgrade should be transparent to the user — either keep the old binary until the new one is verified ready (atomic replace / install-then-remove), or on Windows detect that the binary is in use and wait for the foreground process to exit before replacing, rather than failing outright and leaving a broken install behind.

### Logs / Error Output

```shell
$ ocr version
[ERROR] OpenCodeReview binary not found. Run: npm install -g @alibaba-group/open-code-review
```

On Windows, when `npm i -g` fails it also writes the `~/.opencodereview/update-available` hint (`scripts/update.js:173`), causing every subsequent `ocr` run to print the upgrade nudge (compounds with #697).

### Additional Context

Relevant code locations:

- `bin/ocr.js:43-51` — detached background spawn of `update.js`
- `bin/ocr.js:12-17` — hard-fail branch when binary is not found
- `bin/ocr.js:54` — concurrent background upgrade and foreground binary execution
- `scripts/update.js:164` — `npm i -g` global reinstall (non-atomic delete-then-install)
- `scripts/platform.js:36-58` — `resolveNativeBinary()` lookup logic

Related issue: #697 (upgrade nudge shown even when already on the latest version) — this compounds with the present bug.

Possible fix directions:

1. Have the background upgrade download to a temporary location first and do an atomic `rename(2)` swap after checksum verification, eliminating the delete-then-install gap.
2. On Windows, detect that the binary is in use and wait for the foreground process to exit before replacing (or use a post-restart self-replacement / launcher strategy).
3. During the gap window, when `resolveNativeBinary()` can't find the binary, fall back to a cached copy of the last binary (e.g. kept under `~/.opencodereview/`) instead of hard-failing.

lizhengfeng101:
## Summary

- Add a binary cache fallback to `~/.opencodereview/bin/` so `ocr` remains available while the background auto-upgrade runs `npm i -g`
- Before upgrading, the current binary is copied to the cache; `resolveNativeBinary()` uses it as a last resort when the platform package binary is missing
- After a successful upgrade, the cache is refreshed with the new binary

## Problem

When the background auto-upgrade runs `npm i -g`, npm removes the old platform package before installing the new one. During this gap window, `ocr` fails with "binary not found". On Windows, the running `.exe` cannot be deleted, so the upgrade fails entirely and can leave a permanently broken install.

## Solution

A simple binary cache mechanism:
1. `scripts/update.js` copies the current binary to `~/.opencodereview/bin/` before running `npm i -g`
2. `scripts/platform.js` adds this cache path as the third fallback in `resolveNativeBinary()` (after platform package → legacy path → **cache** → null)

This ensures `ocr` always has a working binary to fall back on, even during the upgrade gap window or after a failed Windows upgrade.

Closes #703

## Test plan

- [x] `make check` passes
- [x] `make test` passes
- [ ] Manual: install old version → trigger upgrade → rapidly invoke `ocr version` during upgrade → should always succeed via cache fallback

### DIFF
--- a/bin/ocr.js
+++ b/bin/ocr.js
@@ -17,6 +17,13 @@ if (!resolved) {
 }
 const binaryPath = resolved.path;
 
+if (resolved.fromStaged) {
+  const ver = resolved.stagedVersion ? ` (v${resolved.stagedVersion})` : "";
+  process.stderr.write(
+    `\x1b[2m[ocr] Using staged binary${ver}; run npm i -g @alibaba-group/open-code-review to reinstall.\x1b[0m\n`
+  );
+}
+
 const hintFile = path.join(os.homedir(), ".opencodereview", "update-available");
 try {
   const hint = JSON.parse(fs.readFileSync(hintFile, "utf8"));
--- a/scripts/platform.js
+++ b/scripts/platform.js
@@ -7,6 +7,10 @@ const os = require("os");
 const IS_WINDOWS = process.platform === "win32";
 const BINARY_FILENAME = IS_WINDOWS ? "opencodereview.exe" : "opencodereview";
 
+const STATE_DIR = path.join(os.homedir(), ".opencodereview");
+const STAGED_BIN_DIR = path.join(STATE_DIR, "staged");
+const VERSION_JSON_PATH = path.join(STAGED_BIN_DIR, "version.json");
+
 const PLATFORM_PKG = {
   "darwin-arm64": "@alibaba-group/ocr-darwin-arm64",
   "darwin-x64": "@alibaba-group/ocr-darwin-x64",
@@ -34,6 +38,41 @@ function getPlatformPackageName() {
   return PLATFORM_PKG[key] || null;
 }
 
+function resolveStagedBinary() {
+  try {
+    const raw = fs.readFileSync(VERSION_JSON_PATH, "utf8");
+    const meta = JSON.parse(raw);
+    if (!meta.version || !meta.stagedAt || !meta.platform) {
+      return null;
+    }
+
+    const currentPlatform = `${process.platform}-${process.arch}`;
+    if (meta.platform !== currentPlatform) {
+      return null;
+    }
+
+    const binPath = path.join(STAGED_BIN_DIR, BINARY_FILENAME);
+    let stat;
+    try {
+      stat = fs.statSync(binPath);
+    } catch (_) {
+      return null;
+    }
+    if (!stat.isFile() || stat.size === 0) {
+      return null;
+    }
+
+    return {
+      path: binPath,
+      fromPlatformPkg: false,
+      fromStaged: true,
+      stagedVersion: meta.version,
+    };
+  } catch (_) {
+    return null;
+  }
+}
+
 function resolveNativeBinary() {
   const pkgName = getPlatformPackageName();
   if (pkgName) {
@@ -41,7 +80,7 @@ function resolveNativeBinary() {
       const pkgDir = path.dirname(require.resolve(`${pkgName}/package.json`));
       const binPath = path.join(pkgDir, "bin", BINARY_FILENAME);
       if (fs.existsSync(binPath)) {
-        return { path: binPath, fromPlatformPkg: true, fromCache: false };
+        return { path: binPath, fromPlatformPkg: true, fromStaged: false };
       }
     } catch (err) {
       if (err.code !== "MODULE_NOT_FOUND") {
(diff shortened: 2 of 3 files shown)


crestonepeaks:
@lizhengfeng101 Thanks for working on #703. We have additional Linux GitHub Actions evidence from a consumer pinned to `alibaba/open-code-review@5679f44a07a17cbd88e8a7341b4f72a4a2b191cf` with `ocr_version: 1.9.10`.

Across three independent runs, the pinned runtime installed and passed its version check before the detached updater disrupted the global package:

- configuration completed, then the review step failed with `ocr: command not found` (exit 127);
- a retry passed installation/version checking, then failed with `Error: Cannot find module '../scripts/platform'`;
- a later run completed installation and configuration, then failed with `Error: Cannot find module '/usr/local/bin/ocr'`; runner cleanup reported an orphaned `npm i @alibaba-...` process.

This appears broader than a missing native binary after the launcher has loaded. PR #704 still requires `../scripts/platform` before its staged-binary fallback and does not protect the global `ocr` shim or the rest of the JavaScript package tree while `npm i -g` replaces them.

As a controlled comparison, we set the launcher-documented `OCR_NO_UPDATE=1` at the action-step level while keeping the same action SHA and OCR 1.9.10 pin. The exact-head run then completed successfully: 2/2 selected items completed, zero failed/waived items, exit code 0, and no binary-gap error or orphan npm updater in the runner cleanup.

Would you be open to extending #704 (or a follow-up) with regression coverage for the missing shim and missing JavaScript-module cases, and considering an action-level no-update option/default when `ocr_version` is explicitly pinned? We are happy to help with tests or a patch if useful, and can provide additional sanitized log excerpts.

weebeeblazin-ui:
https://github.com/user-attachments/assets/59be44aa-d15f-4d91-a2ec-c478bc7eb4ab

weebeeblazin-ui:
muse-symbol-foundation (1).md

PR #716, opened 2026-08-04
feat(config): make per-file token limit configurable

### LINKED ISSUE #707: Expose max_tokens (per-file review prompt ceiling) as user-configurable
## Description

`max_tokens` (the per-file-review-task prompt-size ceiling) is a hardcoded, model-agnostic constant compiled into the engine, with no CLI flag, config key, environment variable, or `llm.extra_body` override that reaches it. On the version we run (built 2026-08-03, postdating both #417 and #421) the literal value is:

```
"MAX_TOKENS": 58888
```

found twice in the compiled binary's embedded static task-config JSON blocks (the main review task and the fix-prompt-generation task), alongside other clearly-static engine defaults (batch size, dedup thresholds, tool-call limits) — not inside any per-model or per-provider metadata structure. `context_window` / `contextWindow` / `context_length` do not appear anywhere in the binary.

## Why this matters

We route real, materially cheaper 1M-context models (e.g. GLM-5.2, 1,048,576-token context window) through this engine specifically to reduce review cost. The 58,888-token ceiling has nothing to do with what the routed model can actually hold — it is a flat cap applied identically regardless of provider/model, so a large real-context model gets throttled down to a fraction of a percent of its own window before it ever sees a file.

Concretely, on our repo, a single per-glob `--rule` file merging JIRA + Confluence + repository guidance + a knowledge-base tree already reaches ~65,500 tokens on its own, before any diff/file content is added. Every file matched by that rule glob blows the ceiling immediately: we observed a `completed_with_errors` run where two files hit the 80% warning (`"prompt tokens (60361) exceed 80% of max_tokens(58888)"`, from the shared threshold #421 introduced) and the file actually carrying the review's substantive content never completed (`"main_task did not complete before stopping"`).

We worked around this from our side by relevance-trimming our own rule-file content down to a smaller budget before invoking the engine — but this is a lossy mitigation for something the engine already computes and reports the exact number for in its own warnings; it should not require reverse-engineering the compiled binary to discover the ceiling exists, let alone that it cannot be changed.

## What we checked (exhaustively) before concluding this is not user-configurable

- **CLI flags** (`ocr review --help`): the only token-related flag is `--max-tokens-budget` — a different, unrelated *total* input+output budget for the whole review run (default `0` = unlimited). It does not touch the per-file `max_tokens` ceiling.
- **Config keys** (`ocr config set <key> <value>` rejects unknown keys against a full enumerated list: `provider, model, providers.<name>.<field>, custom_providers.<name>.<field>, mcp_servers.<name>.<field>, llm.url, llm.auth_token, llm.auth_header, llm.model, llm.protocol, llm.use_anthropic, llm.extra_body, llm.extra_headers, language, telemetry.*`): `max_tokens` and `review.max_tokens` are both rejected as unknown.
- **Environment variables** (complete `OCR_*` list extracted from the binary): `OCR_LLM_TIMEOUT, OCR_LLM_MODEL, OCR_LLM_URL, OCR_LLM_PROTOCOL, OCR_ENABLE_TELEMETRY, OCR_USE_ANTHROPIC, OCR_CONTENT_LOGGING, OCR_LLM_EXTRA_HEADERS, OCR_VIEWER_ALLOWED_HOSTS, OCR_LLM_AUTH_HEADER, OCR_CONFIG_PATH, OCR_LLM_TOKEN` — none relate to a token/context ceiling.
- **`llm.extra_body`** (used elsewhere for provider-specific request tweaks): this is serialized into the outbound LLM API request body — it cannot reach an engine-side check that happens *before* the request is even built, comparing the assembled prompt against the compiled-in constant.

## Suggested fix

Expose `max_tokens` (and/or the 80% warning threshold it already computes against, per #417/#421) as user-configurable — a `--max-tokens-per-file` CLI flag, a `max_tokens`/`review.max_tokens` config key, or an `OCR_MAX_TOKENS` env var would all resolve this cleanly, ideally with a documented interaction with `--max-tokens-budget`. Even better: derive a sane default from the configured model's own advertised context window when one is available, rather than one flat constant for every provider/model.

## Environment

- Engine version: 1.8.6 (npm package `@alibaba-group/open-code-review`), installed via `npm install -g @alibaba-group/open-code-review@1.8.6`
- Confirmed via direct extraction of the installed native binary (`opencodereview`, Mach-O arm64) and its embedded strings
- Related: #417, #421 (the 80% threshold refactor — confirms the mechanism, not the configurability), #409 (related large-MR cost/timeout pain point)

Linxiushen:
## Description

Expose the per-file prompt token ceiling as a saved `max_tokens` setting and a per-run `--max-tokens` override. The resolved value now follows `--max-tokens` > saved configuration > embedded task-template default for both `ocr review` and `ocr scan`.

This keeps the per-file ceiling independent from `--max-tokens-budget`, which caps aggregate usage across a run. Invalid persisted values and negative CLI values fail with actionable errors.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Automated verification on Windows with Go 1.25.5:

- `go test -count=1 ./...`
- `go vet ./...`
- Focused config, flag-parsing, precedence, and persistence round-trip tests
- `git diff --check`

## Checklist

- [ ] I have signed the CLA

## Related Issues

Closes #707

## AI Assistance Disclosure

AI assistance was used during implementation. I reviewed the complete diff and verified the tests locally.

### DIFF
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -291,6 +291,7 @@ type MCPServerConfig struct {
 type Config struct {
 	Provider        string                     `json:"provider,omitempty"`
 	Model           string                     `json:"model,omitempty"`
+	MaxTokens       int                        `json:"max_tokens,omitempty"`
 	Providers       map[string]ProviderEntry   `json:"providers,omitempty"`
 	CustomProviders map[string]ProviderEntry   `json:"custom_providers,omitempty"`
 	Llm             LlmConfig                  `json:"llm,omitempty"`
@@ -356,6 +357,7 @@ func LoadAppConfig(path string) (*Config, error) {
 var supportedConfigKeys = []string{
 	"provider",
 	"model",
+	"max_tokens",
 	"providers.<name>.<field>",
 	"custom_providers.<name>.<field>",
 	"mcp_servers.<name>.<field>",
@@ -427,6 +429,12 @@ func setConfigValue(cfg *Config, key, value string) error {
 		} else {
 			cfg.Model = value
 		}
+	case "max_tokens":
+		maxTokens, err := strconv.Atoi(value)
+		if err != nil || maxTokens <= 0 {
+			return fmt.Errorf("invalid max_tokens %q: must be a positive integer", value)
+		}
+		cfg.MaxTokens = maxTokens
 	case "llm.url", "llm.URL":
 		cfg.Llm.URL = value
 	case "llm.auth_token", "llm.AuthToken":
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -143,10 +143,13 @@ func runConfigUnset(key string) error {
 	if key == "provider" {
 		return unsetActiveProvider(configPath)
 	}
+	if key == "max_tokens" {
+		return unsetMaxTokens(configPath)
+	}
 
 	parts := strings.SplitN(key, ".", 2)
 	if len(parts) != 2 || parts[1] == "" {
-		return fmt.Errorf("unset supports provider, custom_providers.<name>, and mcp_servers.<name>")
+		return fmt.Errorf("unset supports provider, max_tokens, custom_providers.<name>, and mcp_servers.<name>")
 	}
 
 	switch parts[0] {
@@ -155,8 +158,23 @@ func runConfigUnset(key string) error {
 	case "mcp_servers":
 		return unsetMCPServer(configPath, parts[1])
 	default:
-		return fmt.Errorf("unset supports provider, custom_providers.<name>, and mcp_servers.<name>")
+		return fmt.Errorf("unset supports provider, max_tokens, custom_providers.<name>, and mcp_servers.<name>")
 	}
+}
+
+func unsetMaxTokens(configPath string) error {
+	cfg, err := loadOrCreateConfig(configPath)
+	if err != nil {
+		return fmt.Errorf("load config: %w", err)
+	}
+
+	cfg.MaxTokens = 0
+	if err := saveConfig(configPath, cfg); err != nil {
+		return err
+	}
+
+	fmt.Println("Cleared max_tokens; using the embedded template default.")
+	return nil
 }
 
 func unsetActiveProvider(configPath string) error {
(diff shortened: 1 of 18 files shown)


lizhengfeng101 (commented):
Hey, nice work — tests and docs are solid. Two things I'd like to discuss:

**1. `MaxTokens` shouldn't also control LLM output length**

Right now `Template.MaxTokens` is passed directly into `ChatRequest.MaxTokens`, which maps to the API's `max_completion_tokens`. That means when a user sets `max_tokens 200000` to accommodate a model with a larger context window, they're also telling the model it can produce up to 200,000 output tokens per response — that's not the intent.

This field currently serves two roles: the compression/pre-flight threshold baseline (context window ceiling) and the max output passed to the API. These should be independent. I'd suggest this PR only let the configurable value affect compression thresholds and the pre-flight gate, and leave the LLM call's max output at its existing default (or split them into two fields).

**2. Need `ocr config unset max_tokens`**

Once a user runs `ocr config set max_tokens 200000`, there's no way to revert to the template default — `ocr config set max_tokens 0` is rejected by validation, and `ocr config unset` only supports `provider`, `custom_providers.<name>`, and `mcp_servers.<name>`.

Suggest extending `ocr config unset` to handle scalar keys by zeroing the field and saving — `omitempty` will drop it from the JSON automatically:

```go
case "max_tokens":
    cfg.MaxTokens = 0
    return saveConfig(configPath, cfg)
```

Linxiushen:
Addressed both points in 3a2a0ac.

1. The configurable `max_tokens` value now controls only prompt sizing, pre-flight filtering, and compression thresholds. The template's original value is preserved as a runtime-only completion cap, and every review/scan LLM request (including compression and relocation) uses that separate cap. A regression test sets a 200,000-token prompt ceiling and verifies the request still sends the original 58,888 completion limit.

2. `ocr config unset max_tokens` now clears the saved override while preserving unrelated configuration; because the field is `omitempty`, it disappears from the JSON and the embedded template default resumes.

Verification: focused config/flag tests pass; `internal/config/template`, `internal/llmloop`, `internal/agent`, and `internal/scan` all pass. The full `cmd/opencodereview` package compiles and reaches only two existing Windows-specific test failures (Unix absolute-path interpretation and `0600` permission semantics).

lizhengfeng101 (approved):
LGTM

PR #721, opened 2026-08-04
fix(agent): render tool parameters deterministically

### LINKED ISSUE #719: Plan-phase system prompt renders tool parameters in nondeterministic order, so byte-identical requests defeat provider prefix caching
### OpenCodeReview Version

open-code-review v1.8.8 (6dc3eb067) darwin/arm64, built at 2026-08-04T13:09:52Z

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

Other OpenAI-compatible endpoint (DeepSeek). The rendering defect itself is provider-independent; the cost effect applies to providers with prefix caching (e.g. DeepSeek and OpenAI cache automatically).

### Bug Description

The plan-phase system prompt is not byte-stable: `formatToolDefs` renders each tool's parameter list by iterating a Go map without sorting, so the same tool config produces differently-ordered parameter bullets from one request to the next.

The chain, at 6dc3eb067:

1. `internal/config/template/prompts/plan_task_system.md:7` — the `{{plan_tools}}` placeholder sits near the top of the plan-task system message.
2. `internal/agent/agent.go:1471` (`executePlanPhase`) — the placeholder is resolved by calling `formatToolDefs(a.args.PlanToolDefs)` freshly on every per-file plan request.
3. `internal/agent/agent.go:1517` — parameters come from a JSON-decoded `map[string]any` (`fn.Parameters["properties"]`).
4. `internal/agent/agent.go:1527` — `for name, p := range params` iterates that map directly; Go randomizes map iteration order, so the bullet order varies per call:

```go
for name, p := range params {
	suffix := ""
	if required[name] {
		suffix = " (required)"
	}
	...
}
```

For contrast, the API-level function-calling path is already deterministic — `encoding/json` sorts map keys when marshaling — so this is a small gap in the one hand-rolled text rendering, not a systemic issue.

Impact, kept in proportion:

- No semantic impact. The model sees the same bullets in a different order.
- Byte instability has two practical costs. First, on providers with automatic prefix caching, everything from the first reordered bullet to the end of the system prompt — byte-identical content — drops out of the shared prefix whenever two requests draw different orders (in the repro below that is ~1.6k chars of the 5k-char prompt; per-request cost is small but it recurs on every plan request for every caching provider). Second, prompts are not byte-reproducible across identical runs, which makes prompt-level debugging and cache-behavior analysis noisier than it needs to be. Related: #229 (session affinity only helps if the prompt bytes are stable) and #247 (one mechanical source of run-to-run variation that is free to remove).

### Steps to Reproduce

Plan phase runs for files with >= 50 changed lines (`PLAN_MODE_LINE_THRESHOLD`, `internal/config/template/task_template.json:33`), so use files large enough to trigger it. Any configured provider works.

```bash
git init -q ocr-planprompt-repro && cd ocr-planprompt-repro
git commit -q --allow-empty -m init && git tag base
for name in alpha beta gamma; do
  { echo "pub struct ${name}Config {"; for i in $(seq 1 28); do echo "    pub field_$i: u32,"; done; echo "}";
    echo "impl ${name}Config {"; for i in $(seq 1 28); do echo "    pub fn get_$i(&self) -> u32 { self.field_$i }"; done; echo "}"; } > "$name.rs"
done
git add . && git commit -qm "three 60-line files"
ocr review --from base --to HEAD --audience agent
```

Then compare the plan-task system prompts recorded in the session JSONL:

```bash
python3 - <<'EOF'
import json, glob, os, hashlib
f = max(glob.glob(os.path.expanduser('~/.opencodereview/sessions/*ocr-planprompt-repro*/*.jsonl')), key=os.path.getmtime)
for line in open(f):
    e = json.loads(line)
    if e.get('type') == 'llm_request' and e.get('taskType') == 'plan_task':
        c = e['messages'][0]['content']
        print(e['filePath'], hashlib.md5(c.encode()).hexdigest()[:8], len(c))
EOF
```

Since the ordering is random, a single run usually suffices with 3 plan requests; occasionally all draws coincide — re-run once if so.

### Expected Behavior

All plan-task system prompts within a run (and across runs, given unchanged tool config) are byte-identical: same content, same order.

### Logs / Error Output

Observed in one run of the repro above — 3 plan requests, 2 distinct system prompts, identical length, diverging at char 3453:

```
gamma.rs   md5=fad066c6 len=5048
beta.rs    md5=3fb56faa len=5048
alpha.rs   md5=3fb56faa len=5048
```

Unified diff of the beta.rs vs gamma.rs system prompts — the entire difference is one relocated bullet:

```
   Parameters:
+  - case_sensitive: Whether to perform case-sensitive search. Defaults to false.
   - query_name: Filename keyword to search for, supports partial matching. (required)
-  - case_sensitive: Whether to perform case-sensitive search. Defaults to false.
```

### Additional Context

Suggested fix: collect and sort the parameter names before iterating in `formatToolDefs` (alphabetical, or required-first-then-alphabetical) so the rendered text is a pure function of the tool config. One-line-ish change, no semantic effect on prompts.

Sibling report: #718 covers another instance of the same general family (a byte stream that feeds hashing/prompts not being deterministic), in an unrelated subsystem — the two are independent and neither fix covers the other.

Anon2Tokyo:
## Description

Render tool parameters in a deterministic order when building tool descriptions for the plan-stage system prompt.

`formatToolDefs` previously ranged over the JSON schema `properties` map directly, so the rendered parameter order could vary across runs. Sorting parameter names before rendering keeps the generated prompt byte-stable for the same tool definitions, improving prefix/prompt cache reproducibility without changing prompt semantics.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] `go test ./internal/agent`
- [x] `go test -race ./internal/agent`
- [x] `go vet ./...`
- [x] `make build`
- [x] `gofmt -l internal/agent/agent.go internal/agent/agent_test.go` produced no output
## Checklist

- [x] I have signed the CLA

Focused agent tests pass locally, including race detection. No documentation update is needed because this does not change user-facing behavior. I will follow the CLA bot instructions if required.

## Related Issues

Closes #719

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1524,7 +1524,14 @@ func formatToolDefs(toolDefs []llm.ToolDef) string {
 					}
 				}
 			}
-			for name, p := range params {
+			paramNames := make([]string, 0, len(params))
+			for name := range params {
+				paramNames = append(paramNames, name)
+			}
+			sort.Strings(paramNames)
+
+			for _, name := range paramNames {
+				p := params[name]
 				suffix := ""
 				if required[name] {
 					suffix = " (required)"
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha256"
 	"encoding/binary"
@@ -8,7 +9,9 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"maps"
 	"runtime/debug"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -1514,6 +1517,17 @@ func formatToolDefs(toolDefs []llm.ToolDef) string {
 	for _, td := range toolDefs {
 		fn := &td.Function
 		sb.WriteString(fmt.Sprintf("- **%s**: %s\n", fn.Name, fn.Description))
+		if orderedParams, ok := orderedToolParameters(fn.RawDefinition); ok {
+			sb.WriteString("  Parameters:\n")
+			for _, p := range orderedParams {
+				suffix := ""
+				if p.Required {
+					suffix = " (required)"
+				}
+				sb.WriteString(fmt.Sprintf("  - %s: %s%s\n", p.Name, p.Description, suffix))
+			}
+			continue
+		}
 		if params, ok := fn.Parameters["properties"].(map[string]any); ok && len(params) > 0 {
 			sb.WriteString("  Parameters:\n")
 			required := make(map[string]bool)
@@ -1524,13 +1538,7 @@ func formatToolDefs(toolDefs []llm.ToolDef) string {
 					}
 				}
 			}
-			paramNames := make([]string, 0, len(params))
-			for name := range params {
-				paramNames = append(paramNames, name)
-			}
-			sort.Strings(paramNames)
-
-			for _, name := range paramNames {
+			for _, name := range slices.Sorted(maps.Keys(params)) {
 				p := params[name]
 				suffix := ""
 				if required[name] {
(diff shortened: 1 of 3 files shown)


lizhengfeng101:
Thanks for this thorough contribution — the analysis in the PR description is well-written and the test coverage is solid.

I'd like to suggest two refinements before merging:

**1. Preserve original parameter order from `tools.json` instead of alphabetical sorting**

Tool authors intentionally order parameters by importance (e.g. `query` first, optional flags later). While `sort.Strings` does achieve determinism, it discards that semantic ordering. The root issue is that `FunctionDef.Parameters` is typed as `map[string]any`, which loses insertion order during `json.Unmarshal`.

A possible approach: keep the raw JSON bytes available at render time and walk the `properties` object with `json.Decoder` to extract keys in their original document order. Alternatively, we could introduce an explicit `parameter_order` field in `tools.json`. Either way, the goal is: deterministic output that respects the author's intended order.

**2. Consider using `slices.Sorted` (or `slices.Sort`) instead of `sort.Strings`**

Our `go.mod` specifies Go 1.25, so the newer `slices` package is available. If you do keep alphabetical sorting as a fallback (e.g. for dynamically-registered MCP tools where original order isn't available), prefer `slices.Sorted(maps.Keys(params))` — it's more idiomatic for modern Go.

Let me know if you'd like to discuss the design direction for preserving JSON key order — happy to help scope that out.

lizhengfeng101 (changes requested):
@Anon2Tokyo I'd like to suggest two refinements before merging

Anon2Tokyo:
@lizhengfeng101 Thanks for the suggestions. I updated the implementation to preserve the original `properties` order from `tools.json` by carrying the raw tool definition through `BuildToolDefs` and using `json.Decoder` at prompt-render time to walk the `properties` object in document order.
For tool definitions that do not have raw JSON available, such as dynamically generated MCP tools, the renderer falls back to deterministic alphabetical ordering using `slices.Sorted(maps.Keys(params))`.

lizhengfeng101 (approved):
LGTM

PR #726, opened 2026-08-05
feat(benchmark): add Qwen3.8-Max and show version inline

lizhengfeng101:
## Summary

- Add **Qwen3.8-Max** (v1.8.7) benchmark results — F1 23.00%, precision 33.90% (co-silver with Claude-4.6-Opus), recall 17.40%
- Show tool **version as inline sub-text** under the source name in each row, replacing the static footer note
- Remove the footer note — version info is now per-row, making it easier to distinguish entries evaluated on different OCR versions
- Update benchmark screenshots for all 4 languages (en/zh/ja/ru)

## Test plan

- [ ] Verify benchmark table renders correctly with the new Qwen3.8-Max row
- [ ] Verify version sub-text appears under each source name, aligned with the text (not the icon)
- [ ] Verify precision silver medal trophy displays on both Claude-4.6-Opus and Qwen3.8-Max
- [ ] Verify sorting still works correctly across F1/Precision/Recall columns
- [ ] Verify footer note is removed
- [ ] Check mobile scrolling still works


MuoDoo (approved):
LGTM

PR #731, opened 2026-08-05
ci(pages): run tests in Pages workflow

### LINKED ISSUE #730: ci(pages): add npm test step to pages-ci workflow
## Description

The `pages-ci.yml` workflow currently runs `install → lint → typecheck → build → size` but does not execute `npm test`. This means unit tests — including the critical-CSS drift guard added in #610 and the existing `ErrorBoundary.test.tsx` — are never validated in CI.

The test suite is lightweight (8 assertions, ~1.3s runtime, no external dependencies), so adding it has negligible cost but closes a real gap: a contributor could unknowingly desync the inline critical CSS from `index.css` and CI would not catch it.

## Scope

- File(s): `.github/workflows/pages-ci.yml`
- Area: Pages CI pipeline

## Proposed Change

Add a `Test` step after `Lint` (or before `Build`):

```yaml
      - name: Test
        working-directory: pages
        run: npm test
```

## Acceptance Criteria

- [ ] `npm test` step added to `.github/workflows/pages-ci.yml`
- [ ] Step runs after `Install dependencies` and before or after `Build`
- [ ] CI passes on a PR that touches `pages/`

## Context

Discovered while reviewing #610 (inline critical CSS to fix FOUC). That PR added `pages/src/index.html.test.ts` which guards against color drift between the inline `<style>` block and `src/styles/index.css`. Without CI enforcement, the guard only works if contributors remember to run `npm test` locally.

The existing `ErrorBoundary.test.tsx` has the same gap — this fix covers both.

Related: #725 (smoke test for build output) is a separate concern (HTTP-level verification) and can coexist with this change.

aryansk:
## Summary

Add the missing `npm test` step to the Pages CI workflow.

The workflow already installs dependencies, lints, typechecks, builds, and checks bundle size, but it did not run the existing Vitest suite. This step runs the Pages tests after lint and before typecheck/build so regressions such as critical-CSS drift and ErrorBoundary failures are covered in CI.

Closes #730

## Validation

From `pages/`:

- `npm test` — 4 test files, 16 tests passed
- `npm run lint` — passed with 2 pre-existing hook-dependency warnings
- `npm run typecheck` — passed
- `npm run build` — passed with existing webpack asset-size warnings
- `npm run size` — passed (86.67 kB Brotli against a 150 kB limit)
- `git diff --check` — passed

The commit includes a `Signed-off-by` line; the local SSH signing agent was unavailable in this checkout.

### DIFF
--- a/.github/workflows/pages-ci.yml
+++ b/.github/workflows/pages-ci.yml
@@ -33,6 +33,10 @@ jobs:
         working-directory: pages
         run: npm run lint
 
+      - name: Test
+        working-directory: pages
+        run: npm test
+
       - name: Typecheck
         working-directory: pages
         run: npm run typecheck


lizhengfeng101 (approved):
LGTM

aryansk:
Thanks for the LGTM and for taking the time to review and merge this. I am glad the CI change fit the project and that the validation was useful.

PR #737, opened 2026-08-05
fix(tool/code_search): guard option-like refs in buildGrepArgs

### LINKED ISSUE #645: code_search fails in range/commit reviews because git grep treats --end-of-options as a revision

aalhadxx:
**Related:** closes #645, follow-up to #663

## Summary

As discussed in [#663 (comment)](https://github.com/alibaba/open-code-review/pull/663#issuecomment-...), this PR moves the defense-in-depth check for option-like refs **into `buildGrepArgs`** instead of keeping it in `gitGrep`. This keeps the guard co-located with argument construction (the same pattern the other tools follow) and adds a clear comment explaining why `git grep` is the one invocation where we can't use `--end-of-options`.

## What changed

- `internal/tool/code_search.go`:
  - `buildGrepArgs` now rejects refs starting with `-` and returns `nil`.
  - Added a comment noting that `git grep < 2.45` doesn't support `--end-of-options` before the revision.
  - `gitGrep` propagates the `nil` from `buildGrepArgs` as the existing `"Error: ref must not start with '-'`" message.

- `internal/tool/code_search_test.go`:
  - Added `TestBuildGrepArgs_RejectsOptionLikeRef` to assert the guard lives in `buildGrepArgs`.
  - Kept the existing `TestGitGrep_RejectsOptionLikeRef` and `TestGitGrep_OptionLikeRefDoesNotLaunchPager` for end-to-end coverage.

## Why this structure

The codebase already validates refs upstream in `validateReviewRefs`, but the `buildGrepArgs` guard ensures the defense-in-depth is visible right where arguments are built. If anyone ever calls `buildGrepArgs` from a new path, the guard still holds.

### DIFF
--- a/internal/tool/code_search.go
+++ b/internal/tool/code_search.go
@@ -78,7 +78,14 @@ func (p *CodeSearchProvider) buildGrepArgs(searchText string, caseSensitive bool
 	cmdArgs = append(cmdArgs, "-e", searchText)
 
 	if ref := p.FileReader.Ref; ref != "" {
-		cmdArgs = append(cmdArgs, "--end-of-options")
+		if strings.HasPrefix(ref, "-") {
+			// Defense-in-depth: reject option-like refs here even though
+			// validateReviewRefs already verifies the ref upstream.
+			// NOTE: git grep < 2.45 does not support --end-of-options before
+			// the revision, so this is the one git invocation where we can't
+			// rely on that separator.
+			return nil
+		}
 		cmdArgs = append(cmdArgs, ref)
 	}
 
@@ -125,6 +132,9 @@ func (p *CodeSearchProvider) runGitGrep(parentCtx context.Context, cmdArgs []str
 
 func (p *CodeSearchProvider) gitGrep(ctx context.Context, searchText string, caseSensitive bool, usePerlRegexp bool, pathspec []string) (string, error) {
 	cmdArgs := p.buildGrepArgs(searchText, caseSensitive, usePerlRegexp, false, pathspec)
+	if cmdArgs == nil {
+		return "Error: ref must not start with '-'", nil
+	}
 
 	outStr, errStr, err := p.runGitGrep(ctx, cmdArgs)
 
--- a/internal/tool/code_search_test.go
+++ b/internal/tool/code_search_test.go
@@ -33,15 +33,28 @@ func TestBuildGrepArgs_CommitMode(t *testing.T) {
 	p := NewCodeSearch(&FileReader{RepoDir: "/tmp", Ref: "abc1234"})
 	args := p.buildGrepArgs("myFunc", false, false, false, []string{"pkg/"})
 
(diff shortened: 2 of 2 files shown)


lizhengfeng101:
Hi @aalhadxx — this PR is identical to #727 (same branch, same diff), which was already merged earlier today. Closing this as a duplicate. Thanks for the contribution!

PR #741, opened 2026-08-05
ci(pages): add smoke test step to verify build output is servable

### LINKED ISSUE #725: ci(pages): add smoke test step to verify build output is servable
## Description

The `pages-ci.yml` workflow currently runs lint, typecheck, build, and bundle size checks — but never verifies that the build output can actually be served and accessed. Since `pages/` is a React SPA (React Router v6 + webpack), a broken HTML template or missing bundle reference would slip through undetected.

We should add a lightweight smoke test step that starts a static file server against `dist/` and verifies key routes return HTTP 200 with valid bundle references.

## Scope

- File(s): `.github/workflows/pages-ci.yml`, `pages/package.json`
- Area: Pages CI pipeline

## Proposed Approach

1. Add `serve` to `pages/package.json` devDependencies (avoids `npx` download on every run).
2. After the existing `Build` step in `pages-ci.yml`, add a smoke test step:

```yaml
      - name: Smoke test
        working-directory: pages
        run: |
          npx serve dist -s -l tcp://127.0.0.1:3999 &
          SERVER_PID=$!
          sleep 2

          FAILED=0
          for path in "/" "/docs/contributing" "/docs/quickstart" "/features"; do
            BODY=$(curl -sf "http://127.0.0.1:3999${path}")
            if [ $? -ne 0 ]; then
              echo "FAIL: ${path} — not reachable"
              FAILED=1
            elif ! echo "$BODY" | grep -q '\.bundle\.js'; then
              echo "FAIL: ${path} — missing bundle reference"
              FAILED=1
            else
              echo "OK: ${path}"
            fi
          done

          kill $SERVER_PID
          exit $FAILED
```

Key details:
- Uses `serve -s` (SPA mode) to fallback all routes to `index.html`, matching real deployment behavior.
- Checks that responses contain `.bundle.js` references to confirm the build artifact is complete.
- Runs inside the existing `node:24.18.0` container — no additional dependencies or browser binaries needed.
- Estimated added time: ~3 seconds (server start + curl checks).

## Acceptance Criteria

- [ ] `serve` added to `pages/package.json` devDependencies
- [ ] Smoke test step added to `pages-ci.yml` after the Build step
- [ ] At least `/`, `/docs/contributing`, `/docs/quickstart`, `/features` are checked
- [ ] Step correctly fails if a route is unreachable or missing bundle references
- [ ] CI passes on a PR that touches `pages/`

## Context

Discovered while adding documentation changes to `pages/src/content/docs/` — realized there's no validation that the built site is actually servable. This is a lightweight alternative to full browser-based testing (Playwright) that avoids ~400MB browser binary downloads on the self-hosted runner.

HuijungYoon:
## Summary
- Add `serve` as a pages `devDependency` so CI can statically serve `dist/` without downloading on every run.
- After the Build step in `pages-ci.yml`, smoke-test `/`, `/docs/contributing`, `/docs/quickstart`, and `/features` for HTTP 200 responses that include `.bundle.js` references (SPA mode via `serve -s`).
- Closes #725

## Test plan
- [x] Pages CI runs on this PR (`pages/**` path filter)
- [x] Smoke test step passes (all four routes OK)
- [x] Bundle size check still runs after smoke test
- [x] Locally (Node 24.18.0): `cd pages && npm install && npm run build`, then run the smoke script from the workflow

### DIFF
--- a/.github/workflows/pages-ci.yml
+++ b/.github/workflows/pages-ci.yml
@@ -54,9 +54,10 @@ jobs:
           sleep 2
 
           FAILED=0
+          # `serve -s` rewrites SPA routes to index.html, so this verifies
+          # the built shell is reachable and includes bundle references.
           for path in "/" "/docs/contributing" "/docs/quickstart" "/features"; do
-            BODY=$(curl -sf "http://127.0.0.1:3999${path}")
-            if [ $? -ne 0 ]; then
+            if ! BODY=$(curl -sf "http://127.0.0.1:3999${path}"); then
               echo "FAIL: ${path} — not reachable"
               FAILED=1
             elif ! echo "$BODY" | grep -q '\.bundle\.js'; then
@@ -66,6 +67,14 @@ jobs:
               echo "OK: ${path}"
             fi
           done
+
+          # Also check a real static file to ensure dist assets are served.
+          if ! curl -sf "http://127.0.0.1:3999/404.html" >/dev/null; then
+            echo "FAIL: /404.html — static file not reachable"
+            FAILED=1
+          else
+            echo "OK: /404.html"
+          fi
 
           exit $FAILED
 
--- a/.github/workflows/pages-ci.yml
+++ b/.github/workflows/pages-ci.yml
@@ -45,6 +45,30 @@ jobs:
         working-directory: pages
         run: npm run build
 
+      - name: Smoke test
+        working-directory: pages
+        run: |
+          npx serve dist -s -l tcp://127.0.0.1:3999 &
+          SERVER_PID=$!
+          trap 'kill $SERVER_PID 2>/dev/null || true' EXIT
+          sleep 2
+
+          FAILED=0
+          for path in "/" "/docs/contributing" "/docs/quickstart" "/features"; do
+            BODY=$(curl -sf "http://127.0.0.1:3999${path}")
+            if [ $? -ne 0 ]; then
+              echo "FAIL: ${path} — not reachable"
+              FAILED=1
+            elif ! echo "$BODY" | grep -q '\.bundle\.js'; then
+              echo "FAIL: ${path} — missing bundle reference"
+              FAILED=1
+            else
+              echo "OK: ${path}"
+            fi
+          done
+
+          exit $FAILED
+
       - name: Check bundle size
         working-directory: pages
         run: npm run size
--- a/pages/package.json
+++ b/pages/package.json
@@ -55,6 +55,7 @@
     "jsdom": "^30.0.0",
     "postcss": "^8.5.15",
     "postcss-loader": "^7.3.3",
+    "serve": "^14.2.6",
     "size-limit": "^13.0.1",
     "style-loader": "^3.3.3",
     "tailwindcss": "^3.3.5",


lizhengfeng101:
@HuijungYoon Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/alibaba/open-code-review?pullRequest=741) before we can accept your contribution.

HuijungYoon:
@lizhengfeng101  I've signed the CLA! I'm really looking forward to contributing more to this project. Thank you!

lizhengfeng101:
> @lizhengfeng101 I've signed the CLA! I'm really looking forward to contributing more to this project. Thank you!

ok

lizhengfeng101:
Hey @HuijungYoon, nice addition — the smoke step is a sensible lightweight alternative to spinning up Playwright. Two things I'd love your take on before this goes in:

**1. All four routes resolve to the same `index.html` under `serve -s`.** Since this is a React Router SPA, `serve -s` rewrites every unmatched path to `index.html`, and `dist/` only contains `index.html` / `404.html` / the bundles. So `/`, `/docs/contributing`, `/docs/quickstart`, and `/features` all return the *identical* HTML, and the `.bundle.js` assertion is effectively the same check run four times. It'll catch a missing `index.html` or a broken bundle reference, but it won't catch a broken route component (that only surfaces once React Router renders client-side). Not a blocker, but the four-route list reads like it's testing more than it actually is — might be worth either a comment noting the limitation, or adding one assertion against a real static file (e.g. `/404.html`) so we're confirming the server is genuinely reading `dist/`.

**2. Minor: the `if [ $? -ne 0 ]` branch is effectively dead under the Actions default shell.** GitHub runs `run:` steps with `bash --noprofile --norc -eo pipefail`, so `set -e` is active. On a `BODY=$(curl -sf ...)` failure the step aborts *before* reaching the `$?` check, so the friendly "not reachable" message never prints. In practice this rarely bites (given point 1, routes won't individually 404 — only a fully-dead server would fail curl, and that still fails CI on the first `/` request), so I'd call it a cleanup rather than a must-fix. If you want the branch to actually work, moving curl into the tested context does it:

```bash
if ! BODY=$(curl -sf "http://127.0.0.1:3999${path}"); then
  echo "FAIL: ${path} — not reachable"; FAILED=1
elif ! echo "$BODY" | grep -q '\.bundle\.js'; then
  ...
```

Also +1 on the `trap ... EXIT` for cleanup — nice touch over the original issue snippet.

HuijungYoon:
@lizhengfeng101  

Thanks for the thoughtful review 

I pushed a follow-up commit that addresses both items:

1. Added a comment clarifying the SPA limitation with serve -s (the route checks validate shell reachability and bundle references, not client-side route rendering correctness).
2. Updated the curl branch to if ! BODY=$(curl -sf ...) so the “not reachable” path works correctly under GitHub Actions’ bash -e behavior.
3. Added a static asset check for /404.html to confirm we’re actually serving files from dist/, not only exercising SPA fallback.

Really appreciate the careful read and the concrete suggestions.

lizhengfeng101 (approved):
LGTM

PR #747, opened 2026-08-06
test: raise statement coverage to 90% and enforce it in CI

lizhengfeng101:
## Summary

Brings total statement coverage above 90% with test-only changes, and adds a CI gate so it cannot regress.

## What changed

- **New unit tests** across `cmd/opencodereview` and the `internal/*` packages, covering:
  - CLI helpers, flag suggestion, and manifest/warning output formatting
  - Provider TUI handlers (manual entry, custom-provider form, edit/save, delete-confirm, rollback, save-failure reload)
  - Resume/manifest and error branches in `config`, `llm`, `llmloop`, `scan`, `session`, `agent`, `viewer`, `mcp`, `pathutil`, and `telemetry`
- **Coverage gate raised 80% → 90%**:
  - `Makefile` `COVERAGE_THRESHOLD := 90`
  - The "Check coverage threshold" step in `.github/workflows/ci.yml`
- **`.gitignore`**: ignore generated `coverage.out` / `coverage.html`

## Why

The project targets a higher statement-coverage bar. This raises real coverage of previously untested helper, error, and TUI paths, then locks the threshold in CI so future changes must keep coverage at or above 90%.

## Coverage

The authoritative figure is the CI gate, which runs `go test -race ./...` over the whole module and currently reports **90.2%**:

```
go test -count=1 -race -coverprofile=coverage.out ./...
go tool cover -func=coverage.out | grep total:
```

The `make coverage` target reports a slightly higher number (90.6%) because it measures a smaller package set — it omits `-race` and excludes `extensions/` via `PACKAGES`. The two cover different package sets and are not the same measurement; the CI figure governs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,11 +64,11 @@ jobs:
         run: |
           COVERAGE=$(go tool cover -func=coverage.out | grep total: | awk '{print $3}' | sed 's/%//')
           echo "Total coverage: ${COVERAGE}%"
-          if awk "BEGIN {exit !($COVERAGE < 80)}"; then
-            echo "FAIL: Coverage ${COVERAGE}% is below 80% threshold"
+          if awk "BEGIN {exit !($COVERAGE < 90)}"; then
+            echo "FAIL: Coverage ${COVERAGE}% is below 90% threshold"
             exit 1
           fi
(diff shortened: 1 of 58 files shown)


MuoDoo (approved):
LGTM

PR #762, opened 2026-08-07
feat: add QCA delegation integration

hellomypastor:
## Summary

- add a stable, agent-oriented JSON contract (`schema_version: "1"`) for `ocr delegate preview` and `ocr delegate rule`
- strengthen the delegation Skill with explicit file accounting, bounded batching, and `(path, status)` coverage semantics
- add a ready-to-publish QCA Forward template, system prompt, and integration guide
- document QCA as a supported host-agent integration in the project and plugin READMEs

## Why

QCA Forward uses the host agent's LLM. Open Code Review should only provide deterministic file selection and rule resolution in this mode, so users do not need to configure a second OCR LLM endpoint or API key.

### Local

| Check | Result |
| --- | --- |
| `make build` | Passed |
| `go test ./...` | Passed |
| `npm run test:github-actions` | Passed |
| translation sync | Passed |
| `git diff --check` | Passed |
| workspace/range/commit preview in JSON | Passed |
| rule resolution in JSON | Passed |
| existing text output compatibility | Passed |
| invalid and empty `--format` rejection | Passed |

### QCA Forward cloud run

- Session: `sess_00lkc912yy6m8zpyw6b6`
- Template: `tmpl_95e627cbad9552742039accd`
- Skill: `skill_00lkbheh0fyf51na90d7` version 1
- Environment: `env_00jg9y14lbk74b7g7tvw`
- Delegate schema: `1`
- Preview: 12 total files, 3 reviewable, 9 excluded
- Review accounting: 3 reviewed, 0 skipped, 100% coverage
- Confirmed the session used only `ocr delegate preview/rule`; it did not run `ocr review`
- Confirmed no `OCR_LLM_*` or Anthropic credentials were present
- Cloud-side `go test ./cmd/...` passed

[Rendered QCA validation result](https://gist.github.com/hellomypastor/5fe11c5fe4fc95ad82e90bd6a947669c)

## Issues found during validation and addressed

- corrected the QCA custom Skill binding type
- documented that Write/Edit are disabled while Bash read-only behavior remains prompt/runtime-policy enforced
- rejected an explicitly empty output format
- documented replacement of the pinned OCR version placeholder
- made duplicate workspace paths unambiguous by using `(path, status)` as the coverage identity

## Compatibility and security

- text remains the default output format
- JSON output is opt-in through `--format json`
- delegation mode never invokes an OCR-side LLM
- ref validation and existing file-selection/exclusion behavior are reused unchanged

### DIFF
--- a/README.ja-JP.md
+++ b/README.ja-JP.md
@@ -171,6 +171,7 @@ ocr delegate rule src/main.go src/handler.go
   - [Codex](plugins/open-code-review/README.md#codex) — 呼び出し可能なレビュースキルを含むプラグインをインストール
   - [Cursor](plugins/open-code-review/README.md#cursor) — 移植可能なレビュースキルを含むプラグインをインストール
   - [OpenCode](plugins/open-code-review/opencode/README.md) — ネイティブレビュー・ツールとスラッシュコマンドをインストール
+  - [QCA Forward](plugins/open-code-review/qca/README.md) — QCA ホストモデルと公開可能なテンプレートで委任モードを実行
   - [Skill 対応エージェント](https://open-codereview.ai/docs/agent-skill) — 移植可能なエージェントスキルをインストール
 - レビュー実行モード — 連携後、どの LLM がレビューを実行するかを選択
   - [デフォルト（OCR が管理）](https://open-codereview.ai/docs/configuration) — OCR が設定済みの LLM を使用してレビューを実行
--- a/README.ko-KR.md
+++ b/README.ko-KR.md
@@ -171,6 +171,7 @@ ocr delegate rule src/main.go src/handler.go
   - [Codex](plugins/open-code-review/README.md#codex) — 호출 가능한 리뷰 스킬이 포함된 플러그인 설치
   - [Cursor](plugins/open-code-review/README.md#cursor) — 이식 가능한 리뷰 스킬이 포함된 플러그인 설치
   - [OpenCode](plugins/open-code-review/opencode/README.md) — 네이티브 리뷰 도구와 슬래시 명령 설치
+  - [QCA Forward](plugins/open-code-review/qca/README.md) — QCA 호스트 모델과 게시 가능한 템플릿으로 위임 모드 실행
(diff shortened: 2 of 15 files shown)


lizhengfeng101:
Nice work on the delegation JSON contract — the `schema_version` envelope and the `make(..., 0)` guard on the preview arrays are exactly the right call. Two small things I'd love to see tightened before this lands, neither a blocker:

**1. `ruleGroupsJSON` can emit `"files": null`**

In `cmd/opencodereview/delegate_cmd.go`, the rule-group builder copies files like this:

```go
files := append([]string(nil), group.Files...)
```

When `group.Files` is empty this produces a `nil` slice, which marshals to `"files": null` instead of `[]`. That's inconsistent with the "arrays are never null" contract this PR just established for `preview` (where `previewFiles` uses `make([]..., 0)`), and it forces agent consumers to special-case null. In practice a rule group almost always has at least one file, so the impact is low — but since it's brand-new code here, I'd make it match:

```go
files := make([]string, 0, len(group.Files))
files = append(files, group.Files...)
```

A test with an empty-`files` group would nail the contract down, though it's optional.

**2. README translations are out of sync**

The QCA Forward entry was added to `README.md` and `README.zh-CN.md`, but the other three localized READMEs still stop at OpenCode and are missing the new row:

- `README.ja-JP.md`
- `README.ko-KR.md`
- `README.ru-RU.md`

Each carries the same integrations list (Cursor / OpenCode, around line 172–173), so it'd be good to add the QCA Forward line there too and keep the docs from drifting across locales.

Thanks — the CLI path is solid and the default (`text`) output is unchanged, so this is really just polish. 🙏

hellomypastor:
> Thanks for the thoughtful review! Both points have been addressed in `1e1b9d5`:
>
> 1. `ruleGroupsJSON` now initializes `files` as a non-nil empty slice, ensuring the JSON contract consistently emits `"files": []`. I also added a regression test covering an empty rule group.
> 2. Added the QCA Forward integration entry to the Japanese, Korean, and Russian READMEs.
>
> Verified with:
>
> - `go test ./...`
> - `npm run test:github-actions`
> - `git diff --check`
>
> Thanks again for catching these!

lizhengfeng101 (approved):
LGTM

PR #773, opened 2026-08-07
feat(llmloop): backfill comment thinking from turn output

yingjiexu2002:
## Description

Comments parsed from `code_comment` tool calls have a runtime-only `thinking` field that most models never populate (it is deliberately not advertised in the tool schema). This PR backfills it from the model's output of the current turn, so `--format json` output carries the review reasoning:

- Add `ChatResponse.ReasoningContent()` to expose the native reasoning content of a response (OpenAI `reasoning_content` in both streaming and non-streaming paths; Anthropic thinking blocks).
- In the agent loop, compute the turn's thinking text as reasoning content, falling back to the assistant message for models that do not expose reasoning, and backfill it into each collected comment whose `thinking` is empty. Explicit model-provided thinking always wins.
- Terminal output is unchanged (`renderComment` does not render thinking); only the JSON output gains the field.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Added unit tests: backfill uses the turn's thinking text when a comment has no explicit thinking; explicit thinking is preserved; no backfill when the turn has no thinking text. Also verified end-to-end with a local review run (`--format json`) against providers with and without native reasoning output.

## Checklist

- [x] I have signed the CLA

## Related Issues

N/A


lizhengfeng101:
**On the `content` fallback for `thinking`**

The backfill prefers `reasoning_content` and falls back to the turn's assistant message (`content`) when there's no native reasoning. I'd suggest verifying this fallback is actually worth keeping before we ship it.

The concern: for a tool-calling turn, the assistant message is often not real reasoning — it tends to be a short, user-facing preamble like *"I'll now leave a comment on this file"* or *"Here's my feedback"*. Backfilling that into `thinking` (and copying it across every comment in the turn) adds noise to the JSON output rather than the review reasoning we actually want, and it's the part most likely to read as misleading.

Could you run a few review passes against a couple of models **without** native `reasoning_content` and check what `content` actually contains in those turns? If it's mostly these low-value summaries, the fallback buys us little and I'd lean toward only backfilling from `reasoning_content` and leaving `thinking` empty otherwise. If you find that some models do put substantive reasoning in `content`, then keeping the fallback makes sense — ideally noted with an example in the PR description.

Either way, capturing what you observe would make the fallback's necessity clear.

internal/llmloop/loop.go
```diff
@@ -247,8 +247,14 @@ func (r *Runner) RunPerFile(ctx context.Context, messages []llm.Message, newPath
 		taskCompleted := false
 		hasValidResult := false
 
+		// Prefer the model's native reasoning content; fall back to the
+		// assistant message of this turn for models that do not expose it.
+		thinking := resp.ReasoningContent()
+		if thinking == "" {
+			thinking = content
+		}
```
wu21-web:
```suggestion
```

wu21-web: Capture the model's native reasoning content for this turn.

internal/llmloop/loop_execute_more_test.go
wu21-web, line 10:
```suggestion

 "github.com/alibaba/open-code-review/internal/config/template"
```

wu21-web: Tests are needed to lock in the new behavior and regression tests that would fail if the fallback returns.
Tip on L11: this is a package you need to include for new tests, e.g TestRunPerFile_BackfillsThinkingFromReasoningContent
is a test to verify the full wiring

pages/src/content/docs/en/tools.md
wu21-web, line 1: Russian site documentation is missing, use a translator to translate this to `pages/scr/content/docs/ru/tools.md`

internal/llmloop/loop.go
```diff
@@ -247,8 +247,14 @@ func (r *Runner) RunPerFile(ctx context.Context, messages []llm.Message, newPath
 		taskCompleted := false
 		hasValidResult := false
 
+		// Prefer the model's native reasoning content; fall back to the
+		// assistant message of this turn for models that do not expose it.
```
wu21-web:
```suggestion
		// Capture the model's native reasoning content for this turn. Models
		// without a reasoning channel leave it empty.
```

pages/src/content/docs/en/tools.md
```diff
@@ -89,8 +89,11 @@ is optional but encouraged. `path` is a top-level optional override —
 if omitted, the agent injects the file currently under review. The
 agent also injects `path` automatically when the model leaves it out, so
 the model rarely needs to set it explicitly. `thinking` (per-comment)
-captures the model's reasoning and is preserved on the comment but not
-shown in the final review output.
+captures the model's reasoning and is preserved on the comment; OCR
+backfills it from the reasoning content the model emits on the current
+turn (or the turn's message when there is none), and includes it in
+the JSON output (the terminal output does not
+render it).
```
wu21-web, line 96: If you applied the suggestion here: https://github.com/alibaba/open-code-review/pull/773/changes#r3740773775
This means you have to likely update the documentation here.

pages/src/content/docs/ja/tools.md
```diff
@@ -79,8 +79,8 @@ OCR が行番号を自動計算できるようにします。
 `comments` は配列なので、モデルは 1 回のツール呼び出しで複数のコメントを発行できます。`content` と
 `existing_code` は必須です。`suggestion_code` は任意ですが、提供が推奨されます。`path` はトップレベルの任意の上書きで——
 省略すると、agent は現在レビュー中のファイルを注入します。モデルが省略しても agent が自動的に `path` を注入するため、
-モデルが明示的に設定する必要はほとんどありません。`thinking`（コメントごと）はモデルの推論を捕捉し、コメントに保持されますが、
-最終的なレビュー出力には表示されません。
+モデルが明示的に設定する必要はほとんどありません。`thinking`（コメントごと）はモデルの推論を捕捉し、コメントに保持されます。
+OCR はモデルが現在のターンで出力した推論内容（推論内容がない場合はそのターンの本文）で自動的に補完し、JSON 出力に含めます（ターミナル出力には表示されません）。
```
wu21-web: ditto

pages/src/content/docs/zh/tools.md
```diff
@@ -78,8 +78,8 @@ OCR 自动计算行号。
 `comments` 是数组，因此模型可以在一次工具调用中发出多条评论。`content` 和
 `existing_code` 必需；`suggestion_code` 可选但建议提供。`path` 是顶层可选覆盖——
 省略时，agent 会注入当前评审的文件。即便模型省略，agent 也会自动注入 `path`，因此
-模型极少需要显式设置。`thinking`（按评论）捕获模型推理，保留在评论上，但不会
-在最终评审输出中显示。
+模型极少需要显式设置。`thinking`（按评论）捕获模型推理，保留在评论上；OCR 会用
+模型当轮输出的推理内容（无推理内容时用当轮正文）自动回填，并包含在 JSON 输出中（终端输出不渲染）。
```
wu21-web: ditto

wu21-web (changes requested):
You idea is great and we appreciate it. Here are several suggestions.
By the way,  im not sure if this is true, but it is worth considering: JSON output marshals the whole LlmComment and terminal renderComment never renders it.

yingjiexu2002:
> **On the `content` fallback for `thinking`**
> 
> The backfill prefers `reasoning_content` and falls back to the turn's assistant message (`content`) when there's no native reasoning. I'd suggest verifying this fallback is actually worth keeping before we ship it.
> 
> The concern: for a tool-calling turn, the assistant message is often not real reasoning — it tends to be a short, user-facing preamble like _"I'll now leave a comment on this file"_ or _"Here's my feedback"_. Backfilling that into `thinking` (and copying it across every comment in the turn) adds noise to the JSON output rather than the review reasoning we actually want, and it's the part most likely to read as misleading.
> 
> Could you run a few review passes against a couple of models **without** native `reasoning_content` and check what `content` actually contains in those turns? If it's mostly these low-value summaries, the fallback buys us little and I'd lean toward only backfilling from `reasoning_content` and leaving `thinking` empty otherwise. If you find that some models do put substantive reasoning in `content`, then keeping the fallback makes sense — ideally noted with an example in the PR description.
> 
> Either way, capturing what you observe would make the fallback's necessity clear.

Removed. In an opus 4.6 run without thinking, the fallback fired on only 8/23 comments and just 3 of those were valuable — the rest were preambles like "I'll now leave a comment on this file". thinking is now backfilled only from reasoning_content and left empty otherwise; docs and tests updated.

internal/llmloop/loop.go
yingjiexu2002: Done

internal/llmloop/loop_execute_more_test.go
yingjiexu2002: Added in 2b18430: TestRunPerFile_BackfillsThinkingFromReasoningContent drives the full RunPerFile loop via a scripted fake client and asserts the comment gets the turn's reasoning_content; TestRunPerFile_NoFallbackToContent asserts thinking stays empty when only content exists — verified it fails if the fallback is re-added.

pages/src/content/docs/en/tools.md
yingjiexu2002: Synced it.

lizhengfeng101 (commented):
@yingjiexu2002  When multiple tool calls share the same thinking behavior in the same turn, it is recommended to explicitly state in the code comments that this is an intentional design to avoid future maintainers mistakenly thinking it is a bug.

yingjiexu2002:
> @yingjiexu2002 When multiple tool calls share the same thinking behavior in the same turn, it is recommended to explicitly state in the code comments that this is an intentional design to avoid future maintainers mistakenly thinking it is a bug.

Done, documented at both sites in loop.go

wu21-web (approved):

lizhengfeng101 (approved):
LGTM

PR #772, opened 2026-08-07
feat(providers): add siliconflow (GLOBAL) to built-in providers

wu21-web:
## Description
SiliconFlow is rapidly gaining traction as a major AI infrastructure and Model-as-a-Service (MaaS) provider, boasting over 10 million users. This PR adds the SiliconFlow (international) to built-in providers.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## How Has This Been Tested?

- [x] `make test` passes locally
- [x] Manual testing (describe below)
`go test` and `go build`

## Checklist

- [x] My code follows the project's coding style (`go fmt`, `go vet`)
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix is effective or my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have updated the documentation accordingly (if applicable)
- [x] I have signed the CLA

## Related Issues

none

PR #775, opened 2026-08-07
feat(providers): add siliconflow-cn as a built-in model provider

wu21-web:
## Description
This pull request is a follow-up of #772 , adds a siliconflow-cn for chinese users at the sametime.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [x] I have signed the CLA

## Related Issues

none yet


lizhengfeng101:
Same feedback as on #772 — the `Models` list here is identical, so the same concern applies to this preset.

Several entries look like previous-generation models rather than what a new user would want as a first choice:

- `deepseek-ai/DeepSeek-V3` and `deepseek-ai/DeepSeek-R1` are superseded by the V3.2/V4 line already in the same list.
- `Qwen/Qwen3-32B` and `Qwen/Qwen3-14B` are both older and quite small for this workload — review runs depend heavily on reliable tool calling and long-context handling, and small models tend to degrade there.
- `baidu/ERNIE-4.5-300B-A47B` also looks like an older generation.

These presets are exactly what people pick from the provider TUI without thinking twice, so the defaults carry a lot of weight. Could you re-verify the list and confirm:

1. Every model id still resolves on the **CN** endpoint (`https://api.siliconflow.cn/v1`) — the region split with #772 means the two lists should not be assumed identical unless you have checked both.
2. Each listed model supports the streaming + tool-calling behaviour the review loop needs.
3. The list reflects the current generation, and older/smaller entries are either dropped or intentionally kept (a one-line rationale in the PR description would help if intentional).

One process note: this PR and #772 both edit the same `expected` slice in `TestListProviders_Order`, so whichever lands first will leave the other with a conflict. It may be simpler to fold both presets into a single PR, or to confirm the model list once and apply the same result to both.

A shorter list of models you have actually run a review with is more valuable than a longer list that is partly stale. Thanks!

wu21-web:
I will notify you when a merged conflict is resolved.

wu21-web:
Model changes have been synced to this pull request.

internal/llm/providers.go
```diff
@@ -338,6 +338,21 @@ var registry = []Provider{
...
+	{
+		Name:        "siliconflow-cn",
+		DisplayName: "SiliconFlow CN API",
+		Protocol:    ProtocolOpenAIChatCompletions,
+		BaseURL:     "https://api.siliconflow.cn/v1",
+		EnvVar:      "SILICONFLOW_API_KEY",
+		Models: []string{
+			"deepseek-ai/DeepSeek-V4-Pro",
+			"deepseek-ai/DeepSeek-V4-Flash",
+			"meituan-longcat/LongCat-2.0",
```
stay-foolish-forever: same as #772

stay-foolish-forever (approved):

internal/llm/providers.go
wu21-web: Just did that

wu21-web:
@lizhengfeng101 The merge conflicts have been resolved.

lizhengfeng101 (approved):
LGTM

PR #784, opened 2026-08-08
fix(cli): stop preview from creating a review session

zmwangx:
## Description

`runPreview`, `runScanPreview`, and delegate's `preview` each built a throwaway agent via `agent.New` / `scan.NewAgent` purely to reach `Preview`. Both constructors auto-create a session, and `session.New` opens persistence and buffers a `session_start` record, so every preview created a JSONL file under the OCR home. Preview never runs or finalizes a review, so that file was left unfinalized and usually empty.

Make the exported entry point a package-level `Preview(ctx, args)` in both `internal/agent` and `internal/scan`. It builds only what file selection needs, so there is no session, manifest, or runner to leak. The existing bodies stay as unexported methods, keeping the in-package tests (including scan's regression test that `Preview` must not mutate `a.items`).

Deleting the file afterwards was rejected: it would still leak on crash and would keep the wrong abstraction.

Tests assert at the CLI boundary, with a temporary OCR home, that no session store is created by any of the three preview commands. They also neutralize global git config, which git resolves via `XDG_CONFIG_HOME` independently of `HOME`.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Run against `6aa3f50`, a commit already on `main`, using a throwaway `HOME` so the real session store is untouched.

Before this change, a preview leaves a phantom session behind:

```
$ ocr review --commit 6aa3f50 --preview
$ ocr session list
SESSION ID                            MODE  RANGE  FILES  COMMENTS  STATUS   STARTED
1b4ff2d7-22e5-4e4e-856b-ad936a5b7c26  -     -      0      0         aborted  -
```

That row is backed by a 0-byte `.jsonl` file: the `session_start` record was still buffered when the process exited, so the session has no mode, range, or start time and is reported as `aborted`.

With this change, the same invocation creates nothing:

```
$ ocr review --commit 6aa3f50 --preview
$ ocr session list
No sessions found for /path/to/open-code-review
```

`ocr scan --preview` and `ocr delegate preview` behave the same way.

## Checklist

- [x] I have signed the CLA

## Related Issues

None.

---

Disclosure: PR authored by AI with human guidance and careful review.

### DIFF
--- a/cmd/opencodereview/delegate_cmd.go
+++ b/cmd/opencodereview/delegate_cmd.go
@@ -127,7 +127,7 @@ func loadDelegateContext(opts delegateOptions) (*delegateContext, error) {
 
 // preview runs the agent's file-selection logic and returns the preview result.
 func (dc *delegateContext) preview(ctx context.Context) (*agent.DiffPreview, error) {
-	ag := agent.New(agent.Args{
+	return agent.Preview(ctx, agent.Args{
 		RepoDir:    dc.cc.RepoDir,
 		From:       dc.opts.from,
 		To:         dc.opts.to,
@@ -135,7 +135,6 @@ func (dc *delegateContext) preview(ctx context.Context) (*agent.DiffPreview, err
 		FileFilter: dc.cc.FileFilter,
 		GitRunner:  dc.cc.GitRunner,
 	})
-	return ag.Preview(ctx)
 }
 
 // mergeBase computes the merge-base for range mode. Returns "" for other modes.
--- a/cmd/opencodereview/delegate_exec_test.go
+++ b/cmd/opencodereview/delegate_exec_test.go
@@ -68,6 +68,30 @@ func silenceStdout(t *testing.T, fn func()) {
 	fn()
 }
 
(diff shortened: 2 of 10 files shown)


wu21-web (approved):

lizhengfeng101 (dismissed):
LGTM

lizhengfeng101:
Really nice work here. The root-cause analysis is spot on — cutting the problem at the abstraction boundary (package-level `Preview` that never touches session machinery) rather than papering over it with post-hoc cleanup is the right call. I especially appreciate the attention to detail on the test side: isolating `HOME` *and* neutralizing `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` is the kind of thing that prevents mysterious CI flakes months down the road.

Clean fix, solid tests, clear write-up. Merging — thanks!

lizhengfeng101:
@zmwangx Merge conflicts exists

zmwangx:
Thanks, rebased.

lizhengfeng101 (approved):

Issue #787, opened 2026-08-08
Add a provider-wide per-attempt concurrency gate


Gongyl01:
### Problem Statement

Part of #368

OCR uses the official LLM SDK retry loops with `WithMaxRetries(5)`. Those SDKs already own request-local retry eligibility, retry count, backoff, server retry hints, sequential attempts, and cancellation.

OCR has concurrency limits today, but they do not form one provider-wide limit. The current review topology is:

- A file worker holds one `--concurrency` slot while it runs the optional plan request, the main loop (including any grace round), waits for that file's comment jobs, and then runs the review filter. Those foreground requests are serial within the file slot.
- The main loop may start one asynchronous memory-compression job per file conversation. That request can overlap the file's foreground main request.
- Comment post-processing runs in a separate worker pool derived from `--concurrency`. A worker may call the model for re-location after deterministic in-file resolution and cross-file relocation both fail.

Plan and review-filter requests are therefore not independent overflow sources. The aggregate excess comes from overlapping foreground file work, asynchronous compression, and the separate comment pool. Lowering `--concurrency` reduces pressure on more than one path, but it does not establish a single provider-wide maximum across them.

The missing capability is provider-wide admission, not another retry loop.

### Proposed Solution

Keep the official SDK as the sole retry owner and add one provider-scoped, context-aware concurrency gate shared by all `ocr review` LLM request paths in a run.

#### Scope review requests explicitly

- Mark the review run with a dedicated admission-scope context value and preserve it through child contexts, including `context.WithoutCancel` paths.
- Do not use retry-report `RequestMeta` as the admission switch. Retry identity and admission scope are separate concerns, and the existing grace-round request has no `RequestMeta`.
- Gate plan, main-task, grace-round, memory-compression, re-location, and review-filter requests.
- Requests from `ocr scan` and `ocr llm test` do not carry the review admission scope and remain unchanged.

#### Admit one real attempt at a time

Install the gate at the shared SDK middleware boundary, inside the SDK retry loop:

1. Before an eligible attempt reaches the retry observer and transport, acquire one provider permit using the request context.
2. If acquisition is cancelled, return without calling the transport and without leaking waiter or permit state.
3. If the transport returns an error, no response, or a response without a body, release the permit immediately.
4. If a response body is returned, transfer permit ownership to an idempotent body wrapper and release it on EOF or `Close`.
5. Retryable responses are closed by the SDK before it enters backoff, so SDK backoff holds no provider permit.
[...]

Issue #786, opened 2026-08-08
Add trusted resume with explicit provider and model transition lineage


Gongyl01:
### Problem Statement

Part of #368

`ocr review --resume` currently validates the review mode and raw ref strings, then reuses completed checkpoints whose per-file diff fingerprints still match. Changed, failed, or missing items are dispatched again.

However, it does not validate the resolved commits, the complete selected input set, repository identity, provider/model transition, or review-rule identity. A resumed result can therefore combine reused checkpoints and newly executed work even when the overall run identity has changed. Fully failed parent runs are also rejected before they can resume because they contain no completed checkpoint.

### 1. Validate the complete run identity before reuse

- Before reusing any checkpoint, compare the parent and child `source_artifact_sha256` using the #367 algorithm, `SHA-256(sorted[(item_id, fingerprint)])`, together with input mode, repository identity, and review-rule identity (`rule_config_sha256`).
- Any mismatch rejects the whole resume with zero checkpoint reuse, and fails before run creation: no child run, session, or manifest is persisted.
- Raw-ref string comparison is removed in favor of the resolved-identity comparison, so equivalent spellings such as `abc1234` and `abc1234def` no longer reject a resume.
- The parent manifest remains immutable; an accepted child run creates its own manifest under the same input identity.

### 2. Partition reusable and pending work

- Only completed or previously reused checkpoints that are readable and whose fingerprints agree with the parent manifest become `reused` in the child run.
- Failed, incomplete, missing, or corrupted checkpoints are dispatched again.
- A fully failed parent run remains resumable, in which case every selected item is dispatched again.
- Replayed comments are merged into output only and are not inserted into a new LLM context.

### 3. Require explicit provider/model transitions and record lineage

- A provider change is accepted only when the resuming command explicitly supplies `--provider`; a model change under the same provider is accepted only when it explicitly supplies `--model`.
- Changes caused only by configuration files or environment variables are implicit transitions and are rejected.
- An accepted resume writes one `resume_lineage` session event with schema `ocr.resume-lineage/v1`, recording the parent and child run IDs and the source/target provider and model. The transition kind follows from comparing source with target, and configuration changes are already carried by the manifest's `runtime_config_sha256`, so neither is stored again.
- Reused, completed, and remaining coverage comes from the child `ocr.run-manifest/v1`; the lineage event does not duplicate coverage.
- The lineage event does not contain credentials, complete endpoints, local paths, diffs, prompts, or model responses.

## Acceptance criteria

[...]

Issue #368, opened 2026-07-14
Add provider-aware partial retry with preserved lineage


acoliver:
## Problem

Provider failures have different operational meanings. HTTP 429 concurrency/quota responses, HTTP 529 overload, authentication/configuration failures, and network failures should not share one blind retry policy. Manual provider switches also lose the lineage needed to determine whether the same inputs were reviewed.

Across 145 retained OCR attempts, 19 partial runs carried HTTP 429 markers and six carried HTTP 529 markers. PR 306 added resumable sessions and failed-file checkpoints, but provider transitions, changed-input identity, and typed retry policy are not yet complete.

## Desired behavior

Add provider-aware retry of only failed work while preserving immutable input identity and explicit retry lineage.

## Acceptance criteria

- Classify 429, 529, authentication/configuration, timeout, and network failures separately.
- Support a provider-specific concurrency budget.
- Apply bounded exponential backoff with jitter and honor server retry headers where present.
- Retry only failed files/subtasks under the same immutable run manifest.
- Cross-provider fallback is explicit configuration, never implicit.
- Lineage records parent run/session, source and target provider/model, attempt number, changed configuration/inputs, reused files, completed files, and remaining failures.
- A retry cannot create a storm of parallel duplicate requests.
- Terminal completeness remains partial until all selected files complete or are explicitly waived.
- Tests cover same-provider resume, 429 backoff, 529 backoff, authentication failure, explicit provider transition, changed-input rejection, and cancellation.

## Non-goals

- This does not rank providers or hard-code a preferred provider.
- Repository workflows should configure budgets and display status rather than reimplement provider retry internals.

## Related work

PR 306 is the expected substrate for session and failed-file resume behavior.

hanhaiqingchuan:
I'd like to work on this. Could you please assign it to me?

Anon2Tokyo:
Hi @hanhaiqingchuan , I noticed this issue is assigned to you. I have a possible fix for this. Are you still working on it? If not, I'd like to submit a PR.

Gongyl01:
> Hi [@hanhaiqingchuan](https://github.com/hanhaiqingchuan) , I noticed this issue is assigned to you. I have a possible fix for this. Are you still working on it? If not, I'd like to submit a PR.

Hi, thanks for checking. I’m working on this issue in coordination with @hanhaiqingchuan , and the implementation is already underway. I’m planning to submit a PR soon. Thanks for understanding!

[...]

Issue #367, opened 2026-07-14
Emit a complete immutable review manifest and partial-coverage contract


acoliver:
## Problem

OpenCodeReview can produce useful findings while some selected files or subtasks fail. Consumers currently receive aggregate status and warnings, but not a complete machine-readable record of which files were selected, completed, reused, failed, or waived. This allows a partial review to look complete.

A retained cross-repository study observed 117 fully successful runs out of 145 attempts (80.7%), while 142 runs produced some usable output (97.9%). In a matched same-commit comparison, one run reviewed 90 files but reported 16 provider-concurrency subtask failures. Useful output and complete coverage are materially different states.

PR 306 added resumable sessions and file checkpoints, which appears to be the right substrate. This request is a narrow output-contract follow-up, not a parallel persistence system.

## Desired behavior

Expose a versioned immutable run manifest through JSON output and persisted sessions containing:

- canonical run/session ID and parent/resume ID;
- repository identity, requested head, resolved base, and exact range;
- OCR version, provider/model identifier, configured concurrency, and non-secret rule/config hashes;
- selected, completed, reused, failed, and explicitly waived file sets;
- typed failure classification per failed item;
- terminal state: complete, partial, failed, or skipped;
- redacted source artifact checksum and elapsed execution metadata.

A partial result may still publish useful findings, but it must not be represented as ordinary success or as a complete no-findings result.

## Acceptance criteria

- JSON output and the persisted session expose the same coverage values.
- Selected/completed/reused/failed arrays are machine-readable and stable under a versioned schema.
- Mixed-success runs terminate as partial until failed items are resumed successfully or explicitly waived.
- Resume retains the original input identity and links to its parent session.
- Tests cover full success, zero findings, mixed failures, resume, cancellation, and provider transition.
- No credentials, raw secrets, or provider tokens are emitted.

## Related work

- PR 306: resumable sessions and file checkpoints
- PR 179: persisted review results/viewer

Neither currently provides the complete selected-versus-completed coverage contract requested here.

lizhengfeng101:
Thank you for this thorough research and well-structured proposal. The cross-repository study data (80.7% fully successful vs 97.9% producing usable output) makes a compelling case for distinguishing between "has useful findings" and "complete coverage."

We agree this is an important capability and plan to support it. The direction of exposing a machine-readable coverage contract (selected/completed/reused/failed/waived) on top of the existing session infrastructure from PR #306 aligns well with where we want to take the project.

[...]

PR #790, opened 2026-08-08
Feat/llm retry report:SDK retry-attempt observability for review (#785)

### LINKED ISSUE #785: Expose official SDK LLM retry attempts in review output
### Problem Statement

Part of #368

OCR's Anthropic, OpenAI Chat Completions, and OpenAI Responses clients all use the official SDK retry loops with `WithMaxRetries(5)`. The SDKs retry eligible connection failures and HTTP responses internally, so OCR currently sees only the final success or failure. A `main_task` that receives 429, then 529, then succeeds is indistinguishable from a first-attempt success in the final review output.

This makes it difficult to tell whether retries happened, whether they recovered the request, what the provider returned before recovery, or why a request ultimately failed.

### Proposed Solution

Keep the official SDK retry decisions unchanged and add request-level observability:

1. **Observe real attempts.** Install SDK middleware on all three clients and record each real HTTP attempt: status code, provider request ID, `Retry-After`, `x-should-retry`, time to response headers, and the observed gap before the next attempt. The middleware must not consume response bodies or override SDK decisions.
2. **Correct post-header blind spots.** At the client return boundary, correct attempts whose failure becomes visible only after HTTP 200, including unexpected EOF or malformed response decoding, SSE stream interruption or integrity failure, and unsuccessful OpenAI Responses object statuses.
3. **Publish one frozen report.** Aggregate attempts by stable logical-request identity in a concurrency-safe collector, freeze one `RetryReport` at the `ocr review` run boundary, and publish the same snapshot to the terminal and to an optional top-level JSON field named `retry_report` with schema `ocr.llm-retry-report/v1`.

Attempt classification reads only HTTP status and Go error types, never error message text:

| `error_class` | Trigger |
| --- | --- |
| `rate_limited` | HTTP 429 |
| `overloaded` | HTTP 529 |
| `authentication` | HTTP 401/403; the SDK's status-based default does not retry these, although an explicit `x-should-retry: true` response directive may override that default |
| `timeout` | HTTP 408/504 or an error matching `context.DeadlineExceeded`, including parent-context and SDK per-attempt deadlines |
| `network` | DNS/connect/reset errors without a response, or unexpected EOF |
| `provider` | Other explicit HTTP/API errors, including 409 |
| `cancelled` | An error matching `context.Canceled`, including explicit parent-context cancellation |
| `unknown` | Stable fallback when no more specific classification is supported |

Per-request `outcome` is decided when the logical request finishes, rather than inferred from its last attempt: `succeeded`, `recovered`, `failed`, or `cancelled`. In particular, cancellation during SDK backoff is reported as `cancelled` even though the last observed attempt was an error.

## Acceptance criteria

- Anthropic, OpenAI Chat Completions, and OpenAI Responses expose every real SDK HTTP attempt through the shared observer.
- Review requests carry stable, non-secret identity for `plan`, `main_task`, memory compression, re-location, and review filter calls; `scan` and `llm test` requests do not enter the report.
- The report classifies the error classes listed above without inspecting raw error-message text.
- Unexpected EOF, malformed response decoding, post-200 stream failures, and unsuccessful Responses object statuses are not misreported as successful attempts.
- Cancellation during retry backoff produces no additional attempt and finalizes the logical request as `cancelled`.
- Terminal and JSON output are generated from the same frozen snapshot; concurrent runs have deterministic request and attempt ordering.
- A clean first-attempt-success run does not emit `retry_report`; a request with an error or an additional attempt does.
- The report does not contain credentials, authorization headers, prompts, request or response bodies, full endpoint URLs, or raw provider error strings.
- Tests cover 429 recovery, 529 exhaustion, authentication failure under the SDK's default status policy, network recovery, cancellation, EOF recovery, stream/response-status correction, concurrent collection, and output compatibility.

## Documentation follow-up

To avoid documenting an intermediate contract that may change while the remaining #368 sub-issues settle retry policy and resume lineage, the CLI reference update is deferred until the #368 contracts are finalized. The final documentation should cover the terminal retry summary and the optional `retry_report` JSON contract together with any later provider-policy or lineage surfaces.

## Out of scope

- This issue does not change SDK retry counts, backoff, jitter, or `Retry-After` behavior.
- It does not add a configuration-error class, provider-specific policy, provider concurrency budget, or a cross-subtask retry coordinator; those are tracked separately under #368.
- It does not change `--resume`, checkpoints, `ocr.run-manifest/v1`, or provider/model transition behavior.
- Only `ocr review` emits the report; `scan` and `llm test` keep their existing output contracts.

## Status

Implementation is available on `Gongyl01/open-code-review:feat/llm-retry-report` as five layered commits: data model, middleware observation, client-boundary correction, request-identity wiring, and run-boundary publishing. A PR with final test and coverage results will follow.

### Alternatives Considered

_No response_

### Affected Area

Telemetry / Observability

### Additional Context

_No response_

Gongyl01:
## Summary

Closes #785.

Part of #368.

Adds a versioned, immutable `RetryReport` to `ocr review` that records the real HTTP attempts made inside the official Anthropic and OpenAI SDK retry loops. OCR currently sees only the final return value of a logical LLM request, so a request that receives 429, then 529, and finally succeeds is indistinguishable from a first-attempt success.

The report is frozen once after the review run and its background work have finished. Text output renders a compact attempt chain, while `--format json` exposes the same frozen value under the optional top-level `retry_report` field. A clean first-attempt-success run emits no report, so the existing output remains unchanged.

This is the observability slice of #368. It explains what the existing SDK retry loops did; it does not take ownership of retry policy.

- **`internal/llm/retry_report.go`**: adds the `ocr.llm-retry-report/v1` value model and a per-run, concurrency-safe `RetryCollector`. The collector derives attempt numbers and timings, decides request outcomes exactly once, validates aggregate invariants, and freezes requests in deterministic `logical_request_id` order.
- **`internal/llm/retry_observer.go`**: mounts one shared Middleware implementation on Anthropic, OpenAI Chat Completions, and OpenAI Responses clients. It observes each real HTTP attempt without reading response bodies or overriding SDK decisions, recording status, provider request ID, server retry hints, `x-should-retry`, time to response headers, and the measured gap between attempts.
- **`internal/llm/retry_boundary.go` + client boundaries**: correct attempts whose failure becomes visible only after HTTP 200, including unexpected EOF or malformed decoding, interrupted/incomplete SSE streams, and unsuccessful Responses object statuses. Every logical request finalizes on success, error, cancellation, or panic.
- **`internal/llm/retry_meta.go` + review call sites**: stamps stable, non-secret identity on `plan`, `main_task`, memory compression, re-location, and review filter requests. The identity joins the report to existing session task records; `scan` and `llm test` remain outside the report.
- **`cmd/opencodereview`**: freezes the collector at the same run boundary as the manifest and publishes the report exactly once across normal and failure exits. Terminal and JSON output consume the same immutable snapshot.

### Commits (5 layered slices)

| Commit | Slice |
|---|---|
| `6c7525c` | `feat(llm)`: add retry report data layer |
| `5b9509b` | `feat(llm)`: observe retry attempts via SDK middleware |
| `4145d13` | `feat(llm)`: correct attempts and finalize requests at the client boundary |
| `8fec2a3` | `feat(llm)`: stamp request identity on review LLM requests |
| `e41741e` | `feat(cmd)`: publish the frozen retry report at the run boundary |

## Design (key invariants)

- **Observation only.** The official SDK remains the retry owner. This PR does not change `WithMaxRetries(5)`, SDK backoff or jitter, `Retry-After` behavior, or retry admission.
- **One logical request, one final outcome.** Request outcome is decided from the complete attempt sequence, the logical call's return value, and the parent context: `succeeded`, `recovered`, `failed`, or `cancelled`. It is never inferred from the last attempt alone.
- **Observed facts, not guessed text.** Attempt classification reads HTTP status and Go error types only. Raw provider error text is never parsed. A non-2xx status is authoritative; failures discovered after HTTP 200 are revised only when the boundary has typed evidence about the phase.
- **Exactly-once publication.** The collector freezes only after `Agent.Run` has joined background work. Normal result and failure-usage paths cannot publish the same report twice, and an invariant violation suppresses the self-contradictory report instead of emitting partial data. Joining background work before session finalization also closes a pre-existing review-path race that could append `llm_request`/`llm_response` events after the session file was finalized.
- **Deterministic under concurrency.** A per-run collector prevents state leakage between runs. Stable request identity, contiguous attempt numbering, and sorted `logical_request_id` output make concurrent runs reproducible.
- **Additive output contract.** `retry_report` is optional and uses `omitempty`; a run with only clean first attempts keeps its previous terminal and JSON output.
- **Allowlisted diagnostics.** The report may contain provider/model labels, file/task identity, status codes, provider request IDs, retry hints, and timings. It never contains credentials, authorization headers, prompts, request or response bodies, complete endpoint URLs, or raw provider error strings.

### Attempt classification

| `error_class` | Evidence |
|---|---|
| `rate_limited` | HTTP 429 |
| `overloaded` | HTTP 529 |
| `authentication` | HTTP 401/403 |
| `timeout` | HTTP 408/504 or an error matching `context.DeadlineExceeded` |
| `network` | Transport failure or unexpected EOF |
| `provider` | Other explicit non-2xx provider status, or a typed provider stream/status failure |
| `cancelled` | An error matching `context.Canceled`, including explicit parent-context cancellation |
| `unknown` | Stable fallback when a post-200 failure cannot be classified more specifically without parsing message text |

### Output contract

Example terminal output:

```text
LLM retry report: 1/2 requests retried, 2 retries, 1 recovered, 1 failed, 0 cancelled
- internal/agent/agent.go / main_task #1: rate_limited(429) -> overloaded(529) -> success
- internal/llm/client.go / plan #1: authentication(401) -> failed
```

The JSON path is additive:

```text
.retry_report.schema_version == "ocr.llm-retry-report/v1"
```

Failed and cancelled logical requests use separate `failed_requests` and `cancelled_requests` aggregates, while cancelled requests remain listed.

The terminal summary and `retry_report` are generated from the same frozen `RetryReport`; JSON mode emits one JSON document and never mixes in the terminal summary.

### Upstream integration

This branch is synchronized through `upstream/main@62e2b99`. The retry report is published alongside #367's frozen run manifest without changing its coverage or terminal-state contract. Existing scan output, `llm test`, Resume/checkpoint behavior, budget reporting, and session persistence semantics remain unchanged.

## How to test

```bash
make test
make vet

# A clean run keeps the existing contract: retry_report is absent.
ocr review --from main --to feature --format json | jq '.retry_report'

# On a run that encounters a retryable or terminal LLM failure,
# inspect the frozen report and its attempt chains.
ocr review --from main --to feature --format json | jq '.retry_report'
```

## Checklist

- [x] `make test` passes locally (full non-extension package suite with `-race -count=1`)
- [x] `make vet` passes
- [x] `git diff --check upstream/main...HEAD` is clean
- [x] `gofmt -l` reports no changed Go files
- [x] Tests cover 429 recovery, 529 exhaustion, authentication behavior, timeout and cancellation semantics, transport and unexpected EOF recovery, post-200 decode/stream/status correction, panic finalization, concurrent collection, deterministic ordering, and exactly-once terminal/JSON publication
- [x] Clean first-attempt-success output is regression-locked with no `retry_report`
- [x] The emitted schema and terminal rendering are allowlist-tested against secret or raw provider content

## Out of scope (deliberately)

- Provider-specific concurrency limits, retry budgets, retry admission, custom backoff, or a shared coordinator; tracked by #787.
- Trusted Resume, changed-input rejection, or provider/model transition lineage; tracked by #786.
- Automatic provider selection, provider ranking, or implicit cross-provider failover.
- CLI reference updates are deferred until the remaining #368 contracts are finalized, so retry output, provider policy, and Resume lineage can be documented together.


wu21-web:
Thank you, but this is big and hard to review.

Gongyl01:
> Thank you, but this is big and hard to review.

Hi @wu21-web, Thanks for the feedback! To help with the review, here's a breakdown — about **4,670 of the 6,197 added lines (~75%) are tests**. The core implementation is only ~1,500 lines:

**New core files (suggested reading order):**
1. `internal/llm/retry_meta.go` (+154) — retry metadata carried per request
2. `internal/llm/retry_boundary.go` (+141) — retry boundary decisions
3. `internal/llm/retry_observer.go` (+132) — observes retry attempts from the SDK
4. `internal/llm/retry_report.go` (+658) — aggregates observations into the report

**Integration points:**
- `internal/llm/client.go` (+97) and `internal/llm/responses_client.go` (+27) — hook observers into the SDK clients
- `internal/llmloop/loop.go` (+70) — thread retry metadata through the loop
- `internal/agent/agent.go` (+40) — propagate to agents
- `cmd/opencodereview/output.go` (+79) and `cmd/opencodereview/review_cmd.go` (+32) — render the report

lizhengfeng101:
Really solid piece of work — thanks for the layered commits and for writing down the invariants instead of just asserting them in the PR description. A few notes from going through it carefully.

**Things I verified rather than assumed**, since they're load-bearing for the whole design:

- The middleware genuinely sits *inside* the SDK retry loop. `requestconfig.go:419-423` builds `handler` from the middleware chain, and `res, err = handler(req)` is called at line 449 inside `for retryCount := 0; retryCount <= cfg.MaxRetries`. Same shape in openai-go v3. So one observer invocation per real HTTP attempt holds.
- `shouldRetry` (`requestconfig.go:248`) does consult `x-should-retry` ahead of the status code and does not exclude 2xx — your `OutcomeSucceeded` comment about "nothing was recovered from" is describing a real case, not a hypothetical.
- The per-attempt timeout path returns the *derived* ctx's `DeadlineExceeded`, while `finalizeRequest` reads the parent ctx for `parentCancelled`. So an attempt-level timeout correctly lands on `failed` rather than `cancelled`, exactly as your design note claims.
- `go build`, `go vet`, and `go test -race -count=1` across `internal/{llm,llmloop,diff,agent,scan,session}` + `cmd/opencodereview` all pass on my machine.

---

### The one thing I'd like to see changed: `cancelled` is counted in `failed_requests`

`retry_report.go:528` folds `OutcomeCancelled` into `FailedRequests`, and `RetryReport` has no `cancelled_requests` field. That matters more than it looks, because cancelling an in-flight background compression isn't an edge case — it's a designed-for, routine event. Beyond the deferred `cancelPendingCompression` at `loop.go:237`, there are two in-loop call sites (`loop.go:569` and `loop.go:590`): whenever the async job hasn't landed by the time the conversation crosses the warning threshold, you cancel it and fall back to synchronous compression.

The full chain: `asyncCtx = WithTimeout(WithoutCancel(ctx), 5min)` → `cancel()` → the in-flight request aborts, `http.Client.Do` returns `(nil, *url.Error{Err: context.Canceled})` → the observer records an attempt with `StatusCode == 0` classified `cancelled/context` → `finalizeRequest` sees `ctx.Err() == context.Canceled` → `OutcomeCancelled` → `FailedRequests++`, and it gets listed because `outcome != succeeded`.

Net effect: a run with zero retries and zero genuine failures can print

```
LLM retry report: 0/12 requests retried, 0 retries, 0 recovered, 1 failed
- internal/foo.go / memory_compression #1: cancelled -> cancelled
```

The classification itself is right — it really was cancelled. It's the aggregate label that's wrong, and `cancelled -> cancelled` reads oddly too (the attempt has no status code, so `retryAttemptChain` emits the bare class and then appends the request outcome). I'd add `cancelled_requests` and keep `failed_requests` for `OutcomeFailed` only. A user-initiated Ctrl-C on `main_task` is still worth surfacing, so this is about splitting the counter, not hiding anything.

---

### Two latent things — both unreachable today, so entirely your call

I'm flagging these as code-level inconsistencies, not as bugs. I checked reachability on both and couldn't get either to fire in production.

**`freezeErr` reaches the process exit code.** `review_cmd.go:284` returns `errors.Join(emitErr, freezeErr)`, which lands on `main.go:27`'s `os.Exit(1)`. On a fully successful run, stdout would carry a complete `"status": "success"` document while the process exits non-zero — awkward for CI. That said, I walked every `Freeze` error exit and none of them is reachable: `runID` is always non-empty (`session.New` unconditionally does `sessionID := generateUUID()`, `history.go:151`); the violation strings need either duplicate `RequestMeta` identity (impossible — `RequestNo` is allocated under `fs.mu`, scoped per `(FilePath, TaskType)`) or an attempt after the boundary defer (impossible — both the EOF re-call and the streaming path complete before it); `non-2xx without a classification` can't happen because `observeAttempt` is the only production caller and always classifies; and `not finalized` is closed off by `WaitBackground`. Your own `var newRetryCollector` hook exists for exactly this reason, so we agree on the facts. My only suggestion is that a pure-observability invariant breach probably shouldn't be able to fail an otherwise-successful review — a stderr warning would be a softer landing. Not a blocker either way.

**`isErrorStatus` vs the SDKs' `>= 400`.** `retry_report.go:172` treats anything outside 200–299 as proof of failure, while both SDKs use `res.StatusCode >= 400`. I initially thought a 3xx would show up as a false error attempt and flip a successful request to `recovered`, but that doesn't hold: both SDKs populate `GetBody` (`requestconfig.go:401` / `:434`), so `http.Client` follows 301/302/303/307/308 and the middleware only ever sees the post-redirect response. The statuses Go won't follow (300/304/305/306) don't occur against an LLM API — no conditional requests are sent, and Go rejects 305 outright. And even if a 300 did arrive, the SDK would try to JSON-decode it, fail, and `classifyBoundaryError` would land the request on `failed`, not a bogus `recovered`. So: real divergence, no reachable consequence. Might still be worth pinning to `>= 400`, or adding a line explaining why you're deliberately stricter than the SDK, since the "a non-2xx status is authoritative" framing in the PR body reads as if it matches the SDK boundary when it doesn't.

---

### Small stuff

- **`WaitBackground` fixes more than the PR claims credit for.** Because it runs ahead of `finalizeManifest()` and `session.Finalize()`, a background compression goroutine can no longer write `llm_request`/`llm_response` lines after the session file has been finalized. That was a pre-existing race on the review path and you've closed it — worth a sentence in the PR body, since a reviewer skimming for "why is this needed" will only see the retry-report justification.
- **`internal/scan` still has that race.** `scan/agent.go:335,371` call `a.session.Finalize()` but nothing calls `WaitBackground`. Out of scope here, and the nil `NewRequestMeta` comment is clear about why scan opts out of the report — but a line in `WaitBackground`'s doc saying "scan deliberately doesn't join, since it freezes no report" would stop the next reader from thinking it's an oversight.
- **Comment parity on re-location.** `compression.go:230` explains the record-before-request reordering and its resume implication in detail. `loop.go:470` makes the same reordering but only explains the `cm.Path` choice. Behaviour is equivalent (I diffed the old and new `msgs` semantics — `BuildReLocationMessages` returns nil for exactly the cases the old early return covered), so this is purely a cross-reference. Also worth noting that `rlStart` now includes prompt construction and `AppendTaskRecord` in `TaskRecord.Duration` — negligible, but the comment says moving it *would* change what Duration measures, when leaving it in place changed it slightly too.
- **Docs.** The top-level field table in `pages/src/content/docs/{en,ja,ru,zh}/cli-reference.md` doesn't list `retry_report`. It's already missing `manifest` from #367, so this isn't new debt you're creating, and `omitempty` keeps the JSON contract additive — but the terminal block is unconditionally user-visible. Deferring until #786/#787 land is a reasonable call; I'd just open a tracking issue so the backlog doesn't grow another entry.

---

One retraction on my side: I first questioned `outputRetryReportText(os.Stdout, ...)` versus the `stdout.Writer()` used elsewhere. That was wrong — `newQuietHandle` silences the `stdout` package's progress writer, not `os.Stdout`, and result output should bypass it exactly the way `outputTextWithWarnings` and the project summary already do. Your version is correct.

Overall: the only change I'd ask for before merge is splitting `cancelled_requests` out of `failed_requests`. Everything else is comments, docs, or optional tightening.

Gongyl01:
Thanks @lizhengfeng101 for the detailed review. I addressed the merge-blocking point in 5f99e5f: cancelled_requests is now separate from failed_requests, cancelled requests remain visible, and the duplicated cancelled -> cancelled rendering is removed. I also updated the related validation, tests, listing-rule comment, and the PR body’s WaitBackground race note. The two unreachable latent issues and the other optional items remain unchanged to keep this follow-up minimal. make check and the full race-enabled make test pass locally.

lizhengfeng101 (commented):
### Exit code concern: freeze error surfaces as a non-zero exit on a successful review

In `review_cmd.go`, the final return is now:
```go
return errors.Join(emitErr, freezeErr)
```

When the review itself fully succeeds but `Freeze` detects an invariant violation (un-finalized request), the CLI exits non-zero with `"freeze retry report: ..."`. The review result is already published to stdout — the JSON is valid and complete — but the exit code says failure.

I understand this is deliberate (the E2E test asserts it) and the condition is practically unreachable in production (every `CompletionsWithCtx` finalizes in its defer, and `WaitBackground` joins all goroutines). The intent is to surface internal bugs early.

However, this breaks the principle that **observability must not affect the system it observes**. A downstream CI pipeline using `ocr review || fail` would fail for a reason unrelated to review quality, with no actionable remediation for the user.

**Suggestion**: downgrade freeze failure to a warning on stderr and return the review result cleanly:

```go
if freezeErr != nil {
    fmt.Fprintf(os.Stderr, "[ocr] warning: %v (retry report suppressed)\n", freezeErr)
    // do not join into the return error
}
```

If the team prefers to keep it as a hard error for internal-bug detection, consider at minimum:
1. Document it in the CLI reference as a "publish error" distinct from "review failed"
2. Give the exit code a distinct value (e.g. exit 2 vs exit 1) so CI can distinguish

lizhengfeng101 (approved):
LGTM

PR #791, opened 2026-08-08
feat(scan): report token budget stop in JSON summary.budget_exceeded

### LINKED ISSUE #771: scan: expose token budget stop state in JSON output
## Problem

`ocr scan --format json --max-tokens-budget ...` already has an aggregate token budget gate: when the running token usage plus the next file estimate would exceed the configured budget, scan stops dispatching remaining files and records a `token_budget_reached` warning.

However, the JSON summary does not expose this state through `summary.budget_exceeded=true`. For automation users, this makes it hard to distinguish:

- a complete scan with no budget pressure
- a partial scan that stopped because the token budget gate was reached

This is mainly a machine-readable reporting gap rather than a request to change scan scheduling behavior.

## Why this matters

CI, GitHub Actions, dashboards, or other wrappers may consume `--format json` to decide whether a scan fully covered the selected files. If budget exhaustion is only visible as a warning item, consumers need to duplicate warning parsing logic instead of relying on the existing summary field.

A clearer contract would make `summary.budget_exceeded` reflect the aggregate budget gate consistently across review/scan result output.

## Current code path

Relevant code:

- `internal/scan/agent.go:616-625`: scan detects the aggregate budget gate, records `token_budget_reached`, sets local `budgetHit = true`, and breaks dispatch.
- `internal/scan/agent.go:522-544`: the outer scan loop receives `budgetHit` and stops scheduling subsequent batches.
- `internal/scan/agent.go:219-225`: `BudgetExceeded()` is currently hard-coded to return `false` for scan.
- `cmd/opencodereview/shared.go:354-357`: JSON output passes `ag.BudgetExceeded()` into the output layer.
- `cmd/opencodereview/output.go:316-325`: `summary.budget_exceeded` is populated from that value.

In short, scan has the local `budgetHit` signal, but that signal is not carried into the result provider state used by JSON output.

## Expected behavior

When the aggregate token budget gate stops `ocr scan` from dispatching remaining files, JSON output should include:

```json
{
  "summary": {
    "budget_exceeded": true
  },
  "warnings": [
    {
      "type": "token_budget_reached"
    }
  ]
}
```

When scan completes without hitting the aggregate budget gate, the current behavior can stay unchanged: `summary.budget_exceeded` may remain false or be omitted because of `omitempty`.

## Suggested implementation

A narrow implementation could be:

1. Add a `budgetExceeded bool` field to `internal/scan.Agent`.
2. When `dispatchBatch` returns `budgetHit == true`, set `a.budgetExceeded = true` before breaking out of the scan loop.
3. Change `scan.Agent.BudgetExceeded()` to return `a.budgetExceeded` instead of always returning `false`.
4. Add a focused test for `ocr scan --format json --max-tokens-budget ...` verifying:
   - `warnings` still contains `token_budget_reached`.
   - `summary.budget_exceeded == true` when the budget gate stops dispatch.
   - normal successful scans still omit or report false for `budget_exceeded`.

## Non-goals

This does not need to change:

- scan dispatch policy
- exit-code semantics
- comment generation
- warning generation
- manifest coverage behavior

The requested scope is only to expose an already-detected scan state in the existing machine-readable summary field.

chethanuk:
## Description

`ocr scan --format json --max-tokens-budget N` already detects the aggregate budget stop — it records a `token_budget_reached` warning and stops dispatching — but `summary.budget_exceeded` stayed `false`, because `scan.Agent.BudgetExceeded()` was hardcoded to `return false`. Automation consuming the JSON had to parse the warning list to tell a complete scan from one truncated by the budget.

The signal already existed inside scan; it just never reached the result provider.

```text
dispatchBatch (agent.go:586)
  recordWarning("token_budget_reached", …)     :636
  budgetHit = true            ← per-batch local, dies with the call
                              ← now also: a.budgetExceeded = true    :641
  break
      │
      ├─ normal return  :696 ─┐
      ├─ ctx-cancel     :651 ─┤→ dispatchSubtasks
      └─ (err path)           ┘   if err != nil { return }   :538  ← returns first
                                  if budgetHit { break }     :557

emitRunResult → ag.BudgetExceeded() → summary.budget_exceeded
```

**The flag is set where `budgetHit` is set, not at the `break`.** `dispatchBatch` has three exits that carry `budgetHit`, and the ctx-cancel exit at `:651` reaches `dispatchSubtasks`' `err != nil` return at `:538` — which fires *before* the `if budgetHit` check at `:557`. Writing the flag at the break would silently lose it on cancel-after-budget-hit. One write at `:641` covers all three exits.

No mutex or atomic: `budgetHit = true` runs in `dispatchBatch`'s own loop body, before `sem <- struct{}{}` and outside the worker closure, and `dispatchBatch` has one caller in a sequential batch loop — one writer on one goroutine, read only after `Run` returns. `internal/agent/agent.go` stores the same flag as a plain `bool` on the diff-review path. `make test` runs with `-race`, so this is enforced rather than argued.

`budget_exceeded` is diagnostic only and does not touch scan's status semantics: scan publishes no run manifest, so the status stays whatever the warnings imply — `completed_with_warnings` when the gate trips (the `token_budget_reached` warning is what produces that, not this flag), `success` otherwise. Both subtests assert the status explicitly so this cannot start driving it. `omitempty` keeps `budget_exceeded` out of the JSON entirely when the gate does not trip.

### Limitation

The new CLI-level test drives the real `*scan.Agent` and the real `emitRunResult`, not a spawned `ocr` binary — no test in this repo spawns it, and doing so would need live provider credentials. The flag-parsing layer between `parseScanFlags` and `scan.Args.MaxTokensBudget` is covered separately by `cmd/opencodereview/scan_cmd_test.go`.

## Type of Change

## How Has This Been Tested?

`make test` (with `-race`) and `make check` both pass on the current `main`.

- `cmd/opencodereview/scan_budget_json_test.go` (new) — drives a real `*scan.Agent` over eight fixture files through the real budget gate and `emitRunResult`, asserting `summary.budget_exceeded == true` plus the `token_budget_reached` warning at `MaxTokensBudget=120_000`, and that the raw JSON contains no `budget_exceeded` key at all when no budget is set.
- `internal/scan/budget_exceeded_test.go` (new) — table-driven, both directions, through `dispatchSubtasks`.
- Mutation check: deleting only the `a.budgetExceeded = true` line fails both.

## Checklist

- [x] I have signed the CLA

`budget_exceeded` is currently undocumented under `pages/` for the review path too, so documenting it belongs in a separate docs PR rather than this one.

## Related Issues

closes #771

### DIFF
--- a/cmd/opencodereview/scan_budget_json_test.go
+++ b/cmd/opencodereview/scan_budget_json_test.go
@@ -1,0 +1,139 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 alibaba/open-code-review Contributors
+
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/alibaba/open-code-review/internal/config/template"
+	"github.com/alibaba/open-code-review/internal/llm"
+	"github.com/alibaba/open-code-review/internal/scan"
+	"github.com/alibaba/open-code-review/internal/session"
+	"github.com/alibaba/open-code-review/internal/tool"
+)
+
+// fakeScanBudgetClient finishes every file in one round and reports a fixed
+// 50K token usage, so the aggregate budget gate trips deterministically.
+type fakeScanBudgetClient struct{}
+
+func (fakeScanBudgetClient) CompletionsWithCtx(_ context.Context, _ llm.ChatRequest) (*llm.ChatResponse, error) {
+	return &llm.ChatResponse{
+		Choices: []llm.Choice{{
+			Message: llm.ResponseMessage{
+				Role: "assistant",
+				ToolCalls: []llm.ToolCall{{
+					ID:       "1",
+					Type:     "function",
+					Function: llm.FunctionCall{Name: "task_done", Arguments: "{}"},
+				}},
+			},
+			FinishReason: "tool_calls",
+		}},
+		Usage: &llm.UsageInfo{PromptTokens: 50_000, TotalTokens: 50_000},
+	}, nil
+}
+
+// TestScanBudgetJSON is the end-to-end contract for #771: a real *scan.Agent,
+// real enumeration, real budget gate, through the shared emitRunResult
+// pipeline. BudgetExceeded() was hard-coded false, so summary.budget_exceeded
+// never appeared however early the gate cut the run short.
+//
+// The unlimited case asserts on the RAW capture, not the decoded struct:
+// budget_exceeded is omitempty, so false and absent are indistinguishable
+// after Unmarshal.
+func TestScanBudgetJSON(t *testing.T) {
+	cases := []struct {
+		name       string
+		budget     int64
+		want       bool
+		wantStatus string
+	}{
+		// The budget stop must NOT invent a typed status: it stays the
+		// ordinary warning-derived one (output.go leaves out.Status alone).
+		{name: "budget stop sets budget_exceeded", budget: 120_000, want: true, wantStatus: "completed_with_warnings"},
+		{name: "unlimited budget omits the key", budget: 0, want: false, wantStatus: "success"},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv("HOME", t.TempDir())
+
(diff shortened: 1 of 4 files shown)


lizhengfeng101 (approved):
LGTM

PR #792, opened 2026-08-08
fix(pages): keep previous page visible during lazy route transitions

### LINKED ISSUE #788: fix(pages): use startTransition to eliminate black screen flash during page navigation
## Description

When navigating between lazy-loaded pages (`/benchmark`, `/quickstart`, `/docs`, `/blog`), users see a brief **black screen flash**. This happens because `React.lazy()` triggers the `<Suspense>` fallback — a plain black `<div>` with no content:

```tsx
// pages/src/App.tsx, line 78
<Suspense fallback={<div style={{ minHeight: '100vh', background: '#000000' }} />}>
```

The fallback was designed for the **initial app load** scenario, but it also activates during in-app navigation when a chunk hasn't been fetched yet. This creates a jarring visual break — the entire page goes black for a few hundred milliseconds while the JS chunk downloads.

## Proposed Solution

Use React 18's [`useTransition`](https://react.dev/reference/react/useTransition) / `startTransition` to keep the **old page visible** until the new page's chunk has loaded and is ready to render. This is the React-recommended pattern for navigation transitions in SPAs:

- The Suspense fallback should only show during the **first paint** (no prior content to display).
- During **in-app navigation**, the previous route's content should remain visible until the new route is ready.

This is the same pattern used by Next.js App Router and Remix by default.

### Possible approaches

1. Wrap route changes with `useTransition` and show a subtle top progress bar or keep the old page until ready.
2. Use React Router's [`useNavigation`](https://reactrouter.com/en/main/hooks/use-navigation) if migrating to the data router API.
3. Keep the existing `<Suspense>` fallback for the true first-load case, but prevent it from showing during transitions.

## Scope

- File(s): `pages/src/App.tsx`
- Area: Suspense fallback + route transitions
- May also touch: `pages/src/index.tsx` (router setup), `pages/src/components/Navbar.tsx` (link navigation)

## Acceptance Criteria

- [ ] Navigating between pages no longer shows a black screen flash
- [ ] First-time app load still shows an appropriate loading state (spinner or skeleton)
- [ ] Works correctly on both desktop and mobile
- [ ] No regressions on back/forward browser navigation
- [ ] Tests pass (`cd pages && npm test`)
- [ ] Code check passes (`cd pages && npm run lint`)

## Context

The current Suspense fallback intentionally matches the page background (`#000000`) to avoid a white flash on initial load. The issue is that this same fallback is re-triggered on every lazy route transition. The fix should preserve the dark-background initial load behavior while eliminating the flash during navigation.

shaurya703:
Fixes #788

## Problem

Navigating to a lazy-loaded page (`/benchmark`, `/quickstart`, `/docs`, `/blog`) re-triggered the `<Suspense>` fallback — a plain black `<div>` — so the whole page went black while the route's chunk downloaded.

## Approach

Approach 3 from the issue: keep the existing fallback for the true first-load case, but prevent it from showing during in-app transitions.

- New `useTransitionedLocation()` hook: returns a location that trails `useLocation()` by one `startTransition`. `<Routes location={...}>` renders from it, so when the next route suspends, React keeps the current page visible (transition semantics) instead of unmounting to the fallback.
- First paint is unchanged — there is no previous content to keep, so the dark fallback still shows, preserving the intentional no-white-flash initial load.
- `ScrollToTop` now keys off the *displayed* location, so the scroll reset happens when the new page actually appears rather than while the old page is still on screen.
- No router migration needed — works with the existing `<Routes>` API on React Router 6.8.

## Acceptance criteria

- [x] No black flash during navigation — covered by a test: navigate to a lazy route whose chunk is deliberately unresolved, assert the old page is still on screen and the fallback is absent
- [x] First-time load still shows the dark loading state — covered by a test rendering directly at an unresolved lazy route
- [x] Back/forward navigation — covered by a test using router history
- [x] `npm test` — 19/19 pass (5 files, including the new one)
- [x] `npm run lint` — 0 errors (the 2 pre-existing warnings in `HeroSection.tsx`/`MarkdownRenderer.tsx` are untouched); `tsc --noEmit` clean

### DIFF
--- a/pages/src/App.tsx
+++ b/pages/src/App.tsx
@@ -2,9 +2,10 @@
 // Copyright 2026 alibaba/open-code-review Contributors
 
 import React, { Suspense, useEffect } from 'react';
-import { Routes, Route, useLocation } from 'react-router-dom';
+import { Routes, Route } from 'react-router-dom';
 import LandingPage from './components/LandingPage';
 import ErrorBoundary from './components/ErrorBoundary';
+import { useTransitionedLocation } from './hooks/useTransitionedLocation';
 import { useTranslation } from './i18n';
 import FeaturesPage from './pages/FeaturesPage';
 import FeaturesRoutePage from './pages/FeaturesRoutePage';
@@ -14,8 +15,7 @@ const QuickStartPage = React.lazy(() => import(/* webpackChunkName: "quickstart-
 const DocsPage = React.lazy(() => import(/* webpackChunkName: "docs-page" */ './pages/DocsPage'));
 const BlogPage = React.lazy(() => import(/* webpackChunkName: "blog-page" */ './pages/BlogPage'));
 
-const ScrollToTop: React.FC = () => {
-  const { pathname } = useLocation();
+const ScrollToTop: React.FC<{ pathname: string }> = ({ pathname }) => {
   useEffect(() => {
     window.scrollTo(0, 0);
   }, [pathname]);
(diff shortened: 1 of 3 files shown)


lizhengfeng101 (approved):
LGTM

PR #794, opened 2026-08-08
feat(pages): show live npm downloads in highlights stats

lizhengfeng101:
## What

- Add a \`useNpmDownloads\` hook that fetches real monthly download counts from the npm registry API at runtime.
- Surface the live download number in the Highlights stats row, falling back to a static i18n value while the request is loading or if it fails.
- Refresh the Highlights stats copy and ordering, and sync the label / caption / value changes across all four locales (en, zh, ja, ru).

## Why

The download figure was previously a hardcoded static value that drifted from reality. Fetching it live from the npm registry keeps the landing page honest and self-updating, while the static i18n fallback guarantees the section still renders on a slow or blocked network.

## Notes

- The hook \`encodeURIComponent\`s the (scoped) package name before interpolating it into the request URL.
- The fetch is wrapped in an \`AbortController\` with a timeout so the UI degrades to the fallback promptly instead of hanging on \`loading\` when the network is slow or unresponsive.
- No backend involved — the request runs in the visitor's browser against the CORS-enabled npm API, which is compatible with static GitHub Pages hosting.

## Verification

- \`npm run typecheck\` — passes
- \`npm run lint\` — 0 errors (2 pre-existing unrelated warnings)


MuoDoo (approved):
LGTM

PR #801, opened 2026-08-09
fix(agent): make --preview apply the same selection the review applies

### LINKED ISSUE #782: `--preview` reports files as "Will review" that the real run drops pre-dispatch
### OpenCodeReview Version

open-code-review v1.8.10 (adbd4fd) darwin/arm64 built at: 2026-08-08T01:30:30Z

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

### Bug Description

`--preview` applies only the static file filters, but a real review applies an additional per-file prompt-size filter before it freezes coverage. A file can therefore be listed under "Will review" and then never be reviewed.

Preview classifies files with `whyExcluded` (`internal/agent/preview.go:34`), covering `user_exclude`, `unsupported_ext`, `default_path`, `deleted`, and `binary`. A real run additionally calls `filterLargeDiffs` (`internal/agent/agent.go:497`), which drops any diff whose token count exceeds `llmloop.PromptTokenLimit(Template.MaxTokens)` — 80% of `max_tokens`. That runs *before* `registerCoverage` (`internal/agent/agent.go:509`) seals the selected set, so dropped files are absent from the manifest's selected coverage.

Preview cannot apply that filter today even in principle: `runPreview` (`cmd/opencodereview/review_cmd.go:378`) does not pass `Template` into `agent.Args`, so `Template.MaxTokens` is zero on the preview path.

This makes `--preview` unsound as a pre-flight check. It is the natural input for CI admission control ("is this changeset too big to review?"), but its selected set is not the set the review will actually cover.

### Steps to Reproduce

Default configuration, no flags. No API key is needed and no request is made: `filterLargeDiffs` runs before any dispatch; the dummy endpoint only gets past runtime construction.

```sh
mkdir repro && cd repro && git init -q .
git config user.email t@t && git config user.name t
printf 'package main\n\nfunc main() {}\n' > seed.go

# ~186 KB, so its diff exceeds 80% of the default max_tokens (58888 -> 47110)
{ echo "package huge"; for i in $(seq 1 4000); do echo "func Hn$i(a int) int { return a * $i }"; done; } > huge.go
git add -A && git commit -qm "add one large file"

# 1. preview lists huge.go under "Will review"
ocr review --commit HEAD --preview

# 2. the real run drops it before dispatch
OCR_LLM_URL=https://api.invalid.test/v1 OCR_LLM_TOKEN=dummy OCR_LLM_MODEL=fake-model ocr review --commit HEAD
```

### Expected Behavior

Preview's `will_review: true` set should equal the set a fresh, non-resumed, unbudgeted real run registers as selected coverage.

Concretely:

1. Extract the deterministic pre-dispatch selection as one operation returning every input file with a decision and a stable reason, and have both `Agent.Preview` and the real pre-dispatch path consume it. Reimplementing the size check separately in preview would just let the two drift again.
2. Add a stable exclusion reason for oversized diffs, e.g. `too_large`, so consumers need not reproduce the tokenizer or the 80% threshold. The human preview can render it through the existing reason rendering.
3. Give preview the resolved per-file limit without initializing a provider. Since #716 the limit comes from `--max-tokens`, then app config, then the template; `resolveMaxTokens` needs only `AppCfg`, which loads independently of `llm.ResolveEndpointWithOptions`. So preview can resolve the real limit while keeping its current property of requiring no API key.

Explicitly out of scope: aggregate `--max-tokens-budget` exhaustion, LLM failures, and anything else depending on execution. Those cannot be predicted statically, and preview should document that rather than guess.

### Logs / Error Output

```shell
$ ocr review --commit HEAD --preview

Preview: 1 file(s) changed  |  +4001  -0

Will review (1):
  [A]  huge.go              +4001 -0

$ ocr review --commit HEAD
[ocr] 1 file(s) changed, reviewing 1 in /tmp/repro
[ocr] Skipping huge.go (~70047 tokens exceeds 80% of max_tokens(58888))
[ocr] Pre-filtered 1 file(s) exceeding 80% of max_tokens
[ocr] All changed files exceeded the token size limit. Skipping review.
[ocr] Summary: 0 file(s) reviewed, 0 comment(s), ~0 token(s) used, 0s elapsed
Review skipped: no items were selected.
```

### Additional Context

The `WARNING: prompt tokens ...` line above is a different gate: it measures the assembled prompt rather than the diff, and does not drop the file. `ok.go` stays selected. Only `big2.go` leaves the selected set, and only `big2.go` is the divergence.

This looks like the remaining half of #480, which asked that "preview must apply the same relevant selection stages as review execution, or clearly report why later prefilters can change scope". #482 fixed the reporting side by making `files_reviewed` count dispatchable files; the selection sets themselves still diverge, and #480's note that "large-diff filtering also occurs at a different stage" still holds.

Related: #367, #480, #482, #716.

I'm happy to implement this, but it changes how selection is structured, so I'd rather agree on the approach first — in particular whether you prefer a shared pure selection function over having preview call `filterLargeDiffs` directly, and whether `too_large` is the reason string you want.

---

Disclosure: Issue report authored by AI, with human guidance and careful review.

zmwangx:
## Description

`--preview` applied only the static file filters, but a real review applies one more gate before it freezes coverage: any diff whose token count exceeds 80% of `max_tokens` is dropped. That gate runs before the selected set is sealed, so a file could be listed under "Will review" and then never be reviewed. Preview could not apply it even in principle — `runPreview` never passed a template, so the limit was zero on that path.

Rather than teach preview the size check separately — which would just let the two drift again — this turns the deterministic pre-dispatch selection into **one operation** both paths consume:

- `internal/agent/selection.go` holds `selectFiles`, which takes the changed diffs and returns a decision per file: the exclusion reason, or none. It is pure — no output, no mutation, no git or LLM access — so it can run in a preview that builds none of the review runtime.
- `Run` and `Agent.Preview` both call it. The three independently implemented decisions it replaces — the static filter, the size filter, and the reviewable count — are gone; the run's skip logging, its file counts, and preview's entries are all derived from the same decisions.
- The size gate moves from dispatch into shared selection. It still runs before coverage is sealed, while preview now consumes the identical decision.
- The token limit is an input to selection (`Args.Template.MaxTokens`), and zero disables the gate — which is how `delegate preview` opts out.

The net guarantee: preview's `will_review` set is the set a fresh, non-resumed, unbudgeted run registers as selected coverage.

User-visible changes:

- Oversized diffs are reported with a new `too_large` exclusion reason, in the human view and in `--format json`.
- Preview resolves the real per-file limit (`--max-tokens`, then the saved `max_tokens`, then the template) without building an LLM runtime, so it still needs no provider and no API key.
- The run's `N file(s) changed, reviewing M` line no longer counts files the size gate is about to drop.
- Run-level telemetry gains `review.skipped`, carrying the cause as an attribute. `no.files.changed` fires exactly where it did before, so existing consumers are unaffected.
- The skip line for an emptied selection names the count instead of claiming "All changed files", which was wrong whenever the static filters had already removed some.
- `ocr delegate preview` is deliberately unchanged: the host agent reviews with its own context window, not OCR's `max_tokens`.

Out of scope because they are execution-specific or unavailable to preview: aggregate `--max-tokens-budget` exhaustion, resume reuse (`--preview` cannot be combined with `--resume`), and LLM failures. The commit message covers the design and the trade-offs in detail.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

The repro from #782: one commit adding a single ~186 KB file, whose diff exceeds 80% of the default `max_tokens`. No real API key is needed: preview initializes no provider, and the full run uses dummy endpoint settings that are never contacted because the size gate exits before any request.

```sh
mkdir repro && cd repro && git init -q .
git config user.email t@t && git config user.name t
printf 'package main\n\nfunc main() {}\n' > seed.go

{ echo "package huge"; for i in $(seq 1 4000); do echo "func Hn$i(a int) int { return a * $i }"; done; } > huge.go
git add -A && git commit -qm "add one large file"
```

**Before** — preview promises a review that never happens:

```
$ ocr review --commit HEAD --preview

Preview: 1 file(s) changed  |  +4001  -0

Will review (1):
  [A]  huge.go              +4001 -0

$ OCR_LLM_URL=https://api.invalid.test/v1 OCR_LLM_TOKEN=dummy OCR_LLM_MODEL=fake-model ocr review --commit HEAD
[ocr] 1 file(s) changed, reviewing 1 in /tmp/repro
[ocr] Skipping huge.go (~70047 tokens exceeds 80% of max_tokens(58888))
[ocr] Pre-filtered 1 file(s) exceeding 80% of max_tokens
[ocr] All changed files exceeded the token size limit. Skipping review.
[ocr] Summary: 0 file(s) reviewed, 0 comment(s), ~0 token(s) used, 0s elapsed
Review skipped: no items were selected.
```

**After** — preview reports the drop, and the run's own count agrees with it:

```
$ ocr review --commit HEAD --preview

Preview: 1 file(s) changed  |  +4001  -0

Excluded from review (1):
  [A]  huge.go              (too_large)

$ OCR_LLM_URL=https://api.invalid.test/v1 OCR_LLM_TOKEN=dummy OCR_LLM_MODEL=fake-model ocr review --commit HEAD
[ocr] 1 file(s) changed, reviewing 0 in /tmp/repro
[ocr] Skipping huge.go (~70047 tokens exceeds 80% of max_tokens(58888))
[ocr] Pre-filtered 1 file(s) exceeding 80% of max_tokens
[ocr] 1 file(s) exceeded the token size limit; nothing left to review. Skipping review.
[ocr] Summary: 0 file(s) reviewed, 0 comment(s), ~0 token(s) used, 0s elapsed
Review skipped: no items were selected.
```

A regression test pins the property directly: preview's `will_review` set equals the set the run registers as selected coverage, over a changeset that mixes reviewable, extension-filtered, deleted and oversized files.

## Checklist

- [x] I have signed the CLA

## Related Issues

Closes #782.

---

Disclosure: PR authored by AI with human guidance and careful review.


lizhengfeng101:
@zmwangx Right fix — one selectFiles for both paths kills the drift at the source. Please add a test asserting previewMaxTokens resolves the same value the real run applies (that's the one seam left), and rebase onto main. 👍

zmwangx:
@lizhengfeng101 Cool, done. I had to introduce a test hook for `agent.New` in order to test this, which to me is a bit of a smell, haven't thought this through.

lizhengfeng101:
Pushed a small follow-up (a7fa1e4d): I think this smell is worth removing before merge. Dropped the `newReviewAgent` package var and asserted the max_tokens parity by calling `resolveMaxTokens` directly — the same function the run uses — instead of monkeypatching `agent.New`. Same parity property, no global mutable test state, and the seam no longer leaks into `review_cmd.go`. `make check` + `make test` green. Feel free to adjust or revert if you'd rather keep it.

lizhengfeng101 (approved):
LGTM

PR #802, opened 2026-08-09
fix(delegate): guard JSON workflow against old CLI versions

### LINKED ISSUE #800: ocr delegate preview fails with unknown flag: --format (mismatch with SKILL.md added in commit b1c7c6a)
### OpenCodeReview Version

open-code-review v1.8.10 (4bcc95a)

### Operating System

Linux (x86_64)

### Installation Method

npm (global)

### LLM Provider

### Bug Description

Commit b1c7c6a updated open-code-review-delegate/SKILL.md to instruct host agents to execute ocr delegate preview --format json. However, the CLI binary's command parser for delegate preview does not accept --format. Either --format needs to be implemented in the CLI binary for delegate preview, or --format json should be removed from SKILL.md.

### Steps to Reproduce

1. Install the latest `open-code-review` CLI tool and install/enable the `open-code-review-delegate` skill in a coding agent (e.g., QwenCode).
2. Prompt the agent to run a code review using delegation mode.
3. The host agent reads `SKILL.md` and attempts to execute: `ocr delegate preview --format json`
4. Observe the command failure due to an un-recognized `--format` flag.

### Expected Behavior

Either:
1. `ocr delegate preview` should accept the `--format json` flag and return machine-readable JSON output as documented/instructed in `skills/open-code-review-delegate/SKILL.md`.
OR
2. If `--format` is not intended for `ocr delegate preview`, `skills/open-code-review-delegate/SKILL.md` should be updated to remove `--format json` so agents do not invoke invalid CLI arguments.

### Logs / Error Output

```shell
Command: ocr delegate preview --format json
Directory: (root)
Output: Error: unknown flag: --format
Error: (none)
Exit Code: 1
```

### Additional Context

- **PR / Commit reference:** Commit b1c7c6a880cbe7c1c96b16ca0a693a4e63dfd1a9 updated `SKILL.md` and QCA templates to explicitly instruct host agents to pass `--format json`.
- **Agent impact:** When host agents run into this error, they are forced to catch the non-zero exit code, fall back to executing `ocr delegate preview --help`, and re-run the command without `--format json`.

amh1k:
## Description

Fixes the delegation skill/CLI version mismatch where newer skill instructions invoke:

```bash
ocr delegate preview --format json
```

against older OCR binaries that do not support the `--format` flag.

This adds CLI parser regression coverage and updates both delegation skill copies to detect unsupported OCR versions and instruct users to upgrade instead of silently falling back to incompatible text output.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] Manual and automated testing:
  - `go test ./...`
  - `go test ./cmd/opencodereview -run TestDelegateFlags_RegisterJSONFormat -count=1`
  - `make check`
  - `go run ./cmd/opencodereview delegate preview --help`
  - `go run ./cmd/opencodereview delegate rule --help`
  - `ocr delegate preview --format json` output validated as JSON
  - `git diff --check`

## Checklist

- [x] I have performed a self-review of the code
- [x] I have added tests that prove the fix is effective
- [x] New and existing unit tests pass locally
- [x] I have signed the CLA

## Related Issues

Closes #800

### DIFF
--- a/cmd/opencodereview/delegate_helpers_test.go
+++ b/cmd/opencodereview/delegate_helpers_test.go
@@ -6,6 +6,8 @@ package main
 import (
 	"context"
 	"testing"
+
+	"github.com/spf13/cobra"
 )
 
 func TestValidateDelegateOptions(t *testing.T) {
@@ -31,6 +33,42 @@ func TestValidateDelegateOptions(t *testing.T) {
 				t.Errorf("validateDelegateOptions() err = %v, wantErr %v", err, c.wantErr)
 			}
 		})
+	}
+}
+
+func TestDelegateFlags_RegisterJSONFormat(t *testing.T) {
+	var opts delegateOptions
+	cmd := &cobra.Command{
+		Use:           "preview",
+		SilenceUsage:  true,
+		SilenceErrors: true,
+		Args:          cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return validateDelegateOptions(&opts)
+		},
+	}
+	registerDelegateFlags(cmd, &opts)
+	cmd.SetArgs([]string{"--format", "json"})
+
+	if err := cmd.Execute(); err != nil {
+		t.Fatalf("delegate preview rejected --format json: %v", err)
+	}
+	if opts.format != "json" {
+		t.Fatalf("format = %q, want json", opts.format)
+	}
+
+	formatFlag := cmd.Flags().Lookup("format")
+	if formatFlag == nil {
+		t.Fatal("delegate preview did not register --format")
+	}
+	if formatFlag.Shorthand != "f" {
+		t.Fatalf("format shorthand = %q, want f", formatFlag.Shorthand)
+	}
+
+	for _, command := range []*cobra.Command{delegatePreviewCmd, delegateRuleCmd} {
+		if command.Flags().Lookup("format") == nil {
+			t.Errorf("%s did not register --format", command.Name())
+		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


wu21-web (approved):

lizhengfeng101:
Thanks for both the report and the PR, @amh1k — the diagnosis here was spot on. Your version-skew analysis (newer skill instructions reaching an older binary, rather than a missing implementation) is exactly right, and the write-up in #800 made it fast to confirm.

We're going to close this one without merging, and that's not a reflection on the change itself. As raised in #800, we want to settle a more fundamental question first: whether `ocr delegate` should expose a structured output mode at all. The delegation surface was designed so that its primary consumer is the outer host agent rather than a program, and that trade-off is still open. Until the protocol direction is settled, adding version guards to `SKILL.md` risks being reverted along with whatever we decide — and in the meantime the guard would have to be kept in sync across both skill copies and the QCA assets, which is the same drift problem that produced this issue in the first place.

The insight we do want to carry forward is the one your PR is built on: **silently falling back to text output is the wrong failure mode.** The downstream steps consume structured fields like `reviewable_files`, `mode`, and `merge_base`, and recovering those from the human-readable manifest is unreliable — so a version mismatch should fail loudly rather than quietly degrade the review. That needs an explicit answer in whatever shape the redesign takes.

We'll put together a proper design for this module and open it up for discussion — including compatibility in both directions (older CLI × newer skill, and the reverse, which today has no protection at all). We'd be glad to have you in that thread.

Thanks again for the CLA, the added test coverage, and the detailed PR description.

PR #808, opened 2026-08-09
feat(agent): group semantically related files for multi-file review

lizhengfeng101:
## Summary

- Introduce semantic file grouping: related files (e.g., `.go` + `_test.go`, header + implementation) are reviewed together in a single LLM conversation for better cross-file analysis.
- Add `grouping.go` with grouping logic and `enforceMaxFilesPerGroup` (max 10 files/group) + token budget enforcement.
- Revamp plan/review prompts: replace JSON plan format with a structured MUST/SHOULD directive using `[quick]`/`[deep]` cost tags; remove MAY category to reduce wasted investigation rounds.
- Update `main_task_system.md` to encourage "comment as you go" instead of batching comments until all items are investigated.
- Preserve comments on failure: when all groups fail but `CommentCollector` already has comments, return them instead of discarding. Rounds-exhausted groups with existing comments are no longer marked as hard errors.
- Adapt `code_comment` tool to accept a `path` field for multi-file group attribution.

## Motivation

Per-file review misses cross-file inconsistencies (e.g., interface change without updating callers). Grouping related files in one conversation enables the model to catch these issues. The prompt and failure-handling changes address recall regression observed during evaluation (model over-investigating without producing comments, and produced comments being discarded on partial failures).

## Test plan

- [x] `make check` passes (format, vet, license headers)
- [x] `make test` passes (unit tests including new grouping tests)
- [ ] Run evaluation benchmark to confirm recall recovery


css521 (approved):
LGTM

masihmirsali4-sketch:
b17460a6a3de8c5f93f693a1f3a0a17a77ed452c

PR #815, opened 2026-08-09
refactor: fix determinism tests in manifest_hash_test

wu21-web:
## Description
READ:
https://github.com/alibaba/open-code-review/pull/520#discussion_r3744212667
https://github.com/alibaba/open-code-review/pull/520#discussion_r3744227412

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [x] I have signed the CLA

## Related Issues

none

### DIFF
--- a/internal/agent/manifest_hash_test.go
+++ b/internal/agent/manifest_hash_test.go
@@ -41,10 +41,6 @@ func TestRuleConfigSHA256(t *testing.T) {
 	}
 	a := New(Args{SystemRule: base})
 
-	if a.ruleConfigSHA256() != a.ruleConfigSHA256() {
-		t.Fatal("ruleConfigSHA256 is not deterministic")
-	}
-
 	// A changed default rule changes the digest.
 	changed := New(Args{SystemRule: &rules.SystemRule{
 		DefaultRule: "different",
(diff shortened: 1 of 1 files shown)


lizhengfeng101:
Hi @wu21-web, thanks for the contribution and for taking the time to look into this!

Your observation is technically correct — for a correctly implemented pure function, calling it twice with the same receiver will always return the same value, so the condition is always false.

However, we'd like to keep these determinism checks as **regression guards**. They protect against future changes that might accidentally introduce non-determinism (e.g., iterating over a map without sorting keys, inadvertently using timestamps, etc.). The cost of keeping them is essentially zero, but they provide a safety net if the underlying implementation ever changes.

These kinds of invariant tests are common in projects where correctness of hash/digest functions is critical — their value lies not in catching bugs today, but in catching regressions tomorrow.

Thanks again for the thoughtful review of our test code!

wu21-web:
Thanks for the explaination @lizhengfeng101 . I think the author meant this maybe:
```go
a1 := New(baseArgs)
a2 := New(baseArgs)
if a1.runtimeConfigSHA256() != a2.runtimeConfigSHA256() {
    t.Fatal("runtimeConfigSHA256 is not deterministic")
}
```
and for L44-L46 it is the same idea
This would actually verify that two independently constructed but equivalent agents produce the same hash. the current version is dead code that can never run.

wu21-web:
[changes.patch](https://github.com/user-attachments/files/30886881/changes.patch)
I pushed a new fix to this branch, and here is a portable .patch file for you to review

PR #818, opened 2026-08-10
feat(llm): support custom retry status codes via retry_codes config

lizhengfeng101:
## Summary

Some self-hosted LLM clusters misuse HTTP status codes — e.g., returning 403 or 400 for rate limiting instead of the standard 429. This PR adds a `retry_codes` configuration field that lets users specify additional 4xx status codes to trigger exponential backoff retry.

- Adds `retry_codes` field to provider config, legacy llm config, and CLI `config set`
- Implements SDK middleware that injects `x-should-retry: true` response header on configured status codes, reusing the existing SDK retry mechanism (up to 5 retries with exponential backoff)
- Redundant codes (408/409/429) are silently filtered at resolve time and emit a warning at `config set` time, rather than blocking startup
- Fixes `cloneProviderEntry` to copy previously missing fields (`TimeoutSec`, `ExtraHeaders`, `RetryCodes`)

### Configuration example

```json
{
  "provider": "my-cluster",
  "custom_providers": {
    "my-cluster": {
      "api_key": "sk-xxx",
      "url": "https://internal.llm.corp/v1",
      "protocol": "openai",
      "model": "gpt-4o",
      "retry_codes": [403, 400]
    }
  }
}
```

Or via CLI:
```bash
ocr config set custom_providers.my-cluster.retry_codes 403,400
```

## Bugfix: cloneProviderEntry data loss

`cloneProviderEntry` (used by TUI for rollback snapshots) was missing `TimeoutSec`, `ExtraHeaders`, and `RetryCodes`. This caused silent data loss on failed edits — rolling back would discard the user's timeout and extra headers settings. Now fixed with full test coverage for deep-copy isolation and nil preservation.

## Test plan

- [x] `make check` passes (formatting, vet, license headers)
- [x] `make test` passes (all existing + new tests)
- [x] `TestCloneProviderEntry_TimeoutAndRetryCodes` — verifies deep-copy and mutation isolation
- [x] `TestCloneProviderEntry_NilRetryCodesAndExtraHeaders` — pins nil-stays-nil behavior
- [ ] Manual test: configure `retry_codes` with a mock server returning the configured status code, verify retry behavior

### DIFF
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -537,9 +537,12 @@ func setConfigValue(cfg *Config, key, value string) error {
 		}
 		cfg.Llm.ExtraBody = m
 	case "llm.retry_codes", "llm.RetryCodes":
-		codes, err := llm.ParseRetryCodes(value)
+		codes, warnings, err := llm.ParseRetryCodes(value)
 		if err != nil {
 			return err
+		}
+		for _, w := range warnings {
+			fmt.Fprintf(os.Stderr, "[ocr] WARNING: %s\n", w)
 		}
 		cfg.Llm.RetryCodes = codes
 	default:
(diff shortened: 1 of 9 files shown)


wu21-web (approved):
The approach is sound and the wiring is complete. The only real gap is missing tests for the new functions.

wu21-web (approved):

MuoDoo (approved):
LGTM

PR #822, opened 2026-08-10
fix(llm): request stream usage by default with extra_body override

Fanzzzd:
## Description

Streaming requests currently receive token usage only when the provider volunteers it. OpenAI-compatible servers omit usage from streams unless `stream_options.include_usage` is set, so every streamed review silently loses token/cost accounting (`resp.Usage` stays nil and the session totals under-report).

This PR sets `stream_options: {"include_usage": true}` by default when streaming is enabled, while keeping the existing `extra_body` passthrough fully in control:

| `extra_body.stream_options` | wire behavior |
|---|---|
| absent | `{"include_usage": true}` (new default) |
| an object | sent as-is, replaces the default |
| `null` | field omitted entirely — escape hatch for gateways that reject `stream_options` (some older vLLM/Azure builds) |

Non-streaming requests never carry the field regardless of config, mirroring how the `stream` key itself is handled (#647): providers reject `stream_options` unless `stream` is true.

**Behavioral change:** streamed requests now include `stream_options.include_usage: true` by default. Servers that implement stream_options respond with one extra final chunk carrying usage; servers that reject the field can be accommodated with `stream_options: null` in `extra_body`. No change for non-streaming requests.

Split out of #806 (which forced `include_usage` unconditionally, with no escape hatch) so it can be reviewed independently of the replay design discussion there.

## Type of Change

## How Has This Been Tested?

- [x] New HTTP-boundary tests assert the actual request body: `TestOpenAIClient_StreamingRequestsUsageByDefault` (default present + usage captured from the final chunk) and `TestOpenAIClient_StreamOptionsConfigOverrides` (explicit object replaces default / explicit null suppresses the field / non-streaming never sends it)

## Checklist

- [ ] I have updated the documentation accordingly (if applicable) — `extra_body.stream` itself is currently undocumented; happy to add a Streaming section to the configuration docs (en/zh/ja/ru) in a follow-up if you'd like
- [x] I have signed the CLA

## Related Issues

Related to #806 (split out of it).

### DIFF
--- a/internal/llm/client.go
+++ b/internal/llm/client.go
@@ -380,9 +380,21 @@ func (c *OpenAIClient) CompletionsWithCtx(ctx context.Context, req ChatRequest) 
 			// Ask for the final usage chunk by default.
 			params.StreamOptions = openai.ChatCompletionStreamOptionsParam{IncludeUsage: openai.Bool(true)}
 		} else if streamOptions != nil {
-			// An explicit stream_options in extra_body replaces the default.
-			// An explicit null suppresses the field entirely, for gateways
-			// that reject stream_options.
+			// An explicit stream_options in extra_body replaces the default,
+			// but usage stays requested unless include_usage itself is spelled
+			// out: configuring an unrelated stream option must not silently
+			// disable cost accounting. An explicit null suppresses the field
+			// entirely, for gateways that reject stream_options.
+			if object, ok := streamOptions.(map[string]any); ok {
+				if _, has := object["include_usage"]; !has {
+					merged := make(map[string]any, len(object)+1)
+					for key, value := range object {
+						merged[key] = value
+					}
+					merged["include_usage"] = true
+					streamOptions = merged
+				}
+			}
 			opts = append(opts, openaiopt.WithJSONSet("stream_options", streamOptions))
 		}
 		return c.completionsStreaming(ctx, params, opts...)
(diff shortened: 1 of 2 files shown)


internal/llm/client.go
```diff
@@ -365,13 +365,26 @@ func (c *OpenAIClient) CompletionsWithCtx(ctx context.Context, req ChatRequest)
...
+		if streamOptions, ok := c.cfg.ExtraBody["stream_options"]; !ok {
+			// OpenAI-compatible servers omit token usage from streams unless
+			// asked, silently losing cost accounting for streamed requests.
+			// Ask for the final usage chunk by default.
+			params.StreamOptions = openai.ChatCompletionStreamOptionsParam{IncludeUsage: openai.Bool(true)}
+		} else if streamOptions != nil {
+			// An explicit stream_options in extra_body replaces the default.
+			// An explicit null suppresses the field entirely, for gateways
+			// that reject stream_options.
+			opts = append(opts, openaiopt.WithJSONSet("stream_options", streamOptions))
```
stay-foolish-forever, line 398: Maybe you could further check whether `include_usage` is set here to prevent silent disabling of this option.

Fanzzzd: Good catch — an object configuring only an unrelated stream option (e.g. vLLM's `continuous_usage_stats`) would indeed silently drop usage accounting. Fixed in 9d026d7: `include_usage: true` is now merged into any explicit `stream_options` object unless the config spells out `include_usage` itself; the explicit-`null` suppression escape hatch is unchanged. Covered by a new case in `TestOpenAIClient_StreamOptionsConfigOverrides`.

stay-foolish-forever (approved):

PR #823, opened 2026-08-10
fix(llmloop): keep live rounds out of compression

Fanzzzd:
## Description

**Scoped down per review feedback: this PR now fixes only the no-tool retry bound.** The compression-zone defects that were previously bundled here (background compression summarizing the live tail, unreserved frozen-zone budget, template-less truncation) moved to #838 with standalone reproductions.

When the model answers without tool calls, the retry path appended the assistant text + retry nudge directly to `messages`, bypassing compression entirely — a model that keeps chatting without tools grows the prompt unchecked until the provider rejects it.

The fix addresses the root cause in two parts:

1. **`groupIntoRounds` could not see retry messages.** It only recognized `assistant` and `tool` roles, so the user retry nudges belonged to no round: even when compression did run, partition boundaries and token accounting skipped them. A `round` now spans from its assistant message to the next assistant message (or the end of the conversation), covering tool results and interleaved user messages alike — which also simplifies the boundary math to `compressEnd = lastCompressRound.end`.
2. **The retry path now shares the tool rounds' bounds.** `addNextMessage` is generalized to `appendMessagesWithCompression`, taking one whole logical update at a time so thresholds are evaluated once per round and compression never runs between an assistant message and its tool results. Its success predicate is the exact complement of the sync-compression trigger (`<= warnLimit`), so a conversation sitting exactly at the warning threshold is not misreported as a failure.

**Behavioral change:** a retry loop that overflows the budget now stops with `StopCompression` instead of growing until a provider error. No wire-format changes.

## Type of Change

## How Has This Been Tested?

- [x] `TestRunPerFile_NoToolRetryBoundedByCompression` discriminates: with the old retry path restored it fails (retries keep going, `StopMaxRounds`); with the fix it passes (1 LLM call, `StopCompression`)
- [x] `TestGroupIntoRounds_RetryNudgesStayWithTheirRound` pins the new round span on the retry shape; the existing `TestGroupIntoRounds` is updated to the span representation with identical boundaries for assistant+tool sequences

## Checklist

- [ ] I have updated the documentation accordingly (if applicable) (n/a — internal behavior)
- [x] I have signed the CLA

## Related Issues

Related to #838 (the compression-zone defects split out of this PR) and #806 (originally split out of it).

### DIFF
--- a/internal/llmloop/compression.go
+++ b/internal/llmloop/compression.go
@@ -30,11 +30,15 @@ func PromptTokenLimit(maxTokens int) int {
 	return int(float64(maxTokens) * tokenWarningThreshold)
 }
 
-// round groups consecutive messages starting with an assistant message
-// followed by zero or more tool result messages.
+// round is one assistant turn plus everything that follows it up to the
+// next assistant message: its tool results and any interleaved user
+// messages, such as the no-tool retry nudge. Spanning the whole gap keeps
+// token accounting and zone boundaries aligned with the conversation the
+// model actually replays — index-only bookkeeping (assistant + tool roles)
+// left retry messages invisible to both.
 type round struct {
 	assistantIdx int
-	toolIdxs     []int
+	end          int // exclusive index of the message after the round
 }
 
 // partitionResult describes how messages should be split for compression.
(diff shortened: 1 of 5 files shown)


Fanzzzd:
Addressed the reviewer bot's finding in a1d3b78: the success return of `appendMessagesWithCompression` is now `finalCount <= warnLimit`, the exact complement of the `> warnLimit` sync-compression trigger, so a conversation sitting exactly at the threshold is no longer reported as a compression failure. (The strict `<` predates this PR, but since this PR owns the refactor of that function it's the right place to fix it.)

stay-foolish-forever:
Really appreciate your contribution, but the size of the change seems too large than I expected, with multiple fixes or irrelative modifications applied in a single PR. The "no-tool retry path had no token bound" issue emerged from the bug in function `groupIntoRounds` and I did not see any fixes related to it. And your analysis of the issues related to background compression tasks is reasonable, but maybe there is better solution for this.

My advice is to fix the "no-tool retry path" with a neat implementation in this PR, and left the background compression issue in another PR (better with a separate issue too). Thanks again for your impressive analysis the context compression issues and your rich contribution, wish you have a good day!

Fanzzzd:
Thanks for the direction — you're right on both counts, and the PR is reworked accordingly.

**Scope:** this PR is now only the no-tool retry bound (single commit, `internal/llmloop` only). The compression-zone changes moved out to #838, which carries standalone failing-test reproductions for each defect and leaves the design space open — I noted your point that background compression may deserve a better overall solution there.

**`groupIntoRounds`:** agreed, that's where the root cause lives — it only recognized `assistant`/`tool` roles, so the retry nudges belonged to no round and stayed invisible to partition boundaries and token accounting even when compression ran. A `round` now spans from its assistant message to the next one, which covers tool results and interleaved user messages uniformly and simplifies the boundary computation to `lastCompressRound.end` (identical boundaries for ordinary assistant+tool sequences — the existing `TestGroupIntoRounds` cases keep the same cut points). The loop side routes the retry path through the same append+threshold helper as tool rounds, so the bound applies uniformly.

`TestRunPerFile_NoToolRetryBoundedByCompression` discriminates the fix (fails with the old retry path restored, passes with it), and `TestGroupIntoRounds_RetryNudgesStayWithTheirRound` pins the new span on the retry shape.

lizhengfeng101:
@Fanzzzd The direction here is right, but this branch has drifted far from `main` (it still uses the pre-refactor `RunPerFile`/`addNextMessage` and pre-`messageTokens` token accounting), so please rebase onto latest `main` and reconcile the native/reasoning accounting by hand — or, given the size of the conflict, it may be cleaner to close this and open a fresh PR off current `main`.

PR #828, opened 2026-08-10
feat(llm): add dynamic cache breakpoint on last message for Anthropic

yingjiexu2002:
## Description

Mark the last content block of the final message with ephemeral `cache_control` so multi-turn conversations cache the growing history, not just the static system+tools prefix. Each turn then reads the previous full prefix and only writes the new delta.

Previously only two static breakpoints were set (last system block, last tool definition), so everything after the tools — the growing multi-turn conversation — was re-processed as fresh input on every turn of the per-file review loop.

This change adds a dynamic breakpoint on the last content block of the final message (user text, `tool_result`, or `tool_use` — all handled generically via the SDK's `GetCacheControl()` helper). Since request params are rebuilt from the append-only history on every turn, the marker never leaks back into stored history, and the breakpoint simply advances with each turn. Total breakpoints stay at 3, within Anthropic's limit of 4.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Added `TestBuildAnthropicParams_DynamicCacheBreakpoint` covering:
- conversations ending in a `tool_result` block (the typical agent-loop turn shape) — the last block gets `cache_control: {"type":"ephemeral"}` and earlier messages stay unmarked
- conversations ending in user text

Also verified against the Anthropic API in a real multi-turn review session: turns after the first report `cache_read` tokens for the full prior prefix and only write the delta.

## Checklist 

- [ ] I have signed the CLA

### DIFF
--- a/internal/llm/client.go
+++ b/internal/llm/client.go
@@ -833,6 +833,16 @@ func (c *AnthropicClient) buildAnthropicParams(model string, req ChatRequest) (a
 		tools[len(tools)-1].OfTool.CacheControl = anthropic.NewCacheControlEphemeralParam()
 		params.Tools = tools
 	}
+	// Dynamic breakpoint on the latest message so multi-turn history is
+	// cached incrementally: read the full previous prefix, write only the delta.
+	if len(messages) > 0 {
+		last := &messages[len(messages)-1]
+		if len(last.Content) > 0 {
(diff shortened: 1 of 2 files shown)


lizhengfeng101 (approved):
LGTM

PR #836, opened 2026-08-11
fix(action): pin nested action references to full commit SHAs

### LINKED ISSUE #816: Internal actions/* references use floating tags, undermining SHA-pinned invocations
## Problem

Pinning `uses: alibaba/open-code-review@<commit>` does not currently freeze what runs. In the [v1.8.10 action](https://github.com/alibaba/open-code-review/blob/4bcc95acecd629e64ea984c962cfd08dd892a38e/action.yml), the internal actions use floating tags — `actions/checkout@v7`, `actions/setup-node@v7`, `actions/upload-artifact@v4`, `actions/github-script@v9` — so the composite's implementation can change underneath a consumer's SHA pin without any change to their workflow. Consumers have no way to pin these references themselves, which makes the official action hard to use as a required review gate under SHA-pinning policies and encourages vendoring it.

## Suggestion

- Pin the internal `actions/*` references to full commit SHAs with a `# vX.Y.Z` trailing comment.
- Show a fully SHA-pinned invocation in the GitHub Actions example.
- Add a consistency check so a floating tag cannot reappear in `action.yml`.

→ Implemented in #836.

---

*Scope note: this issue originally also covered the `ocr_version: latest` default. That is a separate design decision with an active discussion, now tracked in #839 so this issue can close with the part already agreed.*

Fanzzzd:
## Description

Closes the floating-tag gap in #816, scoped to what was agreed there ("PR1"):

- **`action.yml`**: the four nested `actions/*` references are pinned to full commit SHAs with a trailing `# vX.Y.Z` comment (`checkout@v7 → v7.0.1`, `setup-node@v7 → v7.0.0`, `upload-artifact@v4 → v4.6.2`, `github-script@v9 → v9.0.0`). Each SHA was cross-checked against the exact version tag before embedding.
- **`scripts/verify-action-pins.sh`**: fails when any external `uses:` in `action.yml` is not a 40-hex SHA followed by a `# vX.Y.Z` comment (local `./` references exempt). Wired into CI next to the license check, so the invariant holds on every PR — which also makes it hold at release time, since releases are cut from CI-green commits.
- **`examples/github_actions/README.md`**: new "Reproducible pinning" section demonstrating the fully pinned invocation, and spelling out why the Action SHA alone is not enough — `ocr_version` must be pinned too because the CLI is installed from npm at run time.

Deliberately **not** in this PR: any change to the `ocr_version: latest` default, release-tag lineage, or versioning scheme — that design question now lives in #839. `@main` convenience usage in the example workflow is unchanged.

## Type of Change

- [ ] Documentation update
- [x] CI / Build / Tooling

## How Has This Been Tested?

- `bash scripts/verify-action-pins.sh` passes on the pinned `action.yml`.
- Negative test: reverting one pin back to `actions/checkout@v7` makes the script exit 1 and print the offending line.
- Every embedded SHA verified to equal the commit its `# vX.Y.Z` comment names, via the GitHub API.

- [ ] `make test` passes locally (no Go changes; `go test ./...` unaffected)

## Checklist

- [x] I have signed the CLA

## Related Issues

Fixes #816

The `ocr_version` default discussion continues in #839

### DIFF
--- a/scripts/verify-action-pins.sh
+++ b/scripts/verify-action-pins.sh
@@ -19,6 +19,10 @@ local_ref='uses:[[:space:]]*\./'
 
 bad=""
 for file in "${files[@]}"; do
+  if [ ! -f "$file" ]; then
+    echo "ERROR: $file not found; the pin check cannot run." >&2
+    exit 1
+  fi
   hits="$(grep -nE 'uses:' "$file" || true)"
   [ -n "$hits" ] || continue
   while IFS= read -r line; do
--- a/scripts/verify-action-pins.sh
+++ b/scripts/verify-action-pins.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 alibaba/open-code-review Contributors
+
+# Verify that every external action referenced by the published composite
+# action (action.yml) is pinned to a full 40-hex commit SHA with a trailing
+# "# vX.Y.Z" version comment. A floating tag inside action.yml silently
+# undermines consumers who SHA-pin alibaba/open-code-review itself: the
+# outer pin freezes this repository, but a moved inner tag still changes
+# what actually runs (see issue #816).
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+files=("action.yml")
+pinned='uses:[[:space:]]*[A-Za-z0-9_.-]+/[A-Za-z0-9_./-]+@[0-9a-f]{40}[[:space:]]+#[[:space:]]*v[0-9]'
+local_ref='uses:[[:space:]]*\./'
+
+bad=""
+for file in "${files[@]}"; do
+  hits="$(grep -nE 'uses:' "$file" || true)"
+  [ -n "$hits" ] || continue
+  while IFS= read -r line; do
+    if printf '%s' "$line" | grep -qE "$local_ref"; then
+      continue
+    fi
+    if ! printf '%s' "$line" | grep -qE "$pinned"; then
+      bad="${bad}${file}:${line}"$'\n'
+    fi
+  done <<< "$hits"
+done
+
+if [ -n "$bad" ]; then
+  echo "The following action references are not pinned to a full commit SHA"
+  echo "with a '# vX.Y.Z' comment:"
+  printf '%s' "$bad"
+  echo "Pin them like: uses: owner/repo@<40-hex-sha> # vX.Y.Z"
+  exit 1
+fi
+
+echo "All external action references in ${files[*]} are SHA-pinned."
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,9 @@ jobs:
       - name: Verify license headers
         run: bash scripts/verify-license.sh
 
+      - name: Verify action pins
+        run: bash scripts/verify-action-pins.sh
+
       - name: Check formatting
         run: |
           unformatted="$(gofmt -s -l .)"
--- a/action.yml
+++ b/action.yml
@@ -198,7 +198,7 @@ runs:
 
     - name: Setup Node.js
       if: steps.check_deps.outputs.node_installed != 'true'
-      uses: actions/setup-node@v7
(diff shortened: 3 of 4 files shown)


scripts/verify-action-pins.sh
```diff
@@ -0,0 +1,38 @@
...
+cd "$(dirname "$0")/.."
+
+files=("action.yml")
+pinned='uses:[[:space:]]*[A-Za-z0-9_.-]+/[A-Za-z0-9_./-]+@[0-9a-f]{40}[[:space:]]+#[[:space:]]*v[0-9]'
+local_ref='uses:[[:space:]]*\./'
+
+bad=""
+for file in "${files[@]}"; do
+  hits="$(grep -nE 'uses:' "$file" || true)"
+  [ -n "$hits" ] || continue
```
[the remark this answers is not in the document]
Fanzzzd: Fair point — a missing `action.yml` fell through to the success message. Fixed in a535fec: the script now errors out when the file is absent (verified by running it from a directory without `action.yml`).

lizhengfeng101 (approved):
LGTM

lizhengfeng101 (commented):
Nice work — SHAs all verified, script logic is sound.

One follow-up thought: `release.yml` has publish/signing credentials, so a compromised upstream action there has a wider blast radius than in CI. Would you be open to a follow-up issue to pin `release.yml` the same way and add it to the `files` array in `verify-action-pins.sh`?

scripts/verify-action-pins.sh
```diff
@@ -0,0 +1,46 @@
...
+# "# vX.Y.Z" version comment. A floating tag inside action.yml silently
+# undermines consumers who SHA-pin alibaba/open-code-review itself: the
+# outer pin freezes this repository, but a moved inner tag still changes
+# what actually runs (see issue #816).
+set -euo pipefail
+
+cd "$(dirname "$0")/.."
+
+files=("action.yml")
+pinned='uses:[[:space:]]*[A-Za-z0-9_.-]+/[A-Za-z0-9_./-]+@[0-9a-f]{40}[[:space:]]+#[[:space:]]*v[0-9]'
```
wu21-web, line 17: The regular expression here does not anchor the entire line.
Therefore, it checks whether a compliant fragment appears at any position in this line, rather than `uses:` whether the actual value is compliant. For example, the following floating reference will be incorrectly accepted:
`- uses: actions/checkout@v7 # prior uses: x/y@0123456789012345678901234567890123456789 # v1`

scripts/verify-action-pins.sh
wu21-web, line 1: Version comments only check `v` for a number following the specified value, so `vX.Y.Z` content that doesn't conform to the declared format will still pass.

scripts/verify-action-pins.sh
Fanzzzd: Confirmed — reproduced with exactly your example line before fixing (the old pattern accepted it). In 37fc06f the pattern now covers the whole line (`^…uses: owner/repo@<40-hex> # vX.Y.Z$`), and only lines that are actual `uses:` directives are examined — which also surfaced a latent false positive: prose mentioning `uses: ./` inside the embedded github-script block previously slipped through the local-ref exemption. Verified with a positive/negative matrix: clean file, your adversarial comment line, `# v7`, trailing garbage after the version comment, dash-form list items, `./` local refs, and a missing file.

scripts/verify-action-pins.sh
Fanzzzd: Also correct — the old pattern accepted any comment starting with `v<digit>`. Fixed in 37fc06f: the comment must now be exactly `# vX.Y.Z` followed by end of line, so both a loose `# v7` and trailing content after the version fail the check.

Fanzzzd:
@lizhengfeng101 good call on the blast radius — filed as #841 with the current floating references in `release.yml` enumerated (including the third-party `softprops/action-gh-release`, which runs in the job holding `NPM_TOKEN` and attestation permissions). Happy to send that PR once this one lands, since it just extends the `files` array of the same check.

Also addressed @wu21-web's two review findings in 37fc06f: the pin pattern is now anchored to the entire line (a floating tag can no longer hide behind a pinned-looking comment fragment) and the version comment must be exactly `# vX.Y.Z` — both reproduced before fixing, plus a full positive/negative matrix in the thread replies.

PR #837, opened 2026-08-11
fix(install): detect stale or shadowed ocr command after install

Cooooper:
## Background

Right after installing (e.g. `npm i -g @alibaba-group/open-code-review`), running `ocr config provider` **in the same terminal** can fail with a confusing error coming from an unrelated program — while a newly opened terminal works fine.

Reproduced scenario:

1. The user has some other executable named `ocr` on PATH (in my case a tmux session helper at `~/.local/bin/ocr`) and that terminal ran it once before the install.
2. `npm i -g @alibaba-group/open-code-review` succeeds; the real `ocr` is linked into the npm prefix, which is *earlier* on PATH.
3. Running `ocr config provider` in the same terminal still executes the old program:

   ```
   ❌ 找不到编号或名称为 'config' 的项目。
   ```

   New terminals work fine.

## Root cause

Shells cache command locations (the bash/zsh hash table). Once a session has resolved `ocr` to some path, it keeps using that cached path even after the real `ocr` appears earlier on PATH — only new shells re-resolve. (Verified with a minimal zsh reproduction: same-window run hits the stale cached binary; `rehash` fixes it.) A foreign `ocr` that precedes the install on PATH is the worse variant: it wins in *every* terminal.

The installers previously printed "OpenCodeReview is ready!" without verifying that the `ocr` command actually resolves to this installation, so users got no hint that a different program was answering their first command.

## Fix

All three installers now verify how `ocr` resolves right after installing and print actionable guidance:

- **npm postinstall** (`scripts/install.js`): classifies the resolution and warns on
  - `stale-hash-risk` — ours wins on PATH, but another `ocr` exists that open terminals may have cached → advises `hash -r` / `rehash` / opening a new terminal (the reported bug);
  - `shadowed` — a foreign executable wins in every terminal → advises removing it or reordering PATH;
  - `not-on-path` — global install whose bin dir is missing from PATH, also reporting who occupies the command name, if anything.
  Windows handling follows `PATHEXT` order (like cmd.exe) and compares paths case-insensitively.
- **`install.sh` / `install.ps1`**: the same detection for the shell/PowerShell install flows.

Example output for the reported scenario (validated against the real environment):

```
[WARN]  Another executable named 'ocr' exists on your PATH: /Users/x/.local/bin/ocr.
[WARN]  Terminals opened before this install may have cached that location and keep
[WARN]  running it instead of OpenCodeReview (new terminals are not affected).
[WARN]  If 'ocr' misbehaves in your current terminal, refresh the shell's command cache:
[WARN]    bash/zsh: hash -r    (zsh also: rehash)    or simply open a new terminal
[WARN]  Verify with: ocr version   (expected output: open-code-review vX.Y.Z ...)
```

Note: an installer subprocess cannot clear the parent shell's hash table, so detection plus precise guidance at install time is the fix.

## Tests

- New `scripts/install.test.js` (plain Node + assert, node >= 14, mirroring `check-translation-sync.test.js`): 15 tests covering PATH walking, ours/foreign classification (symlink + npm shim content), PATHEXT ordering, Windows case-insensitivity, and all classification statuses. Wired as `npm run test:install` and a scoped `install-scripts` CI workflow (node:24.18.0 container, mirroring `translation-sync.yml`).
- `install.sh` notice logic verified in all four scenarios (ours-first-with-lookalike / shadowed / not-on-PATH / clean) under `sh -n` and `set -eu`.
- `make check` and `scripts/verify-license.sh` pass; existing `test:github-actions` tests unaffected.
- The change was reviewed with `ocr review --audience agent` and all findings were addressed (single-pass classification, Windows path casing, PATHEXT semantics, trailing-slash PATH entries, workflow concurrency).

### DIFF
--- a/.github/workflows/install-scripts.yml
+++ b/.github/workflows/install-scripts.yml
@@ -1,0 +1,39 @@
+name: install-scripts
+
+# Tests the npm installer's post-install PATH checks (scripts/install.js):
+# shadowed or stale-hash-prone `ocr` resolutions must be detected and
+# reported at install time. Kept out of ci.yml (Go-only container),
+# mirroring the scoped translation-sync.yml workflow.
+
+on:
+  pull_request:
+    paths:
+      - "scripts/install.js"
+      - "scripts/install.test.js"
+      - "scripts/platform.js"
+      - "package.json"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  install-scripts:
+    runs-on: self-hosted
+    timeout-minutes: 10
+    container:
+      # Pinned to an exact 24.x for reproducible builds, matching the
+      # translation-sync.yml pin (avoids the mutable node:24 tag).
+      image: node:24.18.0
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Trust workspace
+        run: git config --global --replace-all safe.directory '*'
+
+      # Plain Node, no deps; runs on node >= 14 (engines field).
+      - name: Test installer PATH-resolution checks
+        run: node scripts/install.test.js
--- a/install.ps1
+++ b/install.ps1
@@ -76,8 +76,27 @@ function Show-PostInstallPathNotice([string]$BinName, [string]$InstallDir) {
         Write-Host "note: $InstallDir is not on your PATH; add it or run $InstallDir\$BinName directly"
         return
     }
-    if (-not (Get-Command $BinName -ErrorAction SilentlyContinue)) {
-        Write-Host "note: open a new shell so $BinName resolves on PATH"
+    # Resolve the bare name so lookalikes (ocr.cmd, ocr.bat, ...) are caught,
+    # matching what the user's shell would actually execute.
+    $bareName = [System.IO.Path]::GetFileNameWithoutExtension($BinName)
+    $expected = Join-Path $InstallDir $BinName
+    $resolved = @(Get-Command $bareName -All -ErrorAction SilentlyContinue)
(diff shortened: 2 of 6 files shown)


wu21-web:
@Cooooper Thank you for auditing this fix, however you didn't use the pull request template, nor does any issue matches this PR, and the issue cannot be reproduced on aarch64 macos 15  or x86 linux 5.10. Please specify the target host with this issue, and edit the description to the use the template.
Cannot reproduce on: Cloud Shell `Linux 5f1e712a5086 5.10.134-19.7.1.al8.x86_64 #1 SMP Fri Jul 17 17:44:41 CST 2026 x86_64 x86_64 x86_64 GNU/Linux`
Consider opening a ticket/issue first before using an agent to fix this, thank you!

https://github.com/user-attachments/assets/85aa5622-fc7c-4c3a-bf1b-0601e8f9ea60

.github/workflows/install-scripts.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  install-scripts:
+    runs-on: self-hosted
+    timeout-minutes: 10
+    container:
+      # Pinned to an exact 24.x for reproducible builds, matching the
+      # translation-sync.yml pin (avoids the mutable node:24 tag).
+      image: node:24.18.0
+    steps:
+      - uses: actions/checkout@v4
```
wu21-web, line 32:
```suggestion
      - uses: actions/checkout@820762786026740c76f36085b0efc47a31fe5020
```
This action version is outdated, upgrade to keep in sync with ci.yml:24, if #836 is merged do this

wu21-web (changes requested):

lizhengfeng101:
Closing this for now due to the prolonged failure to sign the CLA and the lack of response.

PR #852, opened 2026-08-11
feat(cli): add --output flag to write review/scan results to a file

### LINKED ISSUE #851: feat(cli): add --output <path> flag to write results directly to a file
## Background

When `ocr` is driven by an agent (Claude Code, Codex, or any tool-calling agent), large `review`/`scan` runs emit JSON results on stdout. Host agents hard-truncate tool output that exceeds their buffer (limits differ per host, e.g. ~30K chars for Claude Code's Bash tool), which corrupts the JSON mid-stream. The only workarounds today are:

1. Shell redirection to a file (`ocr ... > result.json`), which is fragile across shells — e.g. Windows PowerShell 5's `>` writes UTF-16 and breaks JSON parsing.
2. Re-extracting via `ocr session comments <session-id> --json` after the fact. This works (sessions are fully persisted to `~/.opencodereview/sessions/<repo>/<session-id>.jsonl`), but it is an extra step and requires the agent to know the session id.

## Proposal

Add a `--output <path>` (`-o`) flag to `ocr review` and `ocr scan` (and optionally `ocr session comments`):

- Writes the exact same output that would go to stdout (text or JSON, same encoder) directly to a UTF-8 file.
- `--output -` (or omitting the flag) preserves current stdout behavior, so nothing breaks.
- Print a short notice to **stderr** when writing to a file, e.g. `[ocr] Results written to /path/to/result.json` — stderr is small and never truncated, so agents can reliably discover the path and read the full file with a file-reading tool.

## Why

- **Reliable full output for agents**: file I/O is not subject to the host tool-output buffer, so agents always get the complete JSON without truncation or shell-encoding surprises.
- **Host-agnostic**: no dependence on a particular agent's buffer size or redirection behavior.
- **Simple and backward compatible**: mirrors common CLI conventions (`--output -` for stdout); `text` format unaffected.

## Related

The skill docs currently work around this by recommending shell redirection plus `ocr session comments` recovery (see `skills/open-code-review`); a native `--output` would let the docs recommend the robust path instead.

Happy to implement if the maintainers are open to it.

lidaixingchen:
### Summary

Adds a `--output <path>` / `-o` flag to `ocr review` and `ocr scan` that writes the result (JSON or text) directly to a UTF-8 file, solving the host-agent tool-buffer truncation problem for agent-driven review.

Closes [#851](https://github.com/alibaba/open-code-review/issues/851) — `feat(cli): add --output <path> flag to write results directly to a file`.

When `--output` is set, `ocr` prints `[ocr] Results written to <path>` on **stderr** — stderr is never truncated by host tool buffers, so agents can reliably discover the path and read the full result with a file-reading tool.

### Design

- **`--output <path>` / `-o`**: writes the result to the file; `--output -` (or omitted) keeps current stdout behavior — fully backward compatible.
- **Lazy write**: the file is created only on first Write (via `lazyFileWriter`), never upfront — if the run fails or is interrupted, a pre-existing target file is left untouched instead of being truncated. Fail-fast pre-checks (target is a directory / parent missing) error out with a non-zero exit before any write.
- **strip-ANSI for text mode**: when `--format text --output <file>`, an ANSI-stripping state machine (CSI/OSC/single-ESC, with cross-Write chunk buffering) removes terminal color codes so the file stays readable; stdout behavior and `--format json` output are unchanged (JSON has no ANSI).
- **Error channel untouched**: run-level failure JSON (`emitFailureUsage`) and text-mode warnings stay on **stderr**; the result file holds only the review content.
- **Write failures propagate**: file create/write/close errors are joined into the command's error and exit non-zero.

### Testing

- New tests: `resolveOutputWriter` (stdout/directory/missing-parent branches), `lazyFileWriter` (no-write leaves existing file untouched, no-write creates nothing, write creates file + stderr hint, text strips ANSI, JSON bytes unchanged, create failure), `stripAnsiWriter` state machine (passthrough, CSI, truecolor, OSC, OSC+ST, multi-byte chunk boundaries).
- Adapted all ~27 `emitRunResult` call sites in `emit_run_result_test.go` / `budget_output_test.go` to the new `out` parameter.
- Flag parsing tests for `--output`, `-o`, and `-o -`.
- Full `go build ./...` passes; all `--output`-related tests pass.

### Notes

- Docs: `docs/design/output-flag.md` records the design and decisions (lazy write, ANSI stripping, error-channel policy, first-phase scope — session/delegate outputs excluded).
- Pre-existing Windows test failures (`TestSaveConfig` perm 666/600, provider TUI rollback tests) are environment-specific (Windows lacks Unix permission semantics) and reproduce on `main` unmodified; not introduced by this PR.

> Authored by 历代星辰


wu21-web:
I wonder if a pipe redirect can do the same job

lidaixingchen:
> I wonder if a pipe redirect can do the same job

Good question — `>` can cover the simple interactive case, but the flag exists because a redirect falls short in three ways:

1. **`>` truncates on launch.** The shell truncates the target file the moment the command starts (O_TRUNC), so a failed or interrupted review that never produces output still wipes an existing file to 0 bytes. `--output` uses a lazy writer that creates the file only on the first successful write — a failed run leaves the disk untouched.
2. **Encoding & ANSI.** `--output` writes UTF-8 and, for text format, strips ANSI escape sequences so the result file is clean. A plain `>` on Windows PowerShell 5.1 redirects with a non-UTF-8 default encoding and keeps the color codes.
3. **The real use case is a host agent, not a shell.** The flag exists because when `ocr` runs as a subprocess of an LLM host agent, stdout gets truncated once the output is long, and there's no shell to insert a pipe. `--output` lets the agent read results from a file while keeping stdout clean — and never leaves an empty file behind.

So for interactive shells `>` is usually fine; `--output` just handles those edge cases for the same price.

lizhengfeng101:
@lidaixingchen merge conflicts

lidaixingchen:
@lizhengfeng101 Conflicts resolved and merged with the latest upstream/main. All tests and checks have passed.

lizhengfeng101:
Nice work on the lazy-write design — not truncating an existing file on a failed run is the right call, and the ANSI state machine handles chunk splits more carefully than I expected. A few things to sort out first:

**1. The tiktoken file picked up CRLF.** `internal/llm/bpe_data/cl100k_base.tiktoken` gained 100,256 `\r` bytes (1,681,126 → 1,781,382). Functionally it's harmless — `parseBpeData` does `TrimSpace`, and I checked that both versions parse to an identical map — but it's embedded via `go:embed`, so every binary grows ~100KB, and it blows the PR diff up to 3.6MB. I read the diff twice before realizing SARIF and the retry report weren't actually yours. AGENTS.md also asks for LF explicitly. Since `.gitattributes` marks `*.tiktoken binary`, `--renormalize` won't fix it — needs `git checkout origin/main -- internal/llm/bpe_data/cl100k_base.tiktoken`.

**2. ANSI stripping gets bypassed on an unrecognized `--format`.** `emitRunResult` renders text for anything that isn't `json` or `sarif`, but `resolveOutputWriter` only sets `strip` when the format is exactly `"text"`. Since `review`/`scan` never validate `--format` (only `delegate` does, shared_flags.go:162), `--format txt -o out.txt` writes raw escapes to the file. I reproduced it: `text` → clean, `txt` and `TEXT` → escapes in the file. Easiest fix is `strip: !isMachineReadable(format)` so both sites use one predicate — and probably worth validating `--format` on review/scan too, since silently falling back to text is surprising on its own.

**3. Docs.** `--output` isn't in `pages/src/content/docs/{en,zh,ja,ru}/cli-reference.md`, which is where `--format`/`--audience`/`--preview`/`--exclude` all live. Also, the PR description points at `docs/design/output-flag.md`, but that file isn't in the diff and isn't in the repo — looks like it didn't get committed.

**4. An unterminated OSC can silently eat the rest of the file.** Once the stripper sees `ESC ]` it discards everything until BEL or ST. Comment content goes through `sanitizeTerminal` so it's fine, but the project summary doesn't:

```go
fmt.Fprintf(out, "\n\n──────── Project Summary ────────\n\n%s\n", summary)
```

That's LLM-generated text on the `scan` path (`internal/scan/agent.go:115`). A stray `ESC ]` in there used to just garble a terminal; with the stripper in place it truncates the output file instead. Running `sanitizeTerminal` on the summary would cover it, or you could cap the OSC buffer and fall back to passing bytes through.

Everything builds and the full test suite passes for me, and the `stripAnsiWriter` tests are genuinely thorough — 1 and 2 are the blockers.

lidaixingchen:
Thanks for the thorough review and insightful feedback! All points have been resolved:

1. **Clean tiktoken restored**: Reverted `internal/llm/bpe_data/cl100k_base.tiktoken` directly from `upstream/main` to restore the clean LF binary asset, dropping the bloated CRLF diff.
2. **Safe-by-default format handling & validation**:
   - Added `validateOutputFormat` to `validateReviewOptions` and `validateScanOptions`, rejecting unrecognized formats fail-fast and normalizing valid formats (e.g. `JSON` / padded values) to lowercase.
   - Updated `resolveOutputWriter` to use `strip: !isMachineReadable(format)`, ensuring any non-machine-readable format strips ANSI escape sequences when writing to files.
3. **Documentation added**:
   - Added `--output / -o` documentation and options tables across all localized CLI references (`pages/src/content/docs/{en,zh,ja,ru}/cli-reference.md`).
   - Synced `--output` usage examples to `README.md` and all localized versions (`zh-CN`, `ja-JP`, `ko-KR`, `ru-RU`).
4. **ANSI state machine runaway & truncation defenses**:
   - Sanitized Project Summary output in `emitRunResult` with `sanitizeTerminal(summary)` at the presentation layer.
   - Enhanced `stripAnsiWriter` with safe fallback flushing:
     - **OSC cap**: Added `maxOSCSequenceLength = 4096`; oversized or unterminated OSC sequences flush pending bytes as normal text instead of silently swallowing subsequent stream content.
     - **CSI & ESC intermediate bytes cap**: Added `maxCSISequenceLength = 256` and invalid byte detection (`< 0x20` or non-ASCII `>= 0x80` such as UTF-8 multibyte characters), safely flushing on invalid characters or sequence overflow.
     - Normalized `outputFormat` across `emitRunResult` and `outputPreview` to guarantee consistent dispatch.

All unit tests pass and new test cases covering all fallback paths and flag validation have been added.

lizhengfeng101 (approved):
LGTM

PR #853, opened 2026-08-12
fix(diff): match gitignore directory patterns

RerankerGuo:
## Description

Directory-only `.gitignore` patterns were matched by comparing the complete
pattern body with individual path components. This worked for a literal
basename such as `vendor/`, but not for path patterns or globs:

- `docs/generated/` did not exclude `docs/generated/file.go`;
- `**/generated/` did not exclude generated directories at any depth;
- `build*/` did not exclude matching directory names;
- `/generated/` did not exclude the root directory.

This change matches directory-only patterns against ancestor directories:

- patterns containing `/` match complete ancestor paths relative to the
  repository root;
- slashless patterns match ancestor directory names at any depth;
- leading `/` remains root-anchored;
- the existing doublestar matcher handles glob and globstar syntax.

The file's final path segment is intentionally excluded from consideration, so
a file named `generated` is not treated as a directory.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] `make check`
- [x] `make test`
- [x] `make build`
- [x] `make coverage` - 90.2%, meeting the 90% threshold
- [x] Compared all eight regression cases with `git check-ignore --no-index`
- [x] Manual CLI preview in a repository containing a tracked
  `docs/generated/file.go`

Before the fix, five new regression cases failed. After the fix, all directory
pattern cases pass and match Git's result. The CLI preview excludes the tracked
generated file while retaining a normal modified file.

Self-review used OCR Delegation Mode with the host agent:
`1/1` reviewable production files reviewed, `0` skipped, `0` findings.

## Checklist

- [x] Documentation is not required for this internal matcher fix
- [x] I have signed the CLA

## Related Issues

No existing issue found. Open issue and pull request text, plus all open pull
request changed files, were checked before implementation.

### DIFF
--- a/internal/diff/gitignore_test.go
+++ b/internal/diff/gitignore_test.go
@@ -84,6 +84,34 @@ func TestIsPathExcluded(t *testing.T) {
 	}
 }
 
+func TestIsPathExcluded_DirectoryPatterns(t *testing.T) {
+	tests := []struct {
+		name    string
+		relPath string
+		pattern string
+		want    bool
+	}{
+		{"path pattern", "docs/generated/file.go", "docs/generated/", true},
+		{"path pattern is root relative", "nested/docs/generated/file.go", "docs/generated/", false},
+		{"globstar at root", "generated/file.go", "**/generated/", true},
+		{"globstar nested", "src/generated/file.go", "**/generated/", true},
+		{"component glob", "src/build-cache/file.go", "build*/", true},
+		{"root anchored", "generated/file.go", "/generated/", true},
+		{"root anchored does not match nested", "src/generated/file.go", "/generated/", false},
+		{"file name is not a directory", "src/generated", "generated/", false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := IsPathExcluded(".", tt.relPath, []string{tt.pattern})
+			if got != tt.want {
+				t.Errorf("IsPathExcluded(%q, %q) = %v, want %v",
+					tt.relPath, tt.pattern, got, tt.want)
+			}
+		})
+	}
+}
+
 // allowListGitignore is the "ignore everything, then re-include" idiom from
 // github/gitignore's Go.AllowList.gitignore. Repositories using it are the
 // reason negation patterns cannot be discarded: the leading `*` matches every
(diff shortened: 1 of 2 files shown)


lizhengfeng101 (approved):
Really nice work here! The fix is clean and well-scoped — you clearly dug into how Git actually resolves these patterns rather than just patching the symptom.

A few things I appreciated:

- The `lastSlash < 0` early-return is a subtle but important guard. Easy to miss, glad you didn't.
- Splitting the `matchFullPath` logic by anchored-or-contains-slash vs. bare-name is exactly right per Git's spec, and the doublestar library does the heavy lifting naturally.
- The test matrix covers the cases that matter — especially "file name is not a directory", which is the kind of edge case that bites people six months later.

One thing worth noting (not blocking): the non-negated `*/` pattern was previously a silent no-op (since no segment literally equals `*`), and now it actually works as Git intends. Unlikely to matter in practice since bare `*/` without negation is rare, but it's a behavioral change worth being aware of.

Overall this is solid — careful analysis, minimal diff, good tests. Ship it! 🚢

internal/diff/gitignore_test.go
```diff
@@ -84,6 +84,34 @@ func TestIsPathExcluded(t *testing.T) {
...
+		name    string
+		relPath string
+		pattern string
+		want    bool
+	}{
+		{"path pattern", "docs/generated/file.go", "docs/generated/", true},
+		{"path pattern is root relative", "nested/docs/generated/file.go", "docs/generated/", false},
+		{"globstar at root", "generated/file.go", "**/generated/", true},
+		{"globstar nested", "src/generated/file.go", "**/generated/", true},
+		{"component glob", "src/build-cache/file.go", "build*/", true},
```
wu21-web, line 98:
```suggestion
		{"component glob", "src/build-cache/file.go", "build*/", true},
		{"wildcard directory", "src/file.go", "*/", true},
		{"deep slashless match", "a/b/generated/file.go", "generated/", true},
		{"slash-containing glob", "docs/v2/generated/file.go", "docs/*/generated/", true},
```

- wildcard directory verifies the `*/` behavior specifically changed by this implementation.
- deep slashless match proves slashless patterns inspect every ancestor, not only the immediate parent.
- slash-containing glob proves glob matching works against complete root-relative ancestor paths.

wu21-web (approved):

PR #855, opened 2026-08-12
fix(agent): name the main-loop stop trigger in item failure reasons

### LINKED ISSUE #842: Failed review items cannot say why they stopped: StopEmptyRounds and StopCompression both surface as "main task stopped before completing"
### What happens

When a per-file review stops without completing, the main loop records why (`internal/llmloop/loop.go`):

- `loop.go:288` — `StopEmptyRounds` (“Too many empty retries”)
- `loop.go:299` — `StopCompression` (“Context compression exceeded threshold”)

But `classifyMainLoopStop` collapses both into one opaque string (`internal/agent/agent.go:1015`):

```go
func classifyMainLoopStop(stop llmloop.MainLoopStop) (session.FailureClass, string) {
	switch stop {
	case llmloop.StopMaxRounds:
		return session.FailureBudget, "reached the maximum tool-request rounds without finishing"
	default: // StopEmptyRounds, StopCompression, StopNone
		return session.FailureUnknown, "main task stopped before completing"
	}
}
```

so the JSON result and run manifest report every such item identically:

```json
"failed": [{ "path": "...", "classification": "unknown", "reason": "main task stopped before completing" }]
```

### Why this matters in CI

The two exits call for opposite responses — empty rounds point at the model or the tool loop; a compression stop points at the context budget. But in `--format json` runs the only lines that name the trigger (`loop.go:287`, `loop.go:298`) are written to `stdout.Writer()` after `stdout.Quiet()` has swapped it to `io.Discard` (`internal/stdout/stdout.go:36`, via `cmd/opencodereview/shared.go:275`), and the session checkpoint that keeps the detail never leaves the machine. On an ephemeral CI runner nothing that survives the run says which stop fired.

This is how I hit it: a CI review pipeline had the same file fail on every run with exactly this string in the uploaded result artifact, and there was no way to tell from the artifacts whether to look at the model or at the context budget.

### Reproduction

Deterministic from the code above: both `StopEmptyRounds` and `StopCompression` fall into the `default:` arm, so any run in which an item exits through either guard produces byte-identical `failed[]` entries. Running the same review in terminal mode shows the two distinct `[ocr]` messages that JSON mode discards.

### Suggested direction

Keep the `FailureUnknown` class (no taxonomy change — the comment’s point that these are unclassifiable as *classes* stands), but let the reason string name the trigger, mirroring the `StopMaxRounds` precedent:

- `StopEmptyRounds` → `"stopped after repeated rounds without a usable tool result"`
- `StopCompression` → `"stopped because context compression exceeded its threshold"`
- `StopNone` (non-completed) keeps the current catch-all.

The stop enum already reaches `classifyMainLoopStop`, so this is a two-line change plus tests. Happy to send the PR if the direction sounds right.

Fanzzzd:
Fixes #842.

When a per-file review stops without completing, `classifyMainLoopStop` collapsed `StopEmptyRounds`, `StopCompression` and `StopNone` into one opaque reason, `"main task stopped before completing"`. The two named stops call for opposite responses (model/tool-loop trouble vs. context budget), and in `--format json` runs the `[ocr]` progress lines that would say which one fired are discarded by `stdout.Quiet()` — so on an ephemeral CI runner the manifest reason is the only stop diagnostic that survives, and it said the same thing for every exit. In practice that meant a recurring per-file failure could not be triaged from uploaded artifacts at all (how #842 was found).

**Change** — exactly the two-line shape proposed in the issue, keeping the `FailureUnknown` class untouched:

- `StopEmptyRounds` → `"stopped after repeated rounds without a usable tool result"`
- `StopCompression` → `"stopped because context compression exceeded its threshold"`
- `StopNone` (non-completed) keeps the existing catch-all string, and `StopMaxRounds` is unchanged.

**Test** — `TestClassifyMainLoopStop` (beside the existing `TestClassifyItemError` it mirrors) pins the full stop→(class, reason) mapping and additionally asserts no two stops share a reason string, so a future regression back into a collapsed message fails the table even if the strings are edited.

No behavior change beyond the reason text: classes, exit codes and manifest shape are untouched. Verified with `go vet` and the full `internal/agent` test suite.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1123,14 +1123,20 @@ func classifyItemError(err error) (session.FailureClass, string) {
 
 // classifyMainLoopStop maps a non-error, non-completed main-loop stop to an item
 // failure class and a safe reason. Only the configured max-tool-request budget is
-// a declared budget stop; the empty-round and compression exits are genuine but
-// unclassifiable, so they map to the honest unknown catch-all. Only an explicit
-// budget trigger may use the budget classification.
+// a declared budget stop, so only it may use the budget classification. The
+// empty-round and compression exits keep the unknown class — the FailureClass
+// taxonomy has no fitting category — but the reason names the trigger: in
+// --format json runs the progress lines that would say why an item stopped are
+// discarded, so this string is the only thing that leaves a CI runner.
 func classifyMainLoopStop(stop llmloop.MainLoopStop) (session.FailureClass, string) {
 	switch stop {
 	case llmloop.StopMaxRounds:
 		return session.FailureBudget, "reached the maximum tool-request rounds without finishing"
-	default: // StopEmptyRounds, StopCompression, StopNone
+	case llmloop.StopEmptyRounds:
+		return session.FailureUnknown, "stopped after repeated rounds without a usable tool result"
+	case llmloop.StopCompression:
+		return session.FailureUnknown, "stopped because context compression exceeded its threshold"
+	default: // StopNone
 		return session.FailureUnknown, "main task stopped before completing"
 	}
 }
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1123,22 +1123,21 @@ func classifyItemError(err error) (session.FailureClass, string) {
 
 // classifyMainLoopStop maps a non-error, non-completed main-loop stop to an item
 // failure class and a safe reason. Only the configured max-tool-request budget is
-// a declared budget stop, so only it may use the budget classification. The
-// empty-round and compression exits keep the unknown class — the FailureClass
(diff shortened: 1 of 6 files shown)


lizhengfeng101:
Pushed a follow-up commit to this branch: I moved the reason strings onto `llmloop.MainLoopStop.Reason()` so the scan path (`internal/scan/agent.go`, which discarded the stop and hardcoded a second opaque sentence) renders the same text instead of drifting from review, and so a constant added to the enum later can no longer inherit the collapsed catch-all. Your `TestClassifyMainLoopStop` is unchanged and still passes.

lizhengfeng101 (approved):
LGTM

PR #856, opened 2026-08-12
fix(ci): pin workflow action references to full commit SHAs

### LINKED ISSUE #841: Release workflow references actions by floating tags while holding npm and attestation credentials
## Problem

Follow-up from [#836 review](https://github.com/alibaba/open-code-review/pull/836#pullrequestreview-3550463356): `.github/workflows/release.yml` references third-party actions by floating tags while running with the repository's most sensitive credentials — `secrets.NPM_TOKEN` plus `id-token`/`attestations` permissions for provenance. A compromised or hijacked tag there has a wider blast radius than the same drift in CI: it can exfiltrate the npm token or publish tampered artifacts with valid-looking provenance.

Current floating references (`main`):

| Line | Reference |
|---|---|
| 33, 74, 171 | `actions/checkout@v7` |
| 55 | `actions/upload-artifact@v4` |
| 139, 179 | `actions/download-artifact@v4` |
| 148 | `softprops/action-gh-release@v3` |
| 156 | `actions/attest-build-provenance@v4` |

`softprops/action-gh-release` deserves particular care: it is a third-party (non-`actions/*`) action running in the job that holds release permissions.

## Suggestion

1. Pin every `uses:` in `release.yml` to a full commit SHA with the `# vX.Y.Z` comment, same convention as #836 established for `action.yml`.
2. Add `.github/workflows/release.yml` to the `files` array in `scripts/verify-action-pins.sh` so the invariant is CI-enforced.
3. Optionally extend to the remaining workflows later; `release.yml` first because of the credential exposure.

Happy to send the PR once #836 lands (it builds on the same check script).

Fanzzzd:
Fixes #841. Extends the pin discipline #836 established for `action.yml` to the repository's own workflows — most importantly `release.yml`, which holds the npm token and the attestation id-token, so a moved upstream tag there had credential blast radius.

**Pins.** All 21 external `uses:` references across the 8 workflow files are now full 40-hex commit SHAs with strict `# vX.Y.Z` comments. Resolution was done with [pinact](https://github.com/suzuki-shunsuke/pinact) and then every SHA was independently cross-checked against the commit its version tag resolves to via the GitHub API (12 unique pins, all verified, including the 4 pre-existing `action.yml` pins). No majors were changed — `translation-sync.yml` stays on checkout v4 (pinned to what `v4` resolves to today) — so behavior is byte-identical; only the tags can no longer move.

**Checker coverage.** `verify-action-pins.sh` now includes `.github/workflows/*.yml` in its file set, so CI fails closed on any future floating reference in either the composite action or a workflow.

**One transparency note.** This PR also restores the hardening wu21-web requested in #836's review — full-line anchoring with a strict version comment, the directive pre-filter for commented lines, and the `grep -n` prefix strip. I pushed that rework to the #836 branch as `37fc06f`, but it turns out that push landed **eight minutes after** the merge commit (merge `7e52a4f` at 07:36Z, push at 07:44Z), so main carries the pre-review version of the script. Extending coverage to workflows makes the difference immediately visible: without the pre-filter the check false-positives on `codeql.yml`'s commented `setup-example` example, and without the anchors a floating tag with a pinned-looking fragment in a trailing comment would pass. Sorry for not catching the race at the time — the reviewed checking logic lands here unchanged (only the header comment and the file list differ, as described above), plus one follow-up commit from a post-push adversarial review: flow-mapping `- {uses: …}` entries now fail closed instead of being silently skipped, and the file set covers `*.yaml` workflows too.

**Verified:**
- clean run: `All external action references in action.yml .github/workflows/*.yml are SHA-pinned`
- negative: floating `actions/cache@v6` → exit 1, loose `# v6` comment → exit 1, commented example lines → ignored
- `gh api repos/<owner>/<repo>/commits/<vX.Y.Z>` equals the pinned SHA for all 12 unique pins

cc @wu21-web — this carries your #836 review hardening across the merge race, plus the workflow coverage we discussed in #841.

### DIFF
--- a/scripts/verify-action-pins.sh
+++ b/scripts/verify-action-pins.sh
@@ -16,7 +16,12 @@ set -euo pipefail
 
 cd "$(dirname "$0")/.."
 
-files=("action.yml" .github/workflows/*.yml)
+# nullglob so an absent *.yaml matchless glob vanishes instead of
+# surviving as a literal (which the missing-file guard would then fail on).
+# GitHub executes both .yml and .yaml workflow files.
+shopt -s nullglob
+files=("action.yml" .github/workflows/*.yml .github/workflows/*.yaml)
+shopt -u nullglob
 # The whole line must be a pinned reference: only list/indent syntax before
 # `uses:`, a 40-hex SHA, and a strict `# vX.Y.Z` comment with nothing after
 # it. Without the anchors a floating tag would be accepted whenever a
@@ -24,7 +29,10 @@ files=("action.yml" .github/workflows/*.yml)
 # trailing comment), and a `# v7` comment would satisfy the format the
 # check claims to enforce. Anything unusual (quoted values, extra trailing
 # content) fails closed rather than being guessed at.
-directive='^[[:space:]]*(-[[:space:]]+)?uses:'
+# The directive filter also matches flow-mapping openers ("- {uses: …}"),
+# which the pinned pattern below never accepts — so flow-style entries fail
+# closed as unusual syntax instead of being silently skipped.
+directive='^[[:space:]]*(-[[:space:]]+)?(\{[[:space:]]*)?uses:'
 pinned='^[[:space:]]*(-[[:space:]]+)?uses:[[:space:]]+[A-Za-z0-9_.-]+/[A-Za-z0-9_./-]+@[0-9a-f]{40}[[:space:]]+#[[:space:]]*v[0-9]+\.[0-9]+\.[0-9]+[[:space:]]*$'
 local_ref='^[[:space:]]*(-[[:space:]]+)?uses:[[:space:]]*\./'
 
--- a/scripts/verify-action-pins.sh
+++ b/scripts/verify-action-pins.sh
@@ -4,18 +4,29 @@
 # Copyright 2026 alibaba/open-code-review Contributors
 
 # Verify that every external action referenced by the published composite
-# action (action.yml) is pinned to a full 40-hex commit SHA with a trailing
-# "# vX.Y.Z" version comment. A floating tag inside action.yml silently
-# undermines consumers who SHA-pin alibaba/open-code-review itself: the
-# outer pin freezes this repository, but a moved inner tag still changes
-# what actually runs (see issue #816).
+# action (action.yml) and by this repository's workflows is pinned to a
+# full 40-hex commit SHA with a trailing "# vX.Y.Z" version comment. A
+# floating tag inside action.yml silently undermines consumers who SHA-pin
+# alibaba/open-code-review itself: the outer pin freezes this repository,
+# but a moved inner tag still changes what actually runs (see issue #816).
+# Workflows are held to the same rule because several of them hold
+# credentials — release.yml alone carries the npm token and the
+# attestation id-token (see issue #841).
 set -euo pipefail
 
 cd "$(dirname "$0")/.."
 
-files=("action.yml")
-pinned='uses:[[:space:]]*[A-Za-z0-9_.-]+/[A-Za-z0-9_./-]+@[0-9a-f]{40}[[:space:]]+#[[:space:]]*v[0-9]'
-local_ref='uses:[[:space:]]*\./'
+files=("action.yml" .github/workflows/*.yml)
+# The whole line must be a pinned reference: only list/indent syntax before
+# `uses:`, a 40-hex SHA, and a strict `# vX.Y.Z` comment with nothing after
+# it. Without the anchors a floating tag would be accepted whenever a
+# pinned-looking fragment appeared anywhere in the line (e.g. inside a
+# trailing comment), and a `# v7` comment would satisfy the format the
+# check claims to enforce. Anything unusual (quoted values, extra trailing
+# content) fails closed rather than being guessed at.
+directive='^[[:space:]]*(-[[:space:]]+)?uses:'
+pinned='^[[:space:]]*(-[[:space:]]+)?uses:[[:space:]]+[A-Za-z0-9_.-]+/[A-Za-z0-9_./-]+@[0-9a-f]{40}[[:space:]]+#[[:space:]]*v[0-9]+\.[0-9]+\.[0-9]+[[:space:]]*$'
+local_ref='^[[:space:]]*(-[[:space:]]+)?uses:[[:space:]]*\./'
 
 bad=""
 for file in "${files[@]}"; do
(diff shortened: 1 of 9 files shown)


wu21-web:
It is thoughtful of you to pin these action SHAs. However I’m not convinced that mandatory SHA pinning should be extended to all repository workflows. Major-version tags such as `@v4` and `@v7` are substantially clearer and easier to maintain, while the practical risk of an official action tag such as `actions/checkout@v7` being maliciously moved appears extremely low.

In my view, restricted permissions, trusted actions, protected release jobs, and automated dependency updates provide a more proportionate security approach here. I would prefer keeping the existing composite-action policy separate and not expanding its pin checker to `.github/workflows/`.

wu21-web (changes requested):
@Fanzzzd I think the full shell script to verify action pins should be removed entirely, also from the workflow. You can continue to pin SHAs for actions though. A maintainer can simplify toggle "Requiring SHA-Pinning for all actions" in repo settings and that is far more secure than a shell script. The runner won't even start without a pinned SHA.

Fanzzzd:
Thanks for thinking this through so carefully. +1 on the repo-level SHA-pinning toggle regardless of what happens here — it's real protection for this repo's workflows, and the `uses: ./` false positive it used to have was [fixed in November](https://github.com/orgs/community/discussions/170337).

Two small facts for the maintainers' decision, then I'm happy with whatever is chosen:

- The platform policy isn't documented to reach the references **inside the published `action.yml`** — the part that protects downstream consumers, which is what the checker has guarded since #836. And `release.yml` only executes at release time, so a floating ref there surfaces exactly then; the script catches it at PR time, before the credentials are in play.
- On maintenance: this repo's `dependabot.yml` already has the `github-actions` ecosystem enabled, and Dependabot bumps pinned SHAs and their version comments automatically — so after this one-time change the ongoing cost is the weekly Dependabot review that already happens.

Happy to keep the checker as-is, narrow it to `action.yml` + `release.yml`, or drop the workflow extension in favor of the toggle — maintainers' call. The pins themselves seem uncontroversial either way.

(On the bot's translation-sync note: staying on checkout v4 there is deliberate — this PR pins what's present without changing majors, so it's behavior-neutral. Happy to bump that major in a separate PR.)

wu21-web:
Thanks. I take back my comment on maintenance since dependabot is enabled. Either way is OK, and your shell script look fine, changes seem correct, but I lean towards a simpler switch. Anyway, let us wait for the maintainer's response.

scripts/verify-action-pins.sh
```diff
@@ -4,32 +4,54 @@
...
 
-files=("action.yml")
-pinned='uses:[[:space:]]*[A-Za-z0-9_.-]+/[A-Za-z0-9_./-]+@[0-9a-f]{40}[[:space:]]+#[[:space:]]*v[0-9]'
-local_ref='uses:[[:space:]]*\./'
+# nullglob so an absent *.yaml matchless glob vanishes instead of
+# surviving as a literal (which the missing-file guard would then fail on).
+# GitHub executes both .yml and .yaml workflow files.
+shopt -s nullglob
+files=("action.yml" .github/workflows/*.yml .github/workflows/*.yaml)
+shopt -u nullglob
```
wu21-web:
```suggestion
files=("action.yml")
```

scripts/verify-action-pins.sh
```diff
@@ -4,32 +4,54 @@
...
 set -euo pipefail
 
 cd "$(dirname "$0")/.."
 
-files=("action.yml")
-pinned='uses:[[:space:]]*[A-Za-z0-9_.-]+/[A-Za-z0-9_./-]+@[0-9a-f]{40}[[:space:]]+#[[:space:]]*v[0-9]'
-local_ref='uses:[[:space:]]*\./'
+# nullglob so an absent *.yaml matchless glob vanishes instead of
+# surviving as a literal (which the missing-file guard would then fail on).
+# GitHub executes both .yml and .yaml workflow files.
```
wu21-web:
```suggestion

```

wu21-web (commented):
These can be simplified with "Require actions to be pinned to a full-length commit SHA". I take back my comment about linting `.github/workflows/*`.
> [!CAUTION]
> Accepting these changes below needs "Require actions to be pinned to a full-length commit SHA" to be enabled in `Actions > General` to achieve the same effect.

Fanzzzd:
Both suggestions applied as-is in 2f38dfb, with your `Co-authored-by` (I also reverted the header sentence about workflows that the change made stale). Kept the `action.yml` check because that's the one layer the [repo toggle](https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/) can't see — nested `uses:` in a published composite only run in *consumers'* workflows, which is why e.g. [trivy-action pins internally](https://github.com/aquasecurity/trivy-action/blob/master/action.yaml); [Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies) likewise only scans `.github/workflows/*`.

For whoever merges, per the CAUTION note: please enable **Settings → Actions → General → "Require actions to be pinned to a full-length commit SHA"**. (If PR-time workflow linting is ever wanted back, [zizmor](https://docs.zizmor.sh/audits/#unpinned-uses) is the standard tool.)

wu21-web (approved):
Great. ChatGPT reported no offenses, and also no issues found when I reviewed it myself. Let us wait the maintainer to respond.
> [!CAUTION]
> Accepting these changes below needs "Require actions to be pinned to a full-length commit SHA" to be enabled in `Actions > General` to achieve the same effect.

scripts/verify-action-pins.sh
```diff
@@ -14,22 +14,36 @@ set -euo pipefail
...
+# `uses:`, a 40-hex SHA, and a strict `# vX.Y.Z` comment with nothing after
+# it. Without the anchors a floating tag would be accepted whenever a
+# pinned-looking fragment appeared anywhere in the line (e.g. inside a
+# trailing comment), and a `# v7` comment would satisfy the format the
+# check claims to enforce. Anything unusual (quoted values, extra trailing
+# content) fails closed rather than being guessed at.
+# The directive filter also matches flow-mapping openers ("- {uses: …}"),
+# which the pinned pattern below never accepts — so flow-style entries fail
+# closed as unusual syntax instead of being silently skipped.
+directive='^[[:space:]]*(-[[:space:]]+)?(\{[[:space:]]*)?uses:'
```
wu21-web, line 27: Minor:
I think this suggestion prevent sth like this:
`- "uses": owner/action@v1`
where "uses" is actually quoted       --> Valid YAML

```suggestion
directive="^[[:space:]]*(-[[:space:]]+)?(\\{[[:space:]]*)?['\"]?uses['\"]?[[:space:]]*:"
```

Not really familiar with regexes, so ask an AI before applying this.

wu21-web: About the quotes too, they should be either single or double to align the format (with the comments above).

lizhengfeng101:

lizhengfeng101 (approved):
LGTM

PR #859, opened 2026-08-12
fix: make FileFilter patterns case-insensitive

### LINKED ISSUE #844: Include/exclude patterns containing an uppercase letter never match: FileFilter lowercases the path but not the pattern
### What happens

`FileFilter.IsUserExcluded` and `FileFilter.IsUserIncluded` lowercase the **path** but not the **pattern** before calling `doublestar.Match` (`internal/config/rules/system_rules.go:222-251`):

```go
func (f *FileFilter) IsUserExcluded(path string) bool {
	lowerPath := strings.ToLower(path)
	for _, pattern := range f.Exclude {
		expanded := expandBraces(pattern)
		for _, p := range expanded {
			if matched, _ := doublestar.Match(p, lowerPath); matched {
```

So any include/exclude pattern containing an uppercase letter can never match anything — it is compared against an already-lowercased path. The rule-resolution paths got this right: `Resolve`, `resolveDetail` and `matchProjectRuleEntry` all lowercase **both** sides (`doublestar.Match(strings.ToLower(p), lowerPath)`, e.g. `system_rules.go:135`), which was the point of #75. `FileFilter` implements the same case-insensitive contract only halfway.

### Reproduction

Unit-level (drop into `internal/config/rules`, fails on `main`):

```go
func TestFileFilterCaseAsymmetry(t *testing.T) {
	f := &FileFilter{Exclude: []string{"README.md"}, Include: []string{"**/CHANGELOG.md"}}
	if !f.IsUserExcluded("README.md") {
		t.Error(`exclude pattern "README.md" does not match path "README.md"`)
	}
	if !f.IsUserIncluded("docs/CHANGELOG.md") {
		t.Error(`include pattern "**/CHANGELOG.md" does not match path "docs/CHANGELOG.md"`)
	}
}
```

```
--- FAIL: TestFileFilterCaseAsymmetry
    exclude pattern "README.md" does not match path "README.md"
    include pattern "**/CHANGELOG.md" does not match path "docs/CHANGELOG.md"
```

CLI-level, in any repo whose diff touches a file with an uppercase letter in its name (here `seedVolume.ts`):

```
$ ocr review --from A --to B --exclude "packages/backend/convex/seedVolume.ts"
[ocr] 8 file(s) changed, reviewing 6 ...        # not excluded

$ ocr review --from A --to B --exclude "packages/backend/convex/seedvolume.ts"
[ocr] 8 file(s) changed, reviewing 5 ...
[ocr] Skipping packages/backend/convex/seedVolume.ts — filtered by path/extension rules
```

The same pattern spelled in lowercase works; spelled the way the file is actually named, it is silently ignored. Wildcards mask the bug (`seed*.ts` works because the literal part happens to be lowercase), which is probably why it has survived: common patterns like `**/*.test.ts` are unaffected, while `**/CHANGELOG.md`, `README.md` or an exact `--exclude path/to/SomeFile.ts` silently do nothing — no error, no warning, the file is just reviewed anyway.

### Suggested fix

Lowercase the expanded pattern in both methods, matching what `Resolve` already does:

```go
if matched, _ := doublestar.Match(strings.ToLower(p), lowerPath); matched {
```

Two lines plus a regression test. Happy to send the PR.

Itachi7011:
## Summary

Fixes #844.

`FileFilter.IsUserIncluded` and `FileFilter.IsUserExcluded` lowercased
the file path before matching, but did not lowercase the user-provided
pattern.

This caused patterns containing uppercase letters such as
`README.md` and `**/CHANGELOG.md` to fail to match.

## Changes

- Lowercase expanded exclude patterns before `doublestar.Match`
- Lowercase expanded include patterns before `doublestar.Match`
- Extended `TestFileFilter_CaseInsensitive` with uppercase-pattern
  regression coverage

## Testing

```text
go test ./internal/config/rules/...

### DIFF
--- a/internal/config/rules/system_rules.go
+++ b/internal/config/rules/system_rules.go
@@ -219,6 +219,7 @@ func (f *FileFilter) HasInclude() bool {
 }
 
 // IsUserExcluded reports whether the given path matches any user exclude pattern.
+// The check is case-insensitive: both path and pattern are lowercased.
 func (f *FileFilter) IsUserExcluded(path string) bool {
 	lowerPath := strings.ToLower(path)
 	for _, pattern := range f.Exclude {
@@ -233,6 +234,7 @@ func (f *FileFilter) IsUserExcluded(path string) bool {
 }
 
 // IsUserIncluded reports whether the given path matches any user include pattern.
+// The check is case-insensitive: both path and pattern are lowercased.
 // Returns false when Include is empty (no user include restriction defined).
 func (f *FileFilter) IsUserIncluded(path string) bool {
 	if !f.HasInclude() {
--- a/internal/config/rules/system_rules.go
+++ b/internal/config/rules/system_rules.go
@@ -224,7 +224,7 @@ func (f *FileFilter) IsUserExcluded(path string) bool {
 	for _, pattern := range f.Exclude {
 		expanded := expandBraces(pattern)
 		for _, p := range expanded {
-			if matched, _ := doublestar.Match(p, lowerPath); matched {
+			if matched, _ := doublestar.Match(strings.ToLower(p), lowerPath); matched {
 				return true
 			}
 		}
@@ -242,7 +242,7 @@ func (f *FileFilter) IsUserIncluded(path string) bool {
 	for _, pattern := range f.Include {
 		expanded := expandBraces(pattern)
 		for _, p := range expanded {
-			if matched, _ := doublestar.Match(p, lowerPath); matched {
+			if matched, _ := doublestar.Match(strings.ToLower(p), lowerPath); matched {
 				return true
 			}
 		}
--- a/internal/config/rules/system_rules_test.go
+++ b/internal/config/rules/system_rules_test.go
@@ -678,7 +678,7 @@ func TestFileFilter_IsUserIncluded_EmptyInclude(t *testing.T) {
 func TestFileFilter_CaseInsensitive(t *testing.T) {
 	f := &FileFilter{
 		Include: []string{"src/**/*.java", "**/CHANGELOG.md"},
-		Exclude: []string{"**/generated/**", "README.md"},
+		Exclude: []string{"**/generated/**", "README.md", "**/*.{Go,Java}"},
 	}
 
 	if !f.IsUserIncluded("SRC/Main/Foo.Java") {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by wu21-web — APPROVED** [review]
No issues found.

**Comment by lizhengfeng101** [comment]
@Itachi7011 Nice first contribution — the fix is exactly right, and I appreciate that you kept it minimal instead of refactoring around it. I pulled the branch locally: tests pass, and I confirmed the new assertions genuinely fail once `system_rules.go` is reverted, so the regression coverage is real. I also grepped every `doublestar.Match` call site — these two were the only remaining asymmetric ones, so nothing was missed (`allowlist/allowed_ext.go` lowercases its patterns at init, and `diff/git.go` is case-sensitive on purpose since that's gitignore semantics).

Three small suggestions, none of them blocking:

**1. Use mixed-case paths in the test.** The new assertions pair an uppercase pattern with an all-lowercase path:

```go
if !f.IsUserIncluded("docs/changelog.md") { ... }
if !f.IsUserExcluded("readme.md") { ... }
```

That does catch the regression, but it sidesteps the spelling people actually hit — pattern *and* path both containing uppercase, e.g. `--exclude packages/backend/convex/seedVolume.ts` against a file named `seedVolume.ts`. The repro in #844 is written that way. The real-world spelling still fails before the fix, and additionally pins the user-facing case:

```go
if !f.IsUserIncluded("docs/CHANGELOG.md") { ... }
if !f.IsUserExcluded("README.md") { ... }
```

**2. Add one brace-expansion case.** The only interaction still untested is the *ordering* of `ToLower` and `expandBraces`. You went with expand-then-lowercase — correct, and it mirrors `Resolve` — but nothing currently locks that in. One extra pattern would:

```go
Exclude: []string{"**/*.{Go,Java}"},  // assert "pkg/Main.JAVA" is excluded
```

**3. State the contract in the doc comments.** Neither `IsUserExcluded` nor `IsUserIncluded` mentions being case-insensitive, whereas `allowlist.IsExcludedPath` spells it out ("The check is case-insensitive."). Given this bug was precisely a half-implemented version of that unwritten contract, one line would help keep it from drifting back:

```go
// IsUserExcluded reports whether the given path matches any user exclude pattern.
// The check is case-insensitive: both path and pattern are lowercased.
```

Thanks again — clean, well-scoped patch.

Itachi7011:
Thanks for the detailed review! I appreciate the suggestions. I'll keep these in mind. Glad to hear the fix and regression coverage look good. 🙏

lizhengfeng101 (changes requested):
@Itachi7011  I'd like to see your updates regarding these three small suggestions.

lizhengfeng101 (approved):
LGTM

Itachi7011:
Thanks for the review! I've addressed all three suggestions:

- Updated the regression cases to use mixed-case paths.
- Added brace-expansion coverage with an uppercase pattern.
- Documented the case-insensitive behavior in both FileFilter methods.

I also ran `go test ./...` locally, and all tests pass.

Fanzzzd:
Congrats on landing your first contribution, @Itachi7011 — clean fix, regression test included, merged within a day. Exactly how it should go. Welcome aboard 🎉

PR #865, opened 2026-08-12
feat(git): validate if PATH git version is safe git

wu21-web:
## Description
Simply prints a warning to the user if their git in path version is lower than 2.41

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Compiled it on alibaba cloud linux with has git 2.17.1
```
shell@Alicloud:~/open-code-review$ make build
go build -ldflags "-X main.Version=v0.0.0-7e59d53 -X main.GitCommit=7e59d53 -X main.BuildDate=2026-08-12T11:53:57Z" -o ./dist/opencodereview ./cmd/opencodereview

shell@Alicloud:~/open-code-review$ 
shell@Alicloud:~/open-code-review$ cd dist/
shell@Alicloud:~/open-code-review/dist$ ls
opencodereview
shell@Alicloud:~/open-code-review/dist$ ./opencodereview config provider
warning: git 2.17.1 is older than the minimum supported version 2.41.0. Some commands may not work correctly; consider upgrading git.
Cancelled.
shell@Alicloud:~/open-code-review/dist$ 
```

## Checklist

- [x] I have signed the CLA

## Related Issues

Discussion at: https://github.com/alibaba/open-code-review/discussions/860


lizhengfeng101 (commented):
Thanks for adding the git version check — this is a nice quality-of-life improvement for users on older distros.

A few things I'd love to see cleaned up before merging:

### 1. Who's responsible for printing the warning?

Right now there's a split: `checkGitVersion` writes the "too old" warning directly to stderr itself, but `main.go` handles printing for other error cases (git not found, unparseable output). This means a reader looking at `main.go` sees `!errors.Is(err, ErrGitVersionTooOld)` and has to wonder "wait, why are we silently swallowing the most important error?" — you have to go read the internals to realize it was already printed.

I'd suggest letting `CheckGitVersion` just return the version and error without touching stderr at all, and have `main.go` do all the printing:

```go
ver, err := gitcmd.CheckGitVersion()
if err != nil {
    fmt.Fprintf(os.Stderr, "warning: %v\n", err)
}
```

One place does the I/O, one place does the logic — easier to reason about.

### 2. Every invocation pays for a subprocess spawn

The check runs before `rootCmd.Execute()`, so even `opencodereview --help` or `opencodereview --version` forks a `git --version` process. That's a small but unnecessary latency hit for commands that don't touch git at all.

Would be cleaner to move this into a `PersistentPreRunE` hook on the root command (or only on subcommands that actually need git).

### 3. Returning both a value and an error feels surprising

When the version is too old, `checkGitVersion` returns *both* the version string and an `ErrGitVersionTooOld` — that goes against the usual Go convention where `(value, error)` are mutually exclusive. Callers don't normally expect to use the first return value when `err != nil`.

If you need the version available on the error path, consider embedding it in the error itself:

```go
type VersionTooOldError struct {
    Current GitVersion
    Minimum GitVersion
}

func (e *VersionTooOldError) Error() string {
    return fmt.Sprintf("git %s is older than the minimum supported version %s", e.Current, e.Minimum)
}
```

That way callers who care can `errors.As` into it, and callers who don't just see a nice error message.

lizhengfeng101:
`commandNeedsGit` matches on the leaf `cmd.Name()`, so the git check never fires for `delegate preview`/`delegate rule` (which actually use git) yet fires on the no-op `delegate`/`rules` help commands — please match the top-level command and drop `rules` (it doesn't use git).

lizhengfeng101:
@wu21-web

wu21-web:
I know. I will do it as soon as possible. Currently facing academic issues.

cmd/opencodereview/root.go
```diff
@@ -55,6 +65,23 @@ func init() {
 	rootCmd.AddCommand(completionCmd)
 }
 
+func commandNeedsGit(cmd *cobra.Command) bool {
+	// `ocr --version` / `-V` is handled by the root command's RunE.
+	if v, _ := cmd.Flags().GetBool("version"); v {
+		return false
+	}
```
lizhengfeng101: Nit (non-blocking): dead code. `version` is a root-only local flag, so `ocr --version` already has `cmd == root` and hits the `switch` default (`false`); `review`/`scan`/`delegate` have no `version` flag. Removing it keeps `TestCommandNeedsGit_RootVersionFlag` green — it then covers the default branch.

```suggestion
```

lizhengfeng101 (approved):
LGTM

PR #872, opened 2026-08-12
feat(llmloop): add grace round after tool-request budget exhausted

lizhengfeng101:
## Summary

When the LLM tool-request budget (`MaxToolRequestTimes`) is exhausted during file review, the model may have identified issues but not yet submitted them via `code_comment`. Previously these findings were silently lost.

This PR adds a **grace round**: after the main loop exits on budget, one final LLM call is made with only `code_comment` and `task_done` available. The model can either submit pending findings or cleanly exit — preventing false-positive pressure since `task_done` provides an opt-out.

## Changes

- `runGraceRound`: new method on `Runner` that appends a budget-exhausted notice to the conversation and performs a single LLM call with restricted tools
- `graceRoundToolDefs`: filters `MainToolDefs` to only `code_comment` + `task_done`
- Token usage from the grace round is counted in the runner's totals
- Only triggers on `StopMaxRounds` (not on empty-round or compression stops)

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` all green
- [x] Manual verification: set `MAX_TOOL_REQUEST_TIMES` to 3, confirmed grace round fires and comment is collected

### DIFF
--- a/internal/llmloop/loop.go
+++ b/internal/llmloop/loop.go
@@ -303,8 +303,68 @@ func (r *Runner) RunPerFile(ctx context.Context, messages []llm.Message, newPath
 
 	if stop == StopMaxRounds {
 		fmt.Fprintf(stdout.Writer(), "[ocr] Max tool requests reached for %s.\n", newPath)
+		r.runGraceRound(ctx, messages, newPath, sessionID)
 	}
 	return false, stop, nil
+}
+
+// runGraceRound performs one final LLM call after the tool-request budget is
+// exhausted, giving the model a chance to submit any findings it identified
+// but did not yet report via code_comment.
+func (r *Runner) runGraceRound(ctx context.Context, messages []llm.Message, newPath string, sessionID string) {
+	graceDefs := graceRoundToolDefs(r.deps.MainToolDefs)
+	if len(graceDefs) == 0 {
+		return
+	}
+
+	messages = append(messages, llm.NewTextMessage("user",
+		"Your tool-call budget is exhausted. This is your FINAL round. You may ONLY:\n"+
+			"- Call code_comment to submit any findings you have identified but not yet reported.\n"+
+			"- Call task_done if you have nothing more to report.\n"+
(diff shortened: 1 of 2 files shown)


internal/llmloop/loop.go
```diff
@@ -303,10 +303,76 @@ func (r *Runner) RunPerFile(ctx context.Context, messages []llm.Message, newPath
...
+
+	if ctx.Err() != nil {
+		fmt.Fprintf(stdout.Writer(), "[ocr] Grace round skipped for %s: context cancelled\n", newPath)
+		return
+	}
+
+	resp, err := r.deps.LLMClient.CompletionsWithCtx(ctx, llm.ChatRequest{
+		Model:     r.deps.Model,
+		Messages:  messages,
+		Tools:     graceDefs,
```
stay-foolish-forever, line 334: Restricting the tool list for this final call will likely cost us the whole KV cache prefix.

Placing the tool list in or next to the system prompt is the mainstream practice among current LLMs, precisely so that the prefix cache keeps hitting across a session. Swapping `r.deps.MainToolDefs` for the two-tool subset rewrites that part of the prefix, so everything behind it is invalidated.

The timing makes it expensive: `runGraceRound` only fires on `StopMaxRounds`, i.e. when the conversation has accumulated a full budget's worth of `tool_result` file contents and is at its longest. We re-read that prompt at full price and write a fresh cache entry that nothing downstream will ever read. The new `totalCacheReadTokens`/`totalCacheWriteTokens` counters should show it directly: read ≈ 0, write ≈ the whole prompt.

That said, there's a genuine trade-off here between effectiveness and cost. Narrowing the tool list is a hard constraint, whereas leaving `MainToolDefs` intact and relying on the prose instruction (optionally filtering `resp.ToolCalls()` on the response side) is only a soft one — the model could spend its last turn on a read tool. If the temporary override does measurably improve how often the grace round lands a real finding, then this may simply be a price we have to pay.

stay-foolish-forever (approved):

PR #884, opened 2026-08-13
feat(actions): render review metadata as Shields badges

### LINKED ISSUE #882: OCR Reviews on Github Actions Support Badges
### Problem Statement

I hope the [bug] and other tags the github-actions[bot] produce is a badge, like other code review tools.
A [bug] shields.io badge will be great visually.

Image

### Proposed Solution

At runtime, split the first line (badges) into a array, and for each badge text, if there is a matching badge url, replace it with the badge url instead.
**This is NOT the final solution.**

### Alternatives Considered

Not yet...

### Affected Area

Output / Formatting

### Additional Context

_No response_

Xvvln:
## Summary

- render structured review `category` and `severity` metadata as one standard `flat` Shields badge in GitHub Actions comments
- keep the existing plain-text `buildBadge()` output unchanged for CLI compatibility and safe fallback behavior
- preserve the hidden `ocr-id` marker and use the same graphical rendering for inline comments and summary fallbacks

## Design

The GitHub-only renderer uses the static Shields endpoint format suggested in #882:

```text
https://img.shields.io/badge/<CATEGORY>-<SEVERITY>-<COLOR>?style=flat&labelColor=24292f
```

Only known schema enums are allowed into the URL. Missing or unknown metadata falls back to the existing text badge instead of generating an image URL. The severity colors follow the requested progression: low is green, medium is orange, high is red, and critical is dark red. The Markdown alt text retains both metadata values for image-disabled and assistive clients.

This reads the structured fields directly and does not parse or split the comment body.

## Validation

- `npm run test:github-actions`
- `make test`
- `make check`
- all 8 category x 4 severity combinations covered by unit tests
- all four severity badge URLs verified to return `200 image/svg+xml`
- rendered through GitHub's Markdown API and confirmed to use the Camo image proxy

Fixes #882

### DIFF
--- a/scripts/github-actions/post-review-comments.js
+++ b/scripts/github-actions/post-review-comments.js
@@ -61,6 +61,22 @@ const SEVERITY_RANK = new Map(
   SEVERITIES.map((s, i) => [s, SEVERITIES.length - i])
 ); // critical=4, high=3, medium=2, low=1
 
+// GitHub review comments support Markdown images but do not expose a native
+// severity-pill field to third-party Actions. Render the two structured
+// metadata dimensions as one compact Shields badge instead: category stays on
+// the neutral left segment and severity controls the right-segment color.
+// Values are fixed enums, not arbitrary model output, so no untrusted text is
+// interpolated into the image URL.
+const GITHUB_BADGE_LABEL_COLOR = "24292f";
+// Keep Shields' white text legible while following #882's requested semantic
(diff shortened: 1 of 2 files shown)


Xvvln:
Closing this draft to avoid duplicating the work of the current assignee, as discussed in #882. The implementation and test coverage will remain available here as a reference, and I will be happy to review the assignee’s PR when it is opened.

PR #885, opened 2026-08-13
feat(action): render category/severity badge as a shields.io image

### LINKED ISSUE #882: OCR Reviews on Github Actions Support Badges

chaojixinren:
Closes #882

## What

Replaces the plain-text badge line (`[bug · high]`) in GitHub Actions review and summary comments with a single static shields.io badge, using exactly the format suggested in the issue discussion:

`https://img.shields.io/badge/<category>-<severity>-<color>`

The color is keyed off severity, with `blue` used as a fixed fallback when the severity is missing or unrecognized:

| Severity | Color |
| --- | --- |
| low | green |
| medium | orange |
| high | red |
| critical | darkred |
| *(missing / unknown)* | blue |

## How

- Adds a GitHub Actions-only `buildBadgeImage()` in `scripts/github-actions/post-review-comments.js`; the CLI's plain-text `buildBadge()` output is untouched.
- The badge is built from the structured `category` / `severity` fields (validated OCR enums), not by parsing comment text.
- Any non-empty metadata renders a badge: both fields → `category-severity`, one field → a single segment. The color comes from the severity when known, otherwise a fixed `blue`, so no lookup can ever yield a broken `-undefined` URL and no non-empty finding falls back to plain text.
- Values are sanitized (control chars stripped) and URL-escaped before interpolation.
- The image alt text keeps the plain-text form (`bug · high`), so screen readers and image-load failures degrade to the previous output.
- The hidden `ocr-id` marker stays as the first line; badge placement in `formatComment` / `formatCommentMarkdown` is unchanged.
- No new dependencies, no custom badge rendering.

## Testing

- `npm run test:github-actions` — passes. `testBuildBadgeImage` covers the full severity-color matrix, single-field badges, unknown-value rendering, case-insensitive matching, and control-char sanitization; `testSeverityBadgeColorCoversSeverities` guards the enum ↔ color-map invariant.
- E2E: generated badge URLs for known, single-field, and unknown-value cases from the real code path and verified each returns `200 image/svg+xml` from shields.io; rendered the output through GitHub's Markdown API and confirmed the badges resolve.

### Screenshot
截屏2026-08-13 13 24 23

### DIFF
--- a/scripts/github-actions/post-review-comments.js
+++ b/scripts/github-actions/post-review-comments.js
@@ -1380,6 +1380,73 @@ function buildBadge(comment) {
   return "";
 }
 
+// Shields.io color per severity, following the reviewer-suggested scale in
+// #882 (low -> green, medium -> orange, high -> red). "critical" is not in
+// that three-level scale; it extends it one step hotter (darkred, a standard
+// CSS color name shields.io accepts). Keys match SEVERITIES.
+const SEVERITY_BADGE_COLOR = Object.freeze({
+  critical: "darkred",
+  high: "red",
+  medium: "orange",
+  low: "green",
+});
+// Fallback color when severity is missing or not a known enum member. The
+// reviewer's guidance was to render these rather than fall back to plain text,
+// with any color acceptable; blue reads as informational.
+const CATEGORY_BADGE_COLOR = "blue";
+
+// Escape a value for a shields.io static-badge path segment ("-" separates
+// segments, "_" renders as a space). Our enum values contain neither, but the
+// escape keeps the helper correct if the enums ever grow such a value.
+function shieldsEscape(value) {
+  return encodeURIComponent(value.replace(/-/g, "--").replace(/_/g, "__"));
+}
+
+// Escape Markdown image alt text. The alt is derived from the plain-text badge
+// whose metadata has had control chars stripped but may still carry
+// Markdown-special characters (e.g. a stray "]" from high-temperature model
+// output), which would prematurely close the image's [...] alt span. Escaping
+// "\", "[", and "]" keeps the image intact and the alt rendering to the literal
+// text. "[" is escaped defensively; "]" and "\" are the structural ones.
+function escapeMarkdownAlt(value) {
+  return value
+    .replace(/\\/g, "\\\\")
+    .replace(/\[/g, "\\[")
+    .replace(/\]/g, "\\]");
+}
+
+// Build the category/severity badge as a single shields.io static badge
+// (#882) for surfaces that render Markdown (GitHub review/summary comments),
+// in the format img.shields.io/badge/<label>-<color>. flat is shields.io's
+// default style, so the URL carries no style parameter. The image's alt text
+// is the plain-text badge content from buildBadge (with Markdown-special
+// characters escaped), so screen readers and image-load failures degrade to
+// the previous text form.
+//
+//   both fields -> ![category · severity](…/badge/category-severity-<color>)
+//   one field   -> ![field](…/badge/field-<color>)
+//   neither     -> "" (no badge line)
+//
+// The color comes from the severity when known, otherwise the fixed category
+// color — no lookup can ever yield "undefined". Values are sanitized (control
+// chars stripped) and URL-escaped before interpolation, so rendering an
+// unrecognized value is safe: it is only ever the same text already shown in
+// the plain-text badge. This is a deliberate divergence from strict CLI
+// byte-parity (I6): the CLI keeps the plain-text badge (a terminal cannot
+// render images); the alt text preserves the shared form across surfaces.
+function buildBadgeImage(comment) {
+  const text = buildBadge(comment);
+  if (text === "") return "";
+  const category = sanitizeMetadata(comment && comment.category).trim().toLowerCase();
+  const severity = sanitizeMetadata(comment && comment.severity).trim().toLowerCase();
+  const color = SEVERITY_BADGE_COLOR[severity] || CATEGORY_BADGE_COLOR;
+  const alt = escapeMarkdownAlt(text.slice(1, -1)); // "[bug · high]" -> "bug · high"
+  const label = category && severity
+    ? `${shieldsEscape(category)}-${shieldsEscape(severity)}`
+    : shieldsEscape(category || severity);
+  return `![${alt}](https://img.shields.io/badge/${label}-${color})`;
+}
+
 // Parse the publication policy from the raw opt-in inputs. Returns a normalized
 // policy object, or the NO_ROUTING sentinel when no routing is requested or any
 // value is malformed (fail-open for the policy itself, upholding I1).
@@ -1467,7 +1534,7 @@ function routeComment(comment, policy) {
 // as the first visible line. The code suggestion block is appended if present.
 function formatComment(comment, id) {
   let body = id ? `<!-- ${id} -->\n` : "";
-  const badge = buildBadge(comment);
+  const badge = buildBadgeImage(comment);
   if (badge) body += `${badge}\n`;
   body += comment.content || "";
   if (comment.suggestion_code && comment.existing_code) {
@@ -1483,7 +1550,7 @@ function formatCommentMarkdown(comment, error) {
   // with formatComment), so the heading/reason lines still anchor the comment
   // and existing substring assertions on them are unaffected. The badge is ""
   // for any finding without category/severity metadata.
-  const badge = buildBadge(comment);
+  const badge = buildBadgeImage(comment);
   if (badge) md += `${badge}\n`;
   md += `### 📄 \`${comment.path}\``;
   if (comment.start_line && comment.end_line) {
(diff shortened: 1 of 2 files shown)


chaojixinren:
@Xvvln This PR is now open — your review would be very welcome, as discussed in #882.

scripts/github-actions/post-review-comments.js
```diff
@@ -1380,6 +1380,53 @@ function buildBadge(comment) {
...
+// Shields.io color per severity, following the reviewer-suggested scale in
+// #882 (low -> green, medium -> orange, high -> red). "critical" is not in
+// that three-level scale; it extends it one step hotter (darkred, a standard
+// CSS color name shields.io accepts). Keys match SEVERITIES.
+const SEVERITY_BADGE_COLOR = Object.freeze({
+  critical: "darkred",
+  high: "red",
+  medium: "orange",
+  low: "green",
+});
```
wu21-web, line 1392:
```suggestion
const SEVERITY_BADGE_COLOR = Object.freeze({
  critical: "darkred",
  high: "red",
  medium: "orange",
  moderate: "orange",
  low: "green",
});
```

It will be great to build a big map here, so nothing gets missed.

scripts/github-actions/post-review-comments.js
```diff
@@ -1380,6 +1380,56 @@ function buildBadge(comment) {
...
+  const text = buildBadge(comment);
+  if (text === "") return "";
+  const category = sanitizeMetadata(comment && comment.category).trim().toLowerCase();
+  const severity = sanitizeMetadata(comment && comment.severity).trim().toLowerCase();
+  // Gate rendering on the color map itself (not SEVERITY_RANK) so a severity
+  // added to SEVERITIES without a color entry falls back to plain text
+  // instead of producing a broken "-undefined" shields.io URL.
+  if (!CATEGORIES.includes(category) || !Object.prototype.hasOwnProperty.call(SEVERITY_BADGE_COLOR, severity)) return text;
+  const alt = text.slice(1, -1); // "[bug · high]" -> "bug · high"
+  const url = `https://img.shields.io/badge/${shieldsEscape(category)}-${shieldsEscape(severity)}-${SEVERITY_BADGE_COLOR[severity]}`;
```
wu21-web:
```suggestion
  const color = SEVERITY_BADGE_COLOR[severity];
  if (!CATEGORIES.includes(category) || !SEVERITY_RANK.has(severity) || !color) {
    return text;
  }

  const alt = text.slice(1, -1); // "[bug · high]" -> "bug · high"
  const url = `https://img.shields.io/badge/${shieldsEscape(category)}-${shieldsEscape(severity)}-${color}`;
```
Prevents a future recognized severity without a configured color from generating a broken badge URL

wu21-web (commented):
Thank you for your effort. I not quite familiar with javascript, so cannot review your pull request deep. The logic is sound however.

scripts/github-actions/post-review-comments.js
wu21-web, line 1: Actually, you don't have to falback to nothing.
Because this is also valid:
`https://img.shields.io/badge/<title>-<COLOR CODE>`

The color can be randomly choosed from a array.

scripts/github-actions/post-review-comments.js
wu21-web: > [!NOTE]
> The keys and values are not limited to `moderate: "orange"`.

scripts/github-actions/post-review-comments.js
wu21-web: Related: https://github.com/alibaba/open-code-review/pull/885#discussion_r3772491653

scripts/github-actions/post-review-comments.js
wu21-web: Screenshot 2026-08-13 at 1 53 21 PM

scripts/github-actions/post-review-comments.js
chaojixinren: Good catch on wanting nothing to slip through — but `moderate` isn't actually one of the severity values. The enum is fixed to four levels in both the schema (`internal/config/toolsconfig/tools.json`) and the Go constants (`internal/tool/code_comment.go`): `critical`, `high`, `medium`, `low`. Adding `moderate` would let an unrecognized value through the guard, so I'd rather not include it.

The "nothing gets missed" concern itself is covered by a drift-guard test I added: it asserts every member of `SEVERITIES` has a color entry, and that a severity missing its color degrades to plain text instead of producing a `-undefined` URL.

scripts/github-actions/post-review-comments.test.js
```diff
@@ -1714,6 +1714,79 @@ function testBuildBadgeMatchesCliDegeneration() {
...
+    "![security · critical](https://img.shields.io/badge/security-critical-darkred)"
+  );
+  assert.strictEqual(
+    buildBadgeImage({ category: "performance", severity: "medium" }),
+    "![performance · medium](https://img.shields.io/badge/performance-medium-orange)"
+  );
+  assert.strictEqual(
+    buildBadgeImage({ category: "style", severity: "low" }),
+    "![style · low](https://img.shields.io/badge/style-low-green)"
+  );
```
wu21-web, line 1741:
```suggestion
  assert.strictEqual(
    buildBadgeImage({ category: "bug", severity: "high" }),
    "![bug · high](https://img.shields.io/badge/bug-high-red)"
  );
  assert.strictEqual(
    buildBadgeImage({ category: "security", severity: "critical" }),
    "![security · critical](https://img.shields.io/badge/security-critical-darkred)"
  );
  assert.strictEqual(
    buildBadgeImage({ category: "performance", severity: "medium" }),
    "![performance · medium](https://img.shields.io/badge/performance-medium-orange)"
  );
  assert.strictEqual(
    buildBadgeImage({ category: "style", severity: "low" }),
    "![style · low](https://img.shields.io/badge/style-low-green)"
  );
  assert.strictEqual(
    buildBadgeImage({ category: "style", severity: "moderate" }),
    "![style · low](https://img.shields.io/badge/style-moderate-orange)"
  );
```

As we discussed: https://github.com/alibaba/open-code-review/pull/885/changes#r3772497368
Don't forget to expand the test as well, @chaojixinren .

scripts/github-actions/post-review-comments.js
wu21-web: Agreed, `moderate: "orange"` is like a safe, just prevents the program from weird errors. Sometimes model don't strictly follow to instructions if the temperature is too high.

wu21-web: @chaojixinren I see, R46-R55 doesn't have the `moderate` option, now I understand. This and https://github.com/alibaba/open-code-review/pull/885#discussion_r3772538143 will be skipped, please resolve the threads.

wu21-web (approved):
No overall issues with the current implementation, good job!
Some missing badges for [documentation] don't really matter, having them will be better but it is optional anyway.

scripts/github-actions/post-review-comments.test.js
chaojixinren: Done, thanks!

wu21-web (approved):

chaojixinren:
> No overall issues with the current implementation, good job! Some missing badges for [documentation] don't really matter, having them will be better but it is optional anyway.

Thanks for the review, @wu21-web! I've addressed all your comments and updated the PR description accordingly.

* Added a safe fallback color so unknown severities still render a valid badge instead of producing `-undefined`.
* Any non-empty metadata now renders a badge, including single-field cases.
* Used a fixed blue fallback to keep the output deterministic and testable.
* Updated the severity map and expanded the tests as requested.

The bot's `-undefined` concern is covered by the same fallback-color change.

Would appreciate another look when you have a chance!

scripts/github-actions/post-review-comments.js
```diff
@@ -1380,6 +1380,59 @@ function buildBadge(comment) {
...
+// the plain-text badge. This is a deliberate divergence from strict CLI
+// byte-parity (I6): the CLI keeps the plain-text badge (a terminal cannot
+// render images); the alt text preserves the shared form across surfaces.
+function buildBadgeImage(comment) {
+  const text = buildBadge(comment);
+  if (text === "") return "";
+  const category = sanitizeMetadata(comment && comment.category).trim().toLowerCase();
+  const severity = sanitizeMetadata(comment && comment.severity).trim().toLowerCase();
+  const color = SEVERITY_BADGE_COLOR[severity] || CATEGORY_BADGE_COLOR;
+  const alt = text.slice(1, -1); // "[bug · high]" -> "bug · high"
```
wu21-web:
```suggestion
  const alt = text.slice(1, -1).replace(/\\/g, "\\\\").replace(/\[/g, "\\[").replace(/\]/g, "\\]");
```

This might be better and prevents malformed/injected Markdown resembling.

wu21-web (approved):

scripts/github-actions/post-review-comments.js
chaojixinren: Done!

wu21-web (approved):

stay-foolish-forever (approved):

PR #892, opened 2026-08-13
fix(cmd): report expected arguments when positional count is wrong

### LINKED ISSUE #890: fix(cli): argument-count errors don't name the missing argument or show usage
### OpenCodeReview Version

open-code-review v1.9.2 (5b37b5f) darwin/arm64 built at: 2026-08-12T06:41:47Z

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

Anthropic (Claude)

### Bug Description

<html><head></head><body><p>Leaf commands that accept positional arguments currently fall back to Cobra's raw argument-count error when the number of arguments is incorrect.</p><p>For example, <code inline="">ocr config set</code> requires both a key and a value. If the user forgets the value:</p><pre><code class="language-console">$ ocr config set provider
Error: accepts 2 arg(s), received 1
</code></pre><p>This message is technically correct, but not actionable. It does not tell the user:</p><ul><li><p>which argument is missing,</p></li><li><p>what the expected command form is,</p></li><li><p>whether there are known valid values,</p></li><li><p>or how to discover the correct usage.</p></li></ul><p>Because the root command uses <code inline="">SilenceUsage: true</code>, usage is not printed automatically. The error also does not suggest <code inline="">--help</code>, leaving the exit code as the only additional signal.</p><h3>Inconsistent with existing CLI error handling</h3><p>This falls below the usability standard already established for neighbouring CLI errors.</p><p>For example, misspelled flags are handled by <code inline="">flagErrorWithSuggestion</code> in <code inline="">cmd/opencodereview/flag_suggest.go</code>:</p><pre><code class="language-console">$ ocr review --forma json
Error: unknown flag: --forma

Did you mean this?
      --format
</code></pre><p>Argument-count errors should provide a similarly actionable response.</p><p>The necessary information is already available at the point of failure:</p><ul><li><p>every affected command declares its positional argument signature in <code inline="">Use</code>;</p></li><li><p><code inline="">ocr config set</code> also provides an <code inline="">Example</code>;</p></li><li><p><code inline="">ocr completion</code> declares:</p></li></ul><pre><code class="language-go">ValidArgs: []string{"bash", "zsh", "fish", "powershell"}
</code></pre><p>So, for commands with known valid positional values, that information could also be surfaced instead of returning Cobra's generic message.</p><h3>Affected commands</h3>
Command | Use field | Current error on wrong count
-- | -- | --
ocr config set | set <key> <value> | accepts 2 arg(s), received 1
ocr config unset | unset <key> | accepts 1 arg(s), received 0
ocr rules check | check [flags] <file-path> | accepts 1 arg(s), received 0
ocr session show | show [flags] <session-id> | accepts 1 arg(s), received 0
ocr session comments | comments [flags] <session-id> | accepts 1 arg(s), received 0
ocr delegate rule | rule [flags] <path...> | requires at least 1 arg(s), only received 0
ocr completion | completion <shell> | accepts 1 arg(s), received 0

<h3>Expected behavior</h3><p>When a leaf command receives the wrong number of positional arguments, the error should explain the expected invocation rather than exposing Cobra's raw validator message.</p><p>For example:</p><pre><code class="language-console">$ ocr config set provider
Error: missing required argument &lt;value&gt;

Usage:
  ocr config set &lt;key&gt; &lt;value&gt;
</code></pre><p>And where valid positional values are known:</p><pre><code class="language-console">$ ocr completion
Error: missing required argument &lt;shell&gt;

Valid values:
  bash, zsh, fish, powershell

Usage:
  ocr completion &lt;shell&gt;
</code></pre><p>The exact wording can vary, but the error should give users enough information to correct the command without having to guess or separately run <code inline="">--help</code>.</p><p>This is the same class of CLI error-consistency issue as #641, where parent commands silently printed help and exited with status 0.</p></body></html>

### Steps to Reproduce

```
# 1. Install v1.9.2 (npm -g) or build from source
ocr version   # open-code-review v1.9.2 (5b37b5f) darwin/arm64

# 2. Omit the value of a two-argument command
ocr config set provider
# Error: accepts 2 arg(s), received 1

# 3. Omit the key
ocr config unset
# Error: accepts 1 arg(s), received 0

# 4. Omit the session id
ocr session comments
# Error: accepts 1 arg(s), received 0

# 5. Omit the shell name, whose valid values are declared in ValidArgs
ocr completion
# Error: accepts 1 arg(s), received 0

# 6. Compare with the flag error path, which is helpful
ocr review --forma json
# Error: unknown flag: --forma
#
# Did you mean this?
#     --format
```

### Expected Behavior

Argument-count errors should identify the command, show the expected positional arguments, and point the user toward the correct invocation — in the same spirit as the existing flag suggestion behavior.

For example:

```console
$ ocr config set provider
Error: "ocr config set" requires 2 arguments (<key> <value>), but received 1

Usage:
  ocr config set <key> <value>

Example:
  ocr config set llm.model claude-opus-4-6
  ocr config set provider anthropic

Run 'ocr config set --help' for more information.
```

Where the command defines a known set of valid positional values, those should be surfaced as well:

```console
$ ocr completion
Error: "ocr completion" requires 1 argument (<shell>), but received 0

Valid values: bash, zsh, fish, powershell

Run 'ocr completion --help' for more information.
```

The exit code should remain `1`, and valid invocations should be unaffected.

### Suggested implementation

A natural implementation, consistent with the existing flag-error handling, would be a small helper in `cmd/opencodereview` that wraps `cobra.ExactArgs` and `cobra.MinimumNArgs`.

The helper can build the error from metadata the command already provides:

* the positional-argument signature from `Use`;
* `Example`, where present;
* `ValidArgs`, where present.

That keeps the behavior centralized instead of introducing seven command-specific error strings, and ensures the message stays accurate as command definitions evolve.

Happy to send a PR if this direction looks right.

### Logs / Error Output

```shell
$ ocr version
open-code-review v1.9.2 (5b37b5f) darwin/arm64
built at: 2026-08-12T06:41:47Z

$ ocr config set
Error: accepts 2 arg(s), received 0

$ ocr config set provider
Error: accepts 2 arg(s), received 1

$ ocr config set a b c
Error: accepts 2 arg(s), received 3

$ ocr config unset
Error: accepts 1 arg(s), received 0

$ ocr config unset a b
Error: accepts 1 arg(s), received 2

$ ocr rules check
Error: accepts 1 arg(s), received 0

$ ocr session show
Error: accepts 1 arg(s), received 0

$ ocr session comments
Error: accepts 1 arg(s), received 0

$ ocr delegate rule
Error: requires at least 1 arg(s), only received 0

$ ocr completion
Error: accepts 1 arg(s), received 0
```

### Additional Context

Reproduced with the released `v1.9.2` (`5b37b5f`) binary and confirmed with a source build of `main` at `27a820a`. The behavior is identical in both, so this is not build-specific.

Relevant code:

* `cmd/opencodereview/config_cmd.go:48,59` — `cobra.ExactArgs(2)` / `cobra.ExactArgs(1)`
* `cmd/opencodereview/rules_cmd.go:33` — `cobra.ExactArgs(1)`
* `cmd/opencodereview/session_cmd.go:51,67` — `cobra.ExactArgs(1)`
* `cmd/opencodereview/delegate_cmd.go:73` — `cobra.MinimumNArgs(1)`
* `cmd/opencodereview/completion.go:17-18` — `cobra.MatchAll(cobra.ExactArgs(1), cobra.OnlyValidArgs)`
* `cmd/opencodereview/flag_suggest.go` — existing friendlier handling for the analogous flag-error case
* `cmd/opencodereview/root.go:20-21` — `SilenceUsage` / `SilenceErrors`, which is why Cobra does not print usage automatically
* `cmd/opencodereview/main.go:25` — errors are ultimately printed as a bare `Error: %v`

Related: #641.

chaojixinren:
## What

Cobra's default message for a wrong positional-argument count names neither the command nor what it expects, and the root command sets `SilenceUsage`, so no usage block follows it:

```console
$ ocr config set provider
Error: accepts 2 arg(s), received 1
```

After:

```console
$ ocr config set provider
Error: "ocr config set" requires exactly 2 argument(s) (<key> <value>)

Usage:
  ocr config set <key> <value> [flags]

Example:
  ocr config set llm.model claude-opus-4-6
  ocr config set provider anthropic

Run 'ocr config set --help' for more information.
```

## How

`exactArgs` and `minimumArgs` are drop-in replacements for `cobra.ExactArgs` and `cobra.MinimumNArgs` that build the message from metadata the command already declares: the positional signature in `Use`, plus `Example` and `ValidArgs` where present.

The guidance therefore cannot drift from the command's own `--help` output, and no command carries a hand-written error string. This mirrors how the existing `flagErrorWithSuggestion` handles the analogous flag-error case.

The supplied count is deliberately not echoed back: it adds nothing the user cannot already see in the command they just typed.

Seven commands are wired: `config set`, `config unset`, `rules check`, `session show`, `session comments`, `delegate rule`, and `completion`.

Aliases report the canonical path (`ocr d rule` → `"ocr delegate rule"`), matching what `--help` prints.

## Scope

Message text only. Exit codes are unchanged, and every previously accepted argument count is still accepted.

I diffed the built binary against `main` across 29 invocations: 14 differ in text, 0 differ in exit code, and 15 are byte-identical. The byte-identical cases cover all `NoArgs` commands, unknown subcommands, `OnlyValidArgs` rejections, unknown flags, and all success paths.

Two adjacent gaps are deliberately left out to keep this reviewable and are worth follow-up issues:

1. The `OnlyValidArgs` error (`ocr completion xyz`) does not list valid values despite the command declaring them.
2. The 15 `NoArgs` commands hit the same dead end under `SilenceUsage`.

Both require a different mechanism than a count validator.

## Tests

`TestEveryPositionalCommandUsesFriendlyErrors` walks the command tree and fails if a command declaring positional placeholders still reports Cobra's raw message, so wiring a future command to `cobra.ExactArgs` directly is caught.

I validated this guard destructively: reverting `rules check` to `cobra.ExactArgs(1)` makes it fail as intended.

Verified locally against every CI gate: license headers, action pins, `gofmt -s`, line endings, `go mod tidy`, `go vet`, tests (23 packages, 0 failures), coverage at 90.3% against the 90% threshold, build, and the smoke test's 10 assertions.

`arg_errors.go` is at 100% per-function coverage.

I did not run `govulncheck` because there are no dependency changes and `go.mod` is untouched, or the cross-compile matrix because this is pure Go with no build tags or cgo.

Fixes #890

### DIFF
--- a/cmd/opencodereview/arg_errors.go
+++ b/cmd/opencodereview/arg_errors.go
@@ -1,0 +1,105 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 alibaba/open-code-review Contributors
+
+package main
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+// Positional-argument validators that replace Cobra's raw count message
+// ("accepts 2 arg(s), received 1") with an actionable one built from metadata
+// the command already declares — the positional signature in Use, plus Example
+// and ValidArgs where present — so the guidance cannot drift from the command's
+// own help output. Because the root command sets SilenceUsage, everything the
+// user sees has to travel in the error itself.
+
+// exactArgs behaves like cobra.ExactArgs(n) but reports a wrong count with the
+// command's own usage information.
+func exactArgs(n int) cobra.PositionalArgs {
+	return func(cmd *cobra.Command, args []string) error {
+		if len(args) == n {
+			return nil
+		}
+		return argCountError(cmd, fmt.Sprintf("requires exactly %d argument(s)", n))
+	}
+}
+
+// minimumArgs behaves like cobra.MinimumNArgs(n) but reports too few arguments
+// with the command's own usage information.
+func minimumArgs(n int) cobra.PositionalArgs {
+	return func(cmd *cobra.Command, args []string) error {
+		if len(args) >= n {
+			return nil
+		}
+		return argCountError(cmd, fmt.Sprintf("requires at least %d argument(s)", n))
+	}
+}
+
+// argCountError assembles the guidance for a wrong argument count. The count the
+// user supplied is deliberately not echoed back — it adds nothing they do not
+// already know from the line they just typed.
+func argCountError(cmd *cobra.Command, requirement string) error {
+	var b strings.Builder
+	b.Grow(256)
+
+	path := cmd.CommandPath()
+	fmt.Fprintf(&b, "%q %s", path, requirement)
+	if sig := positionalSignature(cmd); sig != "" {
+		fmt.Fprintf(&b, " (%s)", sig)
+	}
+
+	if len(cmd.ValidArgs) > 0 {
+		b.WriteString("\n\nValid values: ")
+		b.WriteString(strings.Join(validArgNames(cmd.ValidArgs), ", "))
+	}
+
+	b.WriteString("\n\nUsage:\n  ")
+	b.WriteString(cmd.UseLine())
+
+	if example := strings.TrimRight(cmd.Example, "\n"); example != "" {
+		b.WriteString("\n\nExample:\n")
+		b.WriteString(example)
+	}
+
+	b.WriteString("\n\nRun '")
+	b.WriteString(path)
+	b.WriteString(" --help' for more information.")
+
+	return errors.New(b.String())
+}
+
+// positionalSignature extracts the positional-argument part of cmd.Use, e.g.
+// "<key> <value>" from "set <key> <value>" and "<path...>" from
+// "rule [flags] <path...>". Bracketed fields are not placeholders the user
+// types, so "[flags]" and inline enumerations such as "[bash|zsh]" are dropped;
+// enumerated values are reported from ValidArgs instead.
+func positionalSignature(cmd *cobra.Command) string {
+	fields := strings.Fields(cmd.Use)
+	if len(fields) < 2 {
+		return ""
+	}
+
+	parts := make([]string, 0, len(fields)-1)
+	for _, f := range fields[1:] {
+		if strings.HasPrefix(f, "<") {
+			parts = append(parts, f)
+		}
+	}
+	return strings.Join(parts, " ")
+}
+
+// validArgNames strips the "name\tdescription" form Cobra allows in ValidArgs,
+// keeping only the value the user is expected to type. The split matches the one
+// cobra.OnlyValidArgs applies, so the listed values are exactly those accepted.
+func validArgNames(valid []string) []string {
+	names := make([]string, 0, len(valid))
+	for _, v := range valid {
+		names = append(names, strings.SplitN(v, "\t", 2)[0])
+	}
+	return names
+}
--- a/cmd/opencodereview/arg_errors_test.go
+++ b/cmd/opencodereview/arg_errors_test.go
@@ -1,0 +1,295 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 alibaba/open-code-review Contributors
+
+package main
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/spf13/cobra"
+)
+
+// TestArgCountErrors_AreActionable verifies that a wrong positional-argument
+// count reports what the command expects instead of Cobra's raw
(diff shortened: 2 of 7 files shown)


lizhengfeng101 (approved):
LGTM

PR #898, opened 2026-08-14
test(agent): add end-to-end regression test for review item fingerprint stability

### LINKED ISSUE #718: Per-file diff fingerprint depends on the file's position in the patch (trailing-newline artifact), silently defeating --resume reuse
### OpenCodeReview Version

open-code-review v1.8.8 (6dc3eb067) darwin/arm64, built at 2026-08-04T13:09:52Z

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

Other OpenAI-compatible endpoint (DeepSeek) — the defect is provider-independent: the fingerprint is computed before any LLM call.

### Bug Description

The review-item fingerprint is not invariant to the file's **position within the patch**, so `--resume` silently re-reviews files whose content did not change.

`reviewItemFingerprint` hashes the extracted per-file diff text (`internal/agent/agent.go:733` at 6dc3eb067):

```go
func reviewItemFingerprint(mode string, d model.Diff) string {
	sum := sha256.Sum256([]byte(mode + "\x00" + d.OldPath + "\x00" + d.NewPath + "\x00" + d.Diff))
	return fmt.Sprintf("%x", sum)
}
```

but `d.Diff` for the **last** file in a multi-file patch carries a trailing blank line that the same file does not carry when another file's section follows it. When the file set of the range changes (a new file enters the diff), a file whose own hunks are byte-identical can move from tail position to non-tail position (or vice versa), its `d.Diff` gains/loses that trailing line, and its fingerprint flips.

Consequence: on a resumed run, that file misses the fingerprint index and is re-reviewed at full LLM cost even though nothing in it changed. In iterating workflows (review after each push on a growing range) this happens on every run. Any future fingerprint-based dedup or caching inherits the same instability.

I diffed the extracted `<current_file_diff>` payloads for the same unchanged file across two runs (from the session JSONL): the **only** difference was one trailing empty line at the end of the block.

### Steps to Reproduce

Any configured provider works. In an empty directory:

```bash
git init -q ocr-fp-repro && cd ocr-fp-repro
printf 'fn main() {\n    println!("hi");\n}\n' > main.rs
git add . && git commit -qm init && git tag base

# commit 1: modify main.rs — it will be the ONLY (thus last) file in the patch
printf 'fn helper() -> u32 { 41 }\n' >> main.rs
git add . && git commit -qm "change main"
ocr review --from base --to HEAD --audience agent        # run 1

# commit 2: add an unrelated file; main.rs is untouched but no longer last in the patch
printf 'pub fn double(x: u32) -> u32 { x * 2 }\n' > util.rs
git add . && git commit -qm "add util"
ocr review --from base --to HEAD --resume <RUN1_SESSION_ID> --audience agent   # run 2

# control — commit 3: add another file; main.rs stays in a non-tail position
printf 'pub fn triple(x: u32) -> u32 { x * 3 }\n' > zeta.rs
git add . && git commit -qm "add zeta"
ocr review --from base --to HEAD --resume <RUN2_SESSION_ID> --audience agent   # run 3
```

Session IDs are printed on failure hints and available via `ocr session list`; per-item fingerprints are visible in `~/.opencodereview/sessions/<repo>/<session>.jsonl` (`review_item_done` / `review_item_reused` events) and in the `--format json` run manifest.

### Expected Behavior

The fingerprint of a file whose own diff content is byte-identical between two runs should be identical, regardless of which other files entered or left the range. Run 2 should reuse `main.rs`; run 3 should reuse both `main.rs` and `util.rs`.

### Logs / Error Output

Observed (run 2): `main.rs` is re-reviewed — run manifest shows `"reused": []` — and its fingerprint changed although its hunks are identical:

```
run1  review_item_done  main.rs  fp=71ae2305bc49fd54f5dfc23c3dd260648faaa6c211aa1aa333f1a82f99c81b18
run2  review_item_done  main.rs  fp=587e13a2…   (same hunks; only a trailing blank line differs)
```

Unified diff of the two extracted per-file diff payloads for main.rs (run1 vs run2) — the entire difference:

```
@@ -6,4 +6,3 @@
      println!("{}", data);
  }
 +fn helper() -> u32 { ... }
-
```

Control (run 3, resumed from run 2): `main.rs` kept a stable non-tail position and WAS reused with zero LLM calls, confirming the mechanism; `util.rs` moved from tail to non-tail and was re-reviewed despite being unchanged:

```
run3  review_item_reused  main.rs  fp=587e13a2…  sourceSession=<run2>
run3  review_item_done    util.rs  fp=8e99a001…  (was 36a119eb… in run2; content unchanged)
```

### Additional Context

Suggested fix: normalize trailing whitespace/newlines of `d.Diff` before hashing in `reviewItemFingerprint` (or make the patch splitter emit position-independent per-file text). Note the compatibility edge: changing the computation invalidates fingerprints stored in existing sessions, so it may be worth bumping a fingerprint/schema version in the run manifest so a resume across the boundary degrades to a clean re-review instead of silent mismatch.

### LINKED ISSUE #897: test(agent): add end-to-end regression test for review item fingerprint stability
## Description

[#732](https://github.com/alibaba/open-code-review/pull/732) fixed [#718](https://github.com/alibaba/open-code-review/issues/718) by normalizing trailing line endings before hashing a per-file review diff:

```go
// internal/agent/agent.go
func reviewItemFingerprint(mode string, d model.Diff) string {
	diffText := strings.TrimRight(d.Diff, "\r\n")
	sum := sha256.Sum256([]byte(mode + "\x00" + d.OldPath + "\x00" + d.NewPath + "\x00" + diffText))
	return fmt.Sprintf("%x", sum)
}
```

The accompanying test, `TestReviewItemFingerprintIgnoresTrailingLineEndings` (`internal/agent/agent_test.go:562`), feeds hand-written strings straight into the helper and asserts that appending `"\n"`, `"\r\n"` or `"\n\n"` does not change the hash. That is a good unit test of the helper's string handling, but it does not exercise the path where the bug actually occurred: `internal/agent/agent_test.go` never calls `diff.ParseDiffText`, so nothing asserts that fingerprints produced from **real parser output** are position-independent.

That gap matters because the trailing-newline artifact is produced by the parser, not by the helper. `ParseDiffText` splits on `"\n"` and flushes each file section with `strings.TrimSuffix(buf.String(), "\n")` (`internal/diff/parser.go:53` and `:110`), which removes exactly one newline. The trailing run therefore depends on where the file sits in the patch:

```
# two-file patch
a.go => "...@@ -1 +1 @@\n-a\n+A"      # not last: no trailing newline
b.go => "...@@ -1 +1 @@\n-b\n+B\n"    # last: one trailing newline
```

Workspace mode makes the asymmetry larger, because `collectDiffs` joins untracked file diffs with `"\n\n"` (`internal/diff/git.go`):

```
a.go => "...+A\n\n"      # two trailing newlines
b.go => "...+B\n\n\n"    # three trailing newlines
```

A test that goes through `ParseDiffText` pins down the user-visible guarantee — *the same unchanged file keeps the same fingerprint regardless of its position in the patch, so `--resume` reuses it* — instead of the helper's current implementation. If the parser's flush or joining logic is ever reworked, a helper-only test cannot notice; an end-to-end test can.

## Scope

- File(s): `internal/agent/agent_test.go`
- Function/area: `reviewItemFingerprint`, exercised through `diff.ParseDiffText`
- No production code needs to change. `internal/agent` already imports `internal/diff` (`agent.go:28`) and `internal/diff` does not import `internal/agent`, so there is no import cycle. The test lives inside `package agent`, so it can call the unexported `reviewItemFingerprint` directly.

A suggested shape — parse the same two-file patch twice with the target file in different positions, and assert the fingerprint is identical:

```go
func TestReviewItemFingerprintStableAcrossPatchPosition(t *testing.T) {
	section := func(path, from, to string) string {
		return "diff --git a/" + path + " b/" + path + "\n" +
			"--- a/" + path + "\n+++ b/" + path + "\n" +
			"@@ -1 +1 @@\n-" + from + "\n+" + to + "\n"
	}

	// Pre-create the files so finalizeDiff's working-tree read succeeds and the
	// test does not emit spurious "cannot read file" warnings.
	dir := t.TempDir()
	for _, name := range []string{"a.go", "z.go"} {
		if err := os.WriteFile(filepath.Join(dir, name), []byte("new\n"), 0o644); err != nil {
			t.Fatal(err)
		}
	}

	target, other := section("a.go", "old", "new"), section("z.go", "x", "y")

	fingerprintOfA := func(t *testing.T, patch string) string {
		t.Helper()
		diffs, err := diff.ParseDiffText(context.Background(), patch, dir, "", nil)
		if err != nil {
			t.Fatalf("ParseDiffText: %v", err)
		}
		for _, d := range diffs {
			if d.NewPath == "a.go" {
				return reviewItemFingerprint(session.ReviewModeRange, d)
			}
		}
		t.Fatal("a.go missing from parsed diffs")
		return ""
	}

	for _, tc := range []struct {
		name, last, first string
	}{
		// The splitter leaves a trailing newline on whichever file is last.
		{"plain patch", other + target, target + other},
		// Workspace mode joins untracked file diffs with "\n\n", so the trailing
		// run is longer than a single newline.
		{"untracked join", other + "\n\n" + target + "\n\n", target + "\n\n" + other + "\n\n"},
	} {
		t.Run(tc.name, func(t *testing.T) {
			if got, want := fingerprintOfA(t, tc.last), fingerprintOfA(t, tc.first); got != want {
				t.Errorf("fingerprint depends on position in patch:\n last  = %s\n first = %s", got, want)
			}
		})
	}
}
```

Additional imports needed: `os`, `path/filepath`, and `github.com/alibaba/open-code-review/internal/diff`. `context` and `session` are already imported by `agent_test.go`.

Contributors are welcome to structure this differently — driving it through `applyResume` and asserting on `ResumeInfo.ReusedFiles` would assert the same guarantee one level closer to the user-visible behaviour.

## Acceptance Criteria

- [ ] A test in `internal/agent` derives fingerprints from `diff.ParseDiffText` output and asserts they do not depend on a file's position within the patch
- [ ] Coverage includes both a plain multi-file patch and the workspace-mode `"\n\n"` joining case
- [ ] The test fails if `reviewItemFingerprint`'s trailing-line-ending normalization is removed
- [ ] The test emits no spurious `cannot read file` warnings to stderr

## Context

Found while reviewing #732. The fix itself is correct and is already on `main` (`590a6f3`); this issue only covers the missing regression coverage.

The suggested test was verified locally against `main` under Go 1.26.6:

- On `main` as it stands, both sub-cases pass.
- Replacing `strings.TrimRight(d.Diff, "\r\n")` with `strings.TrimSuffix(d.Diff, "\n")` makes the `untracked join` sub-case fail while `plain patch` still passes, confirming the second case is what pins down the multi-newline behaviour.
- Reverting the normalization entirely reproduces #718: the same file yields two different fingerprints depending on its position.

For accuracy: the existing helper test also fails under that `TrimSuffix` variant, via its `crlf` and `extra blank line` cases. The value of the proposed test is not that it catches an extra class of regression in the helper, but that it covers the real parser path and states the guarantee in terms of observable `--resume` behaviour rather than the helper's internals.

### Optional follow-up (not required for this issue)

The positional asymmetry still exists in `d.Diff` itself — only the hash is normalized. Those stray blank lines flow into the prompt sent to the model and into token estimation (`internal/agent/estimate.go:56` and `:76`). Changing `internal/diff/parser.go:53` and `:110` from `strings.TrimSuffix(buf.String(), "\n")` to `strings.TrimRight(buf.String(), "\r\n")` would fix the root cause, and it produces exactly the same fingerprints as the current code, so it would not invalidate any existing session. Anyone who wants to tackle that should open a separate issue or PR; the test described above is worth having either way.

yingjiexu2002:
## Description

Adds the regression test described in #897: `TestReviewItemFingerprintStableAcrossPatchPosition` derives fingerprints from real `diff.ParseDiffText` output and asserts they do not depend on a file's position within the patch.

#732 fixed #718 by normalizing trailing line endings in `reviewItemFingerprint`, but the existing helper-level test feeds hand-written strings and never exercises `ParseDiffText` — the component that actually produces the position-dependent trailing newlines. This test pins the user-visible `--resume` guarantee (same unchanged file keeps the same fingerprint regardless of patch position) on the real parser path:

- **plain patch**: two-file patch with the target file last vs. first (the splitter leaves one trailing newline on the last file)
- **untracked join**: workspace-mode `"\n\n"` joining, where the trailing run is 2-3 newlines — the case only `TrimRight` (not `TrimSuffix`) handles

Test-only change; no production code modified.

## Type of Change

- [ ] Breaking change (fix or feature that changes existing functionality)
- [ ] Documentation update
- [x] CI / Build / Tooling

## How Has This Been Tested?

- New test passes on `main` with both sub-cases (`plain patch`, `untracked join`)
- Mutation check: removing the `strings.TrimRight(d.Diff, "\r\n")` normalization makes the test fail (verified locally, then reverted)
- No spurious `cannot read file` warnings on stderr
- `make test` passes (full suite)
- `make check` passes

## Checklist

- [x] I have added tests that prove the fix/feature works
- [x] I have signed the CLA

## Related Issues

closes #897

### DIFF
--- a/internal/agent/agent_test.go
+++ b/internal/agent/agent_test.go
@@ -6,11 +6,14 @@ package agent
 import (
 	"context"
 	"encoding/json"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
 	"github.com/alibaba/open-code-review/internal/config/template"
 	"github.com/alibaba/open-code-review/internal/config/toolsconfig"
+	"github.com/alibaba/open-code-review/internal/diff"
 	"github.com/alibaba/open-code-review/internal/llm"
 	"github.com/alibaba/open-code-review/internal/model"
 	"github.com/alibaba/open-code-review/internal/session"
@@ -585,6 +588,60 @@ func TestReviewItemFingerprintIgnoresTrailingLineEndings(t *testing.T) {
 	withContextLine.Diff += "\n "
 	if got := reviewItemFingerprint(session.ReviewModeRange, withContextLine); got == want {
 		t.Error("fingerprint ignored a real trailing context line")
+	}
+}
+
+// TestReviewItemFingerprintStableAcrossPatchPosition pins down the --resume
+// guarantee end to end: the patch splitter leaves position-dependent trailing
+// line endings on each file section, so the fingerprint must be derived from
+// real diff.ParseDiffText output with the target file in different positions.
+func TestReviewItemFingerprintStableAcrossPatchPosition(t *testing.T) {
+	section := func(path, from, to string) string {
+		return "diff --git a/" + path + " b/" + path + "\n" +
+			"--- a/" + path + "\n+++ b/" + path + "\n" +
+			"@@ -1 +1 @@\n-" + from + "\n+" + to + "\n"
+	}
+
+	// Pre-create the files so finalizeDiff's working-tree read succeeds and the
+	// test does not emit spurious "cannot read file" warnings.
+	dir := t.TempDir()
+	for _, name := range []string{"a.go", "z.go"} {
+		if err := os.WriteFile(filepath.Join(dir, name), []byte("new\n"), 0o644); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	target, other := section("a.go", "old", "new"), section("z.go", "x", "y")
+
+	fingerprintOfA := func(t *testing.T, patch string) string {
+		t.Helper()
+		diffs, err := diff.ParseDiffText(context.Background(), patch, dir, "", nil)
+		if err != nil {
+			t.Fatalf("ParseDiffText: %v", err)
+		}
+		for _, d := range diffs {
+			if d.NewPath == "a.go" {
+				return reviewItemFingerprint(session.ReviewModeRange, d)
+			}
+		}
+		t.Fatal("a.go missing from parsed diffs")
+		return ""
+	}
+
+	for _, tc := range []struct {
+		name, last, first string
+	}{
+		// The splitter leaves a trailing newline on whichever file is last.
+		{"plain patch", other + target, target + other},
+		// Workspace mode joins untracked file diffs with "\n\n", so the trailing
+		// run is longer than a single newline.
+		{"untracked join", other + "\n\n" + target + "\n\n", target + "\n\n" + other + "\n\n"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			if got, want := fingerprintOfA(t, tc.last), fingerprintOfA(t, tc.first); got != want {
+				t.Errorf("fingerprint depends on position in patch:\n last  = %s\n first = %s", got, want)
+			}
+		})
 	}
 }
 


lizhengfeng101 (approved):
LGTM

PR #870, opened 2026-08-12
feat(telemetry): display session ID in terminal summary output

AllenMuu:
## What

Print the review session UUID on a separate line after the terminal
summary, so users can copy it for `ocr resume` without `--format json`
or filesystem browsing. The line is omitted when the session ID is
empty. JSON output is unchanged.

Closes #866

## Changes (+23/-5, 3 files)

- `internal/telemetry/events.go`: add `sessionID string` param to
  `PrintTraceSummary` (appended at end of signature); when non-empty,
  print `Session ID: <id>` on its own line after the summary.
- `cmd/opencodereview/shared.go`: pass `ag.SessionID()` at the call site.
- `internal/telemetry/events_test.go`: 2 new smoke tests + 3 existing
  tests updated for the new signature.

## Design notes

- **Param vs helper**: a dedicated `PrintSessionID(sessionID string)`
  helper would arguably be cleaner than a 9th positional param on
  `PrintTraceSummary` (separates summary-stats from session-identity).
  Kept the param approach to match the issue's hint and minimize review
  friction; happy to refactor if preferred.
- **Empty-case clarification**: the issue says the empty case is
  "e.g. dry-run / preview modes", but `preview` runs via `runPreview`/
  `outputPreview*` and never reaches `PrintTraceSummary`. The real
  empty case is `Agent.SessionID()` returning `""` when session
  persistence is unavailable (`internal/agent/agent.go`). The
  `if sessionID != ""` guard handles this and is exercised by the
  empty test.

## Verification

- `make check` ✅  ·  `make test` ✅ (all 23 packages)
- `make coverage` -> 90.2% total; `PrintTraceSummary` 100%.
- New tests are smoke-test style (matching the 3 pre-existing
  `PrintTraceSummary` tests); branch coverage 100% (both guard arms).
  Real output assertions need a `stdout.Swap(io.Writer)` helper
  (`stdout.Writer()` captures `os.Stdout` at init) - flagged as
  follow-up, out of scope here.

PR #909, opened 2026-08-14
refactor(telemetry): replace PrintTraceSummary positional params with TraceSummary struct

### LINKED ISSUE #906: refactor(telemetry): reduce PrintTraceSummary positional params and add output assertions
## Description

`PrintTraceSummary` in `internal/telemetry/events.go` currently takes 9 positional parameters (after #870 merged the `sessionID` param). This makes calls hard to read and error-prone as the signature grows. Additionally, all existing tests for this function are smoke tests that only verify "no panic" — they cannot assert the actual printed output because `stdout.Writer()` captures `os.Stdout` at init time and there is no swap/redirect mechanism for tests.

This issue covers two related improvements:

### 1. Refactor the function signature

Current signature (after #870):

```go
func PrintTraceSummary(filesReviewed, commentsGenerated int64,
    inputTokens, outputTokens, totalTokens int64,
    cacheReadTokens, cacheWriteTokens int64,
    duration time.Duration, sessionID string)
```

**Option A (preferred):** Introduce a `TraceSummary` struct and accept it as a single parameter:

```go
type TraceSummary struct {
    FilesReviewed    int64
    CommentsGenerated int64
    InputTokens      int64
    OutputTokens     int64
    TotalTokens      int64
    CacheReadTokens  int64
    CacheWriteTokens int64
    Duration         time.Duration
    SessionID        string
}

func PrintTraceSummary(s TraceSummary) { ... }
```

**Option B:** Extract session printing into a separate `PrintSessionID(id string)` helper and keep the remaining 8-param signature (less ideal long-term but simpler diff).

### 2. Add a `stdout.Swap` test helper for output assertions

The `internal/stdout` package currently exposes `Writer()` and `Quiet()`. A new `Swap(w io.Writer) func()` (or a test-only `CaptureOutput(t *testing.T) *bytes.Buffer`) would allow tests to redirect output and assert content:

```go
// Swap replaces the stdout writer and returns a restore function.
func Swap(replacement io.Writer) func() {
    mu.Lock()
    old := w
    w = replacement
    mu.Unlock()
    return func() {
        mu.Lock()
        w = old
        mu.Unlock()
    }
}
```

Then tests can assert actual output:

```go
func TestPrintTraceSummary_WithSessionID(t *testing.T) {
    var buf bytes.Buffer
    defer stdout.Swap(&buf)()
    PrintTraceSummary(TraceSummary{..., SessionID: "abc-123"})
    assert.Contains(t, buf.String(), "[ocr] Session: abc-123")
}
```

## Scope

- File(s): `internal/telemetry/events.go`, `internal/telemetry/events_test.go`, `internal/stdout/stdout.go`
- Function/area: `PrintTraceSummary`, `stdout` package
- Call site: `cmd/opencodereview/shared.go` (update to use struct)

## Acceptance Criteria

- [ ] `PrintTraceSummary` accepts a struct (or the session line is extracted to a helper)
- [ ] `internal/stdout` exposes a `Swap` (or test-capture) mechanism
- [ ] `PrintTraceSummary` tests assert actual output strings, not just "no panic"
- [ ] All existing tests updated for the new signature
- [ ] Coverage maintained ≥90% (`make coverage`)

## Context

Follow-up from #870 (which closed #866). The PR author flagged both items as out-of-scope improvements. See the [PR description](https://github.com/alibaba/open-code-review/pull/870) "Design notes" section for rationale.

abubakarsiddik31:
## Description

`PrintTraceSummary` had grown to nine positional parameters after the session ID parameter landed in #870, making call sites hard to read and easy to get wrong as the signature grows. This PR implements issue #906, Option A:

- Introduce a `TraceSummary` struct and change `PrintTraceSummary` to accept it as a single parameter. Printed output is byte-for-byte unchanged.
- Add `stdout.Swap(replacement io.Writer) func()` to `internal/stdout`, which swaps the package writer under the existing mutex and returns a restore function. This finally lets tests redirect and assert output written through `stdout.Writer()`.
- Rewrite the five `PrintTraceSummary` tests from smoke tests ("call it and hope it does not panic") into output assertions covering the exact summary line, the cache-token variant, the no-token-details variant, and the presence/absence of the `[ocr] Session:` line.
- Update the single production caller in `cmd/opencodereview/shared.go` to pass the struct with named fields.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

Note: `PrintTraceSummary`'s signature changes, but it is internal to the module (`internal/telemetry`), so no external API breaks.

## How Has This Been Tested?

- New `stdout.Swap` tests cover capture, restore, and nesting with `Quiet()`; the telemetry tests were written first and watched to fail (`undefined: Swap`, `undefined: TraceSummary`) before implementation.
- Full suite passes under the race detector (`make test`), and `make coverage` reports 91.3%, above the 90% threshold.
- `make check` (license headers, English-only scan, `go mod tidy`, `gofmt -s`, `go vet`) passes.

## Checklist

- [ ] I have signed the CLA

## Related Issues

Closes #906

### DIFF
--- a/cmd/opencodereview/shared.go
+++ b/cmd/opencodereview/shared.go
@@ -403,9 +403,17 @@ func emitRunResult(
 	}
 
 	if !machineReadable {
-		telemetry.PrintTraceSummary(ag.FilesReviewed(), int64(len(comments)),
-			ag.TotalInputTokens(), ag.TotalOutputTokens(), ag.TotalTokensUsed(),
-			ag.TotalCacheReadTokens(), ag.TotalCacheWriteTokens(), duration, ag.SessionID())
+		telemetry.PrintTraceSummary(telemetry.TraceSummary{
+			FilesReviewed:     ag.FilesReviewed(),
+			CommentsGenerated: int64(len(comments)),
+			InputTokens:       ag.TotalInputTokens(),
+			OutputTokens:      ag.TotalOutputTokens(),
+			TotalTokens:       ag.TotalTokensUsed(),
+			CacheReadTokens:   ag.TotalCacheReadTokens(),
+			CacheWriteTokens:  ag.TotalCacheWriteTokens(),
+			Duration:          duration,
+			SessionID:         ag.SessionID(),
+		})
 	}
 
 	if outputFormat == "json" {
--- a/internal/stdout/stdout.go
+++ b/internal/stdout/stdout.go
@@ -49,8 +49,9 @@ func Quiet() func() {
 //	var buf bytes.Buffer
 //	defer stdout.Swap(&buf)()
 //
-// Like Quiet, Swap is not designed for concurrent use from multiple
-// goroutines; keep swapping and restoring on one goroutine.
+// Like Quiet, Swap acquires the package mutex for memory safety, but concurrent
+// swaps from multiple goroutines produce non-deterministic restore ordering.
+// Keep swapping and restoring on a single goroutine.
 func Swap(replacement io.Writer) func() {
 	mu.Lock()
 	old := w
--- a/internal/stdout/stdout.go
+++ b/internal/stdout/stdout.go
@@ -41,3 +41,24 @@ func Quiet() func() {
 		mu.Unlock()
 	}
 }
+
+// Swap replaces the stdout writer with replacement and returns a restore
+// function. It lets tests capture output written through Writer().
+// Usage:
+//
+//	var buf bytes.Buffer
+//	defer stdout.Swap(&buf)()
+//
+// Like Quiet, Swap is not designed for concurrent use from multiple
+// goroutines; keep swapping and restoring on one goroutine.
+func Swap(replacement io.Writer) func() {
+	mu.Lock()
+	old := w
+	w = replacement
+	mu.Unlock()
+	return func() {
+		mu.Lock()
+		w = old
+		mu.Unlock()
+	}
+}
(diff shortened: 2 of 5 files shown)


lizhengfeng101 (commented):
**internal/stdout/stdout.go – `Swap` doc comment vs implementation**

The comment says:

> Like Quiet, Swap is not designed for concurrent use from multiple goroutines; keep swapping and restoring on one goroutine.

However, both `Swap` and its returned restore closure acquire `mu.Lock()`, so memory-safety under concurrent access *is* provided. The real hazard is semantic: if two goroutines swap/restore concurrently, the restore ordering becomes non-deterministic (goroutine A may restore goroutine B's writer).

Suggestion: rephrase to clarify the distinction, e.g.:

```go
// Like Quiet, Swap acquires the package mutex for memory safety, but concurrent
// swaps from multiple goroutines produce non-deterministic restore ordering.
// Keep swapping and restoring on a single goroutine.
```

Minor nit — not a blocker.

lizhengfeng101 (changes requested):
@abubakarsiddik31 internal/stdout/stdout.go – Swap doc comment vs implementation

abubakarsiddik31:
Good catch @lizhengfeng101 — reworded the comment to make the memory-safety vs. restore-ordering distinction clear, using your suggested phrasing. Pushed in 51f952c.

lizhengfeng101 (approved):
LGTM

PR #912, opened 2026-08-14
fix(pages): make install-channel command box scrollable

wu21-web:
## Description

Screenshot 2026-08-14 at 9 42 16 PM
The command box in the picture above is not scrollable, so even if the screen is wide enough, the command will never show completely, it will always end will '...'. This PR make the command box scrollable from left to right, the scroll bar now automatically fade after a short time.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

`npm run lint` & npm commands
Here is the final effect:

https://github.com/user-attachments/assets/d473f761-e350-4618-84ad-39b8a2704d92

## Checklist

- [x] I have signed the CLA

## Related Issues

none yet

### DIFF
--- a/pages/src/components/HeroSection.tsx
+++ b/pages/src/components/HeroSection.tsx
@@ -518,13 +518,16 @@ const HeroSection: React.FC = () => {
               }}
             >
               <span
+                className="command-scroll"
                 style={{
+                  flex: 1,
+                  minWidth: 0,
                   fontSize: 13,
                   fontFamily: 'Menlo, monospace',
                   color: 'rgba(255,255,255,0.85)',
                   whiteSpace: 'nowrap',
-                  overflow: 'hidden',
-                  textOverflow: 'ellipsis',
+                  overflowX: 'auto',
+                  overflowY: 'hidden',
                 }}
               >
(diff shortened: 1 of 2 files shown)


lizhengfeng101 (approved):
LGTM

lizhengfeng101 (approved):
LGTM! Clean fix for a real UX annoyance — nice work.

A couple of ideas for a follow-up PR if you're interested:

1. **Consider using `<div>` instead of `<span>`** — `overflow-x: auto` technically only works on block-level elements. It works here because the span is a direct flex child, but it's a bit fragile. If someone later wraps it or changes the parent layout, scrolling could silently break. Swapping to a `<div>` (or adding explicit `display: block`) would make the intent clearer.

2. **Scroll hint with a right-edge gradient** — Now that the ellipsis is gone, there's no visual cue that the command extends beyond the visible area (especially on macOS where the scrollbar auto-hides). A subtle fade using something like `mask-image: linear-gradient(to right, black 90%, transparent)` that disappears once scrolled to the end could help discoverability.

Neither of these is blocking — merging as-is. Thanks for the contribution!

PR #925, opened 2026-08-15
feat(cmd): allow command alias

wu21-web:
## Description

Allows users to set aliases for ocr commands, just like git's
`git config --global alias.co checkout`
now users can also set aliases on ocr with
`ocr alias set <short cmd>  <long cmd>`
e,g
`ocr alias set cp 'config provider'` & `ocr cp` is as same as `ocr config provider` (it don't have to be short)
## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [x] I have signed the CLA


wu21-web:
```
user@Users-Mac-mini open-code-review % mkdir ~/open-code-review
user@Users-Mac-mini open-code-review % cp -R ~/.opencodereview ~/open-code-review 
user@Users-Mac-mini open-code-review % ./dist/opencodereview alias set cp 'config provider'
Alias "cp" set to "config provider"
user@Users-Mac-mini open-code-review % ./dist/opencodereview cp
Cancelled.
user@Users-Mac-mini open-code-review % ./dist/opencodereview alias list
cp = config provider
user@Users-Mac-mini open-code-review % ./dist/opencodereview alias rm cp
Removed alias "cp".
user@Users-Mac-mini open-code-review % ./dist/opencodereview cp         
Error: unknown command "cp" for "ocr"
```

lizhengfeng101:
Thanks for the thorough work here — the implementation itself is solid (careful quoting/tokenizer handling, good validation around reserved names and unresolvable targets, and nice test coverage across all of it).

After discussion, we'd rather not bake command aliasing into the `ocr` binary itself. A few reasons:

- It adds an unconditional config-file read + parse to *every* `ocr` invocation (even `ocr version`, `ocr --help`, etc.), regardless of whether aliases are used, which is a cost we'd rather not pay on the hot path.
- Alias names are only checked against reserved/built-in command names at `ocr alias set` time. Since `config.json` is also meant to be hand-editable (per our own docs), a stored alias key that happens to collide with a real subcommand (e.g. `review`) could silently redirect it with no warning — a footgun we'd have to keep guarding against as the command tree evolves.
- Shell-level aliases/functions (`alias cp='ocr config provider'` in bash/zsh, or the equivalent in fish/PowerShell) already solve this well for the common case, with zero code and zero runtime cost, and they compose naturally with the rest of a user's shell setup.

We think leaving this to the shell is the better tradeoff for now. Really appreciate the effort you put into this — closing it out, but happy to see follow-up ideas if there's a use case shell aliases genuinely can't cover (e.g. syncing aliases across machines via `config.json` itself).

PR #927, opened 2026-08-15
fix(cmd): suppress ANSI color when stdout is not a TTY

### LINKED ISSUE #682: ANSI escape sequences are emitted when stdout is not a TTY
### OpenCodeReview Version

open-code-review v1.8.4 (e78474478) windows/amd64

### Operating System

Windows

### Installation Method

GitHub Release binary

### LLM Provider

### Description

`ocr review` emits ANSI escape sequences even when its output is redirected or piped to another program.

This makes it difficult to use the command in scripts or to publish its output elsewhere (e.g. GitHub issue comments), because the raw escape sequences become part of the output.

### Expected behavior

When stdout is not a TTY, `ocr review` should output plain text without ANSI escape sequences.

Alternatively, please provide an option such as:

* `--no-color`
* `--color=never`
* `--plain`

to explicitly disable ANSI formatting.

### Environment

* OpenCode Review (`ocr`)
* Windows (Git Bash)

### Use case

I would like to pipe the review output directly into GitHub CLI:

```bash
ocr review --commit HEAD | gh issue comment <issue> --body-file -
```

Currently the resulting GitHub comment contains literal escape sequences such as:

```text
�[2m
�[93m
�[0m
```

instead of plain text.

### Steps to Reproduce

```bash
ocr review --commit HEAD | cat -v
```

Output contains ANSI sequences such as:

```text
^[[2m─── packages/... ───^[[0m
^[[93m[maintainability · medium]^[[0m ...
```

or:

```bash
ocr review --commit HEAD | grep -P $'\x1b'
```

which matches the lines containing the colored severity labels.

### Expected Behavior

When stdout is not a TTY, `ocr review` should output plain text without ANSI escape sequences.

Alternatively, please provide an option such as:
* `--no-color`
* `--color=never`
* `--plain`

to explicitly disable ANSI formatting.

### Logs / Error Output

```shell

```

### Environment

* OpenCode Review (`ocr`)
* Windows (Git Bash)

chaojixinren:
## Description

Text output hardcoded ANSI escape sequences with no terminal detection, so piping or redirecting a review leaked raw escapes into the consumer. The case from #682 — piping into `gh issue comment` — produced a comment full of literal `^[[2m` / `^[[93m` / `^[[0m` sequences.

Color is now resolved once per run, in this order:

1. `--no-color` / `--color=never` turn it off
2. `--color=always` forces it on, even through a pipe
3. `NO_COLOR` (any non-empty value, per [no-color.org](https://no-color.org)) and `TERM=dumb` turn it off
4. otherwise it follows whether stdout is a terminal

Explicit flags outrank the environment because a flag is a per-invocation decision while the variable is a standing preference.

Both text paths — review findings and `--preview` — route every escape through `colorize()`, so plain mode keeps all the information the color carried: the diff gutter still shows `+` / `-` / space, and status badges still read `[A]` / `[M]` / etc. Preview counts are padded before colorizing so columns align in either mode.

The flags are persistent on the root command, so `ocr --no-color review` and `ocr review --no-color` are equivalent. An invalid `--color` value is rejected rather than silently treated as `auto`:

```
$ ocr review --color=sometimes
Error: invalid --color value "sometimes": must be one of auto, always, never
```

Notes for reviewers:

- No new external dependency. `github.com/charmbracelet/x/term` was already in the module graph indirectly via lipgloss; this promotes it to a direct require, which is the only `go.mod` change.
- `charmbracelet/colorprofile` was considered for reuse and deliberately not adopted: its `envNoColor` parses `NO_COLOR` with `strconv.ParseBool`, so `NO_COLOR=0` would leave color on, contradicting no-color.org. It also treats an unset `TERM` as dumb.
- The decision lives in `package main` rather than `internal/`, since `internal/` emits no escape sequences at all and `output.go` is the only non-TUI emitter.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

`make check` and `make test` pass. `make coverage` reports 91.3%, above the 90% threshold; `addColorFlags`, `validateColorMode`, `resolveColor`, `colorOn`, `colorize` and `colorf` are each at 100%.

Manual end-to-end runs against a rebuilt binary, counting escape sequences in the output:

| Scenario | Escapes |
| --- | --- |
| piped, `auto` (the #682 report) | 0 |
| piped, `--color=always` | 8 |
| piped, `--no-color` | 0 |
| piped, root-level `ocr --no-color review` | 0 |
| real TTY via `script -q /dev/null`, `auto` | 8 |
| real TTY + `NO_COLOR=1` | 0 |
| real TTY + `TERM=dumb` | 0 |
| `--color=sometimes` | rejected with the error above |

Unit tests cover the precedence chain, both rendering paths in each mode, and the flag wiring through the root command.

## Checklist

- [x] I have signed the CLA

## Related Issues

closes #682


wu21-web (approved):

lizhengfeng101:
Nice fix overall — the TTY-detection core and `--color <auto|always|never>` triad are the right shape, and the test coverage on `resolveColor()`'s precedence chain is solid.

One thing I'd push back on before merging: can we drop `--no-color` and the `NO_COLOR` env var check for now?

- `--no-color` is just an alias for `--color=never` — it doesn't add any capability, only another name for the same outcome. That's one more flag to document and test for zero new behavior.
- `NO_COLOR` isn't as universal a convention as it looks at first glance. Plenty of modern CLIs respect it, but plenty of long-established ones (git, GNU coreutils) don't — it's a strong convention in some corners of the ecosystem, not an industry-wide guarantee. Baking it in now is optimizing for a user we haven't heard from yet.

`--color=never` alone already gives users a complete way to force plain output — nothing is lost in terms of what's achievable, just one less way to say the same thing. If someone actually asks for `NO_COLOR` support down the line, it's a five-minute addition with an obvious precedent to follow (this PR). I'd rather add it when there's a real ask than carry it speculatively.

`TERM=dumb` I'd keep — that one isn't about convention, it's closing a real gap in the TTY check itself (some pty-backed shells report `isatty()==true` but can't render ANSI), so it's part of making `auto` actually correct rather than an extra way to configure the same toggle.

chaojixinren:
> merging: can we drop --no-color and the NO_COLOR env var check for now?
> 
> --no-color is just an alias for --color=never — it doesn't add any

Thanks for your review.
I'll drop `--no-color` flag and `NO_COLOR` env support, keep the `TERM=dumb` handling.
Will update tests and docs in the next commit.

lizhengfeng101:
Appreciate the quick turnaround! I've got the 200-PR eval regression suite for the next release starting shortly, so to make sure this lands before that run, I'll go ahead and push the change myself directly to this branch (drop `--no-color`/`NO_COLOR`, keep `TERM=dumb`, update tests + docs). Hope that works for you — will ping once it's in.

lizhengfeng101 (approved):
LGTM

PR #939, opened 2026-08-16
feat(pages): standardize web crawler policy

wu21-web:
## Description
Standardlize web crawler policy robots.txt, deploy to domain root.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [x] I have signed the CLA

## Related Issues

none

## AI Disclosure
No AI used.


pages/public/sitemap.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
```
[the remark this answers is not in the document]
wu21-web: Closing as not planned, I consider this to be done manually.

pages/public/robots.txt
wu21-web, line 1: This file is the core of the pull request. Suggestions are welcome.

lizhengfeng101:
Thanks for tackling this — the site genuinely needed robots.txt/sitemap.xml. Two things I'd want sorted before merge, plus one question.

## Why move the install scripts?

I'll be honest, I'm not seeing what this buys us. The old setup (real files at the repo root, `cp install.sh _site/` at deploy time) already served both `open-codereview.ai/install.sh` and the raw GitHub URL. Moving them under `pages/` only seems to save the two `paths:` entries in the workflow filter, and it costs us the symlink.

And that cost is real: `raw.githubusercontent.com` serves a symlink's *target path*, not its contents. I checked against your branch:

```
$ curl -fsSL https://raw.githubusercontent.com/wu21-web/open-code-review/robots/install.sh
pages/install.sh          # 16 bytes
```

That URL is exactly what we hand users in all four installation docs (`docs/{en,zh,ja,ru}/installation.md`), so after merge `curl -fsSL .../install.sh | sh` pipes the string `pages/install.sh` into `sh` and dies with "No such file or directory". Same for `irm ... | iex`. That makes this a breaking change rather than the non-breaking one the description claims. (Minor, but also: Windows clones without symlink support check the root files out as 16-byte text files.)

Unless there's a motivation I'm missing, I'd suggest dropping this part entirely — keep the scripts at the root and restore `'install.sh'` / `'install.ps1'` in the workflow `paths:` filter. That leaves this PR purely about the crawler files, which is a nice tight scope.

## The sitemap lists 23 URLs, 22 of which currently 404

The site is a client-side SPA and GitHub Pages serves `404.html` with an actual 404 status:

```
/                            200
/docs                        404
/docs/quickstart             404
/blog/introducing-ocr-blog   404
```

Search engines will reject all of those as "Submitted URL not found," so as it stands the sitemap fills Search Console with errors instead of getting us indexed. Either prerender a real `index.html` per route (we could generate the `HtmlWebpackPlugin` entries from the `DocSlug` union in `content/docs/index.ts` — that fixes site-wide indexability, not just the sitemap), or ship with just `/` for now and do prerendering separately. FWIW every slug in the sitemap checks out against the docs and blog registries, including `cicd` → `integrations/ci.md`, so that part's solid.

## Smaller stuff in the workflow

- `test -f _site/robots.txt && cmp ...` silently passes when the file is missing — `bash -e` doesn't exit on the left side of `&&`. The existing `Verify install scripts` step gets this right with one assertion per line; worth copying that shape.
- `CopyPlugin` already emits `public/*` into `dist/`, so the two new `cp pages/public/... _site/` lines are dead, and they make the `cmp` verify the `cp` rather than the build. If `CopyPlugin` ever breaks we'd still go green. Assert on `pages/dist/robots.txt` instead.
- Folding `apt-get` into the step named "Install dependencies" (with a stray `working-directory: pages`) is a bit surprising, and it adds an `apt-get update` round trip to every deploy. The container already has Node — a few lines of Node, or `npx fast-xml-parser`, drops the apt dependency entirely.
- The sitemap is hand-maintained with no test, and `pages/` already has vitest wired up. A test that diffs the sitemap against the docs/blog slug sets would keep it from drifting — or generate it from that registry at build time.

## Nits

- The new `public/` tree in `pages/README.md` omits `images/`.
- I'd skip the `robots.txt` link in the 5 READMEs and 4 FAQs — it's an ops artifact, not a doc page, and in the FAQs it breaks the surrounding relative-link convention. Crawlers find the sitemap via the `Sitemap:` directive anyway.
- No `<lastmod>`, which would help recrawl on docs and blog. Free if the sitemap becomes generated.
- The ja FAQ line uses ` — ` where its neighbors use `——`.

Nothing here is a security concern — the script blobs moved byte-for-byte (`8f92a5fa`, `d042d64a`), checksum verification and the sudo path are untouched.

wu21-web:
@lizhengfeng101 Answers here:
1.  I **didn't** use symlinks because I thought they can redirect. I moved the install scripts because I consider them to be associated with `pages` (serve for pages as well), so they are a part of the `pages`. I can revert this change if you react with :eyes: . Any raw url, e.g `raw.githubusercontent.com/xxx/xx/install.sh` should be deprecated and removed from the docs, should be replaced with the site url.
2. I will check that soon. Surprised that OCR didn't catch that.
3. All these are true. Tip on the last point, I didn't know how to generate the sitemap.xml on deploy dynamically, no idea. I assume manually maintaining that if no one can help me.
4. Sure. All nits are accurate.

I'm a newcomer and wrote this by myself, so there might be plenty of nits and bugs. So feel free to criticize my code.

wu21-web:
Updates done.

wu21-web:
@lizhengfeng101 I am planning to fully redo this since it is getting messy and more content is written by AI. 
Before I re-write this one question:
Can we just the `sitemap.xml` entirely or you prefer to keep it (that mean auto-building `sitemap.xml`) ?

The pages server has good navigation and generally good crawlers and find and explore the structure without any problem. `sitemap.xml` is optional and I don't really want AI to mess things up.

.github/workflows/deploy-pages.yml
```diff
@@ -47,6 +47,7 @@ jobs:
           cp pages/logo.svg _site/logo.svg
           cp install.sh _site/install.sh
           cp install.ps1 _site/install.ps1
+          cp pages/public/robots.txt _site/robots.txt
 
       - name: Verify install scripts
```
lizhengfeng101: One suggestion covers both fixes: drop the redundant `cp` (webpack CopyPlugin already ships `public/*` to the site root) and rename the step accordingly.

```suggestion

      - name: Verify deployment artifact
```

lizhengfeng101 (approved):
LGTM

PR #945, opened 2026-08-16
feat(action): opt-in completeness-gated checkpoint ranges (#476)

### LINKED ISSUE #476: Add completeness-gated cross-push range checkpoints to the reusable GitHub Action
## Problem statement

The reusable GitHub Action's incremental mode reduces duplicate publication, but it does not reduce review scope. On every pull-request update, the Action still invokes OCR over the full merge-base-to-current-head range.

This makes later pushes increasingly expensive on large pull requests even when the newest change is small.

Concrete production evidence:

- OpenCodeReview run: https://github.com/vybestack/llxprt-code/actions/runs/29946076923
- Job: https://github.com/vybestack/llxprt-code/actions/runs/29946076923/job/89011650080
- Pull request: https://github.com/vybestack/llxprt-code/pull/2610
- Newest push: https://github.com/vybestack/llxprt-code/compare/1436f7bf42e61cb6160bf1e267e602b11b843072...cdd6a6cbd7169894d2ad67c7cb8fc5520d86d4d8

The newest push changed 7 files and 323 lines. OCR reviewed the cumulative range: 63 selected files and 15,700 changed lines. The selected scope was therefore 9 times larger by file count and 48.6 times larger by changed-line count. The OCR step took 46:31 and used 3,981,022 tokens.

The behavior is present in the latest audited release, v1.7.15:

- The Action computes merge-base-to-head and invokes OCR over that full range: https://github.com/alibaba/open-code-review/blob/v1.7.15/action.yml#L201-L246
- The incremental input is described and implemented as existing-comment overlap suppression, not range selection: https://github.com/alibaba/open-code-review/blob/v1.7.15/action.yml#L75-L89 and https://github.com/alibaba/open-code-review/blob/v1.7.15/scripts/github-actions/post-review-comments.js#L150-L165

This is not a request to replace the complete immutable manifest proposed in #367. It is the Action-level range-selection and checkpoint-advancement contract that can consume that completeness state.

## Proposed solution

Add an opt-in, completeness-gated cross-push range mode to the reusable GitHub Action.

The Action should maintain or consume a versioned checkpoint identifying the last compatible, completely reviewed PR head and select its range as follows:

1. Initial, reopened, and ready-for-review events use the full PR merge-base-to-head range.
2. A normal synchronize event uses last-complete-head-to-current-head when:
   - the checkpoint belongs to the same repository and PR;
   - its reviewed head is an ancestor of the current head;
   - its base, OCR version, model identity, rules, and relevant policy are compatible;
   - its coverage and publication state are complete.
3. The Action falls back to a full PR review when state is absent, corrupt, partial, failed, canceled, incompatible, non-ancestor, or associated with a different base.
4. The checkpoint advances only after OCR coverage and required publication/state persistence are complete.
5. A canceled, failed, partially completed, malformed, or ambiguously published run never advances the checkpoint.
6. A same-head rerun becomes an observable no-op unless full mode is explicitly requested.
7. Expose a manual full-review override.
8. Emit the selected mode, reason, checkpoint before/after, exact from/to SHAs, ancestry result, and source run in machine-readable output.

The immutable coverage vocabulary and terminal-state rules should come from #367 rather than introducing a second manifest format.

### Acceptance criteria

- A seven-file follow-up to a previously complete review invokes OCR only for the unreviewed checkpoint-to-head range.
- Two rapid pushes where the first run is canceled cause the surviving run to cover last-complete-head through latest-head; no intervening commit is skipped.
- Force pushes, base changes, incompatible configuration, missing state, and incomplete prior runs deterministically select full fallback and report the reason.
- Raw synchronize-event before is never treated as proof that the preceding range completed.
- Checkpoint updates are authenticated, schema-validated, idempotent, and resistant to stale concurrent writes.
- Existing fork-safety and trusted-workflow guarantees are retained.
- Full-range review remains available for explicit or policy-driven use.

### Use synchronize event before-to-after directly

This is unsafe when an intermediate run is canceled or fails. In the observed workflow, run https://github.com/vybestack/llxprt-code/actions/runs/29937651406 was canceled after 38:42 of OCR work. Treating the next event's before SHA as reviewed would create a coverage gap.

### Use OCR resume

Released per-file resume from #265 / PR #306 is useful for retrying a compatible identical range. It validates the same textual review range and does not establish a durable last-successfully-reviewed PR head across changed ranges.

### Leave checkpointing entirely to every caller

Callers can implement this, but the reusable Action already owns PR range resolution, publication state, overlap suppression, and GitHub write reconciliation. A generic, explicit Action contract avoids incompatible implementations while allowing callers to choose the persistence backend.

### Continue full reviews and only suppress repeated comments

That improves presentation but does not reduce model work, tokens, or elapsed time.

## Affected area

- CLI / Commands
- Output / Formatting
- Configuration
- Telemetry / Observability
- Other: reusable GitHub Action

## Additional context

Related work:

- Reusable Action request and implementation: #235 and #337
- Historical request for since-last-review behavior: #147
- Complete immutable coverage manifest: #367 and PR #379
- Provider-aware retry lineage: #368
- Stable finding identity: #369
- PR-aware prior review context: #59
- Same-range resumable sessions: #265 and PR #306

The llxprt integration issue containing the measured run analysis and local workflow work is https://github.com/vybestack/llxprt-code/issues/2649.

chethanuk:
## Description

Closes #476. The reusable Action's `incremental` input suppresses duplicate comments but does not narrow the range: every push still invokes OCR over the full merge-base to head diff. On the production run cited in the issue, a push of 7 files and 323 lines was reviewed as 63 files and 15,700 lines, taking 46:31 and 3,981,022 tokens.

This adds an opt-in `checkpoint_range` input (default `'false'`). A run that completes records its reviewed head in a base64 marker appended to the sticky summary comment. The next push reads that marker and reviews `checkpoint..head` instead of `merge-base..head`. A second input, `full_review`, forces the full range on demand.

The marker is only trusted when every gate agrees, and any doubt widens back to the full range rather than narrowing. In order: sticky summary off, manual override, reopened / ready_for_review, no summary comment yet, the comment was not written by this action's own app, the marker is absent or malformed or duplicated, the payload is for another PR or another marker version or a run that did not complete, the base ref or merge-base moved, the config fingerprint changed (model, language, `llm_extra_body`, `llm_extra_headers`, `llm_auth_header`, `llm_timeout`, `background`, routing, resolved OCR version, and the contents of both `rule` and `.opencodereview/rule.json`), the checkpoint is not an ancestor of the new head, the commit is not in this clone, a rule file could not be read, or the resolver itself threw. Nine outputs (`range_mode`, `range_reason`, `range_from`, `range_to`, `checkpoint_before`, `checkpoint_after`, `ancestry`, `source_run`, `range_summary`) report which of those happened.

Each axis is hashed as its own field rather than joined into one string, so a value containing the separator cannot shift the field boundaries and let a config change slip through. `llm_extra_headers` enters as a digest rather than a raw value, since header values carry credentials and the fingerprint is published in the checkpoint marker. If the resolved OCR version cannot be read the fingerprint is empty rather than a constant, which the gates above already treat as "do not advance, do not narrow".

Writing the checkpoint is gated separately: it advances only when the manifest reports `terminal_state: complete`, nothing failed to post, and the summary comment actually published. A canceled or half-failed run carries the previous checkpoint forward unchanged, so the next run covers the range it never read. A rerun on an unchanged head reports `same_head_noop` and leaves the existing summary untouched instead of replacing its findings with "No comments generated".

### Limitations

- The trust boundary is repository write permission. GitHub attests who *posted* the summary comment, not that its body is unedited, so anyone who can edit a bot comment can move the checkpoint forward and cause a range to be skipped. A fork contributor, the party `pull_request_target` exists to distrust, posts as a `User` and is rejected unconditionally. On the default `github.token` the author must also be this action's own app, so another bot with write access cannot move the checkpoint; on a custom token the app cannot be named, and any bot identity GitHub attests is accepted. The README states both cases.
- `terminal_state: complete` means nothing the run *selected* failed. Waived items and files excluded before selection (unsupported types, size limits) sit inside that guarantee, so a checkpoint claims "everything this configuration chose to review was reviewed", not "every byte was read".
- The sticky summary is rewritten on each run, so with checkpointing on it reflects the latest range, not the whole PR. A narrowed run states its range in one line at the end of the summary. Inline comments are unaffected.

## Type of Change

## How Has This Been Tested?

`node scripts/github-actions/post-review-comments.test.js` and `node scripts/github-actions/check-translation-sync.test.js` both green, and `action.yml` parses as YAML.

The new coverage is a table over the resolver covering each gate and each ordering between them, plus round-trip cases that write a marker and read it back, force-push and base-change fallbacks, the same-head no-op, and the marker-preservation path that stops a completing run from blanking a checkpoint it could not re-derive. With `checkpoint_range` left at its default, the existing outputs are asserted unchanged and no extra API call is made.

## Checklist

- [x] I have signed the CLA

## Related Issues

Closes #476


chethanuk:
@lizhengfeng101 Please review and merge this.

lizhengfeng101:
Really solid work — the fail-closed ordering is careful and the write/read round-trip test is exactly the one I'd have asked for. Three things I'd like to see addressed before this goes in.

**1. `llm_extra_headers` isn't in the fingerprint.** The other axes look complete, but this one can swap the backend model or route to a different provider entirely while `llm_model` stays byte-identical. That's a config change that should kill the checkpoint and currently doesn't — and it's the one direction this feature can't afford to be wrong in. Worth adding `llm_auth_header` and `llm_timeout` too while you're there (timeout changes shift the partial/complete distribution), though `extra_headers` is the one I'd call a real gap.

**2. The author check trusts any bot, not just this action.** I poked at `isCheckpointAuthorOurs` directly:

```
dependabot[bot]                    -> true
renovate[bot]                      -> true
{type:"Bot", slug:"some-linter"}   -> true
```

Combined with `findSummaryIssueComment` matching on `<!-- ocr-summary -->` alone, the effective trust set is "any Bot or App that can post an issue comment carrying that marker" — wider than the "write-permission holders are trusted" boundary the README states. The fork-contributor case is genuinely closed (`user.type === "User"` is rejected unconditionally, and `testCheckpointRejectsUserToServerAppComment` covers the user-to-server hole nicely), so this isn't the scary direction — but the docs and the code disagree.

The comment says an installation token can't ask GitHub which app it is, which is true in general but not for the default token: `github.token` is always the `github-actions` app. Since `appSlug` already exists, you could pass it when the token is the default and fall back to the current behavior otherwise. If you'd rather not branch on that, then I think the README wording needs to change instead — either is fine, but the current pairing isn't.

**3. The `spec:` fallback doesn't do what its comment claims.** `ocr_version` defaults to `latest`, so `${VERSION_ACTUAL:-spec:${OCR_VERSION}}` collapses to the constant `spec:latest` — meaning if `ocr version` ever fails to print, that axis stops distinguishing versions entirely, which is precisely the failure the comment says it's guarding against. I'd let the failure propagate as an empty fingerprint instead: under your existing gates that means "don't advance, don't narrow", which is the honest answer.

Nothing else is blocking from my side. The `same_head_noop` output zeroing and a few brittle text assertions in the test suite are worth a follow-up, but they can land after.

chethanuk:

chethanuk:

lizhengfeng101 (approved):
nice work！

PR #946, opened 2026-08-16
feat(session): compare findings across two review sessions (#922)

### LINKED ISSUE #922: feat(session): compare findings across two review sessions
## Motivation

After fixing findings and running OCR again, users can inspect each saved
session separately, but cannot quickly see what changed between the two review
results. They have to manually compare two comment lists to answer:

- Which findings are new?
- Which findings are still present?
- Which findings have been resolved?

## Proposal

Add a session comparison command:

```bash
ocr session compare <before-session-id> <after-session-id>
ocr session compare --json <before-session-id> <after-session-id>
```

It would compare two saved review sessions from the same repository and group
the results into:

- `New`
- `Persisting`
- `Resolved`

Example:

```text
Session comparison: 4d2... -> 9a7...

Persisting: 1
New:        1
Resolved:   2
```

The text output could show the file path, severity, category, and comment for
each finding. `--json` would make the same result useful in CI or other tools.

This would make iterative review easier: after a fix-and-rerun cycle, users
could immediately see what still needs attention instead of manually comparing
sessions.

Would this be a useful addition to OCR? If so, may I submit a focused PR for
it?

chethanuk:
## Description

Closes #922. Sessions can be inspected one at a time, but after a fix-and-rerun cycle there is no way to see what changed between two of them short of diffing two comment lists by hand.

Adds `ocr session compare <before-session-id> <after-session-id>` (alias `diff`), with `--json` and `--repo`, grouping findings into **new**, **persisting** and **resolved**.

Two decisions worth calling out:

**Findings are matched on path, category and the offending snippet, not on line numbers.** A rerun after an unrelated edit shifts every finding below it, and a line-keyed match would report all of them as resolved plus new. Whitespace inside the snippet is collapsed first so re-indentation does not break identity either. Matching is a multiset match, so N identical findings before and M after yield min(N,M) persisting and the remainder on the longer side. The persisting bucket carries the after copy, which has the current line numbers. This key stays separate from the SARIF fingerprint in `cmd/opencodereview`, which is the identity of a published Code Scanning alert and cannot be renormalized without re-opening every alert downstream.

**A fourth bucket, `not_reviewed`, holds before-findings whose file the after run never looked at.** Without it, comparing a full scan against a diff-scoped rerun reports everything outside the diff as fixed, which is the one wrong answer this command can give. The set of reviewed paths comes from the after session's manifest coverage, and only the `completed` and `reused` partitions count: `selected` is the intended set, so an interrupted run would claim files it never reached. A legacy session with no manifest falls back to the three-bucket behaviour the issue describes.

Different repositories are an error. Different review modes only warn, on stderr, so `--json` stays pipeable.

### Limitations

- A file renamed between the two runs reads as resolved plus new, since the key holds only the path recorded with the finding. The manifest's `OldPath` could map that if it turns out to matter.
- A finding with no recorded snippet falls back to its prose, so a reworded finding on the same line reads as resolved plus new. A line-number fallback would be worse: it re-breaks the drift case above.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

`go build ./...`, `go vet ./...`, `gofmt -l .` and `go test ./internal/session/... ./cmd/opencodereview/...` are all clean, and `ocr session compare --help` renders the command, its alias and both flags.

Table tests cover the buckets themselves (line drift, re-indentation, duplicate findings on both sides, empty sessions), the manifest gate (missing manifest, a file only in `selected`, a path-less finding), and the command layer (mismatched repositories, the mode warning going to stderr, JSON shape and empty buckets encoding as `[]` rather than `null`). Bucket ordering is asserted so `--json` output is diffable between runs. One more test pins that all four locales of the CLI reference document the command, since those files are hand-synced.

## Checklist

- [x] I have signed the CLA

## Related Issues

Closes #922


chethanuk:

chethanuk:

lizhengfeng101:
@chethanuk I'd suggest keeping only `compare` and dropping the `diff` alias.

lizhengfeng101:
@chethanuk CI is red only because `renderComment` gained an `io.Writer` parameter on main after you branched — after a rebase, `renderComment(c)` at session_cmd.go:354 needs to become `renderComment(c, os.Stdout)`.

chethanuk:
Thank you @lizhengfeng101 I fixed it

lizhengfeng101 (approved):
LGTM

PR #952, opened 2026-08-16
fix(ci): line endings check fails on symbolic links

wu21-web:
## Description
```
=== git ls-files --eol ===
  i/lf    w/lf    attr/                         file.txt
  i/      w/      attr/                         link.txt

  === original CI pipeline ===
  git ls-files --eol | grep -v 'i/lf\|i/none\|i/-text'
  i/      w/      attr/                         link.txt
  grep_exit=0
```
The original step is:
```
  if git ls-files --eol | grep -v 'i/lf\|i/none\|i/-text'; then
    echo 'Line endings must end with LF. Have you normalized line endings?'
    exit 1
  fi
```
Git reports symlinks with empty i/ and w/ fields, so the symlink line does not match any of the allowed patterns. grep still finds that line and exits 0, which makes the if branch run and fail the job.

Now the logic is:
1. Make git reformat line endings
2. Compare with `git diff --quiet --cached`
3. If step 2 exits with non-0, abort and print err msg.

## Type of Change

- [ ] Documentation update
- [x] CI / Build / Tooling

## How Has This Been Tested?

I tested my fix on the failing pull request #939 

## Checklist

- [x] I have signed the CLA

## Related Issues

none

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,8 +55,10 @@ jobs:
       
       - name: Check line endings
         run: |
-          if git ls-files --eol | grep -v 'i/lf\|i/none\|i/-text'; then
+          git add --renormalize .
+          if ! git diff --cached --quiet; then
             echo 'Line endings must end with LF. Have you normalized line endings?'
+            git diff --cached --stat
             exit 1
           fi
         
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
         run: |
           git add --renormalize .
           if ! git diff --cached --quiet; then
-            echo 'Line endings must end with LF. Have you normalized line endings?'
+            echo "::error::Line endings must be LF. Run 'git add --renormalize .' and commit the diff below."
             git diff --cached --stat
             exit 1
           fi


wu21-web:
This is considered urgent for the development of #939

wu21-web:
May I request high priority for this one, @lizhengfeng101 @stay-foolish-forever , thank you.

lizhengfeng101:
Nice catch on the symlink issue — confirmed locally that `git ls-files --eol` reports empty `i/`/`w/` fields for symlinks, which is exactly what breaks the old `grep -v` pattern. The renormalize + diff approach is a solid fix since it delegates the "is this file normalized" judgment to git's own clean filter instead of parsing `ls-files` output.

One small suggestion on the failure message. Right now it's a rhetorical question:

```
echo 'Line endings must end with LF. Have you normalized line endings?'
```

`AGENTS.md` already documents the fix (`git add --renormalize .`), and it's the same command this check uses internally to detect the problem — so it's worth putting directly in the error text instead of making contributors look it up. Also, the other failing steps in this job (`Check formatting`, `go mod tidy`) use `::error::` annotations, which highlights the line in the GitHub UI; this step could match that style:

```yaml
if ! git diff --cached --quiet; then
  echo "::error::Line endings must be LF. Run 'git add --renormalize .' and commit the diff below."
  git diff --cached --stat
  exit 1
fi
```

Non-blocking either way — just closes the loop between "here's the error" and "here's the exact command to fix it."

wu21-web:
That is reasonable. Done.

wu21-web:
@lizhengfeng101 Any updates on this?

lizhengfeng101 (approved):
LGTM

Issue #470, opened 2026-07-23
feat(allowlist): expand supported file types — language experts wanted


lizhengfeng101:
## Description

OpenCodeReview currently supports **71 file extensions** for code review. However, many popular and emerging languages are missing from the allowlist. We're looking for **language experts** to help add proper support for their favorite languages — including file extensions, test-file exclude patterns, and **language-specific review rules**.

This is a **tracking issue**. Pick a language group below, create a sub-issue referencing this one, and submit a PR. Each language group is independent and can be contributed separately.

## Currently Supported

<details>
<summary>71 extensions (click to expand)</summary>

`.java` `.kt` `.kts` `.scala` `.groovy` `.py` `.pyi` `.js` `.jsx` `.ts` `.tsx` `.mjs` `.cjs` `.c` `.h` `.cpp` `.cc` `.cxx` `.hpp` `.hxx` `.cs` `.vb` `.fs` `.go` `.rs` `.rb` `.rake` `.gemspec` `.php` `.swift` `.m` `.mm` `.sh` `.bash` `.zsh` `.fish` `.ps1` `.sql` `.css` `.scss` `.sass` `.less` `.html` `.htm` `.ftl` `.ftlh` `.ftlx` `.astro` `.vue` `.svelte` `.xml` `.yaml` `.yml` `.json` `.toml` `.ini` `.env` `.gradle` `.cmake` `.r` `.lua` `.pl` `.pm` `.ex` `.exs` `.erl` `.hrl` `.ets` `.json5` `.dart` `.tf`

</details>

## Wanted — Language Groups

Pick one (or more!) and open a sub-issue:

| Language Group | Extensions | Test Exclude Patterns Needed | Difficulty |
|---|---|---|---|
| **Zig** | `.zig` | `**/test/*.zig`, `**/*_test.zig` | ★☆☆ |
| **Nim** | `.nim`, `.nims`, `.nimble` | `**/tests/**/*.nim` | ★☆☆ |
| **Haskell** | `.hs`, `.lhs` | `**/test/**/*.hs`, `**/*Spec.hs` | ★★☆ |
| **Clojure** | `.clj`, `.cljs`, `.cljc`, `.edn` | `**/test/**/*.clj` | ★★☆ |
| **Elm** | `.elm` | `**/tests/**/*.elm` | ★☆☆ |
| **OCaml / ReasonML** | `.ml`, `.mli`, `.re`, `.rei` | `**/test/**/*.ml` | ★★☆ |
[...]

PR #959, opened 2026-08-16
feat(rules): add Move review rules

chethanuk:
## Description

`.move` files were never reviewed. The extension is missing from `supported_file_types.json`, so `IsAllowedExt` drops them at the filter gate before any rule can run. `Move.toml` was already reachable (`.toml` is allowlisted) but fell through to the generic `default.md`.

This adds `.move` to the allowlist and registers two rule docs in `path_rule_map`: `**/*.move` → `move.md` for source, `**/Move.toml` → `move_toml.md` for the package manifest. The manifest gets its own short doc for the same reason `cargo_toml.md` and `pom_xml.md` do — what goes wrong there (git deps pinned to `rev = "main"`, named addresses changed in place on a published package, `dep-replacements` drifting between testnet and mainnet) has nothing to do with the source-level checks. Data only, no Go source changes.

The four exclude patterns are scoped by extension (`**/build/**/*.move`, not a bare `**/build/**`). `IsExcludedPath` applies every pattern to every path with no language dispatch, so an unscoped `build/` pattern would also drop Java, Gradle, TypeScript and CMake output. That is why the file already carries `**/Tests/**/*.swift` and `**/tests/**/*.nim` rather than the directory alone. Three negative rows in `TestIsExcludedPath` fail if anyone broadens it later.

Both `*_test.move` and `*_tests.move` are excluded. The singular is as common as the plural in Move and usually sits in `sources/` next to the code rather than under `tests/`, where the directory pattern never reaches it.

### Limitation

`.move` diffs that used to be skipped are now reviewed. Anyone who does not want that opts out with user exclude globs, which are checked before the allowlist.

## Type of Change

## How Has This Been Tested?

Rows appended to the existing table tests, plus one new test:

- `TestIsAllowedExt` — `.move` and `.MOVE`.
- `TestIsExcludedPath` — the four new patterns, plus three negatives (`frontend/build/index.ts`, `build/classes/com/example/Foo.java`, `build/main.go`) that fail if the `build/` pattern loses its extension scope.
- `TestResolve_DefaultRules` — source and manifest paths resolve by heading anchors unique to each doc.
- `TestResolveDetail_SystemMoveTomlPatternPrecedesMove` (new) — pins `Move.toml` to the `**/Move.toml` pattern, including the uppercase spelling.

## Checklist

- [x] I have signed the CLA

## Related Issues

Part of #470.


chethanuk:

wu21-web (approved):
I suggest putting `**/*.move` and `**/Move.toml` together in the site documentation for readability, but it is a minor issue. LGTM.

chethanuk:

chethanuk:

chethanuk:

chethanuk:

lizhengfeng101:
Thanks for the contribution! The rule doc is well-structured and the test coverage looks solid.

Since Move is still a relatively niche language, I'd prefer to hold off on merging for now and wait for community members who actively develop in Move (or language experts) to review the rule doc for accuracy. Smart contract review rules carry high stakes — an incorrect rule could either produce false positives or, worse, miss real vulnerabilities.

I'll leave this open and revisit once we get domain-expert feedback.

internal/config/rules/rule_docs/move_toml.md
```diff
@@ -0,0 +1,18 @@
+#### Move Package Manifest Hygiene
+- Git dependencies pinned to a branch or moving tag rather than a full commit hash: `rev = "main"`, `rev = "mainnet"`, and `rev = "framework/testnet"` all resolve differently over time, so the bytecode published today may not be the bytecode published tomorrow
```
Qiyuanqiii: This would flag official manifests. The Move Reference explicitly allows `rev` to be a branch, tag, or commit, and explains that a committed `Move.lock` records fixed dependency revisions for reproducible builds. The official Aptos package example itself uses `rev = "mainnet"`.

Please narrow this to evidence that resolution is not fixed - for example, a missing or uncommitted lockfile, or an unexpected `Move.lock` resolution change - rather than treating a branch or tag as a defect by itself.

Sources: [Move package reference](https://move-book.com/reference/packages/) | [Aptos package guide](https://aptos.dev/build/smart-contracts/create-package)

internal/config/rules/rule_docs/move_toml.md
```diff
@@ -0,0 +1,18 @@
...
+- Framework dependencies that should share one revision pinned to different revisions, mixing incompatible versions of the same source tree
+- `local =` paths that are absolute or reach outside the repository, which do not resolve on another machine or in CI
+- Dependencies added to `[dependencies]` when they are only used by tests, which publishes them as part of the package
+
+#### Addresses and Publication
```
Qiyuanqiii, line 7: `"_"` is a documented named-address parameter, not inherently an invalid publication address. The Move package reference shows that it can be instantiated by a dependent package, and the official Aptos create-package example deliberately declares `hello_blockchain = "_"`.

Please flag this only when the effective build or publish inputs leave a required address unresolved or resolve it to the wrong address, rather than flagging the placeholder itself.

Sources: [Move package reference](https://move-book.com/reference/packages/) | [Aptos package guide](https://aptos.dev/build/smart-contracts/create-package)

internal/config/rules/rule_docs/move_toml.md
```diff
@@ -0,0 +1,18 @@
...
+- `[addresses]` entries left as the `"_"` placeholder in a package that is meant to be published, which fails or resolves to an unintended address
+- Named addresses changed in place for an already-published package, silently retargeting every reference
```
Qiyuanqiii, line 9: The documented package semantics do not allow these sections to silently leak into an ordinary build: `[dev-dependencies]` and `[dev-addresses]` are included only in dev/test modes.

Please remove this rule or require concrete evidence that the project build explicitly enables a dev/test mode (or uses nonstandard tooling with different semantics).

Source: [Move package manifest reference](https://move-book.com/reference/packages/)

internal/config/rules/rule_docs/move.md
```diff
@@ -0,0 +1,49 @@
...
+- Do not report ability grants that match the surrounding module's established pattern without evidence that the wider grant is reachable and harmful
+
+#### Global Storage and Aborts (Aptos and core Move)
+- `move_to` on an address that may already hold the resource, which aborts, without an `exists<T>` guard or a documented caller contract
+- `move_from`, `borrow_global`, or `borrow_global_mut` on an address that may not hold the resource, without an `exists<T>` check or a typed `error::not_found` abort
+- Overlapping mutable global borrows, or a mutable borrow held across a call that re-enters the same resource
+- `acquires` annotations that no longer reflect the resources a function reaches after the change
+- Aborts raised with a bare integer where the module otherwise uses categorized `error::` constructors, losing the client-visible error category
+
+#### Object Model and Sharing (Sui)
```
Qiyuanqiii, line 16: This describes a successful Owned -> Shared conversion, but Sui documents the opposite behavior: only an object created in the same transaction can be shared, and attempting to share an already-owned object aborts with `ESharedNonNewObject`.

Please rewrite this as a reachable-abort check when a pre-existing owned object can reach `share_object`. Whether a newly created object should be shared is a separate authorization/design question and needs evidence from the intended access policy.

Source: [Sui storage functions - Share](https://move-book.com/storage/storage-functions/)

internal/config/rules/rule_docs/move.md
```diff
@@ -0,0 +1,49 @@
...
+- `transfer::share_object` on a value that was owned, which is irreversible and permanently widens who can mutate it (`share_owned`)
+- `transfer::transfer` versus `transfer::public_transfer` chosen inconsistently with the type's `store` ability and the custody the caller expects
+- A value sent to `ctx.sender()` where returning it to the caller would compose better and keeps the PTB in control of its destination (`self_transfer`)
+- Wrapping a shared or frozen object inside another object, which strands it (`freeze_wrapped`)
+- `&TxContext` on a public function that will need mutation later, forcing a breaking signature change (`prefer_mut_tx_context`)
```
Qiyuanqiii, line 21: Please revise both this rule and the earlier generic underlying-id rule. The `id: UID` layout is an additional Sui-verifier requirement, not a generic `key` rule, so the first half is compiler-enforced on Sui and does not apply to Aptos/core Move. Even on Sui, retaining a UID after destroying the containing object can be intentional (the documented proof-of-deletion pattern), so unconditional `id.delete()` is not always required.

A higher-confidence lifecycle check is deleting a parent UID before removing its dynamic fields or received children, which can leave state permanently inaccessible.

Sources: [Move abilities reference](https://move-book.com/reference/abilities/) | [Sui key ability](https://move-book.com/storage/key-ability/) | [Sui UID lifecycle](https://move-book.com/storage/uid-and-id/) | [Sui dynamic fields](https://move-book.com/programmability/dynamic-fields/)

internal/config/rules/rule_docs/move.md
```diff
@@ -0,0 +1,49 @@
...
+- A `key` struct whose first field is not `id: UID`, or a destroy path that does not call `id.delete()`
+
```
Qiyuanqiii, line 23: For current Sui Move, `public` is already transaction-callable and adding `entry` gives the function no additional permissions or restrictions. That is why the compiler reports `public_entry` as an unnecessary-modifier lint; it does not by itself widen the callable surface.

Please remove this as a security finding or platform-qualify it around a concrete authorization/composability defect rather than the modifier alone.

Sources: [Sui visibility modifiers](https://move-book.com/move-basics/visibility/) | [Sui lint reference](https://move-book.com/testing/linting/)

internal/config/rules/rule_docs/move.md
```diff
@@ -0,0 +1,49 @@
...
+- Ability constraints wider than the function body requires, which needlessly restricts the types callers may instantiate
+- Publicly instantiable `T: store` or `T: key + store` parameters where a caller can substitute an arbitrary type into a privileged path
+- One-time-witness types on Sui not verified as genuine before granting the authority they represent
+- Type parameters used to key storage or events where two distinct instantiations can collide
+
+#### Errors, Events, and Upgrade Compatibility
+- Named `const E*` abort code values changed or renumbered: clients and indexers match on them, so a renumbering is a breaking change even though the code still compiles
+- State changes that indexers depend on made without emitting the corresponding event, or an event's fields changed in place
+- Struct fields or public function signatures changed in ways that violate Sui upgrade compatibility or Aptos `upgrade_policy = "compatible"`
+- `#[test_only]` mint, admin, or fixture functions reachable from a non-test build, or test-only imports leaking into published bytecode
```
Qiyuanqiii: Under the documented Sui compilation model, `#[test_only]` items are compiled only in test mode and are excluded from normal builds. Artifacts built with any mode enabled, including test mode, are explicitly non-publishable. Therefore these items cannot normally leak into published bytecode.

Please remove this rule or gate it on evidence that a nonstandard toolchain bypasses the documented compile-time and publishability checks.

Sources: [Sui compilation modes](https://move-book.com/move-advanced/modes/) | [Move unit-testing reference](https://move-book.com/reference/unit-testing/)

Qiyuanqiii (changes requested):
Thanks for the thorough implementation. I reviewed the proposed default rule documents against the official Move, Sui, and Aptos documentation. The allowlist, routing, and test work looks solid, but several statements in `move.md` and `move_toml.md` currently contradict documented platform semantics or restate compiler-enforced checks as review findings. Because these documents would be injected as the defaults for every Move review, those statements would create systematic false positives.

I am requesting changes only for the seven concrete accuracy and platform-gating issues called out inline. Please revise or remove those rules while preserving the useful Sui/Aptos/core-Move distinction already established by the document.

This is an evidence-based documentation review. I am not presenting it as a substitute for the active Move-developer review requested by the maintainer.

Qiyuanqiii:
While I’ve put forward this change proposal, it is no substitute for a proper review from Move community developers. As @lizhengfeng101 noted, Move is still a relatively niche language. My knowledge of it only comes from what I picked up incidentally while learning Rust — after all, Move’s ownership model draws inspiration from Rust, and the core Move VM is also implemented in Rust.
 
That said, I have no hands-on experience building with Move. It’s a purpose-built language for virtual machines, primarily used for digital tokens and financial-related scenarios. I haven’t worked with it extensively, so this will need further review from someone with deep expertise in the language.

Qiyuanqiii (approved):

PR #960, opened 2026-08-16
feat(allowlist): add Thrift and Cap'n Proto support

chethanuk:
## Description

`.thrift` and `.capnp` files were never reviewed. Neither extension is in `supported_file_types.json`, so `IsAllowedExt` drops them at the filter gate before any rule can run.

This adds both to the allowlist and registers `**/*.thrift` → `thrift.md` and `**/*.capnp` → `capnp.md` in `path_rule_map`. Two docs rather than one shared IDL doc, because almost nothing carries over between them. Thrift breaks when a `required` field is added to a live struct, when a retired field id is reused (there is no `reserved` keyword to hold it open), and when a service method is renamed, since method names travel on the wire. Cap'n Proto breaks when an `@N` ordinal moves, when a `UInt32` is widened to `UInt64` (fixed-width slots, unlike a protobuf varint), and when a struct is renamed with no explicit `@0x...` id pinned, since the id is derived from the name. A merged doc would spend half its lines telling the model which half to ignore. Data only, no Go source changes.

The generated-output excludes are scoped by extension (`**/gen-{cpp,cpp2,go,py,...}/**/*.{c,cc,go,py,...}`, not a bare `**/gen-*/**`). `IsExcludedPath` applies every pattern to every path with no language dispatch, so an unscoped directory glob would drop unrelated files in other languages that happen to sit under a colliding directory name. That is why the file already carries `**/Tests/**/*.swift` and `**/tests/**/*.nim` rather than the directory alone. Three negative rows in `TestIsExcludedPath` fail if anyone broadens it later.

Only the Cap'n Proto C++ header gets a pattern. The generator also emits `.capnp.c++`, but `.c++` was never allowlisted, so those files are already gone by the time `IsExcludedPath` runs.

### Limitation

`.thrift` and `.capnp` diffs that used to be skipped are now reviewed. Anyone who does not want that opts out with user exclude globs, which are checked before the allowlist.

## Type of Change

## How Has This Been Tested?

- `TestIsAllowedExt` — `.thrift` and `.capnp`, both cases.
- `TestIsExcludedPath` — the `gen-*`, `kitex_gen` and Cap'n Proto output patterns, plus negatives (`tools/gen-go/config.yaml`, `site/gen-js/index.html`, `kitex_gen/api/schema.json`) that fail if the patterns lose their extension scope, and `idl/service.thrift` / `schema/addressbook.capnp` to pin that the schemas themselves stay in review.
- `TestResolve_DefaultRules` — `.thrift` and `.capnp` paths resolve by heading anchors unique to each doc.

## Checklist

- [x] I have signed the CLA

## Related Issues

Part of #470.

### DIFF
--- a/internal/config/allowlist/default_exclude_patterns.json
+++ b/internal/config/allowlist/default_exclude_patterns.json
@@ -37,5 +37,11 @@
   "**/tests/**/*.elm",
   "**/vendor/**/*.{jsonnet,libsonnet}",
   "**/test/**/*.zig",
-  "**/*_test.zig"
+  "**/*_test.zig",
+  "**/kitex_gen/**/*.go",
+  "**/*.capnp.h",
+  "**/*.capnp.go",
+  "**/*.capnp.ts",
+  "**/*_capnp.rs",
+  "**/*_capnp.py"
 ]
--- a/internal/config/allowlist/allowed_ext_test.go
+++ b/internal/config/allowlist/allowed_ext_test.go
@@ -79,6 +79,10 @@ func TestIsAllowedExt(t *testing.T) {
 		{".LIBSONNET", true},
 		{".zig", true},
 		{".ZIG", true},
+		{".thrift", true},
+		{".THRIFT", true},
+		{".capnp", true},
+		{".CAPNP", true},
 		{".txt", false},
 		{".md", false},
 		{".png", false},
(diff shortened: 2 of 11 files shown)


chethanuk:

wu21-web (approved):

kefu84510-cmd (approved):

chethanuk:

internal/config/allowlist/default_exclude_patterns.json
```diff
@@ -36,5 +36,12 @@
   "**/tests/**/*.elm",
   "**/vendor/**/*.{jsonnet,libsonnet}",
   "**/test/**/*.zig",
-  "**/*_test.zig"
+  "**/*_test.zig",
+  "**/gen-{cpp,cpp2,c_glib,cocoa,csharp,netstd,dart,erl,go,hs,hs2,java,js,nodejs,kotlin,lua,perl,php,py,py3,rb,rs,swift}/**/*.{c,cc,cpp,cs,dart,erl,go,h,hrl,hs,java,js,kt,lua,m,mm,php,pl,pm,py,pyi,rb,rs,swift,ts}",
```
lizhengfeng101: @chethanuk I feel that the hit rate for this matching rule is quite low and it is prone to accidental deletion; could we remove it?

```suggestion

```

lizhengfeng101 (approved):
LGTM

PR #961, opened 2026-08-16
feat(rules): add Solidity and Vyper review rules

chethanuk:
## Description

`.sol` and `.vy` files were never reviewed. Neither extension is in `supported_file_types.json`, so `IsAllowedExt` drops them at the filter gate before any rule can run, which is why a rule doc on its own would never have been reachable.

This adds both extensions and registers two rule docs in `path_rule_map`: `**/*.sol` → `solidity.md` and `**/*.vy` → `vyper.md`. Data only, no Go source changes.

Two docs rather than one shared EVM doc. The repo splits rule docs by language syntax, not by domain — `c.md` and `cpp.md` stay separate despite covering the same ground, and multi-extension docs like `haskell.md` cover several file forms of a single language, never two languages. Vyper checks arithmetic by default and has neither inheritance nor inline assembly, so a merged doc would put claims into every Vyper review that are false for Vyper.

The six exclude patterns are scoped by extension (`**/test/**/*.sol`, not a bare `**/test/**`). `IsExcludedPath` applies every pattern to every path with no language dispatch, so an unscoped `test/` pattern would drop Julia, Nim and Haskell files as well. `**/*.t.sol` on its own is not enough either: Foundry test helpers and mock contracts carry no suffix, and `test/utils/BaseTest.sol` and `contracts/test/ERC20TestToken.sol` are both real layouts. The directory patterns cover those the way `**/test/**/*.jl` and `**/tests/**/*.nim` already do. Both the singular and the plural spelling of the directory is present for each extension.

`lib/**/*.sol` is root-anchored on purpose. That is where Foundry installs git submodule dependencies, but `**/lib/**/*.sol` would also match `src/lib/Math.sol` and `contracts/lib/SafeMath.sol`, which are a project's own hand-written library code. Three negative rows fail if anyone broadens the pattern later.

### Limitation

Vyper's legacy `.v.py` extension is not covered and cannot be. `extFromPath` takes the last dot of the basename, so `token.v.py` resolves to `.py` and is reviewed as Python. An allowlist entry for `.v.py` would never be looked up.

`.sol` and `.vy` diffs that used to be skipped are now reviewed. Anyone who does not want that opts out with user exclude globs, which are checked before the allowlist.

## Type of Change

## How Has This Been Tested?

- `TestIsAllowedExt` — `.sol`/`.SOL` and `.vy`/`.VY`.
- `TestIsExcludedPath` — the six patterns, the three `lib/` negatives above, and `script/Deploy.s.sol`, which stays reviewable.
- `TestResolve_DefaultRules` — source paths for both extensions resolve by heading anchors unique to each doc.

## Checklist

- [x] I have signed the CLA

## Related Issues

Part of #470.


chethanuk:

wu21-web (approved):

chethanuk:

chethanuk:

chethanuk:

chethanuk:

chethanuk:

lizhengfeng101 (approved):
LGTM

PR #962, opened 2026-08-16
feat(rules): add Jsonnet review rules

chethanuk:
## Description

`.jsonnet` and `.libsonnet` files were never reviewed. Neither extension is in `supported_file_types.json`, so `IsAllowedExt` drops them at the filter gate before any rule can run, which is why a rule doc on its own would never have been reachable.

This adds both extensions and registers one rule doc in `path_rule_map`: `**/*.{jsonnet,libsonnet}` → `jsonnet.md`. Data only, no Go source changes.

One doc for both extensions. `.libsonnet` is a naming convention for files meant to be imported, not a second language: same evaluator, same semantics, same failure modes. Multi-extension docs already cover that case, `haskell.md` for `.hs`/`.lhs` and `nim.md` for `.nim`/`.nims`/`.nimble`.

The one exclude pattern is scoped by extension (`**/vendor/**/*.{jsonnet,libsonnet}`, not a bare `**/vendor/**`). `IsExcludedPath` applies every pattern to every path with no language dispatch, so an unscoped `vendor/` pattern would drop vendored Go and PHP sources repo-wide, whose extensions are allowlisted. Two negative rows, `vendor/github.com/pkg/errors/errors.go` and `vendor/monolog/monolog/src/Logger.php`, fail if anyone broadens it later. `vendor/` is where `jb install` writes jsonnet-bundler dependencies, and that tree is wiped and refetched rather than edited.

No test-file pattern is added. Jsonnet projects split between `test_*.libsonnet`, `tests/*.jsonnet` and `*_test.jsonnet` with no dominant convention, and a glob wide enough to catch all three would silently drop handwritten files.

### Limitation

`.jsonnet` and `.libsonnet` diffs that used to be skipped are now reviewed. Anyone who does not want that opts out with user exclude globs, which are checked before the allowlist.

## Type of Change

## How Has This Been Tested?

- `TestIsAllowedExt` — `.jsonnet`/`.JSONNET` and `.libsonnet`/`.LIBSONNET`.
- `TestIsExcludedPath` — the vendor pattern at the repo root and nested under `jsonnet/`, two non-vendor paths that stay reviewable, and the two Go/PHP negatives above.
- `TestResolve_DefaultRules` — source and library paths resolve by a heading anchor unique to `jsonnet.md`.

## Checklist

- [x] I have signed the CLA

The rule table in `pages/src/content/docs/{en,ja,ru,zh}/review-rules.md` lists the new pattern.

## Related Issues

Part of #470.

### DIFF
--- a/internal/config/allowlist/allowed_ext_test.go
+++ b/internal/config/allowlist/allowed_ext_test.go
@@ -71,6 +71,10 @@ func TestIsAllowedExt(t *testing.T) {
 		{".PO", true},
 		{".pot", true},
 		{".POT", true},
+		{".jsonnet", true},
+		{".JSONNET", true},
+		{".libsonnet", true},
+		{".LIBSONNET", true},
 		{".txt", false},
 		{".md", false},
 		{".png", false},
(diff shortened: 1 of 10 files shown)


chethanuk:

lizhengfeng101:
@chethanuk merge conflicts

chethanuk:
> @chethanuk merge conflicts

Resolved

lizhengfeng101 (approved):
LGTM

wu21-web (approved):

PR #964, opened 2026-08-16

### LINKED ISSUE #947: fix(llm): preserve opaque tool-call fields across OpenAI-compatible turns

vincekent:
## In Plain English

People using OpenCodeReview with certain models could get through the first repository lookup and then have the review stop before it produced a result. Information the model needed for the next step was silently lost between requests, so every review that reached that path failed. The review can now carry that information forward and continue normally, while models that do not use it behave as before.

## Description

OpenAI-compatible providers may attach opaque extension fields to assistant function tool calls and require those fields to be echoed unchanged on the next turn. OCR currently maps tool calls down to only the standard id, type, and function fields, which drops provider-required metadata such as Vertex Gemini's thought signature.

This change:

- captures unknown valid JSON fields from provider-returned function tool calls;
- keeps the standard id, type, and function fields owned by OCR;
- replays the opaque fields through the OpenAI SDK's supported SetExtraFields mechanism; and
- adds a two-request adapter regression test proving a nested thought signature survives response mapping and the next request.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] make test passes locally

Validated from a clean detached clone in the repository's Go 1.25.5 container:

- make check
- make test
- make build
- focused two-turn regression test for TestOpenAIClient_PreservesToolCallExtraFieldsAcrossTurns

The full test command passed. Two existing permission-specific viewer tests were skipped because the container runs as root.

## Checklist

- [x] My code follows the project's coding style (go fmt, go vet)
- [ ] I have signed the CLA

## Related Issues

Closes #947

### DIFF
--- a/internal/llm/client.go
+++ b/internal/llm/client.go
@@ -125,9 +125,10 @@ type Choice struct {
 
 // ToolCall represents a function call requested by the model.
 type ToolCall struct {
-	ID       string       `json:"id"`
-	Type     string       `json:"type"`
-	Function FunctionCall `json:"function"`
+	ID          string                     `json:"id"`
+	Type        string                     `json:"type"`
+	Function    FunctionCall               `json:"function"`
+	ExtraFields map[string]json.RawMessage `json:"-"`
 }
 
 // FunctionCall holds the name and arguments of a tool call.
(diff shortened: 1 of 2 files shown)


vincekent:
Withdrawing this contribution; we will not pursue the required CLA. Thank you.

PR #965, opened 2026-08-17
feat(rules): add Clojure and ClojureScript review rules

chethanuk:
## Description

`.clj`, `.cljs`, `.cljc` and `.edn` files were never reviewed. None of the four extensions is in `supported_file_types.json`, so `IsAllowedExt` drops them at the filter gate before any rule can run, which is why a rule doc on its own would never have been reachable.

This adds all four and registers one rule doc in `path_rule_map`: `**/*.{clj,cljs,cljc,edn}` → `clojure.md`. Data only, no Go source changes.

One doc for all four extensions. ClojureScript is the same language on a different host, and `.cljc` compiles to both, so a Clojure/ClojureScript split has no correct answer for `.cljc`: `path_rule_map` is first-match-wins, so whichever of the two patterns is listed first takes it and the other host's rules never load. Multi-extension docs already cover that case, `haskell.md` for `.hs`/`.lhs` and `nim.md` for `.nim`/`.nims`/`.nimble`. EDN is data rather than code, so it gets its own section: dependency coordinates pinned to a branch instead of an exact `:git/sha`, aliases whose `:main-opts` execute code, and `read-string` where `clojure.edn/read-string` belongs.

Exclude patterns are scoped by extension rather than by bare directory, since `IsExcludedPath` applies every pattern to every path with no language dispatch. Test files are `**/test/**/*.{clj,cljs,cljc}` and `**/*_test.{clj,cljs,cljc}`.

Compiled output needs `**/resources/public/js/compiled/**/*.{js,cljs,cljc,edn}`. That directory is the lein-cljsbuild and lein-figwheel `:output-to`/`:output-dir` convention, and the tree it writes holds the Closure bundle, verbatim copies of every ClojureScript source it compiled, and per-namespace `*.cljs.cache.edn` analysis caches. All four of those extensions are allowlisted, `.js` already before this PR, so without the pattern a diff that touches that tree gets `cljs/core.cljs` reviewed as if someone had written it. `resources/public/js/vendor/analytics.js` is a negative row: hand-written JS one directory over stays reviewable.

`**/.cpcache/**` and `**/.shadow-cljs/**` are the only bare directories. Both names are owned by a tool and never typed by hand, and both trees hold `.edn` files (launchpad writes `.cpcache/launchpad.edn`, shadow-cljs writes `builds/*/*/bundle-info.edn`), so both can actually fire now that `.edn` is allowlisted.

Nothing is added for `target/`. `providerDirIgnoreDirs` in `internal/diff/git.go` already drops it at the diff level, before the allowlist is reached. `out/` is left alone because real Clojure `.gitignore`s root-anchor it as `/out/`, which this matcher cannot express, and `**/out/**` would drop any directory named `out` repo-wide.

### Limitation

Clojure diffs that used to be skipped are now reviewed. Anyone who does not want that opts out with user exclude globs, which are checked before the allowlist.

## Type of Change

## How Has This Been Tested?

- `TestIsAllowedExt` — `.clj`/`.CLJ`, `.cljs`/`.CLJS`, `.cljc`/`.CLJC` and `.edn`/`.EDN`.
- `TestIsExcludedPath` — both test-file patterns at the repo root and nested; the compiled bundle, a copied `cljs/core.cljs`, a `*.cljs.cache.edn` and a subproject-prefixed output path; `.cpcache` and `.shadow-cljs`; and three negatives, `src/myapp/core.clj`, `src/myapp/test_helpers.clj` and the hand-written JS above.
- `TestResolve_DefaultRules` — `.clj`, `.cljs` and `.cljc` paths plus `deps.edn` resolve by heading anchors unique to `clojure.md`.

## Checklist

- [x] I have signed the CLA

The rule table in `pages/src/content/docs/{en,ja,ru,zh}/review-rules.md` lists the new pattern.

## Related Issues

Part of #470.


chethanuk:

wu21-web (approved):

chethanuk:

chethanuk:

chethanuk:

chethanuk:

lizhengfeng101:
Thanks for the contribution! The four-extension coverage (`.clj`, `.cljs`, `.cljc`, `.edn`) and the test cases look good.

I'd like to get a review from someone with deeper Clojure experience before merging — mainly to verify the rule doc's priorities match what matters in real Clojure codebases. Once we get that sign-off, happy to merge.

PR #982, opened 2026-08-17
feat(delegate): add full-file scan plan for host-agent delegation

Tymon-lin:
## Description

Delegation mode lets a host agent drive the review while OCR handles the deterministic engineering. Today it only covers **diff review** — `delegate preview` and `delegate rule` both describe a diff. Auditing an unfamiliar codebase, or a directory with no meaningful diff (exactly what `ocr scan` exists for), has no delegated counterpart, so that path still requires a configured LLM endpoint.

This adds **`ocr delegate scan`**, the delegation counterpart of `ocr scan`. In one call it emits the files scan would read, grouped into the same batches scan would dispatch, plus the resolved review rules. Text and `--format json`.

```bash
ocr delegate scan --path internal/agent --batch by-directory
```

`scan.BuildPlan` reuses `Preview`'s reviewability filter and `Agent.Run`'s batching, so the plan matches what `ocr scan` actually dispatches. Like `Preview` it builds no session or runner, so producing a plan cannot open session persistence.

Two deliberate choices worth reviewing:

- **`filterLargeScans` is not applied.** That ceiling exists to fit a file inside OCR's *own* prompt budget; under delegation the host agent reads the files itself, so applying it would hide large files from a host that has room for them. `Preview` makes the same choice, and a test pins the two together.
- **No git required.** Unlike `delegate preview`, a full-file scan needs no refs, so a plain directory is accepted. A test asserts `delegate scan` accepts one *and* that `delegate preview` still rejects it.

The commits are independent and can be taken separately:

| Commit | |
|---|---|
| `feat(delegate): add full-file scan plan…` | the feature above |
| `fix(suggestdiff): stop folding case…` | independent bug fix, found while dogfooding the new command |
| `refactor(scan)!: group by-directory batches…` | **breaking**, see below — drop this commit if unwanted |

### The suggestdiff fix

`ComputeLineDiff` compared lines with `strings.EqualFold`, so a suggestion whose only change was letter case rendered as a single unchanged **context line showing the OLD text** — `func foo()` → `func Foo()` displayed as though the suggestion changed nothing. In Go that rename is the exported/unexported boundary, one of the most common review suggestions there is.

Now case-sensitive. Whitespace tolerance stays: `ExistingCode` is a model-quoted excerpt whose indentation cannot be trusted, and requiring byte equality would render a whole block as rewritten whenever the model re-indented its quote. This matches `diff.normalizeLine`, which already trims whitespace but never folds case when matching `ExistingCode` against the real file.

The existing `"case insensitive match with whitespace"` case asserted the old behaviour; it is split into the two properties that are now distinct, plus regression tests for both directions.

### The breaking change — happy to drop it

`--batch by-directory` keyed on the **first** path segment, which collapsed every file beneath a single top-level directory into one batch. That is the common repository layout (`internal/`, `src/`, `pkg/`, `lib/`), so on exactly the repos the strategy was meant to serve it grouped nothing:

```
# before: 6 files across 5 packages under internal/
- batches: 1
# after
- batches: 5
```

Keying on the file's full parent directory is what *"groups related files into a single review unit"* was describing. But it does change existing behaviour — runs get more, smaller batches, meaning more sub-agent contexts for `ocr scan` — so it is the **last commit and can be dropped cleanly**. I verified that: with it removed the suite passes and the remaining comments consistently describe the old semantics.

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change) — isolated in the final commit
- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

19 new tests across `internal/scan` and `cmd/opencodereview`, including:

- `BuildPlan` agrees with `Preview` on the scannable set (the two no-LLM views must not diverge)
- excluded files never leak into a batch
- `delegate scan` works outside a Git repo while `delegate preview` still rejects one
- `--batch-size` chunking, `--path` scoping, `--background-file` merging, empty slices marshalling as `[]` not `null`
- by-directory grouping for the single-top-level-directory layout and for nested directories

Manual: ran `ocr delegate scan` against this repo, then performed the delegated review by hand from the emitted plan and rules to confirm the output is actually actionable end to end.

Note: `-race` could not be exercised in my environment (no cgo toolchain available); the plain suite passes and the new code is concurrency-free.

## Checklist

- [x] I have updated the documentation accordingly (docs/delegate in en/zh/ja/ru, the delegate skill and both mirrors, all five READMEs, and a `delegate-scan` Claude Code command)
- [ ] I have signed the CLA


lizhengfeng101:
Closing due to the CLA remaining unsigned for a long time.

PR #983, opened 2026-08-17
fix(tool): cap file_read_diff output to 500 lines (#981)

### LINKED ISSUE #981: file_read_diff output is unbounded and can exhaust the LLM context window
## OpenCodeReview version

Built from source at `ec5f6851d2fb363ec36b9f283910145ef87cf5df` (`upstream/main` on 2026-08-17). The installed CLI used to inspect the environment is `v1.9.3 (c35ddd72)`.

## Operating system

macOS (Apple Silicon), Darwin 25.5.0 arm64.

## Installation method

Built from source.

## LLM provider

Provider-independent. This report covers the size of the tool result before it is sent to an LLM.

## Bug description

`file_read_diff` concatenates the complete diff for every requested path without any output budget, truncation marker, or continuation mechanism. A sufficiently large result can consume or exceed the context available to the next LLM request and interrupt the review loop.

The current implementation appends each complete diff to one `strings.Builder` and returns it directly:

- [`internal/tool/file_read_diff.go`](https://github.com/alibaba/open-code-review/blob/ec5f6851d2fb363ec36b9f283910145ef87cf5df/internal/tool/file_read_diff.go#L54-L73)
- [`file_read_diff` tool schema](https://github.com/alibaba/open-code-review/blob/ec5f6851d2fb363ec36b9f283910145ef87cf5df/internal/config/toolsconfig/tools.json#L160-L182), which exposes only `path_array` and no pagination parameter

This differs from other read-oriented tools such as `file_read`, which limits output and reports `IS_TRUNCATED`.

## Steps to reproduce

1. Save the following temporary test as `internal/tool/file_read_diff_repro_test.go` at commit `ec5f6851d2fb363ec36b9f283910145ef87cf5df`:

```go

package tool

import (
    "context"
    "fmt"
    "strings"
    "testing"
)

func TestReproFileReadDiffUnboundedOutput(t *testing.T) {
    const (
        files        = 10
        changedPairs = 10000
    )

    diffs := make(map[string]string, files)
    paths := make([]any, 0, files)
    for fileIndex := 1; fileIndex <= files; fileIndex++ {
        path := fmt.Sprintf("file-%02d.go", fileIndex)
        var diff strings.Builder
        fmt.Fprintf(&diff, "--- a/%s\n+++ b/%s\n@@ -1,%d +1,%d @@\n", path, path, changedPairs, changedPairs)
        for line := 1; line <= changedPairs; line++ {
            fmt.Fprintf(&diff, "-old line %06d from file %02d\n+new line %06d from file %02d\n", line, fileIndex, line, fileIndex)
        }
        diffs[path] = diff.String()
        paths = append(paths, path)
    }

    provider := NewFileReadDiff(NewDiffMap(diffs))
    result, err := provider.Execute(context.Background(), map[string]any{"path_array": paths})
    if err != nil {
        t.Fatal(err)
    }

    t.Logf("output_bytes=%d output_lines=%d", len(result), 1+strings.Count(result, "\n"))
    if strings.Contains(result, "IS_TRUNCATED") || strings.Contains(result, "NEXT_OFFSET") {
        t.Fatal("unexpected truncation or continuation metadata")
    }
}
```

2. Run the repository-required unit-test entry point:

```shell
make test
```

3. Observe that `TestReproFileReadDiffUnboundedOutput` reports `output_bytes=6000860 output_lines=200051`, with neither truncation nor continuation metadata.
4. Remove the temporary test file after reproducing the issue.

Each synthetic changed-line pair has one removed line and one added line. The following table reports raw bytes and lines; it does not estimate tokens.

| Files | Changed line pairs per file | Input bytes | Output bytes | Output lines | Truncation marker | Paging metadata |
|---:|---:|---:|---:|---:|:---:|:---:|
| 1 | 10,000 | 600,058 | 600,086 | 20,006 | No | No |
| 3 | 10,000 | 1,800,174 | 1,800,258 | 60,016 | No | No |
| 10 | 10,000 | 6,000,580 | 6,000,860 | 200,051 | No | No |
| 10 | 50,000 | 30,000,580 | 30,000,860 | 1,000,051 | No | No |

All requested file markers were present in every result. The output grew linearly with the complete requested diff size; the only overhead was the per-file header and trailing newline.

This is a deterministic source-level reproduction. It does not claim that a specific LLM provider was observed returning a context-overflow error, and byte size is not treated as token count.

## Logs / Error output

N/A. This is a deterministic source-level reproduction and does not call an LLM provider.

## Expected behavior

`file_read_diff` should have a deterministic response budget. When the complete result does not fit:

1. the response should explicitly state that it is partial;
2. omitted content should remain retrievable through a follow-up call; and
3. tests should verify the output bound and continuation behavior for both a single large diff and multiple requested files.

## Proposed direction

Keep the existing `path_array` behavior compatible, add a bounded response with explicit partial-result metadata, and expose a cursor or offset that lets the model retrieve the remainder. The exact budget unit, limit, and continuation schema can be agreed before implementation rather than hard-coded in this report.

## Additional context

The existing `file_read_diff` tests cover small single-file and multi-file results but do not assert any response-size bound. I am happy to implement the agreed behavior and add focused unit tests.

chethanuk:
## Description

Closes #981. `file_read_diff` concatenates the complete diff for every requested path into one `strings.Builder` with no output budget. The result is fed back to the model on every agent round, so a large PR exhausts the context window and interrupts the review loop. The repro on the issue is 10 files at 10,000 changed line-pairs, about 200k lines and 6MB in a single tool result; one file at 50,000 pairs reaches 1M lines.

This caps the output at 500 diff content lines counted across all requested files, prefixes the result with `IS_TRUNCATED: <bool>` and appends a note when the budget runs out. That is the same shape the sibling `file_read` tool already uses: `internal/tool/file_read.go` has its own `fileReadMaxLines = 500`, the same header and the same footer, so both read tools now truncate the same way.

500 rather than something larger because the tool result is re-carried across rounds, not paid for once. `internal/agent/estimate.go` assumes `avgMainRoundsPerFile = 7` for the diff path, so 500 lines at roughly 40 tokens a line is about 20k tokens carried seven times — near 140k effective tokens from one call, which already fills a 128k context. The tool exists to spot-check a suspected issue against a related file, not to dump a diff, and 500 lines covers that.

Headers and the footer are excluded from the count, so a result of exactly 500 content lines is not falsely flagged as truncated. When the budget runs out mid-list the remaining files are dropped rather than half-written, and files with an empty diff consume none of it. The `Error: no files found` and `Error: diff not found` paths are unchanged.

### Limitation

`file_read_diff` is on the core review path, so every review now sees the `IS_TRUNCATED` line and, on large diffs, a truncated result where it previously got the whole thing. On a diff over 500 lines the model has to ask for specific files instead of receiving everything at once.

## Type of Change

## How Has This Been Tested?

`go test ./internal/tool -run TestFileReadDiffProvider_Execute`, then `make check` and `make test`.

`TestFileReadDiffProvider_Execute` gains an assertion that untruncated results carry the `IS_TRUNCATED: false` prefix. `TestFileReadDiffProvider_Execute_Truncation` adds four cases: a 501-line diff truncated to exactly 500 content lines with the footer present; a 500-line diff reported as untruncated with no footer, which pins the off-by-one; two files of 300 lines capped at 500 in total with both headers kept; and an empty diff that does not consume the budget.

## Checklist

- [x] I have signed the CLA

## Related Issues

Closes #981


lizhengfeng101:
Nice work, this looks good to me. One heads-up though: `file_read_diff` is on the core review path, so the cap changes behaviour for every review. I'd like to run a performance evaluation tonight before we merge.

If anything needs tuning after that, could you enable "Allow edits by maintainers" so I can push to your branch directly? Would save us a round-trip. Let me know if you'd rather apply the changes yourself — happy either way.

chethanuk:
Done - enabled "Allow edits and access to secrets by maintainers"

chethanuk:

chethanuk:

chethanuk:

chethanuk:

lizhengfeng101:
This caps output but there's no way to paginate — file_read has start_line/end_line but file_read_diff doesn't, so diffs over 500 lines become permanently unreadable even when requesting a single file. Might need an offset param or similar before shipping this.

chethanuk:
> so I can push to your branch directly? Would save us a round-trip

> diffs over 500 lines become permanently unreadable even when requesting a single file. Might need an offset param or similar before shipping this

@lizhengfeng101 Are you making the changes?

chethanuk:
> so I can push to your branch directly? Would save us a round-trip

> diffs over 500 lines become permanently unreadable even when requesting a single file. Might need an offset param or similar before shipping this

@lizhengfeng101 Are you making the changes?

lizhengfeng101:
> Are you making the changes?

@chethanuk Apologies for the delay; the volume of concurrent requests has been very high lately. I hope you understand, and I will handle this as soon as possible.

chethanuk:
No problem - I understand - Thank you

PR #984, opened 2026-08-17

### LINKED ISSUE #947: fix(llm): preserve opaque tool-call fields across OpenAI-compatible turns

snemetz:
Description
Preserve unknown JSON fields returned on OpenAI-compatible function tool calls and echo them on later assistant-history turns through the OpenAI SDK's supported SetExtraFields mechanism.

OpenAI-compatible Gemini endpoint puts a required thought signature under tool_calls[].extra_content. OCR previously reduced every returned tool call to id, type, and function, so a repository tool succeeded on turn one and the provider rejected turn two with HTTP 400. The fix is provider-independent: opaque extension fields survive the response-to-request round trip, while OCR continues to own id, type, and function; invalid JSON is not forwarded; and the fields remain excluded from normal OCR JSON/session serialization.

Type of Change
 Bug fix (non-breaking change that fixes an issue)
 
How Has This Been Tested?
 make test passes locally
 Manual testing (describe below)

Tests and validation:

go test -count=1 ./internal/llm -skip 'TestShellRCFiles|TestTryShellRC' passes. The skipped tests are existing Unix-home assumptions that fail under Windows because os.UserHomeDir does not follow the test's HOME override.
go vet ./... passes.
Added an HTTP-level regression test: a fake OpenAI-compatible server returns nested opaque extra_content on turn one and rejects turn two unless OCR echoes it unchanged.
Built the patched OCR binary and ran a real three-file review against Vertex AI's official OpenAI-compatible Gemini endpoint. Before this patch, all three files failed on request two with HTTP 400 and coverage was 0/3. With this patch, coverage was 3/3, the review completed, and it found the planted defect. Two additional identical runs also completed 3/3, proving the transport/tool loop is stable.
Ran OCR's required self-review command against this diff; it completed with no findings.
The full Windows suite still contains unrelated permission/Unix-shell failures, and the race target requires a CGO toolchain not present on this host. CI remains the authoritative full race run.
ocr scan no longer generates 400 errors

Checklist
 My code follows the project's coding style (go fmt, go vet)
 I have performed a self-review of my code
 I have added tests that prove my fix is effective or my feature works
 New and existing focused unit tests pass locally with my changes
 
Related Issues
Closes https://github.com/alibaba/open-code-review/issues/947

### DIFF
--- a/internal/llm/client.go
+++ b/internal/llm/client.go
@@ -125,9 +125,10 @@ type Choice struct {
 
 // ToolCall represents a function call requested by the model.
 type ToolCall struct {
-	ID       string       `json:"id"`
-	Type     string       `json:"type"`
-	Function FunctionCall `json:"function"`
+	ID          string                     `json:"id"`
+	Type        string                     `json:"type"`
+	Function    FunctionCall               `json:"function"`
+	ExtraFields map[string]json.RawMessage `json:"-"`
 }
 
 // FunctionCall holds the name and arguments of a tool call.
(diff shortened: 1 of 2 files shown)


snemetz:
Too many windows open. Created PR to wrong repo

PR #953, opened 2026-08-16
feat(allowlist): add Elm (.elm) support

wakqasahmed:
## Summary
- Add `.elm` to the supported file types allowlist
- Add `**/tests/**/*.elm` to the default exclude patterns (Elm's conventional top-level `tests/` directory per elm-test / elm.json test-dependencies convention)
- Add `internal/config/rules/rule_docs/elm.md`, an Elm-specific review rule doc covering The Elm Architecture (Model/Update/View), `Maybe`/`Result` handling instead of exceptions, port/JS interop safety, JSON decoder/encoder correctness, `Debug.log`/`Debug.todo` production hygiene, and `elm.json` semantic versioning
- Register `**/*.elm -> elm.md` in `system_rules.json`
- Add test coverage: allowed-extension matching (`.elm`/`.ELM`), exclude-pattern matching (`tests/**/*.elm`), and rule resolution

Part of the language-allowlist expansion tracked in #470.

Closes #940

## Test plan
- [x] `go test ./internal/config/allowlist/... -run 'TestIsAllowedExt|TestIsExcludedPath' -v` — PASS
- [x] `go test ./internal/config/rules/... -run TestResolve_DefaultRules -v` — PASS
- [x] `gofmt -s -l .` — clean
- [x] `go vet ./...` (excluding /extensions/) — clean
- [x] `scripts/verify-english-only.go` — clean (411 files scanned, no unapproved non-English text)

PR #937, opened 2026-08-16
feat(allowlist): add Zig support

ousamabenyounes:
## Description

Adds built-in Zig support to the review allowlist so `.zig` changes are reviewed with a Zig-specific ruleset instead of being skipped (extension not recognized) or falling back to the generic `default.md` checklist. This is the Zig language-group sub-issue tracked under #470.

- `internal/config/allowlist/supported_file_types.json` — recognize `.zig`
- `internal/config/allowlist/default_exclude_patterns.json` — exclude conventional Zig tests: `**/test/**/*.zig` and `**/*_test.zig`
- `internal/config/rules/rule_docs/zig.md` — new precision-focused Zig review rules (memory safety / illegal behavior across build modes, allocators + `defer`/`errdefer` cleanup, error unions & optionals, comptime/generics/`build.zig`, concurrency & C interop)
- `internal/config/rules/system_rules.json` — map `**/*.zig` → `zig.md`
- extension, test-exclusion, and rule-resolution tests

Note on the exclude patterns: the tracking issue suggested `**/test/*.zig`, but this PR uses the nested `**/test/**/*.zig` for consistency with the already-merged Julia (`**/test/**/*.jl`) and Haskell (`**/test/**/*.hs`) patterns, so tests under `src/test/unit/*.zig` are also excluded. `**/*_test.zig` covers the suffix convention.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## How Has This Been Tested?

- [x] `make test` passes locally
- [x] Full local CI replay in the pinned `golang:1.26.6` container: `gofmt -s -l .` clean, `go vet ./...` clean, `go build ./cmd/opencodereview`, `go test -race -count=1 ./...` all green (total coverage 91.2%), `go mod tidy` leaves `go.mod`/`go.sum` unchanged.

### RED → GREEN

New tests fail on the unmodified base (before the allowlist/rule additions):

```
--- FAIL: TestIsAllowedExt/.zig
    allowed_ext_test.go: IsAllowedExt(".zig") = false, want true
--- FAIL: TestIsExcludedPath/zig_test_directory
    allowed_ext_test.go: IsExcludedPath("test/parser.zig") = false, want true
--- FAIL: TestIsExcludedPath/zig__test_suffix
    allowed_ext_test.go: IsExcludedPath("src/parser_test.zig") = false, want true
--- FAIL: TestResolve_DefaultRules/src/main.zig
[...]

PR #988, opened 2026-08-18
feat(lang): review support for R

wu21-web:
## Description
Adds review support for R.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [x] I have signed the CLA

## Related Issues

None yet

### DIFF
--- a/internal/config/allowlist/allowed_ext_test.go
+++ b/internal/config/allowlist/allowed_ext_test.go
@@ -25,6 +25,8 @@ func TestIsAllowedExt(t *testing.T) {
 		{".phtml", true},
 		{".PHTML", true},
 		{".rs", true},
+		{".R", true},
+		{".r", true},
 		{".ets", true},
 		{".ETS", true},
 		{".json5", true},
@@ -178,6 +180,12 @@ func TestIsExcludedPath(t *testing.T) {
 		{"lhs spec file", "src/ParserSpec.lhs", true},
 		{"lhs root spec file", "ParserSpec.lhs", true},
 		{"lhs non-test", "src/Tutorial.lhs", false},
+
+		// R test files
+		{"r test directory", "tests/parser_test.R", true},
+		{"r nested test directory", "packages/core/tests/unit/parser_test.R", true},
+		{"r non-test", "src/parser.R", false},
+		{"r tests in filename", "src/tests_helper.R", false},
 
 		// Nim test files
 		{"nim test directory", "tests/parser_test.nim", true},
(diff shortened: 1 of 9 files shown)


wu21-web:
## Code review: R-language support wiring (`high` effort)

### Conclusion — functional wiring is complete and correct ✅

Verified by running the full `internal/config/...` test suite and a probe confirming that `foo.R`, `foo.r`, and `src/parser.R` all resolve to the R rule (`r.md`), while `tests/testthat/test-x.R` is correctly excluded. The allowlist, case-insensitive rule map, excludes, `TestSystemRulesIntegrity`, and all four locale docs (en/ja/ru/zh) are wired correctly.

The gaps are in **test/documentation wiring**, not runtime wiring.

### Findings

1. **`internal/config/rules/system_rules_test.go`** — no R case in `TestResolve_DefaultRules`
   Every prior language PR (Elm #953, Zig #937) added a case to `TestResolve_DefaultRules`, the one layer that verifies a resolved path reaches the right rule. R has none, so the `**/*.R` glob is unverified. This isn't academic: the branch's original commit shipped `**/.R` (a glob that matches only a file literally named `.R`, i.e. nothing) and **no test caught it** — it survived 3 commits until the manual fixup `68653aa`. `TestSystemRulesIntegrity` can't catch this because `pattern_validity` accepts `**/.R` as a valid glob and `extensions_are_allowlisted` skips it (globExt requires a base starting with `*.`, and `.R`'s base doesn't). A row like `{"src/foo.R", "R Code Review Principles"}` — as added for elm and zig — would have failed `Resolve("src/foo.R")` immediately.

2. **`internal/config/rules/rule_docs/r.md`** — only rule doc missing a trailing newline
   `elm.md`, `zig.md`, `nim.md`, and `julia.md` all end with `\n`; `r.md` does not. Cosmetic only — `LoadDefault` runs `strings.TrimRight(content, "\n")`, so there's no runtime effect — but it's inconsistent with every sibling doc and will surface as a spurious diff in any future append/concatenate or diff-based docs check.

3. **`pages/src/content/docs/en/review-rules.md`** — docs precedence order disagrees with `system_rules.json`
   The docs list the `**/*.R` → `r.md` row after `**/*.rs` (rust); `system_rules.json` places it after `**/*.jl` (julia). No behavioral impact since `.R` overlaps no other pattern, but the precedence shown in review-rules.md (and the ja/ru/zh copies) no longer matches the authoritative JSON order, which can mislead anyone auditing rule precedence or diffing the tables against the JSON.

lizhengfeng101 (approved):
LGTM

PR #994, opened 2026-08-18
fix(agent): recover from hallucinated file paths

### LINKED ISSUE #989: file_read tool accepts paths not in the current commit, driving review to status: partial
## OpenCodeReview Version

- Reproduced on `v1.9.4 (f7344e79) windows/amd64` (production)
- Also affected on `v1.9.5 (ea1de41a windows/amd64)` (local)
- Installed via `npm install -g @alibaba-group/open-code-review`

## Operating System

Windows (Jenkins agent on Windows Server; reproducible from any platform)

## Installation Method

npm (global)

## LLM Provider / Model

`deepseek` / `deepseek-v4-flash` (also seen to reproduce on other providers per #287)

## Bug description

During `ocr review --commit <sha>` on a focused change, the LLM independently calls its `file_read` tool with paths that **do not exist in the commit at all**. OCR executes the read via `git show <commit>:<path>`, git returns "fatal: path ... does not exist in <sha>", and the failed tool call is reported back to the LLM. When the hallucination is severe enough (or when the LLM is persistent), the review run goes from `status: complete` to `status: partial`, and the Jenkins pipeline fails OCR Review.

Three consecutive failed builds on the same Gerrit Change 4700 (different patchsets rebased on top of each other) all failed with the same root cause, in slightly different ways:

| Patchset | Commit SHA | Hallucinated path (file_read) | File actually exists? |
|---|---|---|---|
| 1 | `b087039` | `DOC/云配置项/project.yct` | **Never existed** (path entirely invented by LLM) |
| 4 | `d86b0f1` | `App/MID/RF/CendricApp/src/phscaCendricCadsApp.c` | **Deleted in this commit** (file lived in history but this PR removed the entire Cendric tree) |
| 4 | `d86b0f1` | `App/MID/RF/CendricSwTimer/src/phscaCendricCadsSwTimer.c` | **Deleted in this commit** (same) |

The hallucination has two distinct mechanisms:

1. **Imagined paths** that were never in the repo (e.g. `DOC/云配置项/project.yct` — possibly extrapolated from the commit subject's mention of "切换 RF 芯片").
2. **Deleted-in-this-commit paths** that the LLM remembers from the project's history / training data (e.g. Cendric legacy files — this commit removes the Cendric tree, but the LLM still calls file_read on its members).

OCR executes the read either way, and each call that fails contributes to driving the run into `partial`.

## Steps to reproduce

The exact conditions are LLM-specific, but a high-rate reproduction recipe:

1. Pick a focused commit that touches a small number of files in one area, AND that **removes or renames** other files elsewhere in the repo.
2. Run `ocr review --commit <sha> --provider deepseek --model deepseek-v4-flash --format json --timeout 30` (or any comparable review).
3. After ~5–10 minutes, observe:
   - `status: partial` instead of `status: complete`
   - In the logs: `✘ file_read failed: file "<hallucinated path>" not found: git show <sha>:<path>: ... fatal: path '<path>' does not exist in <sha>` — one or more such entries
   - No `code_comment` tool call on the actually-changed files; the LLM has burned the LLM budget retrying the missing file or pivoted away

A simpler deterministic repro: hand-craft a prompt that says "review this commit" while seeding the LLM with a "looks like a project config file at `DOC/云配置项/project.yct>"-style misdirection. The model will call `file_read("DOC/云配置项/project.yct")` and OCR will fail it.

## Expected behavior

The `file_read` tool is supposed to be a fallback for the LLM to inspect a file it has been told about (typically the changed files in the diff, or files referenced by changed code). It is **not** supposed to allow free-form exploration of unrelated paths in the commit's history.

Desired properties:

- `file_read` accepts only paths that **actually exist in the current commit** (the diff's `Ref`); a non-existent path is rejected up-front, not after a wasted `git show` call.
- Optionally, `file_read` is constrained to the changed files in the diff (or to a small allowlist provided by the user via `--include`). A focused `ocr review` of a 10-file commit should never let the model fetch a 200-file tree.
- The `--exclude` / `--path` family of flags already exists for `ocr scan`; review mode should expose the same controls (`--include` for review, or a path allowlist knob).
- When a `file_read` call is rejected, OCR should report it as a deterministic structural error, not a transient failure that the retry loop can re-attempt. (Today, a hallucinated path makes `status: partial` because the retry treats it as flaky.)
- `existing_code` matching and the diff/filter pipeline should not be perturbed; the constraint is purely on the input side.

## Proposed direction

Add a path-validity check on `file_read` (and the related read-oriented tools) before the git call, against the same `Ref` the read is intended to resolve into. Two reasonable shapes:

1. **Hard allowlist**: read paths must be a subset of the changed files in the diff (or a user-supplied `--include` set). Anything else returns a clean "path is not in this review scope" and is not retried.
2. **Soft guard**: read paths must be a path that exists at the ref. If it does not, return a one-line "no such file" error and do not contribute to the partial status (or trigger retry).

Either shape kills the hallucination-driven partial-status path. The hard allowlist is closer to what `ocr scan --path` already supports; the soft guard is a smaller change.

A related improvement, separately tracked in #981, is bounding `file_read_diff` output size; the two are complementary and should not be conflated.

## Acceptance criteria

- [ ] A `file_read` request for a path that does not exist in the current commit's tree returns a deterministic structural error and is **not** retried by the partial-status retry loop.
- [ ] A `file_read` request for a path that is not in the diff's changed files (and not in a user-supplied allowlist) is rejected with a clear "out of review scope" message.
- [ ] The 3 reproduction patterns above (imagined path, deleted-in-commit path) no longer drive `ocr review` to `status: partial`.
- [ ] Normal review flow on focused commits is unchanged (existing legitimate reads still work).
- [ ] A focused test exercises each reproduction pattern and asserts the review run ends in `status: complete` with the expected comment set.

## Affected Area

Built-in Tools (file_read, code_search, etc.); review pipeline (partial-status retry policy).

## See also

- **#287** — `file_read` call failures in a monorepo (closed as completed, no functional fix shipped). Different root cause (relative path resolution in subdirectory) but the same symptom: `file_read` produces a "fatal: path ... does not exist" message, and the run goes partial.
- **#981** — `file_read_diff` output is unbounded and can exhaust the LLM context window. Same tool family, different overflow mode.
- **#167** — Near-duplicate comments with wrong filenames. Different category but tangentially related to model/filename mismatches.

## Logs

Sample (Patchset 4, v1.9.4, `deepseek-v4-flash`):

```
11:14:40  [OCR] attempt 1/2 status=partial elapsed=541s, retrying…
11:15:18  [ocr]   ✘ file_read failed: file "App/MID/RF/CendricApp/src/phscaCendricCadsApp.c" not found: git show d86b0f1:App/MID/RF/CendricApp/src/phscaCendricCadsApp.c: exit status 128: fatal: path 'App/MID/RF/CendricApp/src/phscaCendricCadsApp.c' does not exist in 'd86b0f1'
11:15:18  [ocr]   ✘ file_read failed: file "App/MID/RF/CendricSwTimer/src/phscaCendricCadsSwTimer.c" not found: git show d86b0f1:App/MID/RF/CendricSwTimer/src/phscaCendricCadsSwTimer.c: exit status 128: fatal: path 'App/MID/RF/CendricSwTimer/src/phscaCendricCadsSwTimer.c' does not exist in 'd86b0f1'
11:24:25  [ERROR] OCR 未完整结束,status=partial
```

xwil1:
## Summary

Fixes #989.

This PR implements the smaller soft-guard direction described in the issue.

- validate `file_read` paths against the reviewed Git ref before `git show`
- classify missing-at-ref paths as deterministic structural rejections
- after three consecutive rejected reads, use the existing `file_find` provider to return current-ref candidates to the same conversation
- refund at most three invalid-only rounds so the agent has bounded room to return to the review mainline
- preserve the tool schema and append-only conversation history, so provider prefix caching is not invalidated
- preserve Unicode and space-containing paths in `file_find` by using raw NUL-delimited Git output

## Root cause and behavior

`file_read` previously passed a model-supplied path directly to `git show`. A hallucinated or deleted path produced an opaque Git failure while still consuming the main-task tool budget. A persistent model could spend the rest of the review retrying nonexistent paths and leave the item partial without reviewing the changed file.

This change detects that deterministic condition before `git show` and, after a bounded threshold, gives the model concrete paths from the reviewed ref or tells it to return to the current diff. It does not auto-select a candidate and does not report completion on the model's behalf. A model that keeps ignoring the recovery result still ends with `StopMaxRounds`/partial.

The implementation also incorporates the initial automated review feedback:

- failed candidate searches are not cached, so a later recovery batch can retry them
- equivalent model paths such as `./src/main.go`, `src//main.go`, and backslash-separated paths are normalized before an exact file match
- successful candidate results remain cacheable because the reviewed ref is immutable during the run

This PR intentionally implements the smaller soft-guard direction from #989. Restricting reads to diff files or adding a review `--include` allowlist remains separate work.

## Test footprint

The PR adds 420 lines across three focused test files:

- `internal/llmloop/path_recovery_test.go`: 328 added lines
- `internal/tool/file_find_test.go`: 42 added lines
- `internal/tool/file_read_test.go`: 50 added lines

The main loop regression covers three rejected reads followed by a valid read, a review comment, and `task_done`. Additional cases cover invented, deleted, and moved paths; successful-read reset; bounded refunds; retry after a transient candidate-search error; Unicode and space-containing filenames; and equivalent non-canonical paths.

The full PR is 8 files with 704 additions and 12 deletions. A duplicate 325-line command-level test was removed during review; the retained tests exercise the same production runner and tool implementations without duplicating the harness.

## Validation

- `make check`
- `make test`
- `make coverage` (`91.6%`, threshold `90%`)
- `go test -race ./internal/tool -run 'TestRead_GitShow_NormalizesEquivalentPaths|TestReadLines_GitShow_MissingPathIsStructuralError' -count=10`
- `go test -race ./internal/llmloop -run 'TestRecoverInvalidPathsDoesNotCacheFinderErrors|TestRunPerFile_(RecoversCandidatesAfterRepeatedMissingPath|RecoversInventedDeletedAndRenamedPaths|SuccessfulReadResetsMissingPathBatch|MissingPathRefundIsBounded)' -count=10`


Tec-yao:
PR #994 currently handles missing paths only when FileReader reads from a Git ref. In ocr scan mode, FileReader uses ModeWorkspace, so missing files return ordinary os.ErrNotExist errors and are not classified as PathNotAtRefError. As a result, the new invalid-path recovery is not triggered for full workspace scans. Please extend the structural path error classification to readFromDisk and readLinesFromDisk, while preserving errors.Is(err, os.ErrNotExist) compatibility.

PR #1011, opened 2026-08-20
test(llm): cover tool_choice mapping for all three provider protocols

lizhengfeng101:
## Summary

- Add tests for `buildOpenAIParams` (Chat Completions), `buildAnthropicParams` (Anthropic Messages), and `buildResponsesParams` (OpenAI Responses) that assert how `ChatRequest.ToolChoice = "required"` is translated into each provider's own tool-forcing representation.
- Each test also covers the "not applicable" cases: an unmapped value like `"auto"`, an empty `ToolChoice`, and `"required"` with no tools attached (the mapping is gated on `len(tools) > 0` in all three clients).

## Why

The review-filter task (`internal/agent`) relies on `ToolChoice: "required"` so the model is forced to call one of `report_incorrect_comments` / `approve_all_comments` rather than answering in free text. Each protocol translates that string differently:

- Chat Completions: bare string `"required"` on `ToolChoice.OfAuto`
- Anthropic: `{"type":"any"}` via `ToolChoiceAnyParam`
- Responses API: `ToolChoiceOptionsRequired` on `ToolChoice.OfToolChoiceMode`

None of these three mappings had direct test coverage, so a refactor touching any of them could silently break the "must call a tool" guarantee without any CI signal.

## Test plan

- [x] `go test ./internal/llm/... -run ToolChoice -v` — all new subtests pass
- [x] `go build ./...`
- [x] `go vet ./internal/llm/...`
- [x] `go test ./internal/llm/...` — full package suite passes

### DIFF
--- a/internal/llm/client_params_test.go
+++ b/internal/llm/client_params_test.go
@@ -70,6 +70,45 @@ func TestBuildOpenAIParams_Minimal(t *testing.T) {
 	}
 }
 
+// TestBuildOpenAIParams_ToolChoice locks the tool_choice mapping the review-filter
+// task depends on (agent.go sets ChatRequest.ToolChoice="required" so the model
+// cannot silently skip both filterTools). openai.ChatCompletionToolChoiceOptionUnionParam
+// serializes OfAuto inline as a bare string, so this is the field that must carry
+// "required" onto the wire — and only when tools are actually attached.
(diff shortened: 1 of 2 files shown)


MuoDoo (approved):
LGTM. CI is also passed.

PR #1026, opened 2026-08-20
fix(scan): join background memory compression before session finalization

### LINKED ISSUE #1025: fix(scan): join background memory compression before session finalization
### Problem Statement

In `internal/scan/agent.go` (`scan.Agent.Run`), when full-scan mode dispatches subtasks across files, background memory compression jobs may be triggered on `llmloop.Runner` via `triggerAsyncCompression` (registered via `r.bg.Add(1)`).

While `internal/agent/agent.go:385` (`Agent.Run`) explicitly calls `a.runner.WaitBackground()` prior to `a.session.Finalize()`, `internal/scan/agent.go` omits this join. Consequently, if a conversation triggers background compression, the main scan routine can exit and finalize/close the session writer while background goroutines are still executing their HTTP requests, leading to potential goroutine leaks and race conditions against closed session file handles.

### Proposed Solution

Add `a.runner.WaitBackground()` in `scan.Agent.Run` immediately before `a.session.Finalize()`, matching the pattern in `internal/agent/agent.go`.

### Claim Statement

*I have reproduced this issue with a targeted test case and have an idiomatic fix prepared. Please assign this to me, I will submit a PR shortly.*

MeiSiristhebest:
### Problem Description
Fixes #1025

In full-scan mode (`scan.Agent`), subtasks share `llmloop.Runner` for per-file reviews. When a conversation exceeds the memory compression soft threshold (60%), `triggerAsyncCompression` spawns background goroutines registered via `r.bg.Add(1)`.

While `internal/agent/agent.go:385` (`Agent.Run` in diff-review mode) explicitly calls `a.runner.WaitBackground()` before `a.session.Finalize()`, `internal/scan/agent.go` omitted this join barrier. As a result, background compression jobs could outlive the main scan lifecycle and attempt to write records after `session.Finalize()` closes the session file writer.

### Key Implementation Changes
- **`internal/scan/agent.go`**: Call `a.runner.WaitBackground()` in `scan.Agent.Run` prior to calling `a.session.Finalize()`, ensuring full parity with `internal/agent/agent.go`.
- **`internal/scan/agent_test.go`**: Add `TestScanAgent_WaitBackground_NoLeakOnRun` unit regression test.

### Verification & Empirical Evidence
- `go test ./internal/scan/... -v` passes cleanly.
- `go test ./internal/llmloop/... ./internal/agent/...` passes cleanly.
- All unit and integration tests verified locally.

### DIFF
--- a/internal/llmloop/loop.go
+++ b/internal/llmloop/loop.go
@@ -118,8 +118,8 @@ func NewRunner(deps Deps) *Runner {
 // ultimately depends on the LLM client honouring context cancellation, and no
 // additional deadline is imposed here: the job already carries its own
 // timeout.
-// Scan does not currently call this because it freezes no retry report; its
-// analogous session-finalization race is outside this change.
+// Both diff-review and scan modes call this prior to session finalization so
+// background compression goroutines are joined before session_end is written.
 func (r *Runner) WaitBackground() {
 	r.bg.Wait()
 }
--- a/internal/scan/agent.go
+++ b/internal/scan/agent.go
@@ -372,6 +372,12 @@ func (a *Agent) Run(ctx context.Context) ([]model.LlmComment, error) {
 	// Project-level summary runs after all batches; never blocks return.
 	a.maybeRunProjectSummary(ctx, comments)
 
+	// Join background memory compression before anything finalizes the session.
+	// Those jobs are cancelled rather than awaited when a conversation ends, so
+	// their LLM request can still be in flight here; joining them ensures no
+	// background goroutine is left leaking or writing to a finalized session.
+	a.runner.WaitBackground()
+
 	// A persistence failure is a delivery error in its own right: when the scan
 	// also failed, both facts are reported (errors.Join) rather than letting the
 	// dispatch error hide the fact that session_end never reached disk.
--- a/internal/scan/agent_test.go
+++ b/internal/scan/agent_test.go
@@ -416,6 +416,12 @@ func (c *blockingScanCompressionClient) CompletionsWithCtx(ctx context.Context, 
 		}, nil
 	}
 
+	// Round 2+: wait until the background compression goroutine has reached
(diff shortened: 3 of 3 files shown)


lizhengfeng101 (commented):
The fix itself looks right, but the test doesn't actually exercise it.

`TestScanAgent_WaitBackground_NoLeakOnRun` points `RepoDir` at an empty temp dir, so we discover 0 files and take the `reviewable == 0` early return at agent.go:336-346. That branch finalizes and returns at line 341, well before the `WaitBackground()` call you added at line 378. I deleted that line locally and re-ran the test — still passes. So it gives us no protection against the regression it's named after.

Coverage-wise it's a no-op too: `TestRun_NoReviewableFiles` (coverage_test.go:549) already walks the same skip path, and the block containing the new call was already being executed by other tests before this PR (just never with a compression job pending). I'd rather drop the test than leave something that reads like a safety net but isn't.

For a test that would actually catch this, `internal/llmloop/retry_background_test.go` has the pattern worth copying: `newBlockingCompressionClient` with `started`/`release` channels. Roughly: give the template a `MemoryCompressionTask` and a small enough `MaxTokens` that the soft threshold trips, put one real file in the repo, have the fake client close `started` and then block on `<-release` when it sees the `memory_compression` request. Then run `Run` in a goroutine, wait for `started`, assert `Run` hasn't returned, release, and after it returns assert `session_end` is the last line of the session JSONL *and* that the compression `llm_request` record is there. The second half of that assertion matters — without it the test can quietly go vacuous again.

One more thing while you're in here: the `WaitBackground` doc comment in llmloop/loop.go still ends with "Scan does not currently call this because it freezes no retry report; its analogous session-finalization race is outside this change." This PR makes that false, and it's the kind of comment someone will trust later.

lizhengfeng101 (approved):
LGTM

PR #1036, opened 2026-08-22
feat(cli): post review findings to GitHub PRs

### LINKED ISSUE #91: OCR should optionally post its findings to the GitHub PR Issue
### Problem Statement

Ideally OCR is runnable from an action or remote service and it can post its findings to the GitHub PR it was triggered from. Even better would be providing inline code comments and suggestions.

### Proposed Solution

I imagine it'd be a decent amount of work to implement, but maybe not a huge amount with the github MCP providing the API.

### Alternatives Considered

_No response_

### Affected Area

Review Agent / LLM interaction, Output / Formatting

### Additional Context

_No response_

ThalesMMS:
## Description

Add opt-in GitHub pull-request publishing to range reviews:

- introduce `--post-to-pr` and `--github-token`, with `GITHUB_TOKEN` fallback;
- discover the unique open pull request from the reviewed head commit and base branch, including fork remotes;
- post only verified right-side diff locations as inline review comments;
- preserve unanchored, old-side, oversized, or failed inline findings in bounded summary batches;
- revalidate the pull request's head, base, and open state before writes;
- reconcile ambiguous review and issue-comment writes with stable markers to avoid duplicate retries;
- validate GitHub.com and GitHub Enterprise endpoint boundaries before transmitting credentials;
- keep posting provenance out of `LlmComment` and public review JSON, while persisting an internal sidecar in resumable session checkpoints;
- document the complete English CLI workflow, permissions, and Enterprise configuration.

Posting remains disabled unless `--post-to-pr` is explicitly supplied, and it is restricted to `--from`/`--to` range reviews.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Validation performed:

- `make check`
- `make test` with the race detector
- `make coverage` — 90.8%, above the 90% gate
- `make build`
- focused GitHub client and posting regression tests
- pages lint, test, typecheck, build, and bundle-size checks
- mandatory pre-commit OCR review — 0 findings
- end-to-end posting on [ThalesMMS/spark-vllm-docker#14](https://github.com/ThalesMMS/spark-vllm-docker/pull/14#pullrequestreview-4999216849), which created a GitHub review with verified inline comments; the subsequent hardening paths are covered by focused integration tests

## Checklist

- [x] I have signed the CLA

## Known Limitation

An extremely large generated suggestion that must be split across summary-comment batches may display a split Markdown fence. All finding content remains preserved; this was classified as a non-blocking presentation-only edge case during the final re-review.

## Related Issues

Closes #91


internal/model/review.go
```diff
@@ -11,6 +16,7 @@ type LlmComment struct {
 	ExistingCode   string `json:"existing_code,omitempty"`
 	StartLine      int    `json:"start_line"`
 	EndLine        int    `json:"end_line"`
+	Side           string `json:"-"`
```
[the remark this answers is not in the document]
ThalesMMS: Fixed in c3b0cc6. The field now documents RIGHT/LEFT diff-side provenance, its use for internal GitHub posting decisions, and its exclusion from the public review JSON schema.

internal/github/client.go
```diff
@@ -0,0 +1,507 @@
...
+			if pull.Head.SHA == headSHA && pull.Base.Ref == baseBranch {
+				matches = append(matches, pull.Number)
+			}
+		}
+		if len(pulls) < pageSize {
+			break
+		}
+		if page == maxResultPages {
+			return 0, fmt.Errorf("open pull request list exceeds %d entries", pageSize*maxResultPages)
+		}
```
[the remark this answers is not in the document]
ThalesMMS: Leaving this as-is because PR discovery must prove uniqueness across the complete open-PR result set. If the 3,000-entry safety cap is reached, an unseen page may still contain a second PR with the same base and head SHA. Returning the earlier match would violate the exact-target guarantee; failing closed at the bounded pagination limit is intentional.

cmd/opencodereview/review_cmd.go
```diff
@@ -563,3 +618,676 @@ func buildToolRegistry(collector *tool.CommentCollector, fr *tool.FileReader) *t
...
+
+func verifyGitHubPRTarget(ctx context.Context, client *github.Client, reviewedHead, expectedBase string) error {
+	pull, err := client.GetPRInfo(ctx)
+	if err != nil {
+		return fmt.Errorf("read PR target: %w", err)
+	}
+	if pull.Head.SHA != reviewedHead {
+		return fmt.Errorf("PR head changed from reviewed commit %s to %s; rerun the review before posting", reviewedHead, pull.Head.SHA)
+	}
+	if pull.Base.Ref != expectedBase {
```
Qiyuanqiii: [P1] Bind posting to the exact reviewed base

This check only compares the base branch name. A local or fork `main` can resolve to commit A while the discovered upstream PR's `main` resolves to commit B; with the same head H, this accepts the target even though OCR reviewed `merge-base(A,H)..H` and GitHub is showing `merge-base(B,H)..H`. The changed-file inventory does not make that safe: findings outside the PR diff are deliberately emitted as summary comments, and overlapping context lines can also pass the numeric range check. Please capture the PR base SHA and verify its actual merge base (or another exact diff identity) against the sealed reviewed base before every write, and include that identity in the retry marker.

cmd/opencodereview/review_cmd.go
```diff
@@ -563,3 +618,676 @@ func buildToolRegistry(collector *tool.CommentCollector, fr *tool.FileReader) *t
...
+	writeField := func(value string) {
+		_, _ = fmt.Fprintf(h, "%d:", len(value))
+		_, _ = h.Write([]byte(value))
+	}
+	writeField(target.repository.Owner)
+	writeField(target.repository.Name)
+	writeField(fmt.Sprint(target.prNumber))
+	writeField(baseBranch)
+	writeField(reviewedHead)
+	for _, comment := range comments {
```
Qiyuanqiii: [P2] Canonicalize findings before deriving retry markers

`comments` is collected from concurrent per-file reviews, and `CommentCollector.Add` preserves completion order, so the same set of findings can arrive in a different order on a retry. Hashing the slice in that order produces a different run ID (a reverse-order regression changes `0f7d13d933f4029873a30834` to `dc3a0144ccbc01a36bacb872`), causing all existing batch markers to be missed and duplicate PR comments to be written. Canonicalize a posting copy before both hashing and batching, or content-address each batch; sorting only inside this hash would still leave unstable batch contents.

cmd/opencodereview/review_cmd.go
```diff
@@ -563,3 +618,676 @@ func buildToolRegistry(collector *tool.CommentCollector, fr *tool.FileReader) *t
...
+	for start := 0; start < len(ghComments); start += githubReviewBatchSize {
+		end := min(start+githubReviewBatchSize, len(ghComments))
+		body := "OpenCodeReview inline findings (continued)."
+		if start == 0 {
+			body = reviewBody
+		}
+		batchIndex := start / githubReviewBatchSize
+		marker := githubPostingMarker(reviewRunID, "review", batchIndex)
+		body = marker + "\n" + body
+		landed, reconcileErr := githubReviewExists(ctx, ghClient, reviewedHead, marker)
```
Qiyuanqiii: [P2] Resume an existing fallback instead of retrying inline

This preflight only checks the inline-review marker. If an earlier run received a non-ambiguous inline error and successfully posted the remaining findings under a `fallback-N` marker, an identical rerun sees no review marker and retries the inline write; once that error is gone, the same finding is present in both the existing fallback summary and the new inline review. Detect an existing fallback sequence for this batch and reconstruct or resume any missing fallback batches instead of switching back to inline delivery.

Qiyuanqiii (commented):
I found three actionable correctness issues in the PR-posting safety and retry paths.

internal/github/client.go
```diff
@@ -0,0 +1,507 @@
...
+	"os"
+	"strings"
+	"time"
+)
+
+const (
+	apiVersion     = "2022-11-28"
+	pageSize       = 100
+	maxResultPages = 30
+)
```
ThalesMMS, line 24: **[maintainability · medium]**

The pagination logic limits retrieval to 3000 items (maxResultPages * pageSize). If a PR exceeds this, the function will return an error. Consider increasing the limit or handling partial pages more flexibly.

internal/github/client.go
```diff
@@ -0,0 +1,507 @@
...
+}
+
+// ParseRepoInfo parses repository owner and name from a Git remote URL.
+func ParseRepoInfo(remoteURL string) (owner, repo string, err error) {
+	repository, err := ResolveRepository(remoteURL)
+	if err != nil {
+		return "", "", err
+	}
+	return repository.Owner, repository.Name, nil
+}
```
ThalesMMS: **[maintainability · medium]**

ParseRepoInfo now depends on environment variables because it calls ResolveRepository. This might change behavior for callers expecting a pure string parsing function.

ThalesMMS (commented):
## OpenCodeReview Summary

2 comment(s) generated.

- Inline findings: 2
- Summary-only findings: 0

### Severity breakdown:

- **medium**: 2

---
*Generated by OpenCodeReview*

ThalesMMS:
/\ example above

wu21-web:
@ThalesMMS The resolve to badges function is failing.
You need to convert the tag to markdown with badge when posting the review via API.

wu21-web:
Also why are the tags bold?

wu21-web:
Other site docs translations are also missing.

internal/github/client.go
```diff
@@ -0,0 +1,507 @@
...
+	httpClient *http.Client
+}
+
+// Repository is a parsed GitHub repository bound to its validated API
+// configuration.
+type Repository struct {
+	Owner string
+	Name  string
+
+	configuration Configuration
```
wu21-web, line 51: I would move this to the Client struct (R52)

lizhengfeng101:
I've read through the whole thing — the client package, the posting orchestration, and the diff inventory.

Wanting this feature isn't in question, and `internal/github/client.go` is the strongest part of the PR. Treating the server URL, the API endpoint and the remote host as a single credential boundary, and bailing out *before* the token goes on the wire when they disagree, is exactly right. The GitLab-remote rejection test is the kind of negative test I wish more PRs had. Same for matching PRs by head SHA instead of guessing a number off a branch name — correct call for forks, and you tested it.

My main concern is architectural. I want to lead with the reason rather than the preference, because it has already cost you a bug.

The PR reaches into three places that have nothing to do with GitHub:

- `internal/model/review.go` gains a `Side` field on `LlmComment`. `"RIGHT"` / `"LEFT"` is GitHub review API vocabulary, and it now lives in the shared data model.
- `internal/diff/resolver.go` sets it in four places, so the resolver now holds an opinion about posting sides.
- `validateResumeIdentity` becomes `state == nil && !opts.postToPR`, so a feature unrelated to resume borrows resume's sealing path, and `executeReviewContext` picks up an extra `ResolveLineNumbers` call that shifts the ordering of the main flow.

Then the bug. `Side` is tagged `json:"-"`, so it doesn't persist. On `--resume`, reused findings come straight out of the session record into the collector (`internal/agent/agent.go:825-827`) carrying `StartLine`/`EndLine` but no `Side`. `ResolveLineNumbers` early-returns on anything that already has a line number (`internal/diff/resolver.go:37`), so `Side` never gets filled in. Every reused finding then hits the first switch arm and is demoted to a summary comment with the reason "line side provenance is not RIGHT" — even when it sits squarely on an added line. And the combination isn't rejected: you explicitly taught `validateResumeIdentity` to support it and added a test for it.

That's the cost of adding a field to a shared model — you have to answer what it does at every serialization boundary, and one of them got missed.

The fix I'd suggest also deletes code. Derive the side at posting time instead of threading it through the core. You're already fetching `GET /pulls/{n}/files` to build the inventory, so build **two** range tables per file, right and left, and decide:

- in right, not in left → inline, `side=RIGHT`
- in both → summary, the line number is ambiguous
- not in right → summary

That middle case is the one the current design can't express. `resolveFromHunk` has an old-side fallback, so a comment can arrive holding an *old-file* line number, and an old line number can easily land inside a new-side range by coincidence. Today `Side == LEFT` catches that; once side is missing, nothing does, and the comment goes inline on the wrong line. Computing both sides catches it without trusting any upstream provenance — which means resume works, LLM-supplied line numbers work, and `LlmComment` doesn't need a new field.

Concretely, I'd like the 731 lines in `review_cmd.go` to move into `internal/ghpost`, with the dependency edge going one way only: `cmd → ghpost → {github, model}`, and specifically *not* to `session`, `agent` or `diff`. Give it a narrow input type of its own rather than taking `reviewOptions` or `session.ManifestInput`:

```go
type Target struct {
    RepoDir      string
    BaseRef      string // the user's --from
    ResolvedHead string // manifest.Input.ResolvedHead
}

func Post(ctx context.Context, t Target, comments []model.LlmComment, o Options) (Result, error)
```

`review_cmd.go` then becomes a dozen lines — pull three fields off the manifest, map, call, log. With that in place the sealed-vs-manifest cross-check in `validateGitHubPostingManifest` can go as well; `verifyGitHubPRTarget` is the check that actually prevents drift.

One thing I'd argue *against*, in case that's where you'd go next: don't turn this into a separate binary consuming `--json`. The contract is technically already there — `jsonOutput` embeds `Manifest` (`cmd/opencodereview/output.go:288`) and `ManifestInput` already carries `resolved_base` / `resolved_head` / `exact_range` (`internal/session/manifest.go:249`). But that field is typed as the internal `*session.RunManifest`. The moment an external process depends on it, every refactor inside `internal/session` becomes a breaking change. That trades "the core isn't touched" for "the core is frozen," which is worse — and it's a decision about the JSON schema that shouldn't ride along on a feature PR. Get the package boundary right first; moving it out of process later is a packaging change, not a redesign.

Smaller things:

- `NewClient` doesn't validate the configuration, and `Repository` is an exported struct with an unexported `configuration`. `github.Repository{Owner: "a", Name: "b"}` gets you a nil `httpClient` and a panic on the first `Do`. Worth a guard or an error return.
- `NewTestConfiguration` is exported from the production package, and it's the one constructor that permits HTTP and a mismatched API host — the two things the rest of the package exists to prevent. I'd move it to an `internal/github/githubtest` subpackage, or gate it on `testing.Testing()`.
- `ParseRepoInfo` has no callers.
- Posting failure now flows through `errors.Join`, so a successful review that fails to post exits non-zero. Defensible, but CI will read that as a failed review. Worth documenting, or giving it its own exit code.
- In the fallback path the summary header counts only the fallback subset (`inlineCount` is 0), so two summaries from the same run disagree about how many comments were generated. Cosmetic.
- Agreed on your Known Limitation — the split fence is presentation-only, and it doesn't change under any of the above.

ThalesMMS (commented):
## OpenCodeReview Summary

8 comment(s) generated.

- Inline findings: 7
- Summary-only findings: 1

### Findings without a verified inline location

#### `internal/github/client.go`

_Summary reason: line side provenance is not RIGHT._

**[maintainability · medium]**

The hardcoded limit of 30 pages (3000 items) might cause the tool to fail on very large pull requests. Consider making this configurable or using a more robust way to handle pagination.

---

### Severity breakdown:

- **high**: 1
- **low**: 3
- **medium**: 4

---
*Generated by OpenCodeReview*

ThalesMMS (commented):
## OpenCodeReview Summary

6 comment(s) generated.

- Inline findings: 6
- Summary-only findings: 0

### Severity breakdown:

- **high**: 1
- **low**: 3
- **medium**: 2

---
*Generated by OpenCodeReview*

ThalesMMS (commented):
## OpenCodeReview Summary

7 comment(s) generated.

- Inline findings: 6
- Summary-only findings: 1

### Severity breakdown:

- **high**: 1
- **low**: 3
- **medium**: 3

---
*Generated by OpenCodeReview*

cmd/opencodereview/review_cmd.go
```diff
@@ -284,9 +297,13 @@ func executeReviewContext(ctx context.Context, opts reviewOptions) error {
 		if id := ag.SessionID(); id != "" {
 			fmt.Fprintf(os.Stderr, "[ocr] Session: %s (retry with: --resume %s)\n", id, id)
 		}
-		return errors.Join(resultErr, emitErr)
+		return errors.Join(resultErr, emitErr, postErr)
 	}
-	return emitErr
+	return errors.Join(emitErr, postErr)
```
ThalesMMS, line 323: bug · high

The `resultErr` is ignored in the final return statement when `ag.SessionID()` is empty. This could cause failures in the main review process to be silenced.

**Suggestion:**
```suggestion
	return errors.Join(resultErr, emitErr, postErr)
```

ThalesMMS (commented):
## OpenCodeReview Summary

1 comment(s) generated.

- Inline findings: 1
- Summary-only findings: 0

### Severity breakdown:

- **high**: 1

---
*Generated by OpenCodeReview*

ThalesMMS:
Thank you @lizhengfeng101, @Qiyuanqiii, and @wu21-web for the time and care you put into reviewing this PR.

@lizhengfeng101, the posting orchestration now lives in `internal/ghpost`; GitHub-specific side provenance was removed from the shared model and is derived from both sides of the PR diff at posting time, including for resumed findings. Client construction, test-only configuration, fallback totals, and posting-failure documentation were also tightened.

@Qiyuanqiii, posting now verifies the exact reviewed merge base before every write, canonicalizes findings before hashing and batching, and detects and resumes existing fallback sequences instead of retrying inline delivery.

@wu21-web, review metadata is now rendered as non-bold Markdown Shields badges, the existing site translations were synchronized, and repository/client construction is opaque and validated before use. I kept the validated configuration bound to `Repository` rather than moving it literally into `Client`, but the unsafe partial-construction path behind that suggestion has been removed.

The follow-up passes also addressed duplicate stderr output, invalid UTF-8 handling, deterministic marker identity, trailing-slash parsing, and duplicate summary delivery.

I intentionally left the latest `resultErr` finding on the PR as a demonstration of the current OCR output. It is a false positive: the outer `resultErr != nil` branch always returns `resultErr`, regardless of whether `SessionID()` is empty; `SessionID()` only controls whether the resume hint is printed. This demonstration used a quantized DiffusionGemma backend, which likely contributed to the nested control-flow misread.

Qiyuanqiii (approved):
Thanks for the updates — I took another look at the latest commits. The three concerns I raised have all been addressed:
- the posting target is now verified against the exact PR merge base, with the reviewed base also included in the marker identity;
- findings are canonicalized before both marker generation and batching;
- existing fallback sequences are detected and resumed without retrying inline delivery.
The additional handling for repeated summary findings also covers the follow-up edge cases nicely. The implementation and the added regression coverage look good to me now. Thanks for working through these — I have not closely reviewed the questions raised by other reviewers, and for my part, I have no further concerns.

lizhengfeng101:
The restructuring looks right — `internal/ghpost` with a one-way dependency, `review_cmd.go` down to 40 lines, model and resolver untouched.

I need to walk back the premise I gave you, though. I claimed the posting side could derive the diff side on its own without upstream provenance. That was wrong, and your original instinct was better than my correction.

Line numbers alone can't carry side information, because old and new ranges within the same hunk overlap by construction. `classifyLocation` uses `containsRange` for the right side but `overlapsRange` for the left, so any contact with the left range demotes to summary. Your own fixture pins it:

```go
// "@@ -1,1 +1,2 @@\n context\n+added"  →  left={1,1}, right={1,2}
{name: "range crosses boundary", comment: {StartLine: 1, EndLine: 2}, want: locationAmbiguous}
```

That comment sits entirely inside the right range and is still demoted. Scale it to an ordinary mid-file edit — `@@ -100,7 +100,8 @@` gives `left={100,106}`, `right={100,107}`, so only line 107 posts inline. Outside of new files and hunks with a large line-number shift, inline posting is effectively off. Wider blast radius than the resume bug it replaced, because it hits the default path rather than one flag combination.

And the old-side case you were originally defending against is real. `resolveFromHunk` falls back to the old side and emits old-file line numbers — `TestResolveLineNumbers_SingleLineHunkMatch` asserts exactly that, on a deleted line, at old-file line 11. So "just require containment in the right range" isn't safe either: an old line number frequently lands inside a new-side range, and that comment goes inline on whatever occupies that line now.

Which leaves the approach I listed second last time as the one I'd actually take: keep the provenance, keep it out of `LlmComment`. Put a side field on `ghpost`'s own input type and have `cmd` map it from the resolver result. `ghpost` stays free of core concepts, `internal/model` stays clean, and the classification stops guessing. Resume needs that provenance persisted somewhere separate from the public comment schema — that's the part worth designing rather than patching.

One thing that argues against re-deriving it from content: `relocation.go` overwrites `cm.ExistingCode` when re-location succeeds, so by posting time the snippet may be the model's rewrite rather than the verbatim diff excerpt. Matching it back against the patch is not a reliable substitute for provenance.

pages/src/content/docs/en/cli-reference.md
```diff
@@ -180,6 +182,52 @@ OCR computes `merge-base(main, feature-branch)..feature-branch` so you only
...
+requires range mode with both `--from` and `--to`; provide the token with
+`--github-token`, or let the CLI read it from `GITHUB_TOKEN`.
+
+The CLI discovers the unique open pull request whose base branch and reviewed
+head commit match the range. You do not provide a pull-request number. At
+posting time, OCR builds separate inventories for the old and new sides of the
+pull-request diff. Only findings whose entire range is exclusive to the new
+side become inline comments. Old-side, incomplete, and ambiguous ranges
+(including the same line number appearing on both sides) are preserved in the
+pull-request summary.
```
wu21-web:
```suggestion
head commit match the range. You do not provide a pull-request number. During location resolution, OCR records whether each finding came from the old
or new side of the changed file. At posting time, OCR also builds an inventory
of the pull-request diff. A finding becomes an inline comment only when its
provenance is the new side and its entire range is contained in a verified,
complete right-side diff hunk. Old-side findings, findings without side
provenance, invalid ranges, and ranges that cannot be verified from a complete
patch are preserved in the pull-request summary. Overlapping numeric line
ranges alone do not make a finding ambiguous when its side is known.
```

wu21-web: > [!TIP]
> This is a documentation change, OK if you don't apply it.

wu21-web (approved):
I found no issues during my review on the code (no tests were reviewed). AI found 1 documentation issue.

ThalesMMS:
Thank you @lizhengfeng101 for revisiting the side-provenance design, and @wu21-web for reviewing the updated implementation and catching the stale documentation. Now OCR records OLD/NEW provenance during location resolution, keeps it outside `LlmComment`, persists it as an internal sidecar for `--resume`, and passes it through a `ghpost`-owned finding type while still verifying each location against the current pull-request diff. Cross-file findings remain owned by the review item that produced them, all producers and asynchronous workers finish before filters run by final path, and checkpoints are written only after filtering, keeping the current and resumed results consistent. Regression tests cover overlapping old/new ranges, relocation, legacy checkpoints, resumed findings, cross-file ownership, and filter/checkpoint ordering. The CLI documentation now describes this behavior and is synchronized across the English, Chinese, Japanese, and Russian pages. @lizhengfeng101, could you please take another look at the current head?

wu21-web (approved):

pages/src/content/docs/en/cli-reference.md
wu21-web: @ThalesMMS Resolve this thread please.

wu21-web:
@ThalesMMS Rebase main and resolve the merge conflicts, thank you.

ThalesMMS:
@wu21-web rebased, conflicts resolved.

Qiyuanqiii (approved):

ThalesMMS:
(sorry, I didn't mean to request review again)

wu21-web:
@ThalesMMS Hi, pinged you for 2 things:
1. The translations (korean) added just now need to be updated. Rebase the changes and edit the korean site docs.
2. I have no applicable review comments for your pull request.

ThalesMMS:
Thank you, @wu21-web . Done 😊. Do you think it will be merged soon?

wu21-web:
Not really, the maintainer is busy. My pull requests are queued just like yours.

Qiyuanqiii:
> 谢谢你， 完成😊。你觉得它会很快合并吗？

Perhaps you could try @‑mentioning him.

wu21-web:
> > 谢谢你， 完成😊。你觉得它会很快合并吗？
> 
> Perhaps you could try @‑mentioning him.

If it is me I won't tag the maintainer unless there is an emergency.

ThalesMMS:
Who is he?

> > 谢谢你， 完成😊。你觉得它会很快合并吗？
> 
> Perhaps you could try @‑mentioning him.

Who is he? 👀

Qiyuanqiii:
> 他是谁？
> 
> > > 谢谢你， 完成😊。你觉得它会很快合并吗？
> > 
> > 
> > 也许你可以试着@提到他。
> 
> 他是谁？👀

The maintainer is the man with the most commits in the project, and also the coolest XD.

PR #1038, opened 2026-08-22
fix(llm): allow opting out of Anthropic prompt-cache breakpoints

### LINKED ISSUE #700: unrecognizedProperty cache_control
### OpenCodeReview Version

v1.8.5 (582953937) linux/amd64

### Operating System

Linux (x86_64)

### Installation Method

npm (global)

### LLM Provider

Anthropic (Claude)

### Bug Description

```
[ocr] Subtask error for pom.xml: LLM completion error: POST "<redacted>": 400 Bad Request {"errorCode":"INVALID_ARGUMENT","errorName":"LanguageModelService:InvalidRequest","errorInstanceId":"b5d73937-d0f4-4dbc-b5ce-1eb5bb3bd40c","parameters":{"unsafeParams":"{unrecognizedProperty=cache_control}","message":"Request contained an unrecognized field"}}
```
Is it possible to remove this cache_control via configuration?

BTW: `ocr llm test` was successful
```
Source: OCR environment
URL:    <redacted>
Model:  <redacted>
I am Open-Code-Review, a command-line code review assistant developed by Alibaba, designed to help analyze and review code directly in your terminal environment.
✓ Connection test successful
```

### Steps to Reproduce

-

### Expected Behavior

-

### Logs / Error Output

```shell

```

### Additional Context

_No response_

ben7am1n:
## Description

OCR unconditionally attaches `cache_control` breakpoints to three places in every Anthropic request: the last system block, the last tool, and the trailing message. Anthropic-compatible gateways that validate the request body strictly reject the field as an unrecognized property instead of ignoring it, so **every** review request fails:

```
400 Bad Request {"errorCode":"INVALID_ARGUMENT", ...
  "unsafeParams":"{unrecognizedProperty=cache_control}",
  "message":"Request contained an unrecognized field"}
```

`ocr llm test` still passes against such a gateway, because that request carries no breakpoint — which is what makes the failure confusing to diagnose. Until now the only fix was patching the source.

This adds a tri-state `prompt_cache` key:

```bash
ocr config set providers.anthropic.prompt_cache false
ocr config set custom_providers.my-gateway.prompt_cache false
ocr config set llm.prompt_cache false
```

plus `OCR_LLM_PROMPT_CACHE=false` as a global override, applied the same way as `OCR_LLM_TIMEOUT` — it reaches every resolution path, which matters in CI where the endpoint comes from environment variables alone and there is no config file to edit. It overrides the config file in both directions.

Design notes:

- **Unset means enabled.** `*bool` rather than `bool` so a config written before this key existed is indistinguishable from one that explicitly opts in, and `omitempty` cannot silently drop an explicit re-enable.
- **Cost, not correctness.** Every `cache_control` marker is optional to the request's meaning, so disabling it gives up the cache discount and nothing else — reviews still run, tools still work. That is the trade the flag makes deliberate.
- **An unparseable `OCR_LLM_PROMPT_CACHE` is an error**, not a silent default. Anyone setting it is working around a failing endpoint; ignoring a typo would leave them staring at the same 400 with no clue the knob never applied.
- **Anthropic protocol only**, since it is the only path that emits these markers on its own. The OpenAI `prompt_cache_key` route is user-supplied via `extra_body` and already opt-in.

## Type of Change

## How Has This Been Tested?

- [x] `make test` passes locally (all 23 packages, with `-race`)
- [x] `make check` passes (fmt, vet, license headers, english-only)
- [x] `make coverage` passes at 91.4% (threshold 90%)

Unit tests added:

| Test | Covers |
| --- | --- |
| `TestBuildAnthropicParams_PromptCacheDisabled` | Asserts on the **marshalled request body**, since the bug is about the wire format — one leftover breakpoint anywhere reproduces it. Includes a positive control so the negative assertion cannot pass vacuously, and a check that the system prompt, tools and messages all survive. |
| `TestResolveEndpoint_ProviderPromptCache` | Tri-state resolution; the `nil` case pins backward compatibility. |
| `TestResolveEndpoint_LegacyLlmPromptCache` | The legacy `llm` section. |
| `TestResolveEndpoint_PromptCacheEnvOverride` | Env precedence in both directions, plus empty/whitespace being treated as unset. |
| `TestResolveEndpoint_InvalidPromptCacheEnv` | Unparseable value errors and names the variable. |
| `TestNewLLMClient_PromptCacheDisabledForwarded` | The `ResolvedEndpoint` → `ClientConfig` seam, which nothing else would notice breaking. |
| `TestSetConfigValuePromptCache` / `...RoundTrip` | All three config keys, invalid input, and survival across a marshal/unmarshal cycle. |
| `TestCloneProviderEntry_PromptCacheNotAliased` | The TUI clone. `DeepEqual` compares pointees, so it would pass on a shared pointer; without a deep copy a rollback path would mutate the entry it is restoring. |

`TestCloneProviderEntry_CopiesEveryField` caught the new field automatically and its fixture was updated — that guard works exactly as its comment promises.

**Manual end-to-end**, against a local server that mimics the gateway in #700 by returning 400 whenever the body contains `cache_control`:

```
=== WITHOUT the flag (reproduces #700) ===
Error: llm request failed: POST "http://127.0.0.1:8899/v1/messages": 400 Bad Request
{"type":"error","error":{"type":"invalid_request_error",
 "message":"Request contained an unrecognized field: cache_control"}}

=== WITH prompt caching disabled ===
✓ Connection test successful

=== WITH an invalid value ===
Error: resolve LLM endpoint: OCR_LLM_PROMPT_CACHE must be a boolean (true/false, 1/0), got "maybe"
```

I also mutation-checked the main test by forcing `cacheBreakpoints := true`; it fails as expected and prints the offending body.

## Checklist

- [x] I have updated the documentation accordingly — `configuration.md` in all four locales (`en`, `zh`, `ja`, `ru`)
- [ ] I have signed the CLA — will sign as soon as the bot posts

## Related Issues

closes #700

### DIFF
--- a/cmd/opencodereview/config_cmd.go
+++ b/cmd/opencodereview/config_cmd.go
@@ -303,6 +303,10 @@ type ProviderEntry struct {
 	ExtraBody    map[string]any    `json:"extra_body,omitempty"`
 	ExtraHeaders map[string]string `json:"extra_headers,omitempty"`
 	RetryCodes   []int             `json:"retry_codes,omitempty"`
+	// PromptCache is nil when unset, which means enabled. Set it to false for
+	// an Anthropic-compatible gateway that rejects `cache_control` as an
+	// unrecognized field instead of ignoring it.
+	PromptCache *bool `json:"prompt_cache,omitempty"`
 
 	// AWSProfile and AWSRegion pin the credentials and region for providers that
 	// authenticate from the AWS chain (bedrock). Both are optional — without
@@ -353,6 +357,7 @@ type LlmConfig struct {
 	ExtraBody    map[string]any    `json:"extra_body,omitempty"`
 	ExtraHeaders map[string]string `json:"extra_headers,omitempty"`
 	RetryCodes   []int             `json:"retry_codes,omitempty"`
+	PromptCache  *bool             `json:"prompt_cache,omitempty"` // nil = default true; false = omit cache_control
 }
 
 // TelemetryConfig holds telemetry-specific settings.
@@ -414,6 +419,7 @@ var supportedConfigKeys = []string{
 	"llm.extra_body",
 	"llm.extra_headers",
 	"llm.retry_codes",
+	"llm.prompt_cache",
 	"language",
 	"telemetry.enabled",
 	"telemetry.exporter",
@@ -574,8 +580,14 @@ func setConfigValue(cfg *Config, key, value string) error {
 			fmt.Fprintf(os.Stderr, "[ocr] WARNING: %s\n", w)
 		}
 		cfg.Llm.RetryCodes = codes
+	case "llm.prompt_cache", "llm.PromptCache":
+		b, err := strconv.ParseBool(value)
+		if err != nil {
+			return fmt.Errorf("invalid boolean for llm.prompt_cache: %w", err)
+		}
+		cfg.Llm.PromptCache = &b
 	default:
-		return fmt.Errorf("unknown config key: %s\nSupported keys: %s\nProvider fields: api_key, api_key_cmd, url, protocol, model, models, auth_header, extra_body, extra_headers, retry_codes, aws_region, aws_profile\nProtocol values: anthropic, anthropic-bedrock, openai, openai-responses\nMCP server fields: type, command, args, env, url, headers, tools, setup", key, strings.Join(supportedConfigKeys, ", "))
+		return fmt.Errorf("unknown config key: %s\nSupported keys: %s\nProvider fields: api_key, api_key_cmd, url, protocol, model, models, auth_header, extra_body, extra_headers, retry_codes, prompt_cache, aws_region, aws_profile\nProtocol values: anthropic, anthropic-bedrock, openai, openai-responses\nMCP server fields: type, command, args, env, url, headers, tools, setup", key, strings.Join(supportedConfigKeys, ", "))
 	}
 	return nil
 }
@@ -644,6 +656,12 @@ func applyProviderField(providerName string, entry *ProviderEntry, field, key, v
 			fmt.Fprintf(os.Stderr, "[ocr] WARNING: %s\n", w)
 		}
 		entry.RetryCodes = codes
+	case "prompt_cache":
+		b, err := strconv.ParseBool(value)
+		if err != nil {
+			return fmt.Errorf("invalid boolean for %s: %w", key, err)
+		}
+		entry.PromptCache = &b
 	case "aws_region", "aws_profile":
 		normalized, err := normalizeAWSSetting(field, key, value)
 		if err != nil {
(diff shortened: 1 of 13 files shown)


ben7am1n:
Thanks @Qiyuanqiii — reproduced exactly as you described, and you were right that it is not limited to the tests I touched. Pushed in 7e1b5cf.

Your repro now passes:

```
$ OCR_LLM_PROMPT_CACHE=definitely-not-a-bool make test PACKAGES=github.com/alibaba/open-code-review/internal/llm
ok  	github.com/alibaba/open-code-review/internal/llm	6.744s
```

Two things I found while fixing it that are worth flagging:

**1. `clearAllEnv` alone was not sufficient.** Adding `envOCRLLMPromptCache` to the reset list took the failure count from 98 to 20, not to 0. The remainder are the tests that never call the helper — `bedrock_test.go` and friends build a config with `writeConfig` and call `ResolveEndpoint` directly, so they inherit the host environment whole. I did both: the helper entry you asked for, plus a `TestMain` that clears these globals for the package, so isolation no longer depends on a new test remembering which helper to call.

**2. The same gap already existed for the two siblings.** `parseEnvOverrides` validates three variables, and none of the three were in the reset list:

| variable | failures before | after |
|---|---|---|
| `OCR_LLM_PROMPT_CACHE` | 98 | 0 |
| `OCR_LLM_TIMEOUT` | 99 | 0 |
| `OCR_LLM_EXTRA_HEADERS` | 1 | 0 |

`OCR_LLM_TIMEOUT=bogus` fails 99 tests on `main` today, independently of this PR. Since the cause and the fix are identical I covered all three rather than leaving a live one behind, but I am happy to split the two pre-existing ones into their own PR if you would rather keep this one scoped to the variable it introduces — just say the word.

The distinction that makes these three special is worth recording, so I put it in the comments: every other variable in the list merely contributes a value, whereas an unparseable one of these aborts `ResolveEndpointWithOptions` before any strategy runs — which is why one stray export takes out a hundred tests with an error naming a variable the test never mentions.

`make test`, `make check` and `make coverage` (91.4%) all pass.

Qiyuanqiii:
The fix is complete and works well. I have verified it locally and found no new issues

lizhengfeng101:
@ben7am1n merge conflicts

lizhengfeng101:
On the design direction: the explicit opt-in switch is the right call, and I'd argue **against** making this implicit.

`cache_control` is a valid field on the real Anthropic API — the breakage only happens on third-party "compatible" gateways that reject unknown fields instead of ignoring them. For the users who hit it, it's a hard, 100%-reproducible block, but the fix shouldn't be automatic:

- Auto-stripping `cache_control` on a 400 would mean pattern-matching gateway error strings, which vary per vendor — fragile, and it risks masking unrelated failures.
- A static switch is predictable, testable, has zero runtime cost, and is consistent with the existing per-provider escape hatches (`retry_codes`, `extra_body`, `timeout_sec`).

Handing the affected users one switch to turn the markers off is enough — no need for probing or graceful-degradation machinery.

_Optional, non-blocking follow-up:_ since the raw 400 comes from the gateway and doesn't point back to OCR, wrapping that specific error with a hint (e.g. "your gateway rejected `cache_control`; run `ocr config set <provider>.prompt_cache false`") would cut the diagnosis time. Can be a separate PR.

PR #1043, opened 2026-08-23
fix(llm): extract raw response payload for Google Gemini 400 Bad Request errors

### LINKED ISSUE #1042: Opaque "400 Bad Request" error with missing details when using Gemini / OpenAI-compatible endpoints
### OpenCodeReview Version

open-code-review v1.9.8 (9c5e90d0) built at: 2026-08-20T02:32:42Z

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

### Bug Description

When an Google Gemini via OpenAI-compatible endpoint `https://generativelanguage.googleapis.com/v1beta/openai` returns an HTTP 400 Bad Request or other non-2xx error, the CLI and session logs only display `400 Bad Request` without the underlying JSON error details.

Because Google Gemini format their error details in nested fields (`details`, `fieldViolations`) rather than standard top-level fields, the default `*openai.Error.Error()` method omits the response payload. This leaves the user with no actionable error message to understand why the request failed.

### Steps to Reproduce

1. Configure an OpenAI-compatible provider pointing to Google Gemini (`https://generativelanguage.googleapis.com/v1beta/openai`) with a Gemini model (e.g., `gemini-3.6-flash`).
2. Run a review on changes that trigger an upstream parameter or schema validation error:
   `ocr review --background "test background"`
3. The review fails on an LLM call with `400 Bad Request`.
4. Inspect the session file (`~/.opencodereview/sessions/.../*.jsonl`) or terminal error output.
5. Notice the error string only contains `400 Bad Request` with no raw response body or reason explanation.

### Expected Behavior

When the API returns an error response, OpenCodeReview should unwrap and display the raw JSON error payload / status / code from the HTTP response (e.g. via `apiErr.RawJSON()`), allowing users and logs to see the exact rejection reason from the vendor (such as field violations or invalid parameters).

### Logs / Error Output

```shell
[ocr] Subtask error for internal/llm/responses_client.go: LLM completion error: POST "https://generativelanguage.googleapis.com/v1beta/openai/chat/completions": 400 Bad Request
```

### Additional Context

I was trying to setup Google Gemini API with `ocr`. The setup was successful, but running `ocr review ...` was failing and I couldn't figure out why from the logs.

Here's a stdout log before the change (similar is in the jsonl log):

```bash
[ocr] Subtask error for internal/llm/responses_client.go: LLM completion error: POST "https://generativelanguage.googleapis.com/v1beta/openai/chat/completions": 400 Bad Request
```

Here's a stdout log after the change:

```bash
[ocr] Subtask error for internal/llm/client.go: LLM completion error: Google Gemini API error (HTTP 400 Bad Request, model: gemini-3.7-flash), response body: [{
  "error": {
    "code": 400,
    "message": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:file_read_diff` , position 2. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.",
    "status": "INVALID_ARGUMENT"
  }
}
]: POST "https://generativelanguage.googleapis.com/v1beta/openai/chat/completions": 400 Bad Request
```

(As you can see, this hints toward additional issue with Gemini API with thought_signatures).

fkdosilovic:
## Description

When Google Gemini (via OpenAI-compatible endpoint (`https://generativelanguage.googleapis.com/v1beta/openai`) returns an HTTP error (e.g. `400 Bad Request`), the default `*openai.Error` string only reports the HTTP status line without vendor payload details (such as Google RPC `details`, `fieldViolations`, and parameter error codes).

This PR introduces structured error extraction from the OpenAI SDK (`*openai.Error` / `*apierror.Error`), surfacing the raw JSON response payload and HTTP status specifically for Gemini endpoints, while leaving all other providers unmodified.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] Manual testing

For manual testing I ran the `make build` and `./dist/opencodereview review --from main --to HEAD --background "LLM error reporting improvements"` to see if the changes were successful.

## Checklist

- [ ] I have performed a self-review of my code (will do before I switch from Draft to ready)
- [x] I have signed the CLA

## Related Issues

Closes #1042 .

### DIFF
--- a/internal/llm/client.go
+++ b/internal/llm/client.go
@@ -664,7 +664,7 @@ func extractOpenAIErrorDetails(err error) (*openAIErrorDetails, bool) {
 	}
 
 	if raw := apiErr.RawJSON(); raw != "" {
-		details.RawBody = strings.TrimSpace(raw)
+		details.RawBody = limitErrorBodyForLog(raw)
 	} else if apiErr.Response != nil && apiErr.Response.Body != nil {
 		bodyBytes, _ := io.ReadAll(apiErr.Response.Body)
 		_ = apiErr.Response.Body.Close()
--- a/internal/llm/client.go
+++ b/internal/llm/client.go
@@ -272,20 +272,14 @@ func retryCodesMiddleware(codes []int) func(*http.Request, func(*http.Request) (
 	}
 }
 
-// limitErrorResponseBody bounds provider error payloads before the OpenAI SDK
-// reads them into memory. The wrapper delegates Close to the transport body.
-func limitErrorResponseBody(req *http.Request, next func(*http.Request) (*http.Response, error)) (*http.Response, error) {
-	resp, err := next(req)
-	if resp != nil && resp.StatusCode >= http.StatusBadRequest && resp.Body != nil {
-		resp.Body = struct {
-			io.Reader
-			io.Closer
-		}{
-			Reader: io.LimitReader(resp.Body, maxErrorBodyBytes),
-			Closer: resp.Body,
-		}
+// limitErrorBodyForLog bounds the payload included in terminal and session logs
+// without truncating the HTTP response before the SDK parses it.
+func limitErrorBodyForLog(raw string) string {
+	raw = strings.TrimSpace(raw)
+	if len(raw) <= maxErrorBodyBytes {
+		return raw
 	}
-	return resp, err
+	return strings.ToValidUTF8(raw[:maxErrorBodyBytes], "\uFFFD") + "... (truncated)"
 }
 
 // --- Factory ---
--- a/internal/llm/client.go
+++ b/internal/llm/client.go
@@ -421,7 +421,6 @@ func NewOpenAIClient(cfg ClientConfig) *OpenAIClient {
 		openaiopt.WithMaxRetries(5),
 		openaiopt.WithHeader("User-Agent", userAgent("")),
 		openaiopt.WithRequestTimeout(cfg.Timeout),
-		openaiopt.WithMiddleware(limitErrorResponseBody),
 	}
 	if mw := retryCodesMiddleware(cfg.RetryCodes); mw != nil {
 		opts = append(opts, openaiopt.WithMiddleware(mw))
--- a/internal/llm/client.go
+++ b/internal/llm/client.go
@@ -43,6 +43,8 @@ var AppVersion = "dev"
 // shrink it, same as keyCmdTimeout.
 var bedrockConfigLoadTimeout = 60 * time.Second
 
+const maxErrorBodyBytes = 64 * 1024
+
 func userAgent(provider string) string {
 	ua := "open-code-review/" + AppVersion
 	if provider != "" {
@@ -270,6 +272,22 @@ func retryCodesMiddleware(codes []int) func(*http.Request, func(*http.Request) (
 	}
 }
 
+// limitErrorResponseBody bounds provider error payloads before the OpenAI SDK
+// reads them into memory. The wrapper delegates Close to the transport body.
+func limitErrorResponseBody(req *http.Request, next func(*http.Request) (*http.Response, error)) (*http.Response, error) {
+	resp, err := next(req)
+	if resp != nil && resp.StatusCode >= http.StatusBadRequest && resp.Body != nil {
+		resp.Body = struct {
+			io.Reader
+			io.Closer
+		}{
+			Reader: io.LimitReader(resp.Body, maxErrorBodyBytes),
+			Closer: resp.Body,
+		}
+	}
+	return resp, err
+}
+
 // --- Factory ---
 
 // NewLLMClient creates the appropriate client based on the resolved endpoint protocol.
@@ -409,6 +427,7 @@ func NewOpenAIClient(cfg ClientConfig) *OpenAIClient {
 		openaiopt.WithMaxRetries(5),
 		openaiopt.WithHeader("User-Agent", userAgent("")),
 		openaiopt.WithRequestTimeout(cfg.Timeout),
+		openaiopt.WithMiddleware(limitErrorResponseBody),
 	}
 	if mw := retryCodesMiddleware(cfg.RetryCodes); mw != nil {
 		opts = append(opts, openaiopt.WithMiddleware(mw))
@@ -630,6 +649,7 @@ func formatOpenAIError(provider, model string, err error) error {
 			rawBody = strings.TrimSpace(raw)
 		} else if apiErr.Response != nil && apiErr.Response.Body != nil {
 			bodyBytes, _ := io.ReadAll(apiErr.Response.Body)
+			_ = apiErr.Response.Body.Close()
 			apiErr.Response.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
 			rawBody = strings.TrimSpace(string(bodyBytes))
 		}
@@ -666,7 +686,7 @@ func formatOpenAIError(provider, model string, err error) error {
 			parts = append(parts, fmt.Sprintf("response body: %s", rawBody))
 		}
 
-		if len(parts) > 1 || rawBody != "" {
+		if len(parts) > 0 {
 			return fmt.Errorf("%s: %w", strings.Join(parts, ", "), err)
 		}
 	}
--- a/internal/llm/client.go
+++ b/internal/llm/client.go
@@ -534,7 +534,7 @@ func (c *OpenAIClient) CompletionsWithCtx(ctx context.Context, req ChatRequest) 
 		}
 	}
 	if err != nil {
-		return nil, formatOpenAIError(c.cfg.Provider, model, err)
+		return nil, formatGeminiError(c.cfg.Provider, c.cfg.URL, model, err)
 	}
 
 	return c.mapOpenAIResponse(sdkResp), nil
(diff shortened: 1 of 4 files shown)


wu21-web:
I would rather remove Gemini. I added it to the providers and that is a bad idea.


 I am hiding this because this is stale via https://github.com/alibaba/open-code-review/pull/1043#issuecomment-5386080908

fkdosilovic:
> I would rather remove Gemini. I added it to the providers and that is a bad idea.

Ok . I can do that as well ... What's the reasoning behind that supporting Gemini " is a bad idea"?

Have you considered using the official client - https://github.com/googleapis/go-genai ? This way you could support the use of Gemini models via Gemini Developer API and Gemini Enterprise Agent Platform (former VertexAI).

wu21-web:
@fkdosilovic I can't really get the idea of the pull request through the pr desc, is there something wrong that is specific to Google Gemini?

fkdosilovic:
> @fkdosilovic I can't really get the idea of the pull request through the pr desc, is there something wrong that is specific to Google Gemini?

I opened an issue and described the problem I'm trying to solve with this PR - https://github.com/alibaba/open-code-review/issues/1042

wu21-web:
Thank you now I get the idea :+1:

Qiyuanqiii (commented):
Thanks for working on this. After reading #1042 and looking through the implementation, I think recovering the raw provider response when `openai-go` cannot expose useful error details is a worthwhile fix. Preserving the underlying `*openai.Error` through wrapping also looks good.

There are two things I think are worth addressing before merge.

First, `limitErrorResponseBody` is currently installed for every OpenAI-compatible client. Since it runs before `openai-go` parses the response, an error body larger than 64 KiB from OpenAI, DeepSeek, a custom proxy, etc. would also be truncated before the SDK builds its error. That seems broader than the PR's stated goal of leaving other providers unchanged. I think it would be safer to keep the transport response intact and limit only the diagnostic text that OCR eventually renders or logs.

Second, the main regression test uses the normal `{"error": {...}}` response envelope, while the response shown in #1042 is array-wrapped (`[{"error": {...}}]`). The standard object shape is already understood by `openai-go`; the array-wrapped shape is the interesting case because the SDK's normal `"error"` extraction does not find it, which is where the restored `Response.Body` fallback becomes useful. It would be good for the regression test to reproduce that actual response shape so we know it fails before the fix and passes after it.

Given the discussion above about whether Gemini should remain a built-in provider, I also wonder if this could be made provider-neutral: enrich an OpenAI-compatible API error when the SDK has no useful raw payload but the restored response body is available, rather than introducing Gemini-specific transport behavior.

That would keep the #1042 fix useful even if the built-in Gemini provider changes or is removed later.

internal/llm/responses_client.go
```diff
@@ -48,6 +48,7 @@ func NewOpenAIResponsesClient(cfg ClientConfig) *OpenAIResponsesClient {
 		openaiopt.WithMaxRetries(5),
 		openaiopt.WithHeader("User-Agent", userAgent("")),
 		openaiopt.WithRequestTimeout(cfg.Timeout),
+		openaiopt.WithMiddleware(limitErrorResponseBody),
```
wu21-web:
```suggestion
		openaiopt.WithRequestTimeout(cfg.Timeout),
```

internal/llm/client.go
```diff
@@ -268,6 +272,22 @@ func retryCodesMiddleware(codes []int) func(*http.Request, func(*http.Request) (
...
+		resp.Body = struct {
+			io.Reader
+			io.Closer
+		}{
+			Reader: io.LimitReader(resp.Body, maxErrorBodyBytes),
+			Closer: resp.Body,
+		}
+	}
+	return resp, err
+}
```
wu21-web:
```suggestion
// limitErrorBodyForLog bounds the payload included in terminal and session logs
// without truncating the HTTP response before the SDK parses it.
func limitErrorBodyForLog(raw string) string {
	raw = strings.TrimSpace(raw)
	if len(raw) <= maxErrorBodyBytes {
		return raw
	}
	return strings.ToValidUTF8(raw[:maxErrorBodyBytes], "\uFFFD") + "... (truncated)"
}
```

internal/llm/client.go
```diff
@@ -612,6 +634,107 @@ func (c *OpenAIClient) completionsStreamingInner(ctx context.Context, params ope
...
+		Type:     apiErr.Type,
+		Original: err,
+	}
+
+	if details.Status == 0 && apiErr.Response != nil {
+		details.Status = apiErr.Response.StatusCode
+	}
+
+	if raw := apiErr.RawJSON(); raw != "" {
+		details.RawBody = strings.TrimSpace(raw)
```
wu21-web:
```suggestion
		details.RawBody = limitErrorBodyForLog(raw)
```

internal/llm/client.go
```diff
@@ -612,6 +634,107 @@ func (c *OpenAIClient) completionsStreamingInner(ctx context.Context, params ope
...
+	} else if apiErr.Response != nil && apiErr.Response.Body != nil {
+		bodyBytes, _ := io.ReadAll(apiErr.Response.Body)
+		_ = apiErr.Response.Body.Close()
+		apiErr.Response.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
+		details.RawBody = strings.TrimSpace(string(bodyBytes))
```
wu21-web:
```suggestion
		details.RawBody = limitErrorBodyForLog(string(bodyBytes))
```

internal/llm/client_test.go
```diff
@@ -1922,3 +1923,322 @@ func TestAnthropicClient_RetryCodesTriggersRetry(t *testing.T) {
...
+	for _, tt := range tests {
+		got := isGemini(tt.provider, tt.url)
+		if got != tt.want {
+			t.Errorf("isGemini(%q, %q) = %v, want %v", tt.provider, tt.url, got, tt.want)
+		}
+	}
+}
+
+func TestFormatGeminiError_Gemini400BadRequest(t *testing.T) {
+	geminiErrBody := `{"error":{"code":400,"message":"Invalid argument in tool call","status":"INVALID_ARGUMENT","details":[{"@type":"type.googleapis.com/google.rpc.ErrorInfo","reason":"FIELD_VIOLATION","metadata":{"field":"messages.tool_calls"}}]}}`
```
wu21-web:
```suggestion
	geminiErrBody := `[{"error":{"code":400,"message":"Invalid argument in tool call","status":"INVALID_ARGUMENT","details":[{"@type":"type.googleapis.com/google.rpc.ErrorInfo","reason":"FIELD_VIOLATION","metadata":{"field":"messages.tool_calls"}}]}}]`
```

internal/llm/client.go
```diff
@@ -406,6 +427,7 @@ func NewOpenAIClient(cfg ClientConfig) *OpenAIClient {
 		openaiopt.WithMaxRetries(5),
 		openaiopt.WithHeader("User-Agent", userAgent("")),
 		openaiopt.WithRequestTimeout(cfg.Timeout),
+		openaiopt.WithMiddleware(limitErrorResponseBody),
```
wu21-web:
```suggestion
		openaiopt.WithRequestTimeout(cfg.Timeout),
```

fkdosilovic:
> limitErrorResponseBody is currently installed for every OpenAI-compatible client. Since it runs before openai-go parses the response, an error body larger than 64 KiB from OpenAI, DeepSeek, a custom proxy, etc. would also be truncated before the SDK builds its error. That seems broader than the PR's stated goal of leaving other providers unchanged.

You are right. Sorry, I forgot to update this part since I only wanted to enrich log messages for Gemini API, but forgot about the limit.

> I think it would be safer to keep the transport response intact and limit only the diagnostic text that OCR eventually renders or logs.

I like this idea a lot.

> Given the discussion above about whether Gemini should remain a built-in provider, I also wonder if this could be made provider-neutral: enrich an OpenAI-compatible API error when the SDK has no useful raw payload but the restored response body is available, rather than introducing Gemini-specific transport behavior.

(Yeah, this is why I forgot about the 64 KiB limit.) This would also simplify the `ClientConfig` and remove the new attribute (`provider`) introduced in this PR.

wu21-web (approved):
Tests are now passing, thank you!

Qiyuanqiii (approved):

wu21-web (approved):

lizhengfeng101:
@fkdosilovic  merge conflicts~

wu21-web:
I didn't ask for this I believe ...

wu21-web:
Keep this open, we wilI decide which is better.

lizhengfeng101:
Thanks for this, @fkdosilovic — and sorry for the overlap.

Closing as superseded by #1393, which has now been merged. It adds `withProviderErrorBody`, which surfaces the provider's raw response body whenever the SDK leaves an `*openai.Error` without a payload — the same gap you're fixing for Gemini 400s. The merged version applies to every OpenAI-compatible provider (not just Gemini), handles the array-wrapped error document case (#1042), restores the body so downstream consumers still see it whole, and bounds only the diagnostic text.

The one thing your PR did that the merged one does not is parse the Google RPC envelope into structured `details` / `fieldViolations` rather than quoting the raw JSON. If seeing those fields laid out (instead of raw payload) is still valuable, that would be a clean follow-up on top of `withProviderErrorBody` — please open a fresh PR and reference this thread. Thanks for digging into this.

PR #1045, opened 2026-08-23
chore: use CRLF for PowerShell scripts

wu21-web:
## Description
.ps1 files should end with CRLF, the original `.gitattributes` normalized everything with LF, with is a bit absolute for .ps1 which meant to be ran on Windows.

install.ps1 changes not shown in diff.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

CI passes

## Checklist

- [x] I have signed the CLA

## Related Issues

N/A


lizhengfeng101:
Thanks for the PR! Closing this one, as I don't think it's needed:

- text eol=crlf doesn't store CRLF in the repo — Git still normalizes the blob to LF on checkin and only converts to CRLF in the working tree. I verified locally that install.ps1 stays LF in the index and git add --renormalize . produces no diff. So .ps1 is not an exception to the CI check, which makes the message change misleading (also a typo: expect → except).
- Practically, install.ps1 ships via irm ... | iex, where line endings don't matter and PowerShell runs LF fine — so the attribute only affects Windows working-tree checkouts, not the install path.

PR #1051, opened 2026-08-23
fix(action): forward review timeout and complete LLM config

crestonepeaks:
## Description

Harden the reusable composite action while continuing to install the official `@alibaba-group/open-code-review` package:

- expose `review_task_timeout` (default 10, integer range 1-120) and pass it once to `ocr review --timeout` as the per-file/concurrent-task deadline, independently from the LLM HTTP timeout;
- build a complete legacy `llm` configuration so `llm.extra_body` reaches the endpoint selected by the resolver, while resolving the token only through an environment-backed `llm.auth_token_cmd`;
- require and verify OCR v1.9.6+ before using `llm.auth_token_cmd`, with actionable failures for older, prerelease, malformed, or unreadable versions;
- preserve the legacy case-insensitive `llm_use_anthropic` resolver contract while writing a canonical boolean and matching protocol;
- fail closed when the validated task timeout is unavailable;
- neutralize stale provider/static-token/header/retry configuration before applying action inputs;
- default the independent LLM request timeout to the CLI-compatible 300 seconds, including an explicitly empty action input;
- run the hardened 24-test action contract suite in a dedicated GitHub-hosted CI job, including changes to its documented README contract.

## Type of Change

- [x] Documentation update
- [x] CI / Build / Tooling

## How Has This Been Tested?

Exact head `92dd6a2e58ccff5801a1c9f5437e02abdcca2a3c`:

- `npm run test:github-actions` (all existing suites plus 24/24 action contracts)
- `npm run test:update`
- `bash scripts/verify-license.sh`
- `bash scripts/verify-action-pins.sh`
- `actionlint .github/workflows/ci.yml`
- YAML parsing and `git diff --check`
- Go 1.26.6 container: `make check`, full race-enabled `make test`, and `make build`
- independent Luna Max review/fix loops: clean
- repository-required local OCR review loops completed; actionable findings were fixed test-first and incorrect findings were verified against executable Bash behavior, upstream history, maintainer requirements, and official GitHub path-filter documentation

## Checklist

- [x] Existing relevant tests pass locally with my changes
- [x] I have updated the action input descriptions and executable contract
- [x] I have signed the CLA

## Related Issues

No linked issue. This is a focused reusable-action bug fix.


action.yml
```diff
@@ -244,6 +249,29 @@ runs:
...
+        done
+        if [[ "$NORMALIZED_REVIEW_TIMEOUT" == "0" || ${#NORMALIZED_REVIEW_TIMEOUT} -gt 3 ]]; then
+          echo "::error::review_timeout must be between 1 and 120 minutes"
+          exit 1
+        fi
+        REVIEW_TIMEOUT_NUMBER=$((10#$NORMALIZED_REVIEW_TIMEOUT))
+        if (( REVIEW_TIMEOUT_NUMBER < 1 || REVIEW_TIMEOUT_NUMBER > 120 )); then
+          echo "::error::review_timeout must be between 1 and 120 minutes"
+          exit 1
+        fi
```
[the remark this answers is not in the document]
crestonepeaks: Fixed in 81d5073f6ba57a1ed0fc8a6d2cd6b9d97ebd924e. Validation now exports the normalized decimal REVIEW_TIMEOUT through GITHUB_ENV, the review step no longer rebinds the raw input, and a regression contract proves 010 is forwarded as 10. The action contract suite passes 18/18.

action.yml
```diff
@@ -42,6 +42,11 @@ inputs:
   llm_timeout:
     description: LLM request timeout in seconds (mapped to env OCR_LLM_TIMEOUT).
     required: false
+    default: '300'
+  review_timeout:
+    description: Review timeout in integer minutes, from 1 through 120.
```
lizhengfeng101: **`review_timeout` is not a review timeout.**

`ocr review --timeout` binds to `perFileTimeout`, which is passed through as `ConcurrentTaskTimeout` (`cmd/opencodereview/review_cmd.go:206`), and its own flag help says `"concurrent task timeout in minutes"` (`cmd/opencodereview/shared_flags.go:45`). It bounds a single concurrent task / per-file review, not the wall-clock duration of the whole review.

As written, a user who reads "Review timeout in integer minutes, from 1 through 120" will reasonably assume `review_timeout: 120` caps total run time — it does not. Please rename the input to something like `review_task_timeout` and state the per-file/per-task scope in the description, otherwise the 1-120 range reads as a guarantee the action cannot make.

action.yml
```diff
@@ -255,10 +284,36 @@ runs:
...
+        ocr config unset provider
+        ocr config set llm.auth_token ""
+        ocr config set llm.extra_headers ""
+        ocr config set llm.retry_codes ""
+        ocr config set llm.url "$OCR_LLM_URL"
+        ocr config set llm.model "$OCR_LLM_MODEL"
+        ocr config set llm.use_anthropic "$OCR_USE_ANTHROPIC"
+        ocr config set llm.protocol "$OCR_LLM_PROTOCOL"
+        ocr config set llm.auth_header "$OCR_LLM_AUTH_HEADER"
+        ocr config set llm.auth_token_cmd 'printf "%s" "$OCR_LLM_TOKEN"'
```
lizhengfeng101, line 336: **This silently raises the minimum required `ocr` version to v1.9.6, with no declaration or guard.**

`llm.auth_token_cmd` was introduced in f75c43a (#605) and first shipped in tag `v1.9.6` (`git tag --contains f75c43a` → earliest `v1.9.6`). `ocr_version` is a user-facing input and `examples/github_actions/README.md` demonstrates pinning it, so anyone pinned to `<= 1.9.5` now fails here with an unknown-config-key error — and because composite `run` steps default to `set -e`, the whole action aborts on a message that does not point at the version as the cause.

At minimum, document the floor in the `ocr_version` input description. Better would be to fail early with an explicit message when the installed version predates `llm.auth_token_cmd`.

action.yml
```diff
@@ -255,10 +284,36 @@ runs:
...
       run: |
+        case "$OCR_USE_ANTHROPIC" in
+          true)
+            OCR_LLM_PROTOCOL="anthropic"
+            ;;
+          false)
+            OCR_LLM_PROTOCOL="openai"
+            ;;
+          *)
+            echo "::error::llm_use_anthropic must be the lowercase literal true or false"
```
lizhengfeng101: **This is a breaking change for existing callers, and the PR is labelled non-breaking.**

The path this replaces (`tryOCREnv`, `internal/llm/resolver.go:271-273`) accepted `true` / `1` / `yes` case-insensitively, and treated an empty value as `useAnthropic = true`. The new `case` statement admits only the lowercase literals `true` and `false` and hard-exits on everything else.

So a workflow currently passing `llm_use_anthropic: 'TRUE'`, `'1'`, or `'yes'` works today and breaks after this PR. Tightening the contract is defensible, but it should not land as a "Bug fix (non-breaking change)" — either widen the `case` arms to keep accepting the legacy spellings, or call this out as breaking in the PR description and the input docs.

package.json
```diff
@@ -15,7 +15,7 @@
   ],
   "scripts": {
     "postinstall": "node scripts/install.js",
-    "test:github-actions": "node scripts/github-actions/post-review-comments.test.js && node scripts/github-actions/check-translation-sync.test.js",
+    "test:github-actions": "node scripts/github-actions/post-review-comments.test.js && node scripts/github-actions/check-translation-sync.test.js && node scripts/github-actions/action-contract.test.js",
```
lizhengfeng101, line 18: **The new 18-test suite will not gate anything — `test:github-actions` is never invoked in CI.**

Grepping `.github/workflows/` and the `Makefile` finds no reference to `npm run test:github-actions`. The only place any of these files runs is `.github/workflows/translation-sync.yml:38`, which calls `check-translation-sync.test.js` directly. (The `npm run` calls in `pages-ci.yml` belong to the `pages/` package, not this one.)

As a result `post-review-comments.test.js` is already unenforced, and this PR adds a third suite to the same unenforced script. Since the whole value of a contract test is catching a future edit to `action.yml`, please wire `test:github-actions` into `ci.yml` (or a `make check` target) as part of this change.

action.yml
```diff
@@ -42,6 +42,11 @@ inputs:
   llm_timeout:
     description: LLM request timeout in seconds (mapped to env OCR_LLM_TIMEOUT).
     required: false
+    default: '300'
```
lizhengfeng101, line 48: **Docs are not updated for either input change.**

`examples/github_actions/README.md:164` carries the action-inputs table. The new `review_timeout` input is missing from it, and this `llm_timeout` default (previously unset) is not recorded either.

Worth noting for reviewers that `300` is behaviourally a no-op — it matches the client's own fallback of `5 * time.Minute` (`internal/llm/client.go:390-391`, `779`, `845`). Making it explicit is fine, but it is now a documented-looking default in `action.yml` that contradicts the undocumented state of the README table.

action.yml
```diff
@@ -276,7 +331,8 @@ runs:
         OCR_RULE: ${{ inputs.rule }}
       shell: bash
       run: |
-        ARGS=(--from "${MERGE_BASE}" --to "${HEAD_SHA}" --audience agent --format json)
+        export OCR_LLM_TIMEOUT="${OCR_LLM_TIMEOUT:-300}"
+        ARGS=(--from "${MERGE_BASE}" --to "${HEAD_SHA}" --audience agent --format json --timeout "$REVIEW_TIMEOUT")
```
lizhengfeng101: **`--timeout` is appended unguarded, unlike every other optional argument here.**

The three lines directly below all gate on `[ -n "$VAR" ]` before appending. This one expands `$REVIEW_TIMEOUT` unconditionally, so if the value is ever empty — the validation step skipped, or a caller injecting `REVIEW_TIMEOUT` into the job env — the action passes `--timeout ""` and the user gets an opaque cobra parse error instead of the actionable message the validation step was written to produce.

A `[ -n "$REVIEW_TIMEOUT" ]` guard (or an explicit assertion that the validation step ran) would keep this consistent with the surrounding lines.

scripts/github-actions/action-contract.test.js
```diff
@@ -0,0 +1,812 @@
...
+      if (/^  [A-Za-z0-9_]+:\s*$/.test(lines[cursor])) break;
+      const defaultMatch = /^    default:\s*(.*)$/.exec(lines[cursor]);
+      if (defaultMatch) defaultValue = parseScalar(defaultMatch[1]);
+    }
+    inputs[name] = { default: defaultValue };
+  }
+  return inputs;
+}
+
+function parseSteps(text) {
```
lizhengfeng101, line 68: **The hand-rolled YAML extractor fails open, which is the wrong direction for a contract test.**

Four assumptions are load-bearing and none is asserted:

1. Exact indentation — `/^    - name:/`, `/^      run:/`, `/^        /` for body dedent.
2. Literal block scalars only — `/^      run:\s*\|\s*$/` (line 74). Reformatting a step to `run: >-` or `run: >` yields `step.run === undefined`.
3. Single-line `env:` values — the loop at line 98 `break`s on the first non-matching line, so a folded or multi-line env value truncates the map silently.
4. `resolveInputExpressions` only substitutes `${{ inputs.* }}`; any `${{ github.* }}` or `${{ steps.* }}` expression is left in the script verbatim.

`renderedRun` does assert that `step.run` is a string, which catches (2) for steps fetched by name — but `parseSteps` returning an empty `env` map is unasserted, and (1) turns into `stepNamed(...)` returning `undefined`, i.e. a test that stops exercising the thing it is named after. Asserting the expected step count and a non-empty `env` for each step under test would make these failures loud.

Two smaller things in the same file: `configValues`'s `key === "llm"` aggregate-JSON branch is dead code — the action never runs `ocr config set llm '{...}'`; and `escapedRegExp`'s character class lists `[` twice, once unescaped and once escaped.

lizhengfeng101 (commented):
Reviewed the change; the root-cause diagnosis is right. `OCR config file` is tried before `OCR environment` (`internal/llm/resolver.go:129-131`), and the old config only held `llm.extra_body`, so `tryLegacyLlmConfig` returned `ok=false` on the incomplete block and `extra_body` was silently dropped. Completing the `llm` block does fix it — I replayed the full `Configure OCR` command sequence against a locally built binary and `extra_body` now lands in the `llm` block as intended. The 18 contract tests also pass locally.

Below are seven specific points, inline at the relevant lines.

action.yml
crestonepeaks: Addressed in 92dd6a2: the input is now named review_task_timeout, its description states the per-file/concurrent-task scope, validation and GITHUB_ENV use the same name, and the CLI receives the normalized value once via --timeout. Contract tests cover naming, boundaries, normalization, ordering, and forwarding.

action.yml
crestonepeaks: Addressed in 92dd6a2: ocr_version now documents a v1.9.6 minimum, and the install step fails early with an actionable error when the version command fails or reports an older, prerelease, or malformed version. Tests cover 1.9.5 rejection, 1.9.6/1.9.10/1.10.0 acceptance, malformed SemVer, prerelease, build metadata, and command failure.

action.yml
crestonepeaks: Addressed in 92dd6a2: llm_use_anthropic now preserves the legacy resolver contract case-insensitively. An explicitly supplied empty string, true, 1, or yes canonicalizes to true/anthropic; every other supplied value canonicalizes to false/openai. The executable matrix covers the legacy spellings and canonical config output.

package.json
crestonepeaks: Addressed in 92dd6a2: ci.yml now has a dedicated GitHub-hosted github-actions job that runs npm run test:github-actions. Its checkout and setup-node dependencies are commit-pinned, and ordered path filters re-include examples/github_actions/README.md so documentation-contract regressions also trigger the gate.

action.yml
crestonepeaks: Addressed in 92dd6a2: the GitHub Actions README now documents the independent review_task_timeout and llm_timeout defaults/units/semantics, includes a 30-minute task and 900-second request example, and documents the OCR v1.9.6 floor with a supported pinned-version sample. These docs are enforced by the contract suite.

action.yml
crestonepeaks: Addressed in 92dd6a2: Run OpenCodeReview now explicitly fails with an actionable error if the validated REVIEW_TASK_TIMEOUT is missing before constructing ARGS. A contract test executes the run step without the validated value and proves OCR is not invoked.

scripts/github-actions/action-contract.test.js
crestonepeaks: Addressed in 92dd6a2: the harness now fails closed on unsupported step indentation, named and block-style nameless steps, non-literal run scalars, multiline env values, duplicate step names, and unresolved expressions. Required step/run/env topology is asserted, the dead aggregate llm branch was removed, and the regex helper was corrected. All 24 action contracts pass.

crestonepeaks:
@lizhengfeng101 All seven review points are addressed at exact head `92dd6a2e58ccff5801a1c9f5437e02abdcca2a3c`, with a direct reply and test evidence in each thread. Exact-head local gates are clean, including the 24 action contracts and Go 1.26.6 `make check`, full race-enabled `make test`, and `make build`. When convenient, please take a fresh review pass.

action.yml
```diff
@@ -244,21 +252,88 @@ runs:
...
+
     - name: Install OpenCodeReview
       shell: bash
       env:
         OCR_VERSION: ${{ inputs.ocr_version }}
       run: |
         npm install -g "@alibaba-group/open-code-review@${OCR_VERSION}"
         echo "OpenCodeReview installed:"
-        ocr version || true
+        OCR_VERSION_OUTPUT="$(ocr version 2>&1)" || {
```
lizhengfeng101: `2>&1` folds stderr into the captured output and the check below only inspects the first line, so the gate's correctness now depends on stderr being empty at this exact moment. That isn't guaranteed: `bin/ocr.js:27-36` writes the update-available hint to stderr via `console.error` whenever `~/.opencodereview/update-available` exists (written by `scripts/update.js:118`). When it does, the hint becomes the first line and a valid install is rejected:

```
[ocr] A new version (v1.9.11) is available. Run to update:
open-code-review 1.9.10 linux/amd64
::error::Unable to verify the installed OpenCodeReview version; v1.9.6 or newer is required
```

How reachable that is depends on whether `$HOME` survives between jobs on a given runner, which I haven't verified — on GitHub-hosted runners it does not, and for `container:` jobs the runner mounts `_temp/_github_home` in rather than using the image's HOME, so container isolation doesn't settle it either way. But `printVersion()` writes to stdout only, so dropping stderr costs nothing and removes the dependency outright. It also keeps the gate robust against any future non-fatal diagnostic. Worth noting the previous `ocr version || true` could never block the pipeline, so this is the one place the hardening trades availability for strictness without needing to.

No test changes required: the fake `ocr` writes the version to stdout (`action-contract.test.js:228`), so all 24 cases still pass with this applied — I ran them. That's also precisely why the current suite can't catch this, since the fake never writes to stderr. A case that emits a stderr line before the version would pin the fix.

```suggestion
        OCR_VERSION_OUTPUT="$(ocr version 2>/dev/null)" || {
```

action.yml
```diff
@@ -244,21 +252,88 @@ runs:
...
+          echo "::error::Unable to read the installed OpenCodeReview version"
+          exit 1
+        }
+        printf '%s\n' "$OCR_VERSION_OUTPUT"
+        OCR_VERSION_LINE="${OCR_VERSION_OUTPUT%%$'\n'*}"
+        if [[ ! "$OCR_VERSION_LINE" =~ ^open-code-review[[:space:]]+v?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(\+([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?([[:space:]]|$) ]]; then
+          echo "::error::Unable to verify the installed OpenCodeReview version; v1.9.6 or newer is required"
```
lizhengfeng101: For a prerelease the regex above rejects the version (it accepts `+build` metadata but not `-rc.1`, per the `1.9.6-rc.1 → invalid` case in the contract test) and the user gets "v1.9.6 or newer is required". For something like `ocr_version: 2.0.0-rc.1` that message is self-contradictory — the version *is* newer, it's the prerelease suffix that's unsupported.

Rejecting prereleases looks deliberate and I think it's the right call; the message just needs to say so, otherwise it sends people looking for a version problem that isn't there. Line 290 already prints the full output, so dropping the version string from the message loses no diagnostic value.

No test asserts this string (only the "Unable to read" one on line 287, which this leaves alone), and 24/24 still pass with this applied.

```suggestion
          echo "::error::Unable to parse the installed OpenCodeReview version; a stable release v1.9.6 or newer is required (prerelease versions are not supported)"
```

lizhengfeng101 (commented):
Reviewed the diff against the CLI source. Both bugs are real and the root-cause analysis holds up: `tryLegacyLlmConfig` bails out when the block is incomplete (`internal/llm/resolver.go:617`), and because the config-file strategy runs before the env strategy (`resolver.go:128-131`), the old partial block meant `llm.extra_body` was dropped entirely. Writing a complete block is the right fix.

Spot-checks that all came back clean:

- **The v1.9.6 floor is exact.** `llm.auth_token_cmd` landed in f75c43a, and `git tag --contains` puts it first in `v1.9.6`. The other keys used here sit well below the floor (`llm.protocol` v1.7.10, `llm.retry_codes` v1.9.0), so nothing is gated unnecessarily.
- **The `llm_use_anthropic` normalizer matches the CLI contract verbatim** against `tryOCREnv` (`resolver.go:270-279`): empty → anthropic, `true`/`1`/`yes` case-insensitive, everything else openai.
- **`--timeout` semantics** confirmed at `shared_flags.go:194` — concurrent task timeout in minutes, default 10. The README's "not a whole-review wall-clock cap" wording is accurate.
- **The 300s default** matches `client.go:391`.
- **Clearing `llm.extra_headers`/`llm.retry_codes` is safe** — `finalizeResolvedEndpoint` layers the env overrides on top of config, so the env path still works, and the tests pin that ordering.
- 24/24 contract tests pass locally, and still pass with both inline suggestions applied (I verified).

Two small suggestions inline. Neither needs a test change, for the reason noted in the first one.

Non-blocking follow-ups, no need to address in this PR:

1. Re-including `examples/github_actions/README.md` in `paths` pulls the *entire* CI matrix (Go race, Windows, cross-compile, container) for a markdown-only change, when only the new `github-actions` job needs it.
2. `github-actions` is a new job name — branch protection's required checks need it added, or the gate is inert. Easy to miss since nothing errors if it's forgotten.
3. The validation step's leading-zero loop (lines 106-113) is equivalent to `^0*([1-9][0-9]?|1[01][0-9]|120)$` followed by `$((10#$X))`. Current behavior is correct — I checked `0000`, `0121`, and long digit strings — so this is style only.
4. The new job SHA-pins checkout/setup-node while every other job in `ci.yml` uses floating `@v7`. Strictly better, just inconsistent; either pin all or extend `verify-action-pins.sh` beyond `action.yml`.

One coverage note: the fake `ocr` makes every `ocr config set` succeed unconditionally, so the suite can't catch the real CLI rejecting a value. `llm_extra_body: ''` (explicitly empty) still reaches `json.Unmarshal("")` → `invalid JSON` and fails the step. Pre-existing and out of scope here, but given the trouble you went to for an explicitly-empty `llm_timeout`, an `[ -n "$OCR_EXTRA_BODY" ] &&` guard would close the matching gap.

lizhengfeng101 (commented):
One follow-up on the new CI job's runner choice.

GitHub-hosted runners are currently queueing badly for us, and this job is a gate on `action.yml` changes — so a hosted runner makes the contract feedback loop slow in practice. Moving it to `self-hosted` with the `node:24` container we already use elsewhere (`ocr-review.yml`, `deploy-pages.yml`, `release.yml`) avoids the queue and drops the `setup-node` step entirely, which is also one less thing to fetch from github.com on an in-network runner.

.github/workflows/ci.yml
```diff
@@ -112,6 +118,19 @@ jobs:
...
 
+  github-actions:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7
+        with:
+          node-version: '24'
```
lizhengfeng101: Switch to the self-hosted runner with the `node:24` container instead of hosted + `setup-node`.

Why the container rather than `setup-node`: the image tag fixes the Node major exactly as `node-version: '24'` did, so nothing is lost on reproducibility, and it removes a github.com fetch that an in-network runner may have trouble with. `node:24` is Debian-based, so `/bin/bash` and `npm` are both present — the contract harness spawns `/bin/bash` directly, so an Alpine image would break it.

`Trust workspace` mirrors what the other self-hosted container jobs in this file already do, to avoid git's dubious-ownership error after checkout.

See the review comment above for the prerequisite — this suggestion assumes fork-PR workflows require approval.

```suggestion
    runs-on: self-hosted
    timeout-minutes: 10
    container:
      image: node:24
    steps:
      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7

      - name: Trust workspace
        run: git config --global --replace-all safe.directory '*'
```

scripts/github-actions/action-contract.test.js
```diff
@@ -0,0 +1,1124 @@
...
+  assert.match(
+    CI_TEXT,
+    /uses:\s*actions\/checkout@[0-9a-f]{40}\s*#\s*v7/,
+    "the newly added GitHub-hosted contract job must pin checkout to an immutable commit"
+  );
+  assert.match(
+    CI_TEXT,
+    /uses:\s*actions\/setup-node@[0-9a-f]{40}\s*#\s*v7/,
+    "the newly added GitHub-hosted contract job must pin setup-node to an immutable commit"
+  );
```
lizhengfeng101, line 1048: This assertion must change together with the runner suggestion: once the job uses `container: node:24` there is no `setup-node` step, so this would fail.

Replacing it with a container-image assertion keeps the original intent — the Node runtime stays explicitly declared in the workflow rather than inherited from whatever the runner happens to ship — which is the same guarantee the `setup-node` pin was providing.

```suggestion
  assert.match(
    CI_TEXT,
    /container:\s*[\s\S]*?image:\s*node:[0-9]/,
    "the contract job must declare its Node runtime via a pinned container image"
  );
```

lizhengfeng101 (approved):
LGTM

PR #1052, opened 2026-08-24
feat(diff)：增加根据repo和diff运行opencodereview的方法

Strke:
## Description
通过增加--repo和--diff方法，降低了测试ocr review时所需的数据准备空间。

由于使用--from和--to运行ocr review需要完整的下载目标仓库的代码，这会导致我们在测试某个规模较大的代码库时要下载很长时间或占用很多的本地存储。所以增加了基于--repo和--diff的ocr review运行方法，这种方法只需要使用仓库的git浅克隆代码以及pr中的patch，即可适配ocr review的后台运行程序进行执行。在本地测试下来，测试50个仓库所需的存储空间从25.31GB降低到了4.87GB

## Type of Change

- [☑️] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [☑️] Manual testing (describe below)
›准备的数据为
patch文件夹
截屏2026-08-24 11 57 03

代码仓库
截屏2026-08-24 11 59 05

启动命令

截屏2026-08-24 12 01 49

截屏2026-08-24 12 01 16

## Checklist

- [ ] I have signed the CLA

### DIFF
--- a/internal/llm/providers.go
+++ b/internal/llm/providers.go
@@ -87,12 +87,21 @@ var registry = []Provider{
 		BaseURL:     "https://api.openai.com/v1",
 		EnvVar:      "OPENAI_API_KEY",
 		Models: []string{
-			"gpt-5.6-sol",
-			"gpt-5.6-terra",
-			"gpt-5.6-luna",
 			"gpt-5.5",
 			"gpt-5.4",
 			"gpt-5.4-mini",
+		},
+	},
+	{
+		Name:        "openai-responses",
+		DisplayName: "OpenAI Responses API",
+		Protocol:    ProtocolOpenAIResponses,
+		BaseURL:     "https://api.openai.com/v1",
+		EnvVar:      "OPENAI_RESPONSES_API_KEY",
+		Models: []string{
+			"gpt-5.6-sol",
+			"gpt-5.6-terra",
+			"gpt-5.6-luna",
 		},
 	},
 	{
--- a/internal/llm/providers_test.go
+++ b/internal/llm/providers_test.go
@@ -5,6 +5,7 @@ package llm
 
 import (
 	"sort"
+	"strings"
 	"testing"
 )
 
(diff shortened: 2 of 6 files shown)


lizhengfeng101:
lin seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user). @Strke

Strke:
Done, the CLA has been signed.

wu21-web:
It is always better to edit on a branch beside `main`.

Strke:
Thank you for pointing it out. I will take note. ٩( 'ω' )و

lizhengfeng101 (commented):
Nice work on the storage reduction — going from 25GB to 5GB for 50 repos is a big win for the eval workflow.

A few things I'd like to discuss:

---

**1. `NewFileContent` semantic mismatch (want to confirm your intended workflow)**

In `PatchProvider.GetDiff()`, you call `ParseDiffText(ctx, ..., repoDir, "", runner)` with an empty ref. This means `finalizeDiff` reads file content from whatever's on disk in the shallow clone.

In range/commit/workspace modes, `NewFileContent` always represents the file state *after* the reviewed changes. But in patch mode, whether that holds depends entirely on which branch the shallow clone is checked out to:
- Clone of head branch (PR source) → correct, file matches post-change state
- Clone of base branch (e.g. main) → wrong, `NewFileContent` shows pre-change content while the diff says otherwise

The LLM would see contradictory information in the second case. Is your workflow always cloning the head branch? If so, might be worth documenting that constraint (or adding a sanity check like `git apply --check` to catch mismatches early).

---

**2. Missing upfront validation for `--diff` path**

Other path flags like `--repo` get validated early in `validateReviewOptions`. The diff directory is only checked at runtime inside `PatchProvider.GetDiff()`. Would be more consistent to fail fast:

```go
if opts.diffDir != "" {
    info, err := os.Stat(opts.diffDir)
    if err != nil || !info.IsDir() {
        return fmt.Errorf("--diff path %q must be an existing directory", opts.diffDir)
    }
}
```

---

**3. Should `--diff` require `--repo`?**

Right now you can run `ocr review --diff /path/to/patches` without `--repo`, and it silently uses cwd. In the typical shallow-clone-based workflow this feature targets, that's probably never intentional. Worth requiring `--repo` explicitly when `--diff` is set?

---

**4. `PatchProvider.ResolveInput` returns empty `InputResolution`**

This means the manifest won't record any commit info for the run. For traceability (and potential future resume support), consider at least capturing the repo's HEAD — something like:

```go
func (p *PatchProvider) ResolveInput(ctx context.Context) InputResolution {
    if p.runner != nil {
        if out, err := p.runner.Output(ctx, p.repoDir, "rev-parse", "HEAD"); err == nil {
            return InputResolution{ResolvedHead: strings.TrimSpace(string(out))}
        }
    }
    return InputResolution{}
}
```

---

**5. Flag naming: `--diff` → `--patch`?**

`--diff` overlaps with the git concept and reads ambiguously next to `--from`/`--to` (which also produce diffs). Since this is really "external patch files as input", `--patch` would:
- Mirror the internal mode name (`ReviewModePatch`)
- Parallel `--commit` as an alternative diff-source specifier
- Avoid confusion with the diff *output* that every mode produces

```
ocr review --repo /path/to/repo --patch /path/to/patches
```

---

None of these are blocking on their own (the core logic is solid), but #1 and #5 are worth settling before merge. Let me know your thoughts!

PR #1054, opened 2026-08-24
fix(cli): exit 128+signo when the native binary is killed by a signal

### LINKED ISSUE #931: CLI launcher exits 0 when the native binary is killed by a signal (result.status is null)
## Version

@alibaba-group/open-code-review 1.8.10 (also present at current HEAD 9a371c9).

## Bug

In `bin/ocr.js` the launcher forwards the child's exit status as:

```js
process.exit(result.status ?? (result.error ? 1 : 0));
```

When the spawned native binary is terminated by a signal, `spawnSync` returns `status = null` and `error = undefined`, so the launcher exits **0**. There is no `result.signal` branch in the file.

## Reproduction

1. Start any long-running `ocr review`.
2. Send `kill -TERM` (or `-KILL`) to the launcher's **child process** (the native binary, not the node launcher).
3. The launcher exits with code 0. Expected: non-zero, conventionally `128 + signo`.

## Real-world impact

When the OOM killer selects the native binary (typically the largest-RSS process of the pair), downstream automation observes rc=0 with empty stdout and treats the run as a clean success. "Killed" and "clean exit" are indistinguishable at the CLI boundary, which silently corrupts any pipeline that gates on the exit code.

## Suggested fix

```js
const os = require('os');
if (result.signal) process.exit(128 + os.constants.signals[result.signal]);
process.exit(result.status ?? (result.error ? 1 : 0));

### PR DESCRIPTION
**Pull request by nolanchic**
## Description

`bin/ocr.js` forwarded the child's exit status as

```js
process.exit(result.status ?? (result.error ? 1 : 0));
```

When the native binary is terminated by a signal, `spawnSync` returns `status = null` and `error = undefined`, so the launcher exits **0**. Downstream automation gating on the exit code then reads an OOM kill (or any signal death) as a clean success with empty stdout.

This adds the conventional `128 + signo` mapping for signal deaths, keeping the existing `status`/`error` behavior for every other case.

## Changes

- `bin/ocr.js`: new `launcherExitCode(result)` — returns `128 + os.constants.signals[result.signal]` when the child died by signal (falling back to a non-zero `129` for an unrecognized name), otherwise the previous `status ?? (error ? 1 : 0)`. The launcher body is wrapped in a `require.main` guard so the module can be imported by tests without side effects.
- `bin/ocr.test.js`: asserts status passthrough, the spawn-error fallback, the SIGTERM/SIGKILL/SIGINT mappings, and that an unknown signal name still exits non-zero.
- `package.json`: `test:launcher` script alongside the existing `test:update` pattern.

## Test plan

- [x] `node bin/ocr.test.js` — all mappings pass
- [x] `require('./bin/ocr.js')` imports cleanly without running the launcher
- [x] `node bin/ocr.js` with no binary installed still prints the install hint and exits 1 (behavior unchanged)

Fixes #931

### DIFF
--- a/bin/ocr.js
+++ b/bin/ocr.js
@@ -14,6 +14,24 @@ const { resolveNativeBinary } = require("../scripts/platform");
 const { version: packageVersion } = require("../package.json");
 const { shouldShowUpdateHint } = require("../scripts/version");
 
+// Maps a child process result to the launcher exit code.
+// A child killed by a signal has status = null, which must not fall through
+// to a clean 0 — pipelines gating on the exit code would read an OOM kill
+// (or any signal death) as success. Conventional 128+signo is used instead.
+function launcherExitCode(result) {
+  if (result.signal) {
+    return 128 + (os.constants.signals[result.signal] ?? 1);
+  }
+  return result.status ?? (result.error ? 1 : 0);
+}
+
+module.exports = { launcherExitCode };
+
+if (require.main !== module) {
+  // Required as a module (tests); the launcher body below must not run.
+  return;
+}
+
 const resolved = resolveNativeBinary();
 if (!resolved) {
   console.error(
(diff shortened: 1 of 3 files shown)


lizhengfeng101 (approved):
nice work！

PR #938, opened 2026-08-16
fix(llm): serve GPT-5.6 models via the OpenAI Responses API (#559)

ousamabenyounes:
## Description

The built-in `openai` provider preset lists `gpt-5.6-sol`, `gpt-5.6-terra`, and
`gpt-5.6-luna` but uses the Chat Completions protocol (`/v1/chat/completions`).
Selecting any of these models for `ocr review` completes the planning phase and
then fails on the first tool-using subtask with a 400:

> Function tools with reasoning_effort are not supported for gpt-5.6-terra in /v1/chat/completions. Use /v1/responses instead.

The Responses protocol (`ProtocolOpenAIResponses`) already exists in the code,
but no built-in preset paired the GPT-5.6 models with it, and the resolver picks
the protocol strictly from the selected preset (no per-model routing). So every
built-in path to a GPT-5.6 model was broken.

This moves the GPT-5.6 models out of the chat-completions `openai` preset into a
dedicated `openai-responses` preset that uses `ProtocolOpenAIResponses`. A user
can now select a GPT-5.6 model without knowing which OpenAI API protocol is
required. Other `openai` models (`gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`) are
unchanged.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## How Has This Been Tested?

- [x] `make test` / `go test -race ./...` passes locally
- [x] Added unit tests, verified RED on the unmodified base and GREEN with the fix

### Test verification (RED → GREEN)

RED — new tests against unmodified `main` (GPT-5.6 still under the chat-completions `openai` preset):

```
--- FAIL: TestListProviders_Order (0.00s)
    providers_test.go:81: expected 26 providers, got 25
--- FAIL: TestLookupProvider_OpenAIDetails (0.00s)
    providers_test.go:179: Models length = 6, want 3
--- FAIL: TestLookupProvider_OpenAIResponsesDetails (0.00s)
    providers_test.go:191: openai-responses not found
--- FAIL: TestGPT56ModelsUseResponsesProtocol (0.00s)
    providers_test.go:239: provider "openai" offers "gpt-5.6-sol" over "openai"; GPT-5.6 needs "openai-responses"
    providers_test.go:239: provider "openai" offers "gpt-5.6-terra" over "openai"; GPT-5.6 needs "openai-responses"
    providers_test.go:239: provider "openai" offers "gpt-5.6-luna" over "openai"; GPT-5.6 needs "openai-responses"
[...]

PR #1057, opened 2026-08-25
feat(rules): add project review rule for providers.go

lizhengfeng101:
## Summary

Adds a `.opencodereview/rule.json` that enforces review standards specifically for `internal/llm/providers.go` — the built-in provider registry.

The rule covers three dimensions:

- **Style consistency** — inline string literals for all field values (no one-use constants), canonical field order, EnvVar naming convention, protocol constants only
- **Documentation sync** — any provider addition/modification must update the built-in provider table in all four locale docs (en/zh/ja/ru `configuration.md`)
- **Test coverage** — new providers require a `TestLookupProvider_<Name>Details` test; protocol-sensitive models require a regression guard test

Uses `merge_system_rule: true` so the Go system rule (error handling, concurrency, security, etc.) remains active alongside the project-specific constraints.

## Motivation

PR #938 exposed that contributors can introduce style inconsistencies and miss documentation/test updates when modifying the provider registry. This rule automates the enforcement during code review.

## Verification

- `ocr rules check internal/llm/providers.go` confirms the rule resolves correctly (source=project, pattern matches)
- Regression tested against an intentionally violating commit: all three violation categories (package-level constant, missing docs, missing tests) were flagged as high-severity findings

### DIFF
--- a/.opencodereview/rule.json
+++ b/.opencodereview/rule.json
@@ -1,0 +1,9 @@
+{
+  "rules": [
+    {
+      "path": "internal/llm/providers.go",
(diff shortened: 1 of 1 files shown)


wu21-web (approved):

wu21-web:
This is great to avoid changes, for instance in  https://github.com/alibaba/open-code-review/pull/938#pullrequestreview-5008447495

stay-foolish-forever (approved):

PR #1063, opened 2026-08-25
feat(cmd):Add a method to initiate OCR based on a single branch and its corresponding patch

Strke:
## Description

By adding the --repo, --branch, --patch, and --apply-patch methods, the data preparation space required for testing OCR review has been reduced.

Running ocr review with --from and --to requires downloading the complete code of the target repository, which can lead to long download times or significant local storage usage when testing a large codebase. Therefore, a new method for running ocr review has been added, which only requires using the repository's git shallow clone code and the patch in the pr to adapt the ocr review backend running program for execution. Local testing has shown that the storage space required to test 50 repositories has been reduced from 25.31GB to 4.87GB (using the code-review-benchmark dataset)

--repo: directory address of the code repository
--branch: The target branch of the code repository
--patch: Directory address of the .patch file
--apply-patch: Whether to apply the patch to the current branch

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

截屏2026-08-26 01 02 43

截屏2026-08-26 01 04 56

## Checklist

- [x] I have signed the CLA


Strke:
Based on the fact that it seems inconvenient to resolve conflicts on the main branch, I have submitted a new pull request

Strke:
https://github.com/alibaba/open-code-review/pull/1052#pullrequestreview-5006739248

These issues have been corrected

Strke:
https://github.com/alibaba/open-code-review/pull/1052#pullrequestreview-5015444408

---

**1. NewFileContent semantic mismatch (want to confirm your intended workflow)**

In my understanding, `NewFileContent` should be the complete file content of a modified file after the changes are completed.

Currently, I've outlined my logic and modified the code, which involves two execution logics:

With the `--apply-patch` flag: In this case, the code will apply the patch to the latest commit on the branches specified by `--repo` and `--branch`, forming a new commit. This new commit is the context that the larger model can see.

Without the `--apply-patch` flag: In this case, we need to ensure that the latest commit on the branches specified by `--repo` and `--branch` is the commit that applied the patch, so that the larger model can see the context corresponding to that patch.

---

**2. Missing upfront validation for --diff path**

An advance check for --patch has now been added.

---

**3. Should --diff require --repo?**

The `--patch` option can be used alone: 
​​`ocr review --patch /path/to/patches`
 In this case, `--repo` defaults to the top-level directory of the Git repository where the current working directory is located; `--branch` defaults to the HEAD of that repository.

---

**4. PatchProvider.ResolveInput returns empty InputResolution**
Previously, commits were only parsed when explicitly passed `--branch` or `--apply-patch`. Now, whenever `--patch` is used, the base branch is determined first:
- If `--branch` is present: parse the branch
- If `--branch` is not present: parse the current HEAD

Then it is fixed as the commit SHA and passed to `PatchProvider`. This allows `ResolveInput()` to write the SHA to the manifest's `ResolvedHead`, preventing it from being empty.

---

**5. Flag naming: --diff → --patch?**

Now diff has been replaced with patch.

lizhengfeng101 (commented):
Two minor suggestions:

1. **`validateResumeIdentity` variadic**: Nit: the variadic `patchInputs ...*diff.InputResolution` feels like a workaround to avoid touching callers — just make it an explicit param, it's cleaner and only two call sites need updating.

2. **`review_cmd.go` patch logic**: The patch-specific branches scattered through `executeReviewContext` are getting hard to follow — would you mind pulling them into a helper like `resolvePatchInput(ctx, cc, opts)` so the main flow reads top-to-bottom again?

Strke:
> Two minor suggestions:
> 
> 1. **`validateResumeIdentity` variadic**: Nit: the variadic `patchInputs ...*diff.InputResolution` feels like a workaround to avoid touching callers — just make it an explicit param, it's cleaner and only two call sites need updating.
> 2. **`review_cmd.go` patch logic**: The patch-specific branches scattered through `executeReviewContext` are getting hard to follow — would you mind pulling them into a helper like `resolvePatchInput(ctx, cc, opts)` so the main flow reads top-to-bottom again?

1、Regarding the first point, I changed the variable-length parameter `patchInputs ...*diff.InputResolution` to a fixed-length explicit parameter `patchInput *diff.InputResolution`, making the semantics clearer.

2、Regarding the second point, I've moved the patch-related logic into the functions `patchInput, err := resolvePatchInput(ctx, cc, opts)` and `sealedInput, mode := resolveSealedReadState(sealed, patchInput, opts)`.

`patchInput, err := resolvePatchInput(ctx, cc, opts)`: This function applies a patch to the current branch of the current repository and saves it as a commit. It also records the commit hashes before and after applying the patch, and the output patchInput contains these two hash values.

`sealedInput, mode := resolveSealedReadState(sealed, patchInput, opts)`: This function determines which commit we will process next. If a value exists in `sealed` (original mode), the commit processed in the original mode will be used; otherwise, the commit in our `patch` mode will be used.

PR #1066, opened 2026-08-26
fix(llmloop): record grace round requests

Gongyl01:
## Description

The grace round is the final LLM request sent after the main loop reaches its tool-request limit.

This change completes its request-level observability:

- create the `main_task` record before sending the request
- attach request identity using the record's `RequestNo`, allowing retry attempts to be correlated with the request
- record the response, token usage, and actual duration even when no tool call is returned
- record the request error and duration when the grace request fails
- emit the same LLM telemetry as regular main-task requests
- add tests for successful responses and suppressed request errors

## Type of Change

## How Has This Been Tested?

- [x] `make check`
- [x] `make test`
- [x] `make coverage` - 91.4%
- [x] Grace-round response and suppressed-error paths

## Checklist

- [x] I have added tests that prove the fix is effective
- [x] New and existing unit tests pass locally

### DIFF
--- a/internal/llmloop/loop.go
+++ b/internal/llmloop/loop.go
@@ -448,33 +448,47 @@ func (r *Runner) runGraceRound(ctx context.Context, messages []llm.Message, newP
 		return
 	}
 
-	resp, err := r.deps.LLMClient.CompletionsWithCtx(ctx, llm.ChatRequest{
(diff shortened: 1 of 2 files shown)


lizhengfeng101 (approved):
LGTM

PR #1070, opened 2026-08-26
fix(llm): preserve provider reasoning/thinking across assistant turns

### LINKED ISSUE #805: Thinking state is parsed but dropped when assistant tool-call turns are replayed
## OpenCodeReview version

- v1.8.10: `4bcc95acecd629e64ea984c962cfd08dd892a38e`
- Reproduced on current `main`: `f44821d9aa0993e8e3cf90ae3fe12e733e045b15`
- Installation: built from source
- OS used for the deterministic test: macOS (Apple Silicon)
- Affected protocols: OpenAI-compatible chat; the same flattening also affects Anthropic thinking and Responses reasoning items

## Bug description

OpenCodeReview parses provider reasoning/thinking from a model response, but drops its native representation before the next tool-call turn. The orchestration loop reconstructs assistant history from only visible `content` and normalized `tool_calls`.

This has two failure modes:

1. Strict thinking providers can reject the next request because the assistant tool-call message is missing required replay state such as `reasoning_content`.
2. Tolerant gateways may return HTTP 200 but silently lose reasoning continuity, so a successful request does not prove that the model received the previous assistant turn as produced.

When visible content is empty, `ChatResponse.Content()` currently moves the reasoning text into ordinary assistant `content`. That avoids an empty string but does not preserve the original field or its protocol semantics.

## Source trace

The data-loss path is explicit:

1. The response type has [`ResponseMessage.ReasoningContent`](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llm/client.go#L138-L144).
2. Both [streaming](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llm/client.go#L395-L464) and [non-streaming](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llm/client.go#L534-L593) OpenAI-compatible paths populate it.
3. `RunPerFile` reduces the response to [`resp.Content()` and `resp.ToolCalls()`](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llmloop/loop.go#L234-L235), then passes only those values to [`addNextMessage`](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llmloop/loop.go#L292).
4. Request-side [`Message`](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llm/client.go#L46-L55) has no reasoning/replay field.
5. [`NewToolCallMessage`](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llm/client.go#L71-L79) and [`addNextMessage`](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llmloop/loop.go#L487-L519) therefore append only `content` and `tool_calls`.
6. [`ChatResponse.Content()`](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llm/client.go#L154-L165) either drops reasoning when visible content exists or returns it as ordinary content when visible content is empty.

The provider-specific clients show the wider form of the problem:

- Anthropic [parses thinking blocks](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llm/client.go#L825-L895), but assistant replay is rebuilt from only [text and tool-use blocks](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llm/client.go#L707-L741), losing signed/redacted thinking blocks.
- Responses [parses reasoning summaries](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llm/responses_client.go#L200-L279), but replay emits only [assistant text and function calls](https://github.com/alibaba/open-code-review/blob/f44821d9aa0993e8e3cf90ae3fe12e733e045b15/internal/llm/responses_client.go#L133-L157), not the original output items/encrypted reasoning.

## Deterministic reproduction

This source-level regression requires no live model or API endpoint:

```go
func TestAssistantToolCallHistoryPreservesReasoningContent(t *testing.T) {
	emptyContent := ""
	response := &ChatResponse{Choices: []Choice{{Message: ResponseMessage{
		Content:          &emptyContent,
		ReasoningContent: "private reasoning that the provider requires on replay",
		ToolCalls:        []ToolCall{{ID: "call_1"}},
	}}}}

	// Same response-to-history conversion used by llmloop.Runner.
	historyMessage := NewToolCallMessage(response.Content(), response.ToolCalls())
	payload, err := json.Marshal(historyMessage)
	if err != nil {
		t.Fatal(err)
	}
	if !bytes.Contains(payload, []byte(`"reasoning_content":"private reasoning that the provider requires on replay"`)) {
		t.Fatalf("assistant tool-call history dropped reasoning_content: %s", payload)
	}
}
```

Run:

```shell
go test ./internal/llm -run '^TestAssistantToolCallHistoryPreservesReasoningContent$' -count=1
```

Actual result on both revisions:

```text
--- FAIL: TestAssistantToolCallHistoryPreservesReasoningContent
assistant tool-call history dropped reasoning_content:
{"role":"assistant","content":"private reasoning that the provider requires on replay","tool_calls":[...]}
```

## Expected behavior

After an assistant returns tool calls, the next request should contain the complete provider-valid assistant turn that preceded the tool results. The review loop should consume a normalized view for tool execution, but it should not reconstruct provider history from that lossy view.

## Proposed solution: adapter-owned assistant turns

The most durable fix is to introduce a deep module at the provider adapter seam:

```go
type AssistantTurn struct {
	Parts  []Part       // normalized text, reasoning summary, and tool calls
	Replay ReplayState  // opaque provider-native continuation state
}

type ReplayState struct {
	Provider string
	Model    string
	Raw      json.RawMessage
}
```

The exact names are unimportant; the invariants are:

1. The adapter returns one finalized `AssistantTurn`; the loop appends that turn as a whole before appending tool results.
2. `Parts` is the normalized application view used by OCR tools, comments, metrics, and redacted logs.
3. `Replay` is adapter-owned and is the only source used to serialize the next provider request. The loop must not parse, edit, reorder, or regenerate it.
4. Streaming assembly belongs to the adapter and finalization must preserve typed indexes/signatures before a turn becomes replayable.
5. Compression may summarize completed history, but it must treat the latest assistant/tool segment atomically and retain its replay envelope.
6. Persistence/logging should store a redacted projection separately from restricted replay state; hidden reasoning should not be exposed merely to make continuation work.

This shape aligns with current agent protocols:

- [Kimi K3](https://huggingface.co/moonshotai/Kimi-K3/blob/main/README.md#6-model-usage) requires replaying the complete assistant message, including `reasoning_content` and tool calls.
- [Anthropic](https://platform.claude.com/docs/en/build-with-claude/thinking#preserving-thinking-blocks) requires complete, unmodified thinking/redacted-thinking blocks and signatures around tool use.
- [OpenAI Agents/Responses](https://openai.github.io/openai-agents-python/models/) preserves typed reasoning items/encrypted content or uses server-managed continuation rather than flattening them to text.
- [Gemini](https://ai.google.dev/gemini-api/docs/thinking#thought-signatures) uses opaque thought signatures that must accompany later tool turns.
- [OpenCode's provider transform](https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/provider/transform.ts) retains reasoning parts and provider options, then applies provider-specific continuation rules.

PR-Agent is not an implementation reference for this particular problem: its own configuration says the review flow is a [single-shot model call with no tool-use loop](https://github.com/The-PR-Agent/pr-agent/blob/main/pr_agent/settings/configuration.toml#L1803-L1817).

### Incremental rollout

1. Replace the loop's separate `Content()`/`ToolCalls()` history append with a complete assistant turn.
2. Implement OpenAI-compatible replay first: preserve and serialize `reasoning_content` in both streaming and non-streaming paths.
3. Add adapter-native Anthropic blocks/signatures and Responses items/encrypted content without forcing them through one universal string field.
4. Update active-history copying, compression boundaries, and any persistence format that claims to preserve model requests.

## Regression coverage

- A strict two-turn fake server: first response has visible content, `reasoning_content`, and a tool call; second request must replay all three fields.
- Repeat with empty visible content to ensure reasoning is not moved into ordinary `content`.
- Streaming and non-streaming OpenAI-compatible responses.
- Multiple parallel tool calls and matched tool results.
- Anthropic signed/redacted thinking blocks.
- Responses typed reasoning items/encrypted content.
- Compression must not split or strip the active assistant/tool segment.
- Redacted logs must not become the replay source.

## Existing issues and PRs

- #1 was closed after adding `llm.extra_body` so users could disable thinking. That is a useful workaround, but it does not preserve reasoning-enabled tool turns.
- #773 backfills a comment's display/output `thinking` field from the current response. It does not add reasoning to assistant history or replay it on the next request.

I could not find an existing issue that tracks complete provider-native assistant-turn replay across tool calls.

lizhengfeng101:
## Summary

Fixes #805. Assistant tool-call history was rebuilt from `Content()`/`ToolCalls()` alone, so provider reasoning state was silently dropped when replaying a turn on the next request — across all three supported protocols:

- **OpenAI-compatible Chat Completions**: `reasoning_content` was parsed from the response but never sent back on the next request.
- **Anthropic Messages**: `thinking`/`redacted_thinking` blocks were flattened into a display string, losing the `signature` required to replay them (Anthropic rejects a modified/missing signature with a 400).
- **OpenAI Responses**: `reasoning` output items (`encrypted_content`) were dropped entirely from the next request's `input`.

### Approach

Adds `Message.Native` (`NativeTurn`), an opaque per-protocol replay payload built by reusing each SDK's own response→request converter (`anthropic.Message.ToParam()`, each Responses output item's `ToParam()`) instead of hand-rolling a serialization format. Each adapter type-asserts its own payload type before reuse and falls back to the existing reconstruction otherwise — a cross-provider turn or a turn built without a native payload is unaffected.

### Also included

- `Message.ReasoningContent`: a plain-text projection (separate from the opaque `Native` payload) used by the compression summarizer, so a reasoning-only turn no longer renders as a blank message in the summarization prompt.
- Fixed `NativeTurn.EstimatedTokens()` double-counting visible content already counted via `ExtractText()`, which could trigger compression earlier than necessary.
- Guarded the Responses adapter against carrying a dangling `reasoning` item (no accompanying `message`/`function_call`) into `Native`, which risked a 400 on replay.
- `reasoning_content` is now persisted in the session JSONL transcript and surfaced in the session viewer, for audit visibility into turns with no visible answer.
- Made Anthropic extended thinking (`extra_body.thinking`) safe to enable via the existing provider config: it's dropped for a given request when it would conflict with a forced `tool_choice` or exceed that call's `max_tokens`, instead of failing the request outright.
- Removed `GroupingTask`'s hardcoded `max_tokens: 4096` cap in favor of the template's own completion limit — this was the actual trigger for the `max_tokens`/`budget_tokens` conflict above once thinking was enabled.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] Deterministic per-protocol reproduction tests (`internal/llm/native_turn_test.go`) asserting the replayed request body contains the original reasoning payload byte-for-byte
- [x] Verified against real multi-turn runs with Anthropic extended thinking enabled (`extra_body.thinking`): reasoning + tool_use turns replay cleanly across dozens of rounds with zero errors

### DIFF
--- a/internal/agent/grouping.go
+++ b/internal/agent/grouping.go
@@ -111,6 +111,10 @@ func callGroupingLLM(ctx context.Context, diffs []model.Diff, client llm.LLMClie
 		})
 	}
 
+	if maxTokens <= 0 {
+		maxTokens = 4096
+	}
+
 	resp, err := client.CompletionsWithCtx(ctx, llm.ChatRequest{
 		Model:     modelName,
 		Messages:  messages,
--- a/internal/agent/grouping_test.go
+++ b/internal/agent/grouping_test.go
@@ -21,9 +21,11 @@ import (
 type fakeGroupingClient struct {
 	response string
 	err      error
+	gotReq   llm.ChatRequest
 }
 
-func (f *fakeGroupingClient) CompletionsWithCtx(_ context.Context, _ llm.ChatRequest) (*llm.ChatResponse, error) {
+func (f *fakeGroupingClient) CompletionsWithCtx(_ context.Context, req llm.ChatRequest) (*llm.ChatResponse, error) {
+	f.gotReq = req
 	if f.err != nil {
 		return nil, f.err
 	}
@@ -281,6 +283,35 @@ func TestCallGroupingLLM_EmptyResponse(t *testing.T) {
 	_, _, err := callGroupingLLM(context.Background(), diffs, client, "fake", task, 4096, nil)
 	if err == nil {
 		t.Fatal("expected error for empty response")
+	}
+}
+
+// TestCallGroupingLLM_UsesTemplateMaxTokens guards against reintroducing the
+// hardcoded MaxTokens: 4096 this replaced: a small, task-specific cap left no
+// room for a provider config that enables Anthropic extended thinking via
+// extra_body.thinking with a larger budget_tokens, so the grouping call
+// failed with "max_tokens must be greater than thinking.budget_tokens" even
+// though the main review loop's own MAX_TOKENS was large enough.
+func TestCallGroupingLLM_UsesTemplateMaxTokens(t *testing.T) {
+	diffs := []model.Diff{{NewPath: "a.go"}}
+	task := &template.LlmConversation{
+		Messages: []template.ChatMessage{{Role: "user", Content: "{{file_list}}"}},
+	}
+
+	client := &fakeGroupingClient{response: `[{"label":"a","files":["a.go"]}]`}
+	if _, _, err := callGroupingLLM(context.Background(), diffs, client, "fake", task, 32000, nil); err != nil {
+		t.Fatalf("callGroupingLLM: %v", err)
+	}
+	if client.gotReq.MaxTokens != 32000 {
+		t.Errorf("MaxTokens = %d, want 32000 (the template's own limit)", client.gotReq.MaxTokens)
+	}
+
+	client = &fakeGroupingClient{response: `[{"label":"a","files":["a.go"]}]`}
+	if _, _, err := callGroupingLLM(context.Background(), diffs, client, "fake", task, 0, nil); err != nil {
+		t.Fatalf("callGroupingLLM: %v", err)
+	}
+	if client.gotReq.MaxTokens != 4096 {
+		t.Errorf("MaxTokens = %d, want fallback 4096 when the template leaves MAX_TOKENS unset", client.gotReq.MaxTokens)
 	}
 }
 
--- a/internal/llm/client.go
+++ b/internal/llm/client.go
@@ -43,6 +43,14 @@ var AppVersion = "dev"
 // shrink it, same as keyCmdTimeout.
 var bedrockConfigLoadTimeout = 60 * time.Second
 
+// defaultAnthropicMaxTokens is the max_tokens value buildAnthropicParams
+// falls back to when the caller leaves ChatRequest.MaxTokens unset. Extended
+// thinking's budget_tokens must stay strictly below whatever max_tokens is
+// actually sent, so the extra_body.thinking guard in CompletionsWithCtx
+// compares against this same default rather than the raw (possibly zero)
+// ChatRequest.MaxTokens.
+const defaultAnthropicMaxTokens = 8192
+
 func userAgent(provider string) string {
 	ua := "open-code-review/" + AppVersion
 	if provider != "" {
@@ -1307,7 +1315,11 @@ func (c *AnthropicClient) CompletionsWithCtx(ctx context.Context, req ChatReques
 			if req.ToolChoice == "required" {
 				continue
 			}
-			if budget, ok := anthropicThinkingBudgetTokens(v); ok && req.MaxTokens > 0 && budget >= int64(req.MaxTokens) {
+			effectiveMaxTokens := int64(req.MaxTokens)
+			if effectiveMaxTokens <= 0 {
+				effectiveMaxTokens = defaultAnthropicMaxTokens
+			}
+			if budget, ok := anthropicThinkingBudgetTokens(v); ok && budget >= effectiveMaxTokens {
 				continue
 			}
 		}
@@ -1442,7 +1454,7 @@ func (c *AnthropicClient) buildAnthropicParams(model string, req ChatRequest) (a
 
 	maxTokens := int64(req.MaxTokens)
 	if maxTokens <= 0 {
-		maxTokens = 8192
+		maxTokens = defaultAnthropicMaxTokens
 	}
 
 	params := anthropic.MessageNewParams{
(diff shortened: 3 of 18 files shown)


css521 (approved):
LGTM

PR #1071, opened 2026-08-26
fix(ci): bump artifact actions to node24 runtimes

lwiles692:
## Description

GitHub Actions runners are deprecating Node.js 20: steps that still target `node20` get force-run on Node 24 with a deprecation warning (see [the changelog](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)). Consumers of this action currently see:

> Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02

This bumps every `actions/upload-artifact` / `actions/download-artifact` usage in the repo to the Node 24 runtimes:

| File | Before | After |
|---|---|---|
| `action.yml` | `upload-artifact@v4.6.2` (SHA-pinned, node20) | `upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a` (SHA-pinned, `v7.0.1`, node24) |
| `.github/workflows/release.yml` | `upload-artifact@v4` | `upload-artifact@v7` |
| `.github/workflows/release.yml` (×2) | `download-artifact@v4` | `download-artifact@v7` |

Compatibility notes:

- All inputs in use (`name`, `path`, `if-no-files-found`; `pattern`, `merge-multiple`) exist unchanged in v7.
- upload-artifact v7 adds an `archive` input that defaults to `true`, so the existing multi-file zipped upload behavior is preserved.
- v6/v7 of both actions require Actions Runner ≥ 2.327.1. GitHub-hosted runners already satisfy this; only consumers on older self-hosted runners need to update their runner before picking this up.

## Type of Change

- [ ] Documentation update
- [x] CI / Build / Tooling

## How Has This Been Tested?

Verified against `actions/upload-artifact` and `actions/download-artifact` release notes / `action.yml` that the pinned v7.0.1 runs on `node24` and that every input used here is present with the same semantics. Pure workflow/action version bumps — no source code touched.

## Checklist

- [ ] I have signed the CLA

## Related Issues

None — preemptive cleanup driven by the runner-side Node 20 deprecation warning.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,7 +52,7 @@ jobs:
           go build -ldflags "${LD_FLAGS}" -o "${BIN_NAME}" ./cmd/opencodereview
           echo "bin_name=${BIN_NAME}" >> $GITHUB_ENV
 
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@v7
         with:
           name: binary-${{ matrix.goos }}-${{ matrix.goarch }}
           path: ${{ env.bin_name }}
@@ -136,7 +136,7 @@ jobs:
             echo "RELEASE_NOTES_EOF"
           } >> "$GITHUB_OUTPUT"
 
-      - uses: actions/download-artifact@v4
+      - uses: actions/download-artifact@v7
         with:
           pattern: binary-*
           merge-multiple: true
@@ -176,7 +176,7 @@ jobs:
       - name: Install jq
         run: apt-get update && apt-get install -y jq
 
-      - uses: actions/download-artifact@v4
+      - uses: actions/download-artifact@v7
         with:
           pattern: binary-*
           merge-multiple: true
--- a/action.yml
+++ b/action.yml
@@ -372,7 +372,7 @@ runs:
 
     - name: Upload review artifacts
       if: ${{ always() && inputs.upload_artifacts == 'true' }}
-      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
       with:
         name: ocr-review-result-${{ github.run_id }}-${{ github.run_attempt }}
         path: |


wu21-web:
I think there is no need for change really.

lwiles692:
> I think there is no need for change really.

When used in GitHub Actions, a warning is triggered that may confuse users. In fact, this dependency is brought in by the OCR package itself and is not part of the user’s repository. Avoid using outdated versions and upgrade to the officially recommended version.

wu21-web:
> > I think there is no need for change really.
> 
> 
> 
> When used in GitHub Actions, a warning is triggered that may confuse users. In fact, this dependency is brought in by the OCR package itself and is not part of the user’s repository. Avoid using outdated versions and upgrade to the officially recommended version.

@lwiles692 This reply looks like it is generated by AI.

wu21-web:
And also @lwiles692 , please check @Fanzzzd  #856 and it's related itemd. I remember that  it was kept as v4 intentionally.

lwiles692:
> > > I think there is no need for change really.
> > 
> > 
> > When used in GitHub Actions, a warning is triggered that may confuse users. In fact, this dependency is brought in by the OCR package itself and is not part of the user’s repository. Avoid using outdated versions and upgrade to the officially recommended version.
> 
> @lwiles692 This reply looks like it is generated by AI.

This is a real issue I encountered while using OCR in my own repository. At the time, I had some doubts, but after troubleshooting, the conclusion matched my description. Functionally, there is indeed no impact at the moment. If you don’t consider it a problem, feel free to close it.

image

wu21-web:
You can safely ignore this warning, the issue it created is a minor.

Qiyuanqiii (commented):
I took another look at this after a related Node-version discussion in #1182.

I think the motivation for this PR is valid. "actions/upload-artifact@v4.6.2" and "actions/download-artifact@v4.3.0" both declare "runs.using: node20", while the proposed v7 versions run on Node 24.

I also checked #856 because it had been mentioned as a reason that v4 may have been retained intentionally. My reading is that #856 intentionally avoided major-version changes because it was scoped to SHA pinning — it pinned the versions already present to keep that PR behavior-neutral. In particular, the explicit note about deliberately staying on v4 referred to "actions/checkout" in "translation-sync.yml"; I couldn't find a project decision in #856 or the related pinning work requiring the artifact actions to remain on v4.

GitHub's current Node 20 deprecation guidance also recommends users move workflows to action versions that run on Node 24. So while the warning may not have caused an immediate functional failure when this PR was opened, upgrading away from the Node 20 artifact actions looks appropriate.

One small nuance: "upload-artifact" switched its default runtime to Node 24 in v6 already, so v7 is not strictly required just to remove the Node 20 warning. That said, the v7 action still provides the inputs used here, so I don't see that as a problem with the proposed change.

Hope this helps clarify the earlier discussion.

lizhengfeng101 (commented):
Thanks for the bump! A few things to address before merge:

1. **Rebase needed** — the branch conflicts with main (these lines were SHA-pinned after your branch point).
2. **SHA-pin the refs in `release.yml`** to match the project convention:
   - `actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1`
   - `actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7.0.0`
3. **Missed occurrence** — `action.yml:1012` has a second `upload-artifact` ref still on v4.

Runner compatibility is fine (all self-hosted runners are on 2.337.0, well above the 2.327.1 minimum).

wu21-web:
https://github.com/alibaba/open-code-review/pull/1561 has been opened for 6hs if you need an urgent fix. I credited this PR author as co-author.

PR #1075, opened 2026-08-26
fix(tool): support subpath and cross-platform queries in file_find

### LINKED ISSUE #1073: bug(tool): file_find fails for directory-scoped and cross-platform subpath queries
### OpenCodeReview Version

v1.9.8 / main (built from source)

### Operating System

Windows / macOS / Linux

### Installation Method

Built from source

### LLM Provider

### Bug Description

When LLM agents invoke the `file_find` tool to locate files across the project during review or planning (e.g. `query_name: "pkg/util"` or `query_name: "internal/diff"` or `query_name: "src/utils"`), `file_find` fails to match any files and returns `// The file was not found`.

In `internal/tool/file_find.go:47-59`, `Execute` strips all directory components from candidate file paths (`base = f[idx+1:]`) and tests substring inclusion exclusively against the base filename (`base`). If `query_name` contains a path separator or subdirectory prefix, `strings.Contains(base, queryName)` evaluates `strings.Contains("util.go", "pkg/util")`, which is always false.

Furthermore, Windows path separators in `query_name` (e.g. `pkg\util.go`) are not normalized, causing queries to fail on Windows environments.

### Steps to Reproduce

1. In a repository containing nested files (e.g. `pkg/util.go`), execute `file_find` tool with `query_name: "pkg/util"`.
2. Observe that `file_find` returns `// The file was not found`.
3. Execute `file_find` with Windows backslash `query_name: "pkg\\util.go"`.
4. Observe that `file_find` returns `// The file was not found`.

### Expected Behavior

`file_find` should match against both the base filename and the relative path (with normalized forward slashes), allowing both base filename queries (`util.go`) and directory-scoped queries (`pkg/util`, `internal/diff`) to resolve matching files.

### Logs / Error Output

```text
// The file was not found
```

### Additional Context

I have investigated this issue and have a reproducible unit test case and surgical fix ready.
Please assign this issue to me, I will submit a PR shortly.

MeiSiristhebest:
## Description

Fixes #1073

When LLM agents invoke the `file_find` tool to locate files across the project during review or planning (e.g. `query_name: "pkg/util"` or `query_name: "internal/diff"` or `query_name: "src/utils"`), `file_find` previously failed to match any files. `Execute` stripped all directory components from candidate file paths (`base = f[idx+1:]`) and tested substring inclusion exclusively against the base filename (`base`), so `strings.Contains("util.go", "pkg/util")` was always false. Windows backslash separators in `query_name` (e.g. `pkg\util.go`) were also not normalized, causing queries to fail on Windows.

### Root Cause
`internal/tool/file_find.go` matched only against the base filename, ignoring the relative path that directory-scoped queries target.

### Key Changes
1. **Match against the full relative path** (`f`, already forward-slash normalized by `git ls-files`/the walk fallback) instead of the base filename only. Since the base filename is always a suffix of the relative path, full-path matching **subsumes** base-filename matching while also enabling directory-scoped queries such as `pkg/util`, `internal/diff`, and `pkg/`.
2. **Normalize `query_name` path separators** with `strings.ReplaceAll(queryName, "\\", "/")` so Windows-style queries resolve identically cross-platform.
3. **Hoist loop-invariant work out of the loop**: precompute `queryCmp` (the case-folded query) once before the loop, and compute `pathCmp` (the case-folded candidate path) once per file. This directly addresses the OpenCodeReview performance finding on the previous attempt (PR #1074), where `strings.ToLower(queryName)`/`strings.ToLower(normalizedQuery)` were recomputed on every iteration.
4. Added regression test `TestFileFind_PathWithSubdirectory` covering forward-slash subpath queries, Windows backslash queries, directory-prefix queries, and case-insensitive subpath queries.

This supersedes the closed PR #1074, incorporating the OpenCodeReview bot's performance feedback (eliminating per-iteration loop-invariant recomputation) and simplifying the matching logic.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] `make test` / targeted package tests pass locally:
  - `go test -v ./internal/tool` — all unit tests pass cleanly (including the new regression test).
  - `go test ./internal/diff/... ./internal/tool/... ./internal/scan/...` — all pass.
- [x] Fail-first baseline verified: on unmodified `main`, the new `TestFileFind_PathWithSubdirectory` fails with `expected to find pkg/util.go when querying 'pkg/util', got: // The file was not found`; after the fix it passes.
- [x] `go fmt` and `go vet` are clean.

## Checklist

- [x] I have signed the CLA

## Related Issues

Fixes #1073


Qiyuanqiii (commented):
Could we update the model-facing tool schema and the localized tool documentation along with this semantic change? `internal/config/toolsconfig/tools.json` still describes `query_name` as a filename keyword, while the tool docs explicitly say it is matched against the basename rather than the full path. Since this schema is exposed to the model at runtime, the new subpath capability is still contradicted by its caller-visible contract. Please describe it as a substring match against the repository-relative path, mention that both `/` and `\` separators are accepted, and sync the localized docs.

MeiSiristhebest:
Thanks for the review! I've updated the model-facing 	ools.json schema and synchronized all localized documentation (EN, ZH, JA, RU) to describe query_name as matching against repository-relative paths with / and \ separator support in commit 3dffd93.

Qiyuanqiii (approved):
good job

lizhengfeng101:
Nice fix for the subpath query issue — the root cause analysis is spot on.

One concern with the current approach: matching against the full relative path unconditionally changes behavior for simple filename queries. For example, `query_name: "tool"` now returns every file under `internal/tool/` (50+ files), when the agent likely just wanted `tool.go` or `tool_test.go`.

I'd suggest a fallback strategy instead:

1. First, match against the basename (current/old behavior)
2. If no results, fall back to matching against the full relative path

This way:
- Pure filename queries stay precise (no result set bloat)
- Subpath queries like `pkg/util` or `internal/diff` still work (basename match returns nothing → falls back to full path match)
- Directory-scoped queries like `internal/tool/` also work via the fallback
- Fully backward compatible — existing queries return identical results

The backslash normalization and loop-invariant hoisting are both good as-is, just the matching strategy could be more conservative.

MeiSiristhebest:
Thanks @lizhengfeng101! That's a great point about keeping pure filename queries precise. I've updated the implementation in commit b42c409 to adopt the fallback strategy:

1. **First pass**: Match against `basename` (keeps filename queries like `query_name: "tool"` strictly scoped to matching filenames).
2. **Fallback pass**: If the basename match yields no results, fall back to matching against the full relative path (enables subpath queries like `pkg/util`, `internal/diff`, and directory-scoped queries like `internal/tool/`).

Added unit test `TestFileFind_BasenamePrecisionAndFallback` to ensure regression coverage.

internal/config/toolsconfig/tools.json
```diff
@@ -192,13 +192,13 @@
...
+      "description": "Search for matching files in the current project by repository-relative path substring or filename keyword. Use this tool when you cannot find the files you need to view in the current change file list.\n\nThis tool searches for repository-relative paths containing specified keywords or subpaths (accepting both '/' and '\\' separators) in the project directory and returns a list of matching file paths. Search is case-insensitive by default, adjustable via case_sensitive parameter.\n\nNote: This tool only supports returning the first 100 matching file paths; excess will be truncated.\n\nExample:\nInput:\nquery_name: UserService\nOutput:\nsrc/main/java/UserService.java\nsrc/test/java/UserServiceTest.java",
```
lizhengfeng101:
```suggestion
      "description": "Search for matching files in the current project based on filename keywords. When the query contains a path separator ('/' or '\\'), also matches against the full repository-relative path, enabling directory-scoped queries. Use this tool when you cannot find the files you need to view in the current change file list.\n\nThis tool searches for filenames containing specified keywords in the project directory and returns a list of matching file paths. Search is case-insensitive by default, adjustable via case_sensitive parameter.\n\nNote: This tool only supports returning the first 100 matching file paths; excess will be truncated.\n\nExample:\nInput:\nquery_name: UserService\nOutput:\nsrc/main/java/UserService.java\nsrc/test/java/UserServiceTest.java\n\nInput:\nquery_name: pkg/util\nOutput:\npkg/util.go\npkg/util_test.go",
```

internal/config/toolsconfig/tools.json
```diff
@@ -192,13 +192,13 @@
...
       "parameters": {
         "type": "object",
         "properties": {
           "query_name": {
             "type": "string",
-            "description": "Filename keyword to search for, supports partial matching."
+            "description": "Substring or subpath to search for in repository-relative paths, accepting both '/' and '\\' separators. Supports partial matching."
```
lizhengfeng101:
```suggestion
            "description": "Filename keyword to search for, supports partial matching. If the query contains '/' or '\\', falls back to matching against the full repository-relative path (e.g. 'pkg/util' or 'internal/diff/')."
```

lizhengfeng101 (approved):
LGTM

PR #1086, opened 2026-08-27
feat(allowlist): add Handlebars and Mustache support

### LINKED ISSUE #1081: feat(allowlist): add Handlebars and Mustache support
### Problem Statement

OpenCodeReview does not currently include `.hbs` or `.mustache` in its supported file types, so Handlebars and Mustache templates are excluded from code review.

This is a focused subtask of #470.

### Proposed Solution

Add focused support for Handlebars and Mustache templates:

- Add `.hbs` and `.mustache` to `internal/config/allowlist/supported_file_types.json`.
- Add positive, case-insensitive, and negative allowlist coverage, including verification that unrelated extensions remain unsupported.
- Do not add a default test-file exclusion pattern because Handlebars and Mustache projects do not share a broadly accepted template-test path convention.
- Add a focused review rule for Handlebars/Mustache templates.
- Register rule mappings for both extensions in `internal/config/rules/system_rules.json`.
- Add rule-resolution tests.
- Keep the contribution limited to these two extensions.

The review rule will focus on escaped versus unescaped output, HTML/template injection, triple-stash usage, missing variables and helpers, partial/context mistakes, and other common template correctness and security issues.

I will follow the contribution guide and run `make check`, `make test`, and `make build` before submitting the pull request.

### Alternatives Considered

Adding only the file extensions and letting these templates fall back to the generic default review rule was considered. This would make the files reviewable, but it would miss Handlebars/Mustache-specific correctness and security risks such as unescaped output, triple-stash usage, helper resolution, partial context, and template injection.

Renaming template files to `.html` is also not appropriate for existing Handlebars and Mustache projects.

### Affected Area

Review Rules, Configuration

### Additional Context

Related to #470.

I claimed this focused subset in:
https://github.com/alibaba/open-code-review/issues/470#issuecomment-5435190602

Scope is limited to `.hbs` and `.mustache`. Other template engines remain outside this contribution.

I did not find an existing Handlebars/Mustache issue or open pull request when checking the repository.

yepYoung:
## Description

Add first-class review support for Handlebars and Mustache templates as the focused subtask tracked in #1081:

- recognize `.hbs` and `.mustache` files case-insensitively;
- keep template files under `test/` and `tests/` reviewable because these ecosystems do not share a reliable default template-test convention;
- route both extensions to a focused review rule covering escaping boundaries, template and partial selection, missing values and helpers, context mistakes, structured output, and render-time performance;
- add positive, uppercase, exclusion, and rule-resolution coverage;
- document the built-in mapping in all five review-rules locale pages.

The contribution stays limited to these two extensions. Other template engines remain outside this PR.

## Type of Change

- [ ] Breaking change (fix or feature that causes existing functionality to change)
- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Validation performed on macOS/arm64 with Go 1.26.5:

- `make check`
- `make test` (race detector and `-count=1`)
- `make build`
- `make coverage` (91.1%, above the 90% threshold)
- Pages lint, 32 tests, typecheck, production build, bundle-size check, and HTTP smoke checks
- Translation-sync tests and documentation locale check
- `git diff --check upstream/main...HEAD`

The repository-required `ocr review --audience agent --background <context>` command was attempted, but this local environment has no configured LLM endpoint. `ocr review --preview` completed successfully, and I performed a separate line-by-line self-review.

## Checklist

- [x] I have added tests that prove the feature works
- [x] I have signed the CLA

## Related Issues

Closes #1081

Part of #470


internal/config/allowlist/allowed_ext_test.go
```diff
@@ -161,6 +167,10 @@ func TestIsExcludedPath(t *testing.T) {
 		// Prisma schemas have no conventional default test-file exclusion.
 		{"prisma schema", "prisma/schema.prisma", false},
 
+		// Handlebars/Mustache templates have no conventional default test-file exclusion.
+		{"handlebars template in tests directory", "tests/templates/card.hbs", false},
+		{"mustache template in test directory", "test/templates/email.mustache", false},
```
wu21-web: Why doesn't this block have a item where `want` is `true`? It is optional but all other blocks have at least one item that has `want` set to `true`.

wu21-web: A example suggestion might be:

```suggestion
        // Handlebars/Mustache templates have no conventional default test-file exclusion.
		{"handlebars template in tests directory", "tests/templates/card.hbs", false},
		{"mustache template in test directory", "test/templates/email.mustache", false},
		{"mustache files", "src/source.mustache", true},
		{"handlebars source files", "src/code.hbs", true},
```

internal/config/allowlist/supported_file_types.json
```diff
@@ -47,6 +47,8 @@
   ".ftl",
   ".ftlh",
   ".ftlx",
+  ".hbs",
+  ".mustache",
```
wu21-web, line 51: Is a default exclude pattern needed for those files?

For example:
```json
"**/test/**/*.{hbs,mustache}",
"**/tests/**/*.{hbs,mustache}",
"**/spec/**/*.{hbs,mustache}"
```

internal/config/allowlist/allowed_ext_test.go
```diff
@@ -37,6 +37,12 @@ func TestIsAllowedExt(t *testing.T) {
 		{".FTLH", true},
 		{".ftlx", true},
 		{".FTLX", true},
+		{".hbs", true},
+		{".HBS", true},
+		{".mustache", true},
+		{".MUSTACHE", true},
+		{".hbsx", false},
+		{".mustachex", false},
```
wu21-web: Why are there unwanted extensions here?
I mean if those unwanted extensions are not in the `allowed_extensions.json` in the first place. There is no need to be over explicit.
This should be handled by the `{"", false},` currently on R103 in your branch.
Deleting these 2 items also aligns the format.

```suggestion

```

internal/config/allowlist/allowed_ext_test.go
yepYoung: Thanks for pointing this out. I added `want: true` cases for `.hbs` and `.mustache` under the existing generic `**/fixtures/**` exclusion, while keeping ordinary `test/templates` paths at `false`.

In `TestIsExcludedPath`, `want: true` means the path is excluded from review, so using ordinary `src/*.hbs` or `src/*.mustache` source files as the true cases would skip the templates this PR is adding support for. The fixture cases now cover the excluded side without changing the intended behavior.

internal/config/allowlist/supported_file_types.json
yepYoung: Good question. I did not add extension-specific exclusions for all `test`, `tests`, or `spec` directories. #470 lists no required default exclusion for Template Engines, and this repository already excludes generic `fixtures`, `testdata`, and `__snapshots__` paths.

A broad directory rule could exclude real Handlebars/Mustache templates based only on their location. The updated tests therefore keep ordinary `test/templates` paths reviewable while proving that templates under existing fixture paths are still excluded. If the maintainers prefer the broader exclusion as a project-level policy, I’m happy to adjust it.

internal/config/allowlist/allowed_ext_test.go
yepYoung: Agreed. I removed the `.hbsx` and `.mustachex` cases. The shared unsupported-extension cases already cover the false path, and the updated block now matches the surrounding test style.

wu21-web (approved):

lizhengfeng101 (approved):
LGTM

PR #1089, opened 2026-08-27
fix(tool): prevent path traversal bypass and normalize paths in code_search and code_comment

### LINKED ISSUE #1088: bug(tool): prevent path traversal bypass and normalize paths in code_search and code_comment
### OpenCodeReview Version

`main` at `eda2548` (built from source)

### Operating System

Windows (also impacts cross-platform path handling across Windows/Linux/macOS)

### Installation Method

Built from source

### LLM Provider

Other OpenAI-compatible endpoint / Anthropic

### Bug Description

Two cross-platform path handling issues exist in `internal/tool`:

1. **Path traversal validation bypass in `code_search`**:
   In `internal/tool/code_search.go`, `hasTraversalPathComponent` only checks for `/` path separators:
   ```go
       for _, part := range strings.Split(pathspec, "/") {
           if part == ".." {
               return true
           }
       }
       return false
   }
   ```
   When `file_patterns` contain Windows-style backslashes (such as `..\pkg` or `pkg\..\internal`), `strings.Split` fails to separate path segments by `\`, allowing path traversal validation to be bypassed. Furthermore, `file_patterns` are forwarded directly to `git grep` without normalizing `\` to `/`, which causes `git grep` pathspec matching to fail.

2. **Silent comment dropping due to unnormalized path in `code_comment`**:
   In `internal/tool/code_comment.go`, `parseCommentsInner` directly sets `cm.Path` from the tool arguments without normalizing path separators or redundant prefixes (e.g., `pkg\util.go`, `./pkg/util.go`, or `pkg//util.go`).
   Downstream in `internal/agent/agent.go`, comments are looked up via `CommentCollector.CommentsForPath(d.NewPath)` where `d.NewPath` is the canonical forward-slash path (e.g. `pkg/util.go`). Because `CommentsForPath` performs strict equality (`cm.Path == path`), any comment generated with backslashes or `./` is silently dropped and never included in the review results.

### Steps to Reproduce

1. Call `CodeSearchProvider.Execute` with `file_patterns: ["..\\pkg"]`.
   - Observe that `hasTraversalPathComponent` returns `false` and does not reject the pattern with `"Error: file_patterns must not contain .."`.
2. Call `CodeCommentProvider.Execute` (or `ParseComments`) with `path: "pkg\\util.go"` or `path: "./pkg/util.go"`.
   - Store comments in `CommentCollector`.
   - Call `CommentCollector.CommentsForPath("pkg/util.go")`.
   - Observe that 0 comments are returned because the path was not normalized.

### Expected Behavior

1. `hasTraversalPathComponent` should normalize `\` to `/` before checking for `..`, and `file_patterns` should be normalized to forward slashes before being passed to `git grep`.
2. `parseCommentsInner` should normalize comment paths (`\` to `/`, stripping leading `./`, cleaning redundant slashes) so comments reliably match `CommentsForPath(d.NewPath)`.

### Additional Context

This aligns `code_search` and `code_comment` with the cross-platform path normalization patterns recently established in `file_find.go` (#1075).

MeiSiristhebest:
## Description

This PR fixes two cross-platform path handling issues in `internal/tool`:

1. **Fix path traversal bypass & normalize pathspecs in `code_search`**:
   - `hasTraversalPathComponent` now normalizes backslashes `\` to `/` before splitting on `/`, preventing traversal sequences like `..\pkg` or `pkg\..\internal` from bypassing the `..` safety check.
   - Normalizes `file_patterns` items from `\` to `/` before passing them to `git grep` so pathspec matching works consistently on all platforms.

2. **Normalize comment paths in `code_comment`**:
   - `parseCommentsInner` now canonicalizes comment paths (`\` to `/`, stripping leading `./`, resolving `.`, and trimming redundant slashes via `normalizeCommentPath`).
   - Ensures that review comments produced with Windows backslashes or leading `./` correctly match `CommentCollector.CommentsForPath(d.NewPath)` during review compilation instead of being silently dropped.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- Added `TestParseComments_NormalizePath` in `internal/tool/code_comment_test.go` to verify normalization of `\`, `./`, and redundant slashes.
- Added backslash test cases to `TestCodeSearchProvider_Execute_RejectsTraversalPattern` in `internal/tool/code_search_test.go`.
- Verified targeted unit tests and full package tests pass cleanly (`go test -v ./internal/tool/...`).

- [x] Unit tests pass locally

## Checklist

- [x] I have signed the CLA

## Related Issues

Closes #1088

### DIFF
--- a/internal/tool/code_search.go
+++ b/internal/tool/code_search.go
@@ -40,7 +40,7 @@ func (p *CodeSearchProvider) Execute(ctx context.Context, args map[string]any) (
 			if hasTraversalPathComponent(s) {
 				return "Error: file_patterns must not contain ..", nil
 			}
-			patterns = append(patterns, s)
+			patterns = append(patterns, strings.ReplaceAll(s, "\\", "/"))
 		}
 	}
 
@@ -99,7 +99,8 @@ func (p *CodeSearchProvider) buildGrepArgs(searchText string, caseSensitive bool
 }
 
 func hasTraversalPathComponent(pathspec string) bool {
-	for _, part := range strings.Split(pathspec, "/") {
+	norm := strings.ReplaceAll(pathspec, "\\", "/")
+	for _, part := range strings.Split(norm, "/") {
 		if part == ".." {
 			return true
 		}
--- a/internal/tool/code_search_test.go
+++ b/internal/tool/code_search_test.go
@@ -437,15 +437,27 @@ func TestCodeSearchProvider_Execute_WithFilePatterns(t *testing.T) {
 	dir := setupTestRepo(t)
 	p := NewCodeSearch(&FileReader{RepoDir: dir, Mode: ModeWorkspace})
 
-	got, err := p.Execute(context.Background(), map[string]any{
-		"search_text":   "Util",
-		"file_patterns": []any{"pkg/"},
-	})
-	if err != nil {
-		t.Fatal(err)
+	tests := []struct {
+		name    string
+		pattern string
+	}{
+		{name: "forward slash", pattern: "pkg/"},
+		{name: "backslash", pattern: "pkg\\"},
 	}
-	if !strings.Contains(got, "util.go") {
-		t.Errorf("expected util.go in result, got: %s", got)
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			got, err := p.Execute(context.Background(), map[string]any{
+				"search_text":   "Util",
+				"file_patterns": []any{test.pattern},
+			})
+			if err != nil {
+				t.Fatal(err)
+			}
+			if !strings.Contains(got, "util.go") {
+				t.Errorf("expected util.go for pattern %q, got: %s", test.pattern, got)
+			}
+		})
 	}
 }
 
--- a/internal/tool/code_search_test.go
+++ b/internal/tool/code_search_test.go
@@ -460,6 +460,9 @@ func TestCodeSearchProvider_Execute_RejectsTraversalPattern(t *testing.T) {
 		{name: "leading parent", pattern: "../pkg", want: "Error: file_patterns must not contain .."},
 		{name: "middle parent", pattern: "pkg/../internal", want: "Error: file_patterns must not contain .."},
 		{name: "trailing parent", pattern: "pkg/..", want: "Error: file_patterns must not contain .."},
+		{name: "leading parent backslash", pattern: `..\pkg`, want: "Error: file_patterns must not contain .."},
+		{name: "middle parent backslash", pattern: `pkg\..\internal`, want: "Error: file_patterns must not contain .."},
+		{name: "trailing parent backslash", pattern: `pkg\..`, want: "Error: file_patterns must not contain .."},
 	}
 
 	for _, test := range tests {
--- a/internal/tool/code_comment.go
+++ b/internal/tool/code_comment.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"path"
 	"strings"
 
 	"github.com/alibaba/open-code-review/internal/model"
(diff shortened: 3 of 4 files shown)


Qiyuanqiii (approved):
good

lizhengfeng101:
Thanks for the PR. The hardening in `hasTraversalPathComponent` makes sense as defense-in-depth, and the `normalizeCommentPath` logic is clean.

That said — I ran a quick scan across 174 historical review sessions on this repo (1580 `code_search` calls, 174 `code_comment` calls). Zero instances of backslash paths, `./` prefixes, redundant slashes, or `..` in either tool. The LLM has never produced a malformed path in practice here.

A few questions:

1. Did you actually hit this in the wild (dropped comments, failed pathspec match), or is this purely from reading the code?
2. For `code_search`: git on Linux/macOS doesn't treat `\` as a separator, so even if the check is bypassed, `git grep -- '..\pkg'` matches nothing real. The actual security boundary is git itself, not our string check. Do you have a scenario where this matters beyond the theoretical?
3. Minor: `path.Clean` resolves `..` silently — an LLM hallucinating `pkg/../secret.go` gets normalized to `secret.go` rather than rejected. Intentional?

Not blocking — the code is correct and the tests are good. Just want to calibrate whether this addresses an observed issue or a hypothetical one.

internal/tool/code_search_test.go
wu21-web, line 462: You are missing coverage.

```suggestion
func TestCodeSearchProvider_Execute_WithFilePatterns(t *testing.T) {
	dir := setupTestRepo(t)
	p := NewCodeSearch(&FileReader{RepoDir: dir, Mode: ModeWorkspace})

	tests := []struct {
		name    string
		pattern string
	}{
		{name: "forward slash", pattern: "pkg/"},
		{name: "backslash", pattern: "pkg\\"},
	}

	for _, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			got, err := p.Execute(context.Background(), map[string]any{
				"search_text":   "Util",
				"file_patterns": []any{test.pattern},
			})
			if err != nil {
				t.Fatal(err)
			}
			if !strings.Contains(got, "util.go") {
				t.Errorf("expected util.go for pattern %q, got: %s", test.pattern, got)
			}
		})
	}
}
```

MeiSiristhebest:
Thanks for the feedback and discussion @lizhengfeng101!

To share what we actually observed testing locally in a native Windows environment:

While the CLI doesn't crash thanks to the existing fallbacks, path discrepancies cause subtle runtime friction on Windows:
1. **Rule matching fallback**: Passing Windows paths (e.g. via Tab-completion) causes `doublestar.Match` to miss language globs (`**/*.ts`), quietly falling back to `system/default` instead of language-specific rules.
2. **Empty retries in `llmloop`**: When backslash patterns in `code_search` hit Git pathspecs, searches return empty, causing `llmloop` to trigger `consecutiveEmptyRounds` retry turns. That ends up burning redundant LLM tokens and adding latency.
3. **`path.Clean`**: This was just meant as a forgiving reconciler for harmless relative paths so valid findings don't get dropped by `CommentsForPath`, while hard security and boundary checks remain at the tool perimeter.

Also, I have incorporated @wu21-web's test coverage suggestion in the latest commit (`1c80093`). Appreciate the great review and collaboration!

wu21-web:
@MeiSiristhebest Also based on your branch name, I think this more of a bug than a 0day security pothole.

MeiSiristhebest:
> @MeiSiristhebest Also based on your branch name, I think this more of a bug than a 0day security pothole.

Haha fair point @wu21-web! I started with a 0day exploration branch and just kept working on it. Totally agree it's a cross-platform bug fix. Thanks!

internal/tool/code_search.go
```diff
@@ -40,7 +40,7 @@ func (p *CodeSearchProvider) Execute(ctx context.Context, args map[string]any) (
 			if hasTraversalPathComponent(s) {
 				return "Error: file_patterns must not contain ..", nil
 			}
-			patterns = append(patterns, s)
+			patterns = append(patterns, strings.ReplaceAll(s, "\\", "/"))
```
lizhengfeng101, line 44:
```suggestion
			// Treat backslashes as separators so Windows pathspecs match.
			patterns = append(patterns, strings.ReplaceAll(s, "\\", "/"))
```

lizhengfeng101 (approved):
LGTM

PR #1097, opened 2026-08-28
docs(skill): prefer --output flag over shell redirection

yingjiexu2002:
## Description

The agent skill taught shell redirection (`> /tmp/ocr_out.txt 2>&1`) as the way to avoid losing review comments to tool-output truncation. Since #852 (v1.10.0), `ocr review` natively supports `--output <path>` / `-o`, which writes results directly to a UTF-8 file and fits agents better than shell redirection.

This PR updates the skill in four places, in both the canonical and the plugin mirror copy.

- Register `--output` / `-o` in the Argument handling catalog, with an upgrade gate for older CLIs (on `unknown flag: --output`, stop the review and ask the user before upgrading with `@latest`; never fall back to redirection or plain stdout silently)
- Switch the Output mode guidance to `--output`, and align the Gotchas reminder with it
- Add an `unknown flag: --output` Troubleshooting entry that mirrors the gate

Documentation only, no code changes.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Verified both SKILL.md copies stay identical except the intentional mirror note, the flag descriptions match the CLI help and the cli-reference docs, and line endings are LF. Also tested end-to-end with Claude Opus 4.6, which followed the updated skill and used `--output` for the review run. On an older CLI (pre-v1.10.0), the agent stopped the review and asked whether to upgrade instead of silently falling back to stdout.

## Checklist

- [x] I have signed the CLA

## Related Issues

Follow-up to #852 (the `--output` flag).

### DIFF
--- a/plugins/open-code-review/skills/open-code-review/SKILL.md
+++ b/plugins/open-code-review/skills/open-code-review/SKILL.md
@@ -51,7 +51,7 @@ ocr review --audience agent --background "business context here" [user-args]
 - **Timeout**: default timeout is 15 minutes per file; adjust with `--timeout <minutes>`
 - **Concurrency**: default concurrency is 8 file workers; reduce with `--concurrency <n>` if rate limits are hit
 - **Preview mode**: use `--preview` or `-p` to preview which files will be reviewed without running the LLM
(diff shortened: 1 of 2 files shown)


lizhengfeng101:
Since the gate keys on the literal `unknown flag: --output`, I'd drop the `-o` shorthand from the agent-facing guidance — pflag reports `unknown shorthand flag: 'o' in -o` for the short form ([flag.go#L1035](https://github.com/spf13/pflag/blob/v1.0.5/flag.go#L1035)), so an agent that happens to pick `-o` on a pre-v1.10.0 CLI would slip straight past both the bullet and the Troubleshooting entry.

yingjiexu2002:
Done

lizhengfeng101 (approved):
LGTM

PR #1101, opened 2026-08-28
feat(viewer): mark comments as fixed/ignored and hide them

### LINKED ISSUE #825: Feature request: viewer: mark comment as Fixed/Solved/Ignore and hide
### Problem Statement

After fixed the issue mentioned in review comment, i don’t want to see the comment again till I fix the rest comments, it should be helpful if we can mark the comment as fixed/solved/ignore/… to hide this comment. And it should support toggle on/off a checkbox to hide/display marked comments too. 

### Proposed Solution

- Mark and hide.
- toggle on / off to hide / display

### Alternatives Considered

_No response_

### Affected Area

WebUI Viewer

### Additional Context

_No response_

nitishagar:
## What

Implements #825 in the reduced, read-only form agreed in review: in the `ocr viewer` session page, each review comment card gains **Fixed / Ignored** buttons (mutually exclusive; setting one replaces the other, **Clear** undoes) and the toolbar gains a **Hide marked** checkbox (default on, persisted per browser) plus **Clear all marks**. Walk the findings, mark as you fix, hide what is done, toggle back anytime — exactly the "leave and come back tomorrow" workflow from the issue.

**The viewer stays read-only.** Marks are browser-side presentation state in `localStorage`, keyed per session page. The server renders cards unmarked and accepts no writes: no write API, no sidecar file, no new request surface. `fixed` and `solved` from the issue were synonyms, so they collapse to one state while nothing is on disk.

## How

- **Identity (server-side)**: each comment gets a stable `MarkID` rendered into `data-mark-id`, derived from its owning session record's `uuid` + index within the record — reload-stable and collision-free; legacy uuid-less records fall back to a full-field sha256 with an occurrence counter so exact duplicates stay distinct. Identity stays on the server (session files are immutable); only the mark *state* is browser state.
- **State (client-side)**: `session.js` keeps a `{markID → state}` map in `localStorage` under a per-session-path key. Stored marks replay on load (before the first filter pass, so counts are right), are validated against the known state set — corrupt or tampered storage degrades to the unmarked view — and compose with the existing severity/category filters through the same group-visibility path.
- **UI**: chip per card follows the filled-badge pattern used by the category/severity badges (background + color, with a dark-mode override); Fixed / Ignored / Clear buttons sit at the bottom of each card; the toggle and Clear-all sit on the existing filter bar with live `marked/hidden` counts.

## Semantics

- One state per comment, mutually exclusive, replace-on-set.
- Marks are **session-scoped and browser-scoped**: another browser or machine sees the session unmarked, and re-running a review produces a new session that starts unmarked.
- Hide-marked defaults **on** (hide-as-you-fix); also presentation state in localStorage.

## Tests

Identity (stability, duplicates, uuid-less fallback, unicode fields) via `LoadSession`; rendering (cards carry `data-mark-id` + buttons, no server-rendered state, collapsed `solved` absent); a mux-level guard asserting `POST`/`DELETE`/`PUT`/`PATCH` against `/marks` all 404, so a write API cannot quietly return. Full suite green under `-race`; `make check` and the docs translation-sync check pass. Docs updated in all five languages (en/zh/ja/ko/ru) plus the README bullets.

Closes #825


internal/viewer/store.go
```diff
@@ -238,6 +240,11 @@ type ReviewComment struct {
 	EndLine        int
 	Category       string // bug, security, performance, maintainability, test, style, documentation, other
 	Severity       string // critical, high, medium, low
+	// MarkID is the stable identity used to bind viewer marks (fixed /
+	// solved / ignored) to this comment. Session files are immutable, so the
+	// identity is reload-stable: record uuid + comment index for records that
+	// carry a uuid, a full-field hash plus an occurrence counter otherwise.
+	MarkID string `json:"-"`
```
wu21-web: Can you please clean up those code comments?

```suggestion
	MarkID string `json:"-"`
```

README.md
wu21-web, line 1: The documentation changes are not complete. At the very least, add the details about this feature to the corresponding site documentation section. Currently deployed at `https://open-codereview.ai/docs/viewer`

internal/viewer/handler.go
```diff
@@ -78,5 +85,97 @@ func handleSession(w http.ResponseWriter, r *http.Request, root, repo, sessionID
...
+		return
+	}
+
+	var req marksRequest
+	body := http.MaxBytesReader(w, r.Body, marksMaxBodyBytes)
+	dec := json.NewDecoder(body)
+	dec.DisallowUnknownFields()
+	if err := dec.Decode(&req); err != nil {
+		http.Error(w, fmt.Sprintf("invalid request body: %v", err), http.StatusBadRequest)
+		return
```
wu21-web: EOF after the first object
```suggestion
	if err := dec.Decode(&req); err != nil {
		http.Error(w, fmt.Sprintf("invalid request body: %v", err), http.StatusBadRequest)
		return
	}
	if err := dec.Decode(&struct{}{}); err != io.EOF {
		http.Error(w, "request body must contain exactly one JSON object", http.StatusBadRequest)
		return
	}
```

internal/viewer/static/session.js
```diff
@@ -92,5 +128,110 @@ document.querySelectorAll('.response-text').forEach(function(el) {
...
+    // Marks: state lives server-side in a sidecar store; the response echoes
+    // the authoritative state, which is what gets applied to the card.
+    function marksURL() {
+        return window.location.pathname.replace(/\/$/, '') + '/marks';
+    }
+
+    // The server whitelists states; this mirrors the same set so a tampered
+    // or buggy response can never grow an arbitrary class name on the chip.
+    const knownMarkStates = { fixed: true, solved: true, ignored: true };
+
```
wu21-web:
```suggestion
    // The server whitelists states; this mirrors the same set so a tampered
    // or buggy response can never grow an arbitrary class name on the chip.
    const knownMarkStates = { fixed: true, solved: true, ignored: true };
    let marksMutationQueue = Promise.resolve();

    function enqueueMarksMutation(operation) {
        const result = marksMutationQueue.then(operation);
        marksMutationQueue = result.catch(function() {
        });
        return result;
    }
```

internal/viewer/handler.go
```diff
@@ -4,8 +4,11 @@
 package viewer
 
 import (
+	"encoding/json"
 	"fmt"
+	"mime"
 	"net/http"
+	"os"
 	"path/filepath"
```
wu21-web, line 9:
```suggestion
import (
	"encoding/json"
	"fmt"
	"io"
	"mime"
	"net/http"
	"os"
	"path/filepath"
```

wu21-web (changes requested):
1. Concurrent UI mutations can leave the page inconsistent with disk
2. Documentation changes incomplete
3. Too many code comments
4. Strict JSON parsing accepts trailing JSON values

lizhengfeng101:
Thanks @nitishagar — the `MarkID` design is genuinely good (record `uuid` + index is exactly the right identity, and the legacy sha256 fallback is a thoughtful touch), and the security work is above the usual bar.

But I have to hold the line on the architecture, and I owe you a better explanation than the issue thread gave. `ocr viewer` is deliberately read-only: it holds no state of its own and is a pure function of the JSONL on disk. That's what keeps it dependency-free, lets us promise session data is never mutated by viewing it, and keeps the browser-facing surface of a tool that exposes users' source code down to GETs. This PR crosses that line, and it doesn't uncross — after merge we own a write API, a CSRF surface, a versioned sidecar format, a global write lock, orphanable temp files, and `.marks.json` accumulating in `~/.opencodereview/sessions/` with no reclaim path (we have no session cleanup command today).

**Keep the feature, drop the server.** Re-reading @iredmail's scenario — *"i may leave and come back to continue"* — the persistence he needs is "close the tab, come back tomorrow." `localStorage` keyed per session gives that, plus exactly the session-scoped semantics you already argued for correctly. So: keep `MarkID` in `store.go` and `data-mark-id` in the template (server-side identity is the right call), keep the buttons, chip, toggle, and filter composition. Drop `marks.go`, `handleMarks`, the route, and the tests that only exist for them — roughly 150 lines instead of 1200, and "read-only" becomes true rather than argued. Your own split already points here: hide-marked is in `localStorage` as presentation state, and the marks are too.

Two smaller things for the reduced version:

- Collapse `fixed`/`solved` into one state — they're synonyms for a review comment, and the issue's `fixed/solved/ignore/…` is examples, not a spec. Free to change now, not after values are on disk.
- The chip sets `border-color` on `.comment-badge`, which has no `border` declaration, so no border renders; and there's no dark-mode override the way every `.cat-*` badge has. Match the filled-badge pattern at `style.css:1045`.

The thread stalled with us asking for a screenshot and going quiet — not a fair signal to build 1200 lines against, and that's on us. **This isn't a no.** I want the feature and I'd like you to land it; if you're up for the reduced form I'll review promptly. If you'd rather not, say so and I'll take the `MarkID` design from here with credit to you.

Also please update `pages/src/content/docs/{en,zh,ja,ko,ru}/viewer.md` — the READMEs are synced but the docs pages aren't.

iredmail:
Thanks for making it happen guys. :)

nitishagar:
Thanks a lot for taking the time to explain all this — the read-only argument finally clicked for me, and you're right that localStorage covers what the issue was actually asking for. The reduced version is pushed: server-side marks API and sidecar file gone (kept the MarkID design), fixed/solved collapsed into one state, chip now uses the filled-badge pattern with a dark-mode override, and the docs pages are updated in all five languages — happy to adjust anything else.

lizhengfeng101:
The reduced form is the right shape — server-side identity, browser-side state, and `MarkID` (record `uuid` + index) holds up: I checked `writeReviewItemRecord` and both `review_item_done` and `review_item_reused` get a fresh `generateUUID()`, so `uuid#index` can't collide within a session. Two things I'd want fixed before merge, plus three smaller ones.

**1. The mark-state allowlist doesn't actually reject anything (`session.js`)**

`knownMarkStates` is an object literal, so `knownMarkStates[state]` hits `Object.prototype`: `"toString"`, `"constructor"`, and `"valueOf"` all pass validation. Put `{"<id>":"toString"}` in localStorage and you get `card.dataset.mark = "toString"`, a chip labelled `toString`, and class `mark-toString` — instead of degrading to the unmarked view the PR description promises. Not exploitable (it's `textContent`, and none of those keys contain spaces so no class injection), but the guard isn't doing its job.

Worth noting this came in as the fix for the bot's "validate state against an allowlist" comment, so it's the one thread that shouldn't be resolved as addressed. An array plus `indexOf`, or `hasOwnProperty.call`, closes it in both call sites.

**2. `TestStartServer_HasNoWriteRoutes` doesn't test what it says**

Four separate problems: `StartServer` resolves its root from `SessionsRoot()`, so the `t.TempDir()` fixture is dead code and the test reads the developer's real `~/.opencodereview/sessions`. The 404s come from `handleRepos`'s `r.URL.Path != "/"` branch and arrive for every method, while `POST /r/{repo}/{sessionID}` — a route that does exist — returns 200, since nothing in the viewer checks `r.Method`. It also leaks a goroutine and a bound port for the rest of the binary, which the neighbouring `TestStartServer_AddrInUse` explicitly documents avoiding. And the reserve-then-release port dance is a TOCTOU under parallel CI.

Extracting `newMux(root string) *http.ServeMux` and table-testing it with `httptest` fixes all four and lets you assert the thing you actually care about: how the real routes behave under POST.

**3. `hidden` in the toolbar counts two different things**

`total - visibleCount` folds mark-hiding into severity/category filtering, so filtering to Critical reads "3 marked, 20 hidden". Count the mark-hidden cards specifically. Same for the empty state — `No comments match this filter.` is wrong when the reason is marks.

**4. Failed writes are invisible**

`saveStoredMarks` only `console.error`s, so marks look applied and vanish on reload. This is the bot's "no user-facing feedback" comment in a new costume rather than something the server removal solved — worth reusing the `data-marks-count` span for a one-line warning.

**5. `Clear all` writes `{}` instead of removing the key**

Which means a key per session, forever, on a fixed `localhost:5483` origin. One of the arguments against the server-side version was `.marks.json` accumulating with no reclaim path; the same thing happens here, and once the quota fills it lands straight on #4's silent failure. `Object.keys(marks).length === 0 ? removeItem : setItem` covers it.

nitishagar:
Thanks @lizhengfeng101 — all five pushed as an amend to the same commit (a11fde0):

**1. Allowlist now actually rejects.** `knownMarkStates` is the array `[fixed, ignored]` and both call sites (`loadStoredMarks`, `applyMarkState`) guard with `indexOf(...) !== -1`, so `toString`/`valueOf`/`constructor` and any future `Object.prototype` member degrade to the unmarked view. Relatedly, the marks map is now built with `Object.create(null)` (every return site, plus Clear-all), so a stored own `__proto__` key replays as inert data instead of a prototype set.

**2. The no-write-routes test now tests what it says.** Extracted `newMux(root)` from `StartServer` and registered the document routes with GET-only patterns (`GET /{$}`, `GET /r/{repo}`, `GET /r/{repo}/{sessionID}`) — so POST/PUT/DELETE/PATCH now get the ServeMux's own 405 + `Allow` (the "down to GETs" surface is enforced, not just argued), unknown paths still 404, and `GET /{$}` keeps the root exact-match. `TestMux_HasNoWriteRoutes` table-drives the real mux with `httptest` against a real `t.TempDir` fixture: full write-method × route cross-product → 405, HEAD/GET positives → 200, unknown write-looking path → 404. No ports, no goroutines, no `SessionsRoot()`.

**3. `hidden` counts only mark-hidden cards** — cards the active filters would otherwise show. Filtering to one severity with a marked card outside it now reads e.g. "1 marked, 0 hidden" instead of folding filter-excluded cards in. The empty state distinguishes causes too: "All matching comments are hidden by marks." vs "No comments match this filter."

**4. Failed persists are visible.** `saveStoredMarks` sets a flag on failure (save or removal) and the toolbar span renders "… — not saved (storage unavailable)" until the next successful persist clears it; in-page marks keep working meanwhile.

**5. Empty marks leave no residue.** `saveStoredMarks` does `removeItem` when the map is empty instead of persisting `{}` — verified in devtools that Clear-all leaves no key.

All of this was exercised end-to-end in a browser against a fixture session (poisoned `toString` states rejected, `__proto__` inert, quota forced full → warning appears and later clears, Clear-all removes the key), and `go test ./...` is green.

internal/viewer/server.go
```diff
@@ -107,6 +82,44 @@ func StartServer(addr, openMode string) error {
...
+// patterns (which also serve HEAD), so the ServeMux itself answers any other
+// method with 405 + Allow before a handler runs. The root pattern matches
+// exactly "/" via {$}; every other unmatched path gets the ServeMux's 404.
+// New routes must register method-qualified patterns to keep this contract
+// testable (TestMux_HasNoWriteRoutes).
+func newMux(root string) *http.ServeMux {
+	mux := http.NewServeMux()
+
+	// Static assets.
+	mux.Handle("/static/", http.StripPrefix("/static/", http.FileServer(http.FS(staticFS()))))
```
lizhengfeng101: `http.FileServer` answers non-GET/HEAD with 200 rather than 405, and the static route is the only one here without a method qualifier, so it leaves a gap in the read-only contract `TestMux_HasNoWriteRoutes` claims.

```suggestion
	mux.Handle("GET /static/", http.StripPrefix("/static/", http.FileServer(http.FS(staticFS()))))
```

internal/viewer/handler_test.go
```diff
@@ -213,3 +213,108 @@ func TestHandleSession_EmptyCWD(t *testing.T) {
...
+		{"DELETE repo route", http.MethodDelete, "/r/repo", http.StatusMethodNotAllowed},
+		{"PATCH repo route", http.MethodPatch, "/r/repo", http.StatusMethodNotAllowed},
+		{"POST root", http.MethodPost, "/", http.StatusMethodNotAllowed},
+		{"PUT root", http.MethodPut, "/", http.StatusMethodNotAllowed},
+		{"DELETE root", http.MethodDelete, "/", http.StatusMethodNotAllowed},
+		{"PATCH root", http.MethodPatch, "/", http.StatusMethodNotAllowed},
+		{"OPTIONS root", http.MethodOptions, "/", http.StatusMethodNotAllowed},
+		{"GET session route still served", http.MethodGet, "/r/repo/s1", http.StatusOK},
+		{"HEAD session route still served", http.MethodHead, "/r/repo/s1", http.StatusOK},
+		{"GET repo route still served", http.MethodGet, "/r/repo", http.StatusOK},
```
lizhengfeng101, line 309: Cover the static route as well, otherwise dropping the `GET ` prefix later still leaves this guard test green (apply together with the previous suggestion — these rows fail without it).

```suggestion
		{"OPTIONS root", http.MethodOptions, "/", http.StatusMethodNotAllowed},
		{"POST static asset", http.MethodPost, "/static/session.js", http.StatusMethodNotAllowed},
		{"PUT static asset", http.MethodPut, "/static/session.js", http.StatusMethodNotAllowed},
		{"DELETE static asset", http.MethodDelete, "/static/session.js", http.StatusMethodNotAllowed},
		{"GET session route still served", http.MethodGet, "/r/repo/s1", http.StatusOK},
		{"HEAD session route still served", http.MethodHead, "/r/repo/s1", http.StatusOK},
		{"GET repo route still served", http.MethodGet, "/r/repo", http.StatusOK},
		{"GET static asset still served", http.MethodGet, "/static/session.js", http.StatusOK},
```

lizhengfeng101 (approved):
LGTM

PR #1123, opened 2026-09-01
fix(rules): route .cxx and .hxx files to C++ rules

DreamZhongJu:
## Description

`.cxx` and `.hxx` are already accepted by the file allowlist, but the built-in C++ rule mapping only covers `.cpp`, `.cc`, and `.hpp`. As a result, these files fall back to the default review rule.

This change extends the existing C++ rule mapping to include `.cxx` and `.hxx`, and adds a regression test for both extensions.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Validated on Ubuntu 24.04 with Go 1.25.6, CGO enabled, and GCC 13.3:

- `go test -race -count=1 ./...`
- `go vet ./...`
- `go run scripts/verify-english-only.go`
- `bash scripts/verify-license.sh`
- `go build ./cmd/opencodereview`

`ocr rules check` now reports the C++ rule pattern for both `.cxx` and `.hxx` files instead of `default`.

## Checklist

- [x] I have signed the CLA

## Related Issues

None.


internal/config/rules/cpp_extension_rules_test.go
wu21-web, line 1: Is there a reason why you separated this test?

internal/config/rules/system_rules.json
```diff
@@ -22,7 +22,7 @@
     "**/*.{ts,js,tsx,jsx}": "ts_js_tsx_jsx.md",
     "**/*.{kt}": "kotlin.md",
     "**/*.rs": "rust.md",
-    "**/*.{cpp,cc,hpp}": "cpp.md",
+    "**/*.{cpp,cc,cxx,hpp,hxx}": "cpp.md",
     "**/*.c": "c.md",
```
wu21-web, line 26: This is minor, but please include them in the docs

internal/config/rules/cpp_extension_rules_test.go
DreamZhongJu: No strong reason—I separated it to keep the regression focused, but the existing `TestResolve_DefaultRules` table is a better fit. I’ve moved the `.cxx` and `.hxx` cases there and removed the standalone test file.

internal/config/rules/system_rules.json
DreamZhongJu: Good catch. I’ve updated the built-in rule mapping in all five localized `review-rules.md` files.

lizhengfeng101 (approved):
LGTM

Issue #1121, opened 2026-08-31
Workspace review skips staged newly added files


acoliver:
## Environment

- OCR: v1.11.1
- Mode: workspace
- Output: `--audience agent --format json`
- Git state: one modified tracked file and seven staged newly added files

## Command

```sh
ocr review --audience agent --format json --timeout 15 --concurrency 4 --background "..."
```

## Expected

Workspace mode reviews staged, unstaged, and untracked changes, including newly added files.

## Actual

The run completed successfully but selected only the modified tracked file:

```json
{
  "message": "Review complete: 0 finding(s) across 1 selected item(s).",
  "summary": { "files_reviewed": 1 },
  "groups": [{ "label": "src/ui_gpui/mod.rs", "files": ["src/ui_gpui/mod.rs"] }]
}
```

The seven staged additions were absent from `manifest.coverage.selected`. They included six text files and one 3,478-line Rust file, so the omission was not caused only by binary detection. The command exited 0 and stderr was empty.

`git status --short` immediately before the run showed:

```text
M  src/ui_gpui/mod.rs
A  src/ui_gpui/vendor/gpui_base_selection/LICENSE-APACHE
A  src/ui_gpui/vendor/gpui_base_selection/auto_scroll.rs
A  src/ui_gpui/vendor/gpui_base_selection/global_state.rs
A  src/ui_gpui/vendor/gpui_base_selection/mod.rs
A  src/ui_gpui/vendor/gpui_base_selection/text_boundary.rs
A  src/ui_gpui/vendor/gpui_base_selection/text_selection.rs
A  src/ui_gpui/vendor/mod.rs
```

The run manifest reported workspace mode with `resolved_base` equal to HEAD.

Parikalp-Bhardwaj:
I traced this to filterDiffs reapplying .gitignore after git collects staged changes. I am working on a focused fix that preserves tracked and staged files while still excluding ignored untracked files, with a regression test.

PR #1126, opened 2026-09-01
fix(diff): preserve staged ignored files in workspace

Parikalp-Bhardwaj:
## Summary

Fixes workspace reviews dropping staged files that match `.gitignore` patterns.

## Changes

- Stop reapplying `.gitignore` rules to diffs already collected by Git.
- Continue excluding ignored untracked files.
- Preserve existing internal directory exclusions.
- Add a regression test covering tracked, staged, and ignored untracked files together.

Related #1121

### DIFF
--- a/internal/diff/git.go
+++ b/internal/diff/git.go
@@ -385,15 +385,17 @@ func matchGitignoreDirectory(relPath, pattern string) bool {
 }
 
 // filterDiffs removes diffs whose file paths are excluded.
+//
+// .gitignore governs untracked files only; untrackedFilesList
+// already screens those before they become diffs.
 func (p *Provider) filterDiffs(diffs []model.Diff) []model.Diff {
-	patterns := p.loadGitignorePatterns()
 	var result []model.Diff
 	for _, d := range diffs {
 		path := d.NewPath
 		if path == "/dev/null" {
 			path = d.OldPath
 		}
-		if !p.isPathExcluded(path, patterns) {
+		if !p.isPathExcluded(path, nil) {
 			result = append(result, d)
 		}
 	}
--- a/internal/diff/git_test.go
+++ b/internal/diff/git_test.go
@@ -538,3 +538,55 @@ func TestCommitDiffMergeCommitReviewsFirstParentDiff(t *testing.T) {
 		t.Error("NewFileContent is empty: content was not read at the merge commit")
 	}
 }
+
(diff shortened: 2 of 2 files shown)


lizhengfeng101:
Thanks for the fix @Parikalp-Bhardwaj — the underlying bug is still present on `main`, so this is still worth landing. But the branch can't be merged as-is: it's branched from an old base and the `filterDiffs` function it edits no longer exists.

**What changed on main:** `filterDiffs` was refactored and renamed to `partitionDiffs` in #1223 / #1242, which split diffs into `Included` / `Excluded`. That refactor only moved the code — it did **not** fix this bug. `partitionDiffs` still reapplies `.gitignore` to already-collected diffs, and worse, gitignore-matched staged files now fall through to *neither* `Included` nor `Excluded` (silently dropped). I confirmed this by running your regression test against current `main` — it fails:

```
staged new file matching .gitignore was dropped; got map[src/app.go:true]
```

**Could you please:**
1. Rebase onto latest `main`.
2. Move the fix into `partitionDiffs` (`internal/diff/git.go`): drop the `isPathExcluded` gitignore check so a force-added ignored file lands in `Included`, while keeping the `isProviderDirExcluded` handling intact.
3. Re-target your regression test to the rebased code.

The test you added already covers the right cases, so it should transfer directly.

Parikalp-Bhardwaj:
Thanks for the guidance @lizhengfeng101  

Rebased #1126 onto the latest `main` and moved the fix into `partitionDiffs`.

It now keeps force-added staged files that match `.gitignore`, while ordinary ignored untracked files are still filtered earlier.

Provider-directory handling and ordering are unchanged.

I also updated the #1300 ordering test, since it expected a tracked ignored file to be dropped. With this fix, tracked files matching `.gitignore` stay included.

Let me know if you want this behavior limited only to force-added staged files.

Parikalp-Bhardwaj:
@lizhengfeng101 since `partitionDiffs` is shared by workspace, range, and commit modes, this also means the current working tree's `.gitignore` will no longer filter diffs that Git already returned in range/commit mode.

That overlaps a bit with #1487, but I haven't marked it as fixed because that, issue also discusses whether the reviewed ref's ignore rules should be used.

#1510/#1511 look complementary here — they add the separate frozen `--staged` mode, while this PR fixes the existing diff partitioning behavior.

PR #1135, opened 2026-09-02
fix(action): support explicit llm_protocol and preserve protocol env override (#1134)

### LINKED ISSUE #1134: bug(action): Configure OCR step hardcodes llm.protocol to openai, breaking openai-responses protocol
### OpenCodeReview Version
v1.11.2 (and all versions since v1.10.0 / PR #1051)

### Operating System
Linux (x86_64)

### Installation Method
GitHub Action (composite action `action.yml`)

### LLM Provider
OpenAI (`openai-responses` / Responses API)

### Bug Description
In `action.yml`, the `Configure OCR` step added in PR #1051 unconditionally overwrites `llm.protocol` with `"openai"` whenever `llm_use_anthropic` is not set to true:

```bash
NORMALIZED_USE_ANTHROPIC="$(printf '%s' "$OCR_USE_ANTHROPIC" | tr '[:upper:]' '[:lower:]')"
case "$NORMALIZED_USE_ANTHROPIC" in
  ""|true|1|yes)
    OCR_USE_ANTHROPIC="true"
    OCR_LLM_PROTOCOL="anthropic"
    ;;
  *)
    OCR_USE_ANTHROPIC="false"
    OCR_LLM_PROTOCOL="openai"
    ;;
esac
ocr config unset provider
...
ocr config set llm.protocol "$OCR_LLM_PROTOCOL"
```

This causes two major issues:
1. It unsets any configured provider and ignores `OCR_LLM_PROTOCOL` passed via workflow step `env:` (e.g. `OCR_LLM_PROTOCOL: openai-responses`).
2. When calling OpenAI with the Responses API format (e.g. `llm_extra_body: '{"reasoning":{"effort":"medium"}}'` or when using reasoning models with function tools), all requests are forced to `/v1/chat/completions` instead of `/v1/responses`, resulting in `400 Bad Request` from OpenAI API (`Unknown parameter: 'reasoning'` or `Function tools with reasoning_effort are not supported for model in /v1/chat/completions`).

Prior to PR #1051 (e.g. in `v1.9.10`), `Configure OCR` did not touch `llm.protocol` or unset the provider, allowing `OCR_LLM_PROTOCOL: openai-responses` to be passed via environment variables directly to the CLI.

### Steps to Reproduce
1. In a GitHub Actions workflow, invoke `alibaba/open-code-review` with `llm_url: https://api.openai.com/v1`, `llm_use_anthropic: "false"`, `env: OCR_LLM_PROTOCOL: openai-responses`, and `llm_extra_body: '{"reasoning":{"effort":"medium"}}'`.
2. Inspect the step log: `Configure OCR` executes `ocr config set llm.protocol openai`.
3. In `ocr review`, all API calls fail with `POST "https://api.openai.com/v1/chat/completions": 400 Bad Request {"message": "Unknown parameter: 'reasoning'."}`.

### Expected Behavior
`action.yml` should allow using the `openai-responses` protocol (and any custom protocol) either by:
1. Preserving `OCR_LLM_PROTOCOL` when set in the environment:
```bash
if [ -z "$OCR_LLM_PROTOCOL" ]; then
  case "$NORMALIZED_USE_ANTHROPIC" in
    ""|true|1|yes) OCR_LLM_PROTOCOL="anthropic" ;;
    *)             OCR_LLM_PROTOCOL="openai" ;;
  esac
fi
```
2. Or exposing a dedicated `llm_protocol` input in `action.yml`.

### Logs / Error Output
```
POST "https://api.openai.com/v1/chat/completions": 400 Bad Request {
  "message": "Unknown parameter: 'reasoning'.",
  "param": "reasoning",
  "code": "unknown_parameter"
}
```

bbubakanov:
## Description

In `action.yml`, the `Configure OCR` step introduced in #1051 unconditionally set `llm.protocol` to `"openai"` whenever `llm_use_anthropic` was false, discarding any custom/Responses protocol set via environment variables (e.g. `OCR_LLM_PROTOCOL: openai-responses`). This caused reviews against models requiring the OpenAI Responses API (such as models with reasoning effort or function tools) to fail with `400 Bad Request` against `/v1/chat/completions`.

This PR:
1. Adds an optional `llm_protocol` input to `action.yml` (mapped to `OCR_LLM_PROTOCOL`), allowing workflows to explicitly configure protocols like `openai-responses`.
2. Resolves `llm.protocol` using `OCR_LLM_PROTOCOL` when provided (either via `llm_protocol` input or step `env: OCR_LLM_PROTOCOL`), and only falls back to `anthropic` / `openai` based on `llm_use_anthropic` when omitted.
3. Includes `OCR_FP_LLM_PROTOCOL` in the checkpoint range cache fingerprint so changing the protocol invalidates stale checkpoint hashes.
4. Forwards `OCR_LLM_PROTOCOL` into the `Run OpenCodeReview` step environment.
5. Adds contract tests verifying both explicit `llm_protocol` input and `OCR_LLM_PROTOCOL` environment override resolution.

## Type of Change

- [ ] Documentation update
- [x] CI / Build / Tooling

## How Has This Been Tested?

- [x] `npm run test:github-actions` (all 26 action contract tests pass, including new `llm_protocol` contract tests)
- [x] `npm run test:update` passes
- [x] `npm run test:launcher` passes

## Checklist

- [x] I have signed the CLA

## Related Issues

Closes #1134

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -17,13 +17,20 @@ inputs:
   llm_model:
     description: Model name (mapped to env OCR_LLM_MODEL).
     required: true
+  llm_protocol:
+    description: >-
+      LLM protocol: openai, openai-responses, anthropic, or a custom protocol name
+      (mapped to env OCR_LLM_PROTOCOL). If omitted, falls back to the protocol
+      selected by llm_use_anthropic.
+    required: false
   llm_use_anthropic:
     description: >-
       Selects the LLM protocol (mapped to env OCR_USE_ANTHROPIC). An explicitly supplied empty
       string, true, 1, or yes selects Anthropic
       case-insensitively; every other value selects the OpenAI-compatible
-      protocol, preserving the CLI environment contract.
-    required: true
+      protocol, preserving the CLI environment contract. Ignored when llm_protocol is set.
+    required: false
+    default: 'false'
   llm_auth_header:
     description: Custom auth header name (mapped to env OCR_LLM_AUTH_HEADER).
     required: false
@@ -379,6 +386,7 @@ runs:
       env:
         OCR_LLM_URL: ${{ inputs.llm_url }}
         OCR_LLM_MODEL: ${{ inputs.llm_model }}
+        OCR_LLM_PROTOCOL: ${{ inputs.llm_protocol }}
         OCR_USE_ANTHROPIC: ${{ inputs.llm_use_anthropic }}
         OCR_LLM_AUTH_HEADER: ${{ inputs.llm_auth_header }}
         OCR_EXTRA_BODY: ${{ inputs.llm_extra_body }}
@@ -389,13 +397,14 @@ runs:
         case "$NORMALIZED_USE_ANTHROPIC" in
           ""|true|1|yes)
             OCR_USE_ANTHROPIC="true"
-            OCR_LLM_PROTOCOL="anthropic"
+            DEFAULT_PROTOCOL="anthropic"
             ;;
           *)
             OCR_USE_ANTHROPIC="false"
-            OCR_LLM_PROTOCOL="openai"
+            DEFAULT_PROTOCOL="openai"
             ;;
         esac
+        RESOLVED_PROTOCOL="${OCR_LLM_PROTOCOL:-$DEFAULT_PROTOCOL}"
         ocr config unset provider
         ocr config set llm.auth_token ""
         ocr config set llm.extra_headers ""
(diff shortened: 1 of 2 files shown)


lizhengfeng101:
@bbubakanov Thanks a lot for this, @bbubakanov — and sorry to close it. We're consolidating the llm_protocol work onto #1192, which covers the same bug (#1134) with a few behaviors
this PR doesn't, so I'm closing this one to avoid overlap rather than because of any problem with your approach.

For transparency, the specific points that led us to #1192:

- use_anthropic / protocol consistency. This PR sets llm.use_anthropic from the llm_use_anthropic input and llm.protocol from RESOLVED_PROTOCOL independently, so
llm_protocol: anthropic with llm_use_anthropic: 'false' writes a config where the two disagree. #1192 mirrors use_anthropic from the resolved protocol (matching how ocr 
config set llm.protocol behaves), so they can't drift apart.
- Checkpoint fingerprint. Here OCR_FP_LLM_PROTOCOL reads the raw inputs.llm_protocol and is added to the hash unconditionally. That invalidates every existing checkpoint
once on upgrade, and — because it captures the input rather than the resolved value — a protocol switched purely through an inherited OCR_LLM_PROTOCOL env var doesn't
change the fingerprint, so a stale checkpoint can be reused across a protocol change (fail-open). #1192 appends only when a protocol is actually set (legacy checkpoints
stay valid) and fingerprints the resolved value (an env-only switch still invalidates, fail-closed).
- Validation & docs. #1192 rejects an unknown protocol with the accepted list, normalizes case, and updates the Actions README (usage example + the config_changed row),
which we'd want to land alongside the input.

Your PR did get the core env-override resolution right, and the reproduction in #1134 was solid — that context was genuinely useful in evaluating the alternatives. Thank
you for the contribution.

PR #1139, opened 2026-09-02
fix(ci): restore npm provenance by publishing from GitHub-hosted runners

### LINKED ISSUE #1137: npm releases lack provenance
### OpenCodeReview Version

1.11.2

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

Anthropic (Claude)

### Bug Description

npm package releases lack provenance, but used to have provenance and so mise fails to install

```
mise use -g npm:@alibaba-group/open-code-review
npm:@alibaba-group/open-code-review@1.11.2 resolving                                                                                                                               ◡
mise ERROR Failed to install npm:@alibaba-group/open-code-review@latest: aube install failed: failed to resolve dependencies
  caused by: trust downgrade for @alibaba-group/open-code-review@1.11.2 (trustPolicy=no-downgrade): earlier published version 1.1.10 had provenance attestation but this version has no trust evidence

This is a supply-chain trust failure, not an ordinary version-resolution error. An earlier release had stronger trust evidence than the selected release. This can indicate a compromised or tampered release; it can also happen when a maintainer manually publishes, backports outside the trusted workflow, skips provenance for convenience, or uses a registry that strips metadata.

Before bypassing, inspect the package's npm release, source tag/commit, publisher identity, and tarball; compare the metadata with npmjs.org. Confirm the release is expected and nothing appears tampered with, then report inconsistent evidence to the relevant upstream owner. Package-release drift belongs with the maintainer; metadata present on npmjs.org but missing from a proxy or mirror belongs with that registry operator.

Only after review, add the narrowest affected `<package>@<version>` to `trust_policy_excludes` for this tool, e.g.:
  "npm:@alibaba-group/open-code-review" = { version = "latest", trust_policy_excludes = ["<package>@<version>"] }
A bare package name exempts every version. `mise settings npm.shell_out=true` uses the npm CLI and bypasses this check entirely, so it should be a last resort.

Investigation guide and known exceptions: https://aube.jdx.dev/security#trust-policy
mise ERROR Version: 2026.8.14 macos-arm64 (2026-08-25)
mise ERROR Run with --verbose or MISE_VERBOSE=1 for more information
```

### Steps to Reproduce

n/a

### Expected Behavior

npm packages published with provenance, ie: have the green tick... see https://docs.npmjs.com/viewing-package-provenance

### Logs / Error Output

```shell

```

### Additional Context

_No response_

nitishagar:
## Summary

Restores npm provenance attestation for all published packages (`@alibaba-group/open-code-review` plus the six `@alibaba-group/ocr-*` platform packages) by moving the lightweight `npm-publish` job back to a GitHub-hosted runner with `id-token: write` and passing `--provenance` to both `npm publish` invocations.

Fixes #1137.

## Background

Provenance was added in 365c9ee (OIDC trusted publishing) and worked while publishes ran on `ubuntu-latest` (last attested release: 1.3.1). When the release pipeline moved to self-hosted runners (76b4d5a), `npm publish --provenance` started failing with E422 — npm provenance rejects OIDC tokens minted outside GitHub-hosted runners — so f577742 dropped the flag and the `id-token` permission. Every release since v1.3.4 has published without attestation, which downstream trust tools (e.g. mise/aube `no-downgrade`) now flag as a supply-chain trust downgrade, breaking installs.

## Change

- `npm-publish` job: `runs-on: self-hosted` → `ubuntu-latest` (the build and release jobs stay self-hosted, untouched).
- `npm-publish` permissions: add `id-token: write` (keeps `contents: read`).
- Both `npm publish` commands gain `--provenance`.

The `node:24` container, `.npmrc`/`NODE_AUTH_TOKEN` auth, and the already-published skip guard are unchanged. A publish whose OIDC token is unavailable now fails loudly instead of silently shipping without provenance.

## Validation

- `make test` green (including the `internal/release` tests that pin release.yml content).
- YAML parse check; grep-verified: exactly one `runs-on` changed, `--provenance` appears on both publishes, `id-token: write` scoped to the two jobs that need it.
- The green tick itself is observable on npmjs.org after the next tag release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -161,12 +161,13 @@ jobs:
 
   npm-publish:
     needs: release
-    runs-on: self-hosted
+    runs-on: ubuntu-latest
     timeout-minutes: 15
     container:
       image: node:24
     permissions:
       contents: read
+      id-token: write
     steps:
       - uses: actions/checkout@v7
 
@@ -214,7 +215,7 @@ jobs:
             if [ "$already" = "$VERSION" ]; then
               echo "Skip: ${pkg_name}@${VERSION} already published"
             else
-              npm publish "$pkg_dir" --access public
+              npm publish "$pkg_dir" --access public --provenance
               echo "Published: ${pkg_name}@${VERSION}"
             fi
           done
@@ -237,7 +238,7 @@ jobs:
           if [ "$already" = "$VERSION" ]; then
             echo "Skip: ${pkg_name}@${VERSION} already published"
           else
-            npm publish --access public
+            npm publish --access public --provenance
             echo "Published: ${pkg_name}@${VERSION}"
           fi
         env:


lizhengfeng101:
Thanks for digging into this — the root-cause analysis is accurate and #1137 is a real problem.

That said, I'm going to close this one. We moved the release pipeline to self-hosted runners specifically because GitHub-hosted queue times had gotten severe enough to hold up tag releases, and putting `npm-publish` back on `ubuntu-latest` reintroduces exactly that dependency on the critical path — a queued publish job blocks the release even though the job itself is cheap.

I'd rather fix provenance without giving up self-hosted. I'll keep #1137 open and look at it from that angle.

Appreciate the writeup either way — it saved me the archaeology.

Issue #972, opened 2026-08-17
一直报git show失败的错误


hubo2024:
### OpenCodeReview Version

v1.9.5

### Operating System

Windows

### Installation Method

npm (global)

### LLM Provider

Other OpenAI-compatible endpoint

### Bug Description

ocr review命令正常
ocr review --commit 0b335be72cb7e342c115ff3ffadbe741a4715377 时一致报git show 失败。
但是我在同一个窗口直接运行：git show 0b335be72cb7e342c115ff3ffadbe741a4715377 又是正常的。
唯一一点是我使用的windows，然后语言是日语的，等于是日语系统，跟这个有关系吗？

<img width="736" height="381" alt="Image" src="https://github.com/user-attachments/assets/5aeb777e-4c02-476c-9744-6e18cd366c31" />

### Steps to Reproduce

我直接在我的项目根目录，cmd运行官网的ocr review --commit 0b335be72cb7e342c115ff3ffadbe741a4715377命令，就一直报错

### Expected Behavior

如果是日语系统的问题，希望也能正常使用

### Logs / Error Output

```shell

```

bailu-ZZ:
From the current code path, `--commit` runs `git show` with `--diff-merges=first-parent` and `--end-of-options`, so this looks more like a Git version compatibility issue than a Japanese Windows locale issue. Please share `git --version`, the complete `ocr review` error including stderr, and the output of the following command.

```shell
git -c core.quotepath=false show --no-ext-diff --no-textconv --find-renames --src-prefix=a/ --dst-prefix=b/ --no-color --diff-merges=first-parent -U3 --end-of-options <commit>
```

The project currently requires Git >= 2.41. If this command reports an unknown option, upgrading Git should explain the behavior. If Git meets that requirement, the full error will help narrow down the Windows-specific cause.

PR #1140, opened 2026-09-02
fix(diff): propagate untracked file listing errors

hyy321:
## Description

In workspace mode, untracked files are collected with `git ls-files --others --exclude-standard`.

Previously, a failure from this command was handled the same way as a successful command that returned no files:

```go
if err != nil || out == "" {
    return nil, nil
}
```

As a result, callers could not distinguish between "there are no untracked files" and "Git failed to enumerate untracked files". If the command fails after the tracked workspace diff has already been collected, workspace diff collection may continue without knowing that the untracked-file portion of the input could not be collected.

This PR separates these two cases. `git ls-files` is now executed through `runGitSplit`, and failures are propagated with `gitFailure`, consistent with the error handling used by other Git operations in the diff provider. A successful command with an empty result still means there are no untracked files, as before.

A regression test verifies that a failed Git invocation in `untrackedFilesList` returns an error instead of being silently treated as an empty result.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Added `TestUntrackedFilesListPropagatesGitFailure`, which forces the Git invocation to fail and verifies that `untrackedFilesList` returns an error.

Verified locally with:

```bash
make check
make test
make build
```

## Checklist

- [x] I have signed the CLA

## Related Issues

No related issue. This was found while reviewing the workspace diff error-handling path.

### DIFF
--- a/internal/diff/git_error_test.go
+++ b/internal/diff/git_error_test.go
@@ -134,6 +134,17 @@ func TestGetDiff_WorkspaceFailureSurfacesFallbackMessage(t *testing.T) {
 	}
 }
 
+func TestUntrackedFilesListPropagatesGitFailure(t *testing.T) {
+	repo := filepath.Join(t.TempDir(), "missing-repo")
+
+	provider := NewWorkspaceProvider(repo, nil)
+
+	_, err := provider.untrackedFilesList(context.Background())
+	if err == nil {
+		t.Fatal("expected untrackedFilesList to return git error")
+	}
+}
+
 // shimGit puts a fake `git` at the front of PATH for the duration of the test.
 func shimGit(t *testing.T, body string) {
 	t.Helper()
--- a/internal/diff/git_error_test.go
+++ b/internal/diff/git_error_test.go
@@ -134,6 +134,11 @@ func TestGetDiff_WorkspaceFailureSurfacesFallbackMessage(t *testing.T) {
 	}
 }
 
+// TestUntrackedFilesListPropagatesGitFailure separates the two cases the old
+// `if err != nil || out == ""` folded together: a repository with no untracked
+// files, and git failing to enumerate them at all. The second one used to
+// return an empty list, so a workspace review carried on with the untracked
+// half of its input missing and no way for the caller to notice.
 func TestUntrackedFilesListPropagatesGitFailure(t *testing.T) {
 	repo := filepath.Join(t.TempDir(), "missing-repo")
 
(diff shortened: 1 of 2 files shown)


lizhengfeng101:
A missing repo dir fails at exec, so `stderr` is empty and the test never exercises the part this PR changes — it would pass against the old `runGit` version too, as long as the error were returned at all.

`shimGit` in the same file gives you a real `git` failure with something on stderr:

```go
shimGit(t, "echo 'some/path.txt'\necho 'fatal: shim refused' >&2\nexit 128\n")

_, err := NewWorkspaceProvider(t.TempDir(), gitcmd.New(2)).untrackedFilesList(context.Background())
// err == nil -> fatal
// want "git ls-files failed" and "shim refused" in err.Error()
// and no "some/path.txt": ls-files writes repo paths to stdout, which must stay out of the error
```

That last assertion is worth having on its own, next to `TestGetDiff_FailureQuotesStderrNotStdout`.

internal/diff/git_error_test.go
```diff
@@ -134,6 +134,17 @@ func TestGetDiff_WorkspaceFailureSurfacesFallbackMessage(t *testing.T) {
...
+func TestUntrackedFilesListPropagatesGitFailure(t *testing.T) {
+	repo := filepath.Join(t.TempDir(), "missing-repo")
+
+	provider := NewWorkspaceProvider(repo, nil)
+
+	_, err := provider.untrackedFilesList(context.Background())
+	if err == nil {
+		t.Fatal("expected untrackedFilesList to return git error")
+	}
+}
```
lizhengfeng101, line 156: Two small additions, both committable from here.

The `err == nil` check already does the main job — I verified it fails against the old `if err != nil || out == ""` version, so the regression is genuinely pinned. What it does not pin is *which* error: swap `gitFailure` for a bare `fmt.Errorf("...: %w", err)` and the test still passes with git's diagnosis gone, which is #972 coming back. One `strings.Contains` closes that. The doc comment just matches the rest of this file.

```suggestion
// TestUntrackedFilesListPropagatesGitFailure separates the two cases the old
// `if err != nil || out == ""` folded together: a repository with no untracked
// files, and git failing to enumerate them at all. The second one used to
// return an empty list, so a workspace review carried on with the untracked
// half of its input missing and no way for the caller to notice.
func TestUntrackedFilesListPropagatesGitFailure(t *testing.T) {
	repo := filepath.Join(t.TempDir(), "missing-repo")

	provider := NewWorkspaceProvider(repo, nil)

	_, err := provider.untrackedFilesList(context.Background())
	if err == nil {
		t.Fatal("expected untrackedFilesList to return git error")
	}
	// Pins gitFailure rather than any error: a bare fmt.Errorf would satisfy the
	// check above while dropping git's own diagnosis, which is the #972 regression.
	if got := err.Error(); !strings.Contains(got, "git ls-files failed") {
		t.Errorf("error %q lost the operation name", got)
	}
}
```

`strings` is already imported. Verified: `gofmt` clean, test passes as written, and fails with `lost the operation name` when `gitFailure` is swapped out.

Disregard my earlier comment about rewriting this around `shimGit` — the stdout-leak property it would have covered is already pinned by `TestGetDiff_FailureQuotesStderrNotStdout`, so that was polish, not coverage.

lizhengfeng101 (approved):
LGTM

hyy321:
Thanks for the clarification and the suggested change. I’ve synced it and understand the distinction now: the original test already covers the swallowed-error regression, while the additional assertion ensures the gitFailure context is preserved.
For similar changes in the future, I’ll take a closer look at the relevant code path and nearby tests before deciding whether additional coverage is needed.

PR #1145, opened 2026-09-02
fix(launcher): forward SIGINT/SIGTERM signals to child process

### LINKED ISSUE #1142: fix(launcher): bin/ocr.js does not forward signals, orphaning the ocr process on cancel
## Description

`package.json` maps the `ocr` bin to `bin/ocr.js`, so on an npm install the process named `ocr` on `PATH` is the Node launcher, not the Go binary. The launcher runs the binary with `spawnSync` and installs no signal handlers (`bin/ocr.js:80`). Node therefore applies the OS default disposition to SIGTERM/SIGINT and dies immediately, while the Go child — a separate process that was never signalled — keeps running, reparented to init.

Reproduced with the launcher's exact shape:

```
$ node -e 'require("child_process").spawnSync("/bin/sh",["-c","sleep 30"],{stdio:"ignore"});' &
node=72449  child=72451     (both in pgid 72445)

$ kill -TERM 72449
node: Terminated: 15
RESULT node=DEAD
RESULT child=ALIVE (orphaned)
  PID  PPID  PGID COMM
72451     1 72445 sleep     <- reparented to init, still running
```

Group-directed senders are unaffected: `spawnSync` does not detach, so a signal sent to the process group reaches the Go binary directly. That covers shells, CI runners, and the opencode plugin, which kills the group on purpose (`plugins/open-code-review/opencode/open-code-review.ts:163`).

PID-directed senders are the problem:

- **`docker stop`** signals PID 1 only. With the launcher as the entrypoint, the review keeps running for the full grace period and everything is then SIGKILLed.
- **The VS Code extension** calls `this.current.kill('SIGTERM')` (`extensions/vscode/src/extension/services/CliService.ts:163`), which signals that one PID. Node dies, `proc.on('close')` fires, and the 3s SIGKILL escalation is cleared — it would have targeted an already-dead PID anyway. The orphaned Go process keeps reviewing and keeps spending tokens until it next writes a progress line to a pipe whose read end node just closed, at which point SIGPIPE kills it outright. That death is unmanaged: no run manifest, no `session_end`, no resumable session.

So "cancel review" does not deterministically cancel the review, and #1111's graceful SIGTERM handling never applies to the officially recommended install path — the same one `action.yml:348` uses.

## Scope

- File: `bin/ocr.js`
- Function/area: the `spawnSync` invocation and `launcherExitCode`
- Tests: `bin/ocr.test.js`

## Acceptance Criteria

- [ ] The launcher uses `spawn` rather than `spawnSync` and forwards SIGINT and SIGTERM to the child
- [ ] The launcher waits for the child and propagates its exit status, preserving the existing `launcherExitCode` contract (128+N when the child itself dies by a signal)
- [ ] Killing the launcher by PID leaves no orphaned `ocr` process
- [ ] `stdio: "inherit"` behavior is preserved, including for the interactive subcommands (`ocr provider`, `ocr config`)
- [ ] A test covers signal forwarding and exit-status propagation

## Context

Found while reviewing #1111, which makes the Go side handle SIGTERM gracefully. That work is correct, but it only reaches the binary when the signal is group-directed, or when `ocr` on `PATH` is the native binary (as with `install.sh`). Fixing the launcher is what makes it apply to `npm install -g` users and to the extension's cancel button.

The launcher has to forward rather than rely on process groups: both the extension and `docker stop` signal a single PID by design.

chaojixinren:
### Problem

The Node.js launcher wrapper (`bin/ocr.js`) was not forwarding signals to the Go binary child process, causing orphaned processes when the parent received SIGINT or SIGTERM.

**Reproduction:**
1. Run `npx @alibaba/open-code-review --staged` (via Node.js launcher)
2. Press Ctrl+C to interrupt
3. The launcher exits, but the Go process continues running in the background

**Root cause:**
- The launcher used `spawnSync` (synchronous/blocking call)
- When SIGINT is received, the launcher exits immediately without forwarding the signal
- The Go child process becomes orphaned (re-parented to PID 1) and continues running
- Users think the process stopped, but it's still consuming resources

### Solution

**Changes:**
- Replace `spawnSync` with `spawn` for asynchronous child process handling
- Add signal handlers for SIGINT and SIGTERM that forward signals to the child process
- Properly propagate child exit status including signal-based termination (exit code 128+N per Unix convention)
- Add integration tests to verify signal forwarding behavior
- Spawn the launcher in its own process group (`detached`) in the VS Code extension so cancellation can kill the whole CLI tree, and use `taskkill /T /F` for tree termination on Windows
- Run the launcher tests in CI (new `test:launcher` step in `ci.yml` and `action-contract.yml`)

**Flow after fix (POSIX):**
1. User presses Ctrl+C → Launcher receives SIGINT
2. Launcher forwards SIGINT to Go child process via `child.kill(signal)`
3. Go process handles the signal and exits gracefully
4. Launcher waits for child to exit and propagates the exit status

**Windows behavior:**
- Console Ctrl+C already reaches both attached processes, so the launcher keeps itself alive with a no-op SIGINT handler and lets the binary exit on its own (calling `child.kill("SIGINT")` would map to an abrupt `TerminateProcess`).
- The VS Code extension cancels via `taskkill /T /F` to terminate the whole process tree.

### Testing

**Added tests (`bin/ocr.test.js`):**
- ✅ SIGINT forwarding integration test
- ✅ SIGTERM forwarding integration test
- ✅ Exit code propagation test

**Added tests (`extensions/vscode/src/extension/services/__tests__/CliService.cancel.test.ts`):**
- ✅ POSIX group kill escalation and EPERM fallback
- ✅ Windows `taskkill` tree kill and failure fallbacks

**Verification:**
```bash
# Node.js launcher tests
node bin/ocr.test.js
# ✓ SIGINT forwarding test passed
# ✓ SIGTERM forwarding test passed
# All signal forwarding tests passed

# Go unit tests
make test
# PASS

# Go e2e tests
go test -v -race -count=1 ./cmd/opencodereview -run ".*e2e.*"
# 9/9 tests passed

# Code quality checks
make check
# check passed

# Build verification
make build
# Success
```

Checklist

- [x] Follows Conventional Commits format
- [x] Added tests to verify the fix
- [x] All tests pass (unit + e2e)
- [x] Code quality checks pass (make check)
- [x] Build succeeds (make build)
- [x] SPDX license headers included
- [x] PR is focused on a single issue

Resolves: #1142

### DIFF
--- a/.github/workflows/action-contract.yml
+++ b/.github/workflows/action-contract.yml
@@ -10,7 +10,11 @@ on:
     branches: [main]
     paths:
       - 'action.yml'
+      - 'bin/ocr.js'
+      - 'bin/ocr.test.js'
       - 'package.json'
+      - 'scripts/platform.js'
+      - 'scripts/version.js'
       - 'scripts/github-actions/**'
       - 'examples/github_actions/README.md'
       - '.github/workflows/action-contract.yml'
@@ -18,7 +22,11 @@ on:
     branches: [main]
     paths:
       - 'action.yml'
+      - 'bin/ocr.js'
+      - 'bin/ocr.test.js'
       - 'package.json'
+      - 'scripts/platform.js'
+      - 'scripts/version.js'
       - 'scripts/github-actions/**'
       - 'examples/github_actions/README.md'
       - '.github/workflows/action-contract.yml'
@@ -48,3 +56,6 @@ jobs:
 
       - name: Test GitHub Actions contracts
         run: npm run test:github-actions
+
+      - name: Test launcher contracts
+        run: npm run test:launcher
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,6 +139,9 @@ jobs:
       - name: Test
         run: go test -count=1 ./...
 
+      - name: Test Node launcher
+        run: npm run test:launcher
+
       - name: Build
         run: go build -o opencodereview.exe ./cmd/opencodereview
 
--- a/bin/ocr.js
+++ b/bin/ocr.js
@@ -5,7 +5,7 @@
 
 "use strict";
 
-const { spawnSync, spawn } = require("child_process");
+const { spawn } = require("child_process");
 const path = require("path");
 const fs = require("fs");
 const os = require("os");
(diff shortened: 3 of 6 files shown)


chaojixinren:
The changes in `extensions/vscode/src/extension/services/CliService.ts` and `extensions/vscode/src/extension/services/__tests__/CliService.cancel.test.ts` are companion changes for the VS Code cancellation path.

After `bin/ocr.js` was changed to forward SIGTERM, the existing 3-second fallback in `CliService.cancel()` could still kill only the launcher if the child process did not terminate in time, leaving the OCR process orphaned. The implementation therefore creates a dedicated POSIX process group and force-kills the entire group on timeout. On Windows, it uses `taskkill /T /F` to terminate the complete process tree. The test file covers these process-group, fallback, and Windows process-tree behaviors.

I realize these two files go beyond the issue’s explicitly stated scope of `bin/ocr.js` and `bin/ocr.test.js`. If you feel this is out of scope for this PR, please let me know.

lizhengfeng101 (approved):
LGTM

PR #1151, opened 2026-09-03
fix(pages): fix duplicate heading navigation in tools page TOC

chaojixinren:
### Problem

The tools page contains repeated section headings such as `Schema`, `Output`, and `Constraints`. These headings were assigned identical DOM IDs, causing:

- Every duplicate TOC entry to navigate to the first matching section.
- Multiple duplicate entries to appear highlighted at the same time.

### Solution

- Generate unique heading IDs per document by suffixing repeated IDs, e.g. `schema`, `schema-2`.
- Share the same ID resolution logic between TOC extraction and Markdown rendering.
- Account for headings outside the visible TOC levels so anchor IDs remain aligned.
- Add regression tests covering repeated headings and hidden heading-level collisions.

### Verification

- Frontend tests: 23 passed
- TypeScript typecheck passed
- Frontend build passed
- `make check` passed
- `make test` passed

### DIFF
--- a/pages/src/components/MarkdownRenderer.test.tsx
+++ b/pages/src/components/MarkdownRenderer.test.tsx
@@ -4,6 +4,7 @@
 import { render, screen } from '@testing-library/react';
 import { describe, expect, it } from 'vitest';
 import { LanguageProvider } from '../i18n';
+import { extractHeadings } from '../utils/extractHeadings';
 import MarkdownRenderer from './MarkdownRenderer';
 
 describe('MarkdownRenderer heading IDs', () => {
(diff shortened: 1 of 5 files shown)


lizhengfeng101:
Could you add a test that asserts `extractHeadings` and `MarkdownRenderer` produce the same ID sequence for the same markdown? Right now nothing catches it if the two walks drift apart (setext headings and `~~~` fences already do), and a one-slot drift turns every later suffixed anchor into a dead link.

chaojixinren:
@lizhengfeng101 Thanks for pointing this out. I added an integration test that compares the heading IDs produced by `extractHeadings` with the IDs rendered by `MarkdownRenderer` for the same Markdown input.

The test covers repeated headings, setext headings, and `~~~` fenced code blocks. I also updated both paths to use the same Markdown token walk, preventing future parser drift from misaligning suffixed anchor IDs.

lizhengfeng101 (approved):
LGTM

PR #1144, opened 2026-09-02
fix(tool): report tool call failures

Gongyl01:
## Description

Tool execution failures were not consistently observable in terminal output or the final JSON report.

In particular, `code_search` converted `git grep` execution errors into successful string results. With Git 2.34.1, the unsupported `--max-count` option caused `git grep` to exit with status 129, but the terminal displayed the tool call as successful and the final JSON only counted the invocation.

This PR:

- propagates actual `git grep` execution failures from `code_search`
- preserves exit code 1 with empty stderr as the normal “No matches found” result
- removes the verbose `git grep` usage text while retaining the actionable error
- records failures through the common registered-tool execution path, covering existing and future registered tools
- persists failed tool calls in session JSONL with `ok: false`
- adds `failure`, `failure_by_tool`, and ordered `failure_details` to the final JSON output
- keeps tool arguments out of failure details

Example JSON output:

```json
"tool_calls": {
  "total": 5,
  "by_tool": {
    "code_search": 2,
    "file_read": 3
  },
  "failure": 1,
  "failure_by_tool": {
    "code_search": 1
  },
  "failure_details": [
    {
      "tool_call_number": 2,
      "tool_name": "code_search",
      "file_path": "src/Base/Parameter.cpp",
      "error": "git grep failed: exit status 129: error: unknown option `max-count'"
    }
  ]
}
```

## Testing

- `make check`
- `make test`
- Manually reproduced and verified the failure with Git 2.34.1.

PR #1152, opened 2026-09-03
feat(tool): include arguments in failure details

Gongyl01:
## Description

#1144 made registered-tool execution failures observable in terminal output, session JSONL, and the final JSON report.

However, each `failure_details` entry only identifies the failed tool call and its error. It does not include the arguments that triggered the failure, so JSON consumers must correlate the failure with session records to reconstruct the original input.

This PR:

- adds `arguments` to `ToolFailureDetail`
- captures the raw JSON-encoded arguments directly from `llm.ToolCall.Function.Arguments`
- includes the arguments in both normal result JSON and failure-usage JSON
- preserves empty and malformed argument payloads without parsing or re-serializing them
- verifies argument propagation through registered-tool failures, shared result output, failure output, and session persistence
- leaves text and SARIF output unchanged because they do not expose `failure_details`

Example JSON output:

```json
"failure_details": [
  {
    "tool_call_number": 2,
    "tool_name": "code_search",
    "file_path": "src/Base/Parameter.cpp",
    "arguments": "{\"search_text\":\"BaseParameter\",\"file_patterns\":[\"src/\"]}",
    "error": "git grep failed: exit status 129: error: unknown option `max-count'"
  }
]
```

The arguments are intentionally preserved as the raw string returned by the LLM. They are not normalized, redacted, or truncated, matching the argument representation already stored in session JSONL. The field remains present when the raw value is empty because an empty argument payload is itself useful failure context.

## Testing

- `make check`
- `make test`

### DIFF
--- a/cmd/opencodereview/budget_output_test.go
+++ b/cmd/opencodereview/budget_output_test.go
@@ -195,6 +195,7 @@ func TestEmitFailureUsage_JSONEmitsStructuredRecord(t *testing.T) {
 			ToolCallNumber: 1,
 			ToolName:       "file_read",
 			FilePath:       "missing.go",
+			Arguments:      `{"file_path":"missing.go"}`,
 			Error:          "file not found",
 		}},
 	}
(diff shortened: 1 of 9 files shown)


lizhengfeng101 (approved):
LGTM

PR #1157, opened 2026-09-03
feat(action): expose effort and max_tokens_budget inputs

### LINKED ISSUE #1147: feat(action): expose --effort and --max-tokens-budget as action inputs
### Problem Statement

The composite action builds the `ocr review` command line itself and exposes only a subset of the CLI's knobs (`--concurrency`, `--timeout`, `--background`, `--rule`). Two review-shaping flags are unreachable from GitHub Actions:

- `--effort low|medium|high` selects the review-round preset (`MAX_REVIEW_ROUNDS` = 1/2/3). Action users are locked to the `medium` default of two rounds per file; on large diffs the second pass re-reads each file from scratch, so the round count is the biggest single lever on token consumption and wall-clock time.
- `--max-tokens-budget N` caps aggregate token usage across the whole review: dispatch stops once exceeded, partial results still publish, and the run exits 0. It is the only built-in circuit breaker against runaway token spend, and action users cannot set it at all — the review command populates the budget solely from the CLI flag (`review_cmd.go`), with no config key or env-var fallback, so there is no workaround.

### Proposed Solution

Add two optional inputs to `action.yml` and pass them through when set:

- `effort`: `low` | `medium` | `high`; empty keeps the CLI default. Either append `--effort "$VALUE"` in the Run step, or write it via `ocr config set effort` in the Configure OCR step (the config-key fallback already exists and merges by key).
- `max_tokens_budget`: non-negative integer; empty or 0 = unlimited; when non-empty, append `--max-tokens-budget "$VALUE"` to the review args.

Both values change what a run covers and costs, so they should join the checkpoint config fingerprint alongside `llm_timeout` — a budget change alters which runs complete vs. stop early, the same reasoning the fingerprint already applies to timeouts.

### Alternatives Considered

- Pre-seeding `effort` into `~/.opencodereview/config.json` before the action runs works today (the action's `ocr config set` calls merge by key), but it is an undocumented side channel that is invisible in the workflow file.
- For `--max-tokens-budget` there is no workaround short of bypassing the action's Run step entirely.
- Forking the action to add two flag pass-throughs.

### Affected Area

Other (GitHub composite action)

### Additional Context

Happy to send a PR; both changes are a few lines in `action.yml` plus input docs.

ylcn91:
## Description

The composite action builds the `ocr review` command itself and exposes only `--concurrency`, `--timeout`, `--background` and `--rule`, so two flags were unreachable from a workflow: `--effort`, which picks the review-round preset, and `--max-tokens-budget`, the only aggregate cap on a review's token spend (`review_cmd.go` gives it no config or environment fallback).

This adds both as optional inputs, appended to the review args only when non-empty, so an unset input keeps the CLI default: the configured effort (otherwise medium) and an unlimited budget. An explicit `0` budget is forwarded rather than dropped and means unlimited, matching the flag's own documentation.

Both are validated before the npm install, so a typo fails with an annotation in seconds instead of after the whole review setup. `effort` accepts low/medium/high in any case, because `ocr review` lower-cases the preset itself; the budget must be a non-negative integer without leading zeros, because the CLI parses integer flags with base-0 `strconv`, where `010` would silently mean 8.

`effort` also joins the checkpoint fingerprint: a checkpoint recorded by a low-effort run must not narrow a later high-effort one. `max_tokens_budget` deliberately stays out of it, since a run that hits its budget is partial and only a complete run should advance a checkpoint.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- `npm run test:github-actions`: action-contract cases for both inputs (unset, set, `0` budget, invalid values, case handling of `effort`) and the fingerprint change in `post-review-comments.test.js`.
- `make test` on this branch (Go packages plus the node script tests).
- Only `action.yml`, the scripts under `scripts/github-actions/` and `examples/github_actions/README.md` change; there is no Go change.

## Checklist

- [ ] I have signed the CLA

## Related Issues

Closes #1147

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -67,6 +67,22 @@ inputs:
   rule:
     description: Path to a custom rules JSON file passed to `ocr review --rule`.
     required: false
+  effort:
+    description: >-
+      Review effort preset passed to `ocr review --effort`: low, medium or
+      high, case-insensitively. It selects how many review rounds the agent
+      runs. Unset or empty appends no flag and keeps the CLI's own resolution,
+      which is the configured effort or otherwise medium.
+    required: false
+  max_tokens_budget:
+    description: >-
+      Aggregate token cap (input + output) for the whole review, passed to
+      `ocr review --max-tokens-budget`. A non-negative base-10 integer with no
+      leading zeros: dispatch stops once the budget is exceeded, the files that
+      were skipped are reported as failed(budget), and the partial result is
+      still published. 0 means unlimited, and so does leaving it unset or
+      empty, which appends no flag at all.
+    required: false
   upload_artifacts:
     description: >-
       Upload raw JSON result and stderr as workflow artifacts. Must be the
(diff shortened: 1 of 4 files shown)


lizhengfeng101:
Superseded by #1154, which landed `effort` and `max_tokens_budget` and closed #1147. Closing as duplicate — thanks for the contribution!

Issue #1141, opened 2026-09-02
fix(review): make cancellation shutdown atomic and time-bounded


lizhengfeng101:
## Description

`ocr review` handles SIGINT — and SIGTERM once #1111 lands — by cancelling the context so the run closes through its normal defer chain. The core of that path is solid: `finalizeManifest()` and `session.Finalize()` run inside `ag.Run` (`internal/agent/agent.go:390-396`), before any cleanup that could stall, so `--resume` stays usable no matter what happens later in the shutdown.

Three narrower gaps remain in the shutdown tail. All three predate #1111 — that PR only widens the set of scenarios that reach them, from an interactive Ctrl-C to `docker stop`, CI cancellation and editor cancel buttons.

### 1. The `--output` report is not written atomically

`lazyFileWriter` opens the target with `os.Create` on the first write (`cmd/opencodereview/shared.go:504`), truncating in place. The type exists precisely so a run that produces no output leaves an existing file untouched, so the effective guarantee for a target path today is "complete report, or previous content". That guarantee disappears the moment the first byte lands: a write failure, or a SIGKILL arriving inside the emit window, leaves a truncated JSON or SARIF document at a path other tools read. A truncated SARIF is the worst case, since consumers treat one upload as the complete alert set.

Writing to a temp file in the same directory and renaming on `Close` restores the guarantee for every failure mode.

This also settles a user-visible question that is currently decided by the open mode rather than on purpose: a cancelled run's partial-coverage report replaces a previous complete report at the same path. That may well be the desired behavior — it should just be a decision.

### 2. Shutdown can take ~15s per stdio MCP server, sequentially

`review_cmd.go` closes MCP clients in a `defer` loop, one at a time. Each `Close` reaches the SDK's `pipeRWC.Close` (`modelcontextprotocol/go-sdk@v1.7.0/mcp/cmd.go:69`), which is bounded but generous: close stdin, wait `TerminateDuration`, send SIGTERM, wait again, send SIGKILL, wait again. `internal/mcp/client.go:41` constructs `CommandTransport` without setting `TerminateDuration`, so the SDK default of 5s applies at each of the three stages — up to ~15s for one unresponsive server, multiplied by the number of configured stdio servers.

[...]

PR #1160, opened 2026-09-03
fix(output): write reports atomically

dvd233:
## Description

`lazyFileWriter` currently truncates an existing `--output` file as soon as the first byte is written. A write failure or forced termination during rendering can therefore leave JSON/SARIF consumers with a partial document at a path that previously held a complete report.

This change restores the old-or-new file guarantee by:

- writing to a collision-resistant temporary file in the target directory;
- preserving `os.Create`-style permissions for new files and the mode of an existing target;
- preserving write-through behavior for symlinked output paths;
- syncing and closing the temporary file before atomically renaming it into place;
- removing temporary files after write, sync, close, or rename failures; and
- printing `Results written to` only after the rename succeeds.

The cancellation policy is unchanged: a successfully rendered cancelled-run report can still replace an older report. This PR only addresses atomic persistence.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Validated on Windows/amd64 with Go 1.25.5:

- `make fmt`
- `make check`
- `make test` (full race-enabled suite)
- `make build`
- `make coverage` (90.5% total; 90% threshold)
- real `ocr review --preview --format json --output` replacement, including JSON parsing and success-hint verification
- forced process termination after writing the temporary file but before `Close`; the previous target remained byte-for-byte intact and only the identifiable `.ocr-out-*` temporary file contained the uncommitted bytes

Tests cover commit-on-close, replacement timing, new/existing file modes, rename failure, injected write failure, symlink preservation, ANSI stripping, unchanged JSON bytes, and temporary-file cleanup.

## Checklist

- [x] I have updated the documentation accordingly (not applicable: no CLI syntax or configuration changed)
- [ ] I have signed the CLA

## Related Issues

Addresses item 1 of #1141.


cmd/opencodereview/shared.go
```diff
@@ -589,22 +674,54 @@ func writeOutError(out io.Writer) error {
...
+	}
+	if err := w.file.Close(); err != nil {
+		cleanupErr := cleanupOutputTemp(w.file, w.tempPath)
+		w.file = nil
+		w.tempPath = ""
+		w.closeErr = errors.Join(fmt.Errorf("close output file %s: %w", w.path, err), cleanupErr)
+		return w.closeErr
+	}
+	w.file = nil
+	if err := os.Rename(w.tempPath, w.commitPath); err != nil {
```
wu21-web, line 708:
```suggestion
	if err := replaceOutputFile(w.tempPath, w.commitPath); err != nil {
```

wu21-web (commented):
@dvd233

lizhengfeng101:
The temp+rename approach is right, but `defer closeOut()` (review_cmd.go:118) registers first and therefore runs last — so the rename now happens *after* the MCP shutdown loop, widening the SIGKILL window from milliseconds to the ~15s/server that issue #1141 item 2 says will overrun `docker stop`, which turns a rare truncated report into a routinely lost one (and invalidates that item's "the report [is] already on disk" premise); calling `closeOut()` explicitly right after `emitRunResult` succeeds and keeping the defer as a fallback fixes it for free, since Close is already idempotent.

dvd233:
Thanks for catching the defer ordering. Fixed in [`7ae6442`](https://github.com/alibaba/open-code-review/commit/7ae6442c8b0e34d88804836ba605fb83060a8a8e):

- `finishOutput` now commits the report immediately after `emitRunResult` succeeds, before deferred MCP shutdown.
- The defer remains as an exact-once fallback for early returns and emit failures; close errors still propagate.
- Added `TestReviewOutputCommittedBeforeMCPShutdown`, which reads and JSON-decodes the output at the MCP shutdown boundary, so the old defer order fails deterministically.

Validation:

- `make check`
- `make test` (`-race`)
- `make coverage` — 90.4% (90% threshold)
- `make build`
- all current GitHub CI and CodeQL checks pass

cmd/opencodereview/shared.go
dvd233: Thanks. `replaceOutputFile` doesn't currently exist in this package or elsewhere in the repository. Is there a specific implementation or platform behavior you want it to provide beyond `os.Rename`? Go's Windows implementation calls `MoveFileEx` with `MOVEFILE_REPLACE_EXISTING`, and `TestLazyFileWriter_ReplacesExistingFileOnlyOnClose` passes in the current Windows CI job. I can add a wrapper if it serves a concrete semantic or testing purpose, but I don't want to introduce a pass-through helper based on the name alone.

lizhengfeng101 (approved):
LGTM

PR #1162, opened 2026-09-04
Honor timeout_sec for openai-go's ResponseHeaderTimeout (#1161)

### LINKED ISSUE #1161: model config timeout_sec does not override openai-go's ResponseHeaderTimeout
### OpenCodeReview Version

1.11.3

### Operating System

Linux (ARM64)

### Installation Method

GitHub Release binary

### LLM Provider

### Bug Description

If the connected endpoint has slow prefill (or deode if streaming is not configured) this timeout will cause a retry even if the request is still being processed. Separately of timeout_sec.

Configuration

{
  "provider": "litellm",
  "model": "GLM-5.3",
  "providers": {
    "litellm": {
      "api_key": "<redacted>",
      "url": "https://llm-proxy.example.com",
      "protocol": "openai",
      "model": "GLM-5.3",
      "timeout_sec": 360000
    }
  }
}

OCR_LLM_TIMEOUT was not set.

Session entry:

{
  "taskType": "plan_task",
  "request_no": 1,
  "duration_ms": 3614491,
  "error": "Post \"https://llm-proxy.example.com/chat/completions\": http2: timeout awaiting response headers"
}

The duration matches six ten-minute attempts plus retry backoff:

3614.491 seconds ≈ 6 × 600 seconds + backoff

Root cause

OCR passes timeout_sec through openaiopt.WithRequestTimeout, but the SDK also installs an independent HTTP transport timeout:

const defaultResponseHeaderTimeout = 10 * time.Minute

OCR does not override the SDK’s HTTP client. It also configures WithMaxRetries(5), resulting in six attempts.

Suggested fix

Provide a custom HTTP client through openaiopt.WithHTTPClient and derive ResponseHeaderTimeout from cfg.Timeout, or set it to zero and rely on the request context. This should cover both Chat Completions and Responses clients.

### Steps to Reproduce

Non-streaming or slow  prefill endpoint configured with a long enough timeout_sec that it wont trigger.

Scanned repo must include files large enough to not be processed within the 600s window.

ocr scan --timeout 0 

The planning request among others failed with:

http2: timeout awaiting response headers

### Expected Behavior

Expected behavior

An explicit timeout_sec should also permit waiting that long for response headers, or OCR should expose a separate response-header timeout.

### Logs / Error Output

```shell
http2: timeout awaiting response headers
```

### Additional Context

Relation to PR #1159

PR #1159 addresses the five-minute request default when no environment or provider timeout is configured. This case already has an explicit timeout_sec, and the PR does not modify the SDK’s separate ten-minute transport timeout.

thejdubb02:
## Description

The OpenAI and OpenAI Responses clients set `WithRequestTimeout(cfg.Timeout)` but never supplied an HTTP client, and `NewAnthropicClient` had the same gap: both openai-go and anthropic-sdk-go hardcode a 10-minute `ResponseHeaderTimeout` in their default HTTP client (`default_http_client.go` in each), applied unless a client is supplied via `WithHTTPClient`. So on a slow-prefill endpoint the response headers timed out at 10 minutes regardless of `timeout_sec` (#1161).

The impact was worse than one stall: the 10-minute header timeout surfaces as `res == nil`, which `shouldRetry` treats as retryable, so a slow endpoint burned up to six attempts of 10 minutes each. For the `timeout_sec: 900` example in `pages/src/content/docs/en/configuration.md` that is roughly 60 minutes of dead time instead of 15.

This supplies an HTTP client whose `ResponseHeaderTimeout` is `cfg.Timeout` plus a small margin, via `WithHTTPClient`, in `NewOpenAIClient`, `NewOpenAIResponsesClient` and `NewAnthropicClient`. The margin keeps the request context deadline (`WithRequestTimeout`) the first to fire, so a real timeout returns via `ctx.Err()` with no retry. The shared helper clones `http.DefaultTransport` and leaves a wrapped transport untouched, mirroring each SDK's own default-client construction.

`NewAnthropicBedrockClient` is deliberately left alone (with a comment): `bedrock.WithConfig` carries `WithoutEnvironmentDefaults`, so `NewClient` skips `DefaultClientOptions` and never installs the 10-minute default there; adding the override would impose a new cap and pull Bedrock out of the shared connection pool.

## Type of Change

## How Has This Been Tested?

- [x] `make test` passes locally (with `-race`)
- `TestHTTPClientWithHeaderTimeout` asserts the helper sets `ResponseHeaderTimeout` to the timeout plus the margin on a cloned transport (not the shared `http.DefaultTransport`), and leaves it unset for a non-positive timeout.
- `TestConstructorsInstallHeaderTimeoutClient` asserts each of the three constructors actually installs the client with `cfg.Timeout`, so removing a `WithHTTPClient` line fails the suite rather than passing on the helper's isolated test. A timing-based test is impractical (the 10-minute default cannot fire in a fast test). `go build ./...`, `go vet`, and `gofmt` are clean.

## Checklist

- [ ] I have updated the documentation accordingly (no docs change needed)
- [x] I have signed the CLA

## Related Issues

Closes #1161


lizhengfeng101:
Diagnosis and fix look right — verified that `defaultHTTPClient()` is wired into `DefaultClientOptions` on the normal path, and that caller options are appended after defaults, so the override lands.

Two things before merging:

**1. `NewAnthropicClient` has the same bug.** `anthropic-sdk-go` hardcodes the identical 10-minute `ResponseHeaderTimeout` (`client.go:54` → `default_http_client.go:14`), and `client.go:930` only passes `WithRequestTimeout`. So `timeout_sec > 600` is still capped there. Worth fixing in this PR — the helper returns a plain `*http.Client`, so it's one extra line.

**2. `NewAnthropicBedrockClient` does *not* need it.** `bedrock.WithConfig` carries `option.WithoutEnvironmentDefaults()`, and `HasWithoutEnvironmentDefaults` sees through `Join` (`option/requestoption.go:514`), so `NewClient` skips `DefaultClientOptions` entirely and `HTTPClient` stays `http.DefaultClient` — no header timeout at all. Adding one there would be a new restriction, not a fix. A short comment at `client.go:990` saying why it's absent would save the next reader the same dig.

**Minor:** setting `ResponseHeaderTimeout` exactly equal to the request timeout makes both deadlines expire together. If the header timeout wins the race it surfaces as a connection error with `res == nil`, which `shouldRetry` treats as retryable — up to 5 more attempts, each with a fresh full timeout, instead of failing immediately on `ctx.Err()`. `timeout + 30*time.Second` keeps the context deadline deterministically first; the test assertion would move to `want + margin`.

thejdubb02:
Thanks for the careful review. All three addressed in the new commit.

1. NewAnthropicClient: fixed. Confirmed the same path: anthropic-sdk-go's DefaultClientOptions installs defaultHTTPClient() (default_http_client.go, the same 10-minute ResponseHeaderTimeout), and NewClient applies it here because this constructor passes no WithoutEnvironmentDefaults. It now uses the shared helper.

2. NewAnthropicBedrockClient: left as-is, with a comment at that constructor explaining why. bedrock.WithConfig returns an option.Join carrying WithoutEnvironmentDefaults, HasWithoutEnvironmentDefaults sees through the Join, so NewClient takes the WithEnvironmentProduction branch and never installs the default client. Adding the override there would impose a new cap rather than remove one.

3. Margin: adopted. ResponseHeaderTimeout is now the request timeout plus a 30s margin, so the WithRequestTimeout context deadline is the first to fire and a header-timeout win no longer surfaces as a retryable nil-response error. The test assertion moved to want + margin. I also left ResponseHeaderTimeout unset when the timeout is zero or less, so a no-timeout config is not silently bounded (the callers clamp to a positive value first, so this only guards a direct call).

Renamed openAIHTTPClient to httpClientWithHeaderTimeout since the Anthropic path shares it now. go build, go vet and go test ./internal/llm/ all pass.

lizhengfeng101:
Nice work on round two — all three points from the earlier review landed cleanly, and leaving Bedrock alone with a comment is the right call. I re-checked that path: `HasWithoutEnvironmentDefaults` recurses through `joinedOption` (`option/requestoption.go:519`), so `NewClient` skips `DefaultClientOptions` and `HTTPClient` falls back to `http.DefaultClient` — there's no header timeout there to remove, and cloning the transport would also pull Bedrock out of the shared connection pool. Good instinct.

Two things before merge, plus two nits.

**1. The description now contradicts the diff.**

The body still says the Anthropic client "is out of scope for this issue," but the diff fixes `NewAnthropicClient` — which was the first review's main request. It also names the test `TestOpenAIHTTPClientHonorsTimeout`, now `TestHTTPClientWithHeaderTimeout`. This gets squashed into the commit message, so it would land in history denying what the commit does.

While you're editing it: the description undersells the fix. It's not only that `timeout_sec` was ignored — the 10-minute header timeout surfaced as `res == nil`, which `shouldRetry` treats as retryable, so a slow endpoint burned six attempts of 10 minutes each. With the margin, the context deadline fires first and the `ctx.Err()` check at `requestconfig.go:594` returns immediately with no retry. For the `timeout_sec: 900` example in `pages/src/content/docs/en/configuration.md` that's roughly 60 minutes down to 15. Worth saying out loud.

**2. Nothing tests that the constructors actually install the client.**

The test exercises the helper in isolation, so deleting the `WithHTTPClient` line from `responses_client.go` would keep the suite green. That is precisely the failure mode this PR already hit once: round one missed `NewAnthropicClient`, tests passed, and a human caught it. I'm not asking for a timing test — you're right that it isn't practical. Cheapest version is to make the helper a package-level var and assert each of the three constructors calls it with `cfg.Timeout`.

**3.** Nit: the helper lives under `// --- OpenAIClient ---`, but `NewAnthropicClient` uses it too. Renaming `openAIHTTPClient` to `httpClientWithHeaderTimeout` acknowledged it's shared; the placement didn't follow. `bedrockConfigLoadTimeout` (`client.go:44`) is a better neighbour.

**4.** Nit: `retry_boundary_test.go` is a more natural home than `client_test.go` — `TestBoundaryRetryAfterOutlivingAttemptTimeoutIsFailed` already covers the per-attempt-timeout-versus-retries semantics.

thejdubb02:
Thanks, all four addressed.

1. Description: rewritten. It no longer says the Anthropic client is out of scope (the diff fixes NewAnthropicClient), the test name is corrected, and I added the retry-storm framing: the header timeout surfaces as res == nil, shouldRetry retries it, so timeout_sec: 900 was roughly 60 minutes of dead time instead of 15. Bedrock's exemption is stated too.

2. Constructor wiring test: added TestConstructorsInstallHeaderTimeoutClient. httpClientWithHeaderTimeout is now a package-level var, and the test spies on it and asserts each of NewOpenAIClient, NewOpenAIResponsesClient and NewAnthropicClient calls it exactly once with cfg.Timeout. Deleting a WithHTTPClient line now fails the suite, which is the round-one miss you called out.

3. Placement: the const and the helper moved up next to bedrockConfigLoadTimeout, out of the OpenAIClient section, since all three constructors share it.

4. Test home: the helper test moved to retry_boundary_test.go next to TestBoundaryRetryAfterOutlivingAttemptTimeoutIsFailed.

make test (with -race), go build, go vet and gofmt are clean.

lizhengfeng101 (approved):
LGTM

Issue #1155, opened 2026-09-03
`ocr review --timeout` does not extend the five-minute LLM request deadline


acoliver:
## Environment

- `open-code-review v1.11.3` (`b3badd376`), darwin/arm64
- JSON agent-mode range review
- One selected Rust file, `zai-anthropic` provider, `glm-5.3`, concurrency 1

## Reproduction

Run a single-file review with an outer timeout larger than five minutes:

```sh
ocr review \
  --audience agent \
  --format json \
  --provider zai-anthropic \
  --model glm-5.3 \
  --concurrency 1 \
  --timeout 30 \
  --from BASE \
  --to HEAD \
  --exclude 'all,other,changed,files'
```

The selected file was 291 lines. Planning completed on the first run, but the main request failed after exactly five minutes. The manifest reported:

```json
{
  "classification": "timeout",
  "reason": "file review exceeded its time limit",
  "duration_to_headers_ms": 300002
}
```

Resuming the session with the same `--timeout 30` started planning again rather than reusing the completed plan. That planning request also failed at exactly five minutes before OCR moved to the main request.

Grouped reviews showed the same `duration_to_headers_ms` of about 300,000 ms despite `--timeout 600`.

## Expected behavior

`--timeout`, documented in minutes as the concurrent task timeout, should either:

1. apply to the provider request deadline needed to complete that task, or
2. document the independent five-minute provider-request ceiling and retry requests that time out before the task deadline.

A resume should reuse a completed plan when the file's main review was the failed stage.

## Impact

A review can repeatedly fail to cover one changed file even when the caller grants a larger task timeout. The JSON correctly marks coverage as failed, but there is no CLI setting that allows the file review to finish.

zhuangzhewei09:
Hi, I'd like to investigate the timeout-handling part of this issue. I checked the reported revision and found that per-request timeouts are already configurable through OCR_LLM_TIMEOUT and provider timeout_sec, independently of --timeout. I'll first reproduce the request-versus-task deadline behavior and verify timeout retries, then propose a focused fix with regression tests and clearer CLI diagnostics/help. In particular, a request timeout may currently be reported as the file's task deadline being exceeded. I'd keep completed-plan reuse as a separate follow-up. Is anyone already working on this scope, and do maintainers prefer keeping the two timeout controls independent or changing how --timeout interacts with the request deadline?

PR #1173, opened 2026-09-05
feat(agent): add --on-grouping-failure to stop a review on a failed grouping (#1164)

### LINKED ISSUE #1164: Let a grouping failure stop the review instead of falling back to per-file dispatch
## Problem Statement

When the grouping LLM call fails, OCR falls back to one review subtask per file and reports it with a single stdout line (`internal/agent/grouping.go:95-99`):

```go
groups, usage, err := callGroupingLLM(...)
if err != nil {
    fmt.Fprintf(stdout.Writer(), "[ocr] LLM grouping failed (%v), falling back to per-file dispatch\n", err)
    return groupDiffsResult{groups: toSingleFileGroups(diffs), usage: usage}
}
```

No run through the GitHub Action can see that line. `action.yml:668` always passes `--audience agent`, which routes `[ocr]` progress to `io.Discard` through `stdout.Quiet()` (`shared.go:338-346`, `internal/stdout/stdout.go:33-37`). The flag is not an action input, so no user can change it.

Nothing else states what happened:

- There is no `grouping.failed` telemetry event to pair with the existing `grouping.skipped` (`grouping.go:158-165`).
- The `--json` result has no field naming the strategy that produced `groups`.
- The `groups` array narrows it and cannot name it. A grouping failure, a `GroupingPerFile` plan, and a template with no `GROUPING_TASK` all emit one group per file, labelled with its path (`grouping.go:373-382`). An `enforceGroupTokenBudget` split is the one case the labels do separate, because it appends `" (split: <path>)"` (`grouping.go:351`).
- `retry_report` does carry the failed grouping request. Its listing rule lists any logical request that retried, saw an error, or did not succeed, whatever the task type (`internal/llm/retry_report.go:535-542`). In our incident it was the only trace that named the failure. It proves the call failed, and it does not say what OCR then decided.

The fallback is expensive, and a caller pays for all of it before it can react. Our run turned a 188-file change set into 188 review conversations, and each one is entitled to `--timeout` minutes and `MAX_REVIEW_ROUNDS` rounds. The `ocr review` call alone took 47m40s inside a `timeout-minutes: 50` job, so under 2m20s was left for every other step, and it still exited 0. The result JSON arrives too late to help, because `shared.go:696-708` writes it after the agent run returns.

## Proposed Solution

1. Add a policy flag, `--on-grouping-failure=fallback|abort`. With `abort`, review stops before it dispatches any subtask, and it exits non-zero. Keep `fallback` as the default so current behaviour does not change. `--max-tokens-budget` shows that a flag can already carry an exit-code contract in its help text (`cmd/opencodereview/shared_flags.go:58`), so the surface exists. Its own rule points the other way: it exits 0 on a partial run.
2. Report the partition in the result JSON. A `grouping` block would carry the strategy behind `groups` (`llm`, `bundle_all`, `per_file`) and, when the LLM call failed, the reason. `GroupingStrategy.String()` already produces those names. This one is for diagnosis after a run, not for prevention.
3. Emit a `grouping.failed` telemetry event next to `grouping.skipped`, with the same attribute set plus the error class.

Item 1 is the one we need. A CI job cannot stop a fan-out it never learns about, and a non-zero exit is all our workflow needs to fail fast.

## Alternatives Considered

- Read the stdout line. It does not survive `--audience agent`, which the action always passes, and a log string carries no stability contract.
- Write the grouping-failure line to stderr. OCR already sends `[ocr]` diagnostics there whatever the audience (`internal/telemetry/events.go:125`, `cmd/opencodereview/output.go:88`), and the action captures stderr to a file and prints it, so a one-line change at `grouping.go:97` would surface it. It still leaves a workflow gate parsing a log line, and it still pays for the whole fan-out first.
- Infer the failure from `retry_report`. It reports the request, not the decision: a failed `grouping_task` entry proves the call failed, and it still leaves the resulting partition unstated. It also arrives only after the run, which is the same problem the result JSON has.
- Cancel the CI run from the workflow. The workflow cannot see the fallback, so there is nothing to trigger the cancel on. With a non-zero exit there is also nothing left to cancel.
- Set `--max-tokens-budget` low enough to cut a per-file fan-out short. That truncates coverage instead of preventing a partition we did not ask for. The undispatched files land in `coverage.failed` as `budget`, so `status` reads `partial` and our gate does catch it, but only after we pay for the part that ran. It is also out of reach from the action today, since #1147 is still open on exposing it as an input.

## Affected Area

- Review Agent / LLM interaction
- CLI / Commands
- Telemetry / Observability

## Additional Context

Read on `v1.11.1`, the version this run used: the workflow pins `alibaba/open-code-review@v1.11.1` and passes `ocr_version: "1.11.1"`. Commit `8d023aa` is the same commit as that tag, so every line number above matches the version that produced the failure. We run OCR through the GitHub Action on a label, with a gate step that reads the result JSON and fails the job unless the review finished. That gate is why the fallback matters to us: today it can turn a degraded review into a green check.

### The run this comes from

The run is on a private GitHub organisation, so there is no shareable URL. The uploaded `ocr-result.json` holds the whole story. OCR reviewed 188 files as 188 groups of one file each, which is the per-file fallback. It cost 21.4M tokens and 47m40s and produced 5 comments, and `status` reads `complete`, so our gate step passed a degraded review. The `retry_report` records exactly one failure out of 1357 requests:

```json
{
  "model": "qwen/qwen3.8-max",
  "file_path": "__grouping__",
  "task_type": "grouping_task",
  "request_no": 1,
  "outcome": "failed",
  "attempts": [
    { "attempt": 1, "outcome": "error", "error_class": "timeout",
      "failure_phase": "context", "status_code": 200,
      "duration_to_headers_ms": 1308 }
  ]
}
```

That block is the only trace that names the failure. The 188 single-file groups show the shape and not the cause. It proves the grouping call failed, and it does not say that the partition became per-file. Whether the grouping call should also retry or carry its own deadline is a separate question, and we are not raising it here.

chethanuk:
Closes alibaba/open-code-review#1164.

## Description

When the LLM grouping call fails, `ocr review` prints a line and reviews every
file on its own. The run then reports `terminal_state=complete` and exits `0`,
so in CI a degraded review is indistinguishable from a clean one. Under
`--audience agent` — which `action.yml` passes — even the printed line goes to
`io.Discard`, so nothing at all records that the partition failed.

`groupDiffsResult` (`internal/agent/grouping.go`) carried `groups` and `usage`
and no failure field, so the error `callGroupingLLM` returns could not leave
`groupDiffs`. Its one caller, `dispatchSubtasks`, fanned out the per-file groups
with nothing to check.

`groupDiffsResult` now carries the failure alongside the fallback groups. It is
set only on the LLM error path, so the three paths that decide a partition
without a round trip — one file, no grouping task, below the thresholds — stay
successes and can never trip an abort.

`dispatchSubtasks` applies a new `--on-grouping-failure` policy flag:

- `--on-grouping-failure=fallback` (default, and the zero value existing
  `agent.Args` callers pass) reviews per file as before, but records a
  `grouping_failed` warning, which reaches text, JSON and SARIF output through
  `Warnings()`.
- `--on-grouping-failure=abort` records a `run_failure` and returns the error,
  so the run ends `failed` with exit `1`.

`abort` is not the default deliberately: a transient grouping timeout would
otherwise redden every CI run.

Three details worth calling out:

- The abort path checks `ctx.Err()` first. `run_failure` is first-cause-wins and
  this return is upstream of the existing post-dispatch context gate, so without
  the check a Ctrl-C or an elapsed deadline during grouping would be filed as
  `unknown` instead of `cancelled`/`timeout`.
- `ctx.Err()` alone isn't enough, though: `groupResult.failure` can also wrap
  `context.DeadlineExceeded`/`context.Canceled` from the *per-request* timeout
  (`option.WithRequestTimeout`), which is scoped below the outer `ctx` and
  leaves `ctx.Err()` nil. That case is now recognized with `errors.Is` on the
  wrapped error — the same pattern `classifyItemError` already uses for the
  per-file path — instead of falling through to `unknown`.
- The remaining catch-all class is `unknown` rather than a new
  `RunFailureGrouping`. The enum is documented as closed on a frozen v1
  manifest schema, so widening it is a maintainer's call, not something a bug
  fix should do.

A grouping failure also now emits a `grouping.failed` telemetry error event,
carrying `grouping.skipped`'s attribute set so the two are comparable.

The flag is registered on `review` only — `ocr scan` never calls `groupDiffs`.
Documented in all five locales of `cli-reference.md`, with the existing exit-code
`1` row extended rather than a new one added.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Both new test files fail to **compile** without the fix, on the fields it adds:

```
$ go vet ./internal/agent/ ./cmd/opencodereview/   # tests present, fix absent
vet: internal/agent/grouping_test.go:788:14: result.failure undefined (type groupDiffsResult has no field or method failure)
vet: cmd/opencodereview/flags_test.go:272:12: opts.onGroupingFailure undefined (type reviewOptions has no field or method onGroupingFailure)
```

A per-request timeout wrapped in `groupResult.failure` while the outer `ctx`
is still live used to be misclassified as `unknown`; a new table-driven test,
`TestDispatchSubtasks_OnGroupingFailureAbort_PerRequestTimeout`, fails without
that fix and passes with it:

```
$ go test -count=1 -run 'TestDispatchSubtasks_OnGroupingFailureAbort_PerRequestTimeout' -v ./internal/agent/...  # fix reverted
    grouping_test.go:1010: run failure = &{Classification:unknown ...}, want nil (pending cause, not a run failure)
    grouping_test.go:1021: item ... classification = "unknown", want "timeout" (not "unknown")
--- FAIL: TestDispatchSubtasks_OnGroupingFailureAbort_PerRequestTimeout (0.41s)

$ go test -count=1 -run 'TestDispatchSubtasks_OnGroupingFailureAbort_PerRequestTimeout' -v ./internal/agent/...  # fix restored
--- PASS: TestDispatchSubtasks_OnGroupingFailureAbort_PerRequestTimeout (1.07s)
```

Six new test functions, 18 subtests, all passing with the race detector on:

```
$ go test -race -count=1 ./internal/agent/... ./cmd/opencodereview/...
ok  	github.com/alibaba/open-code-review/internal/agent	330.118s
ok  	github.com/alibaba/open-code-review/cmd/opencodereview	115.536s
```

```
$ make check
All source files have valid license headers.
No unapproved non-English text in 466 scanned source files.
check passed
```

The coverage gate, baselined on a clean checkout of `af7f922` first (91.0%):

```
$ make coverage
total:	(statements)	91.1%
PASS: Coverage 91.1% meets 90% threshold
```

Neither new function is uncovered — `emitGroupingFailed` and
`validateOnGroupingFailure` are both at 100.0%.

End to end through the CLI:

```
$ ocr review --help | grep on-grouping-failure
      --on-grouping-failure string   what a failed LLM grouping call does: fallback (review each file alone) or abort (stop with exit 1) (default "fallback")

$ ocr review --on-grouping-failure skip --commit HEAD; echo "exit=$?"
Error: invalid --on-grouping-failure value "skip": must be 'fallback' or 'abort'
exit=1

$ ocr scan --help | grep -c on-grouping-failure
0
```

## Known limitation

A semantically empty grouping response — `[]`, or paths matching no diff —
returns nil from `parseGroupingResponse` and gets per-file groups without an
error, so there is nothing for `abort` to catch there. That is a separate issue.

## Checklist

- [ ] I have signed the CLA — pending: sign when the bot prompts on the actual upstream PR

## Related Issues

Closes #1164.


lizhengfeng101:
Thanks for the thorough work here — the test discipline is good, and one of your findings is more valuable than the flag itself.

Closing this, because the premise underneath it doesn't hold.

**`complete` is accurate.** `computeTerminal` (`internal/session/manifest.go:941`) derives the state from coverage alone, and coverage is per file. Every file was reviewed and none failed, so `complete` means "nothing was left uncovered" — it never claimed anything about how the partition was decided. Details in #1164.

**Recording a fallback as `run_failure` cuts against an existing decision.** `agent.go:674-679` deliberately declines `SetRunFailure` for budget truncation so the terminal state stays coverage-derived, and a grouping fallback sits on the same side of that line. That's also why `RunFailureUnknown` doesn't fit — as you sensed when you declined to add `RunFailureGrouping`. The reason isn't unknown; it just isn't a run failure.

**Your second bullet found the actual bug.** You noticed `groupResult.failure` can wrap `context.DeadlineExceeded` from `option.WithRequestTimeout` while the outer `ctx.Err()` is nil, and handled it as a classification detail. That *is* the root cause: `groupDiffs` gets a bare `ctx` (`agent.go:630`), the `--timeout` budget is computed afterwards at `agent.go:647` and only wraps review subtasks, so grouping is bounded by the client's per-request default of 5 minutes (`internal/llm/client.go:503-504`) — the same number a single ordinary completion gets, with no scaling for the change set it must partition.

`abort` makes no grouping call succeed; it only fails sooner. Once grouping carries a deadline that scales, the flag has little left to do. Your own "Known limitation" points the same way: an empty-but-valid grouping response returns nil, so `abort` can't catch it either.

chethanuk:
Fair — I under-scoped this. Thanks for the review @lizhengfeng101 

I read `complete` as a grouping verdict when it's a coverage one. `computeTerminal` says
so plainly, and `agent.go:673-679` had already drawn the line I was arguing against, deliberately.

On the deadline: PR body lines 39-44 already said `groupResult.failure` can wrap
`context.DeadlineExceeded` from `option.WithRequestTimeout`, scoped below the outer ctx,
leaving `ctx.Err()` nil. I shipped that as a classification fix, `unknown` → `timeout`.
You read the same observation as the design bug — grouping has no deadline of its own and
doesn't scale with the change set it has to partition. Same evidence, bigger conclusion,
and yours is the better read.

You're right about `RunFailureUnknown` too. I half-sensed it when I declined to add
`RunFailureGrouping` and then reached for `unknown` anyway.

One thing worth leaving on the record. "`abort` only fails sooner" holds for the
green-check problem, but not for the cost one. Points 2 and 3 let a gate fail the job,
but they land in `emitRunResult`, after `ag.Run` has already returned — the 47m40s and
21.4M tokens are spent by then. A grouping deadline that scales fixes the timeout path; grouping
can still fail on a provider 500 or a malformed JSON body, and then it's the full 188-way
fan-out again with nothing to stop it. Not an argument for the flag as I built it, just
the hole that's left after both fixes land.

On 2 and 3, since you said they stand independently:

- `grouping.failed` is already written on this branch — `telemetry.ErrorEvent(ctx,
  "grouping.failed", err, ...)` at `grouping.go:187`, carrying `grouping.skipped`'s
  attribute set so the two are comparable.
- The `grouping` block in the result JSON isn't. `GroupingStrategy.String()` already emits
  `llm`/`bundle_all`/`per_file`, so it's mostly plumbing plus the failure reason.

Happy to send those two as a separate PR with the flag dropped. And if you'd rather take the grouping deadline itself, that's okay too

hyy321:
Hi @lizhengfeng101, I went back through the discussion here and checked the current main. The grouping deadline issue you pointed out still appears to be unresolved, so I traced the timeout path a bit further.
The timeout issue identified here is that groupDiffs has no grouping-specific deadline that scales with the change set. Separately, #1155 reports that --timeout does not raise the LLM client's per-request deadline, and #1159 is currently addressing that mismatch.
So #1159 would raise the request deadline on the default timeout path, while a separate follow-up would give grouping an appropriately scaled deadline of its own. Without that request-timeout change, a longer grouping context could still be cut off earlier by the client's default per-request timeout.
I'd like to work on that grouping deadline follow-up. Since #1159 is still open, I can test both pieces together locally. If #1159 is still unmerged when the fix is ready, would you prefer a combined PR, or a separate follow-up PR after #1159?

PR #1174, opened 2026-09-05
feat(viewer): export a session to a self-contained HTML file (#1167)

### LINKED ISSUE #1167: Provide an option to export session to a static html
### Problem Statement

I hope there's a way to export the session report to a static html. I use ocr in GitLab merge request pipeline. I want to archive a html report after all things are done and we can debug review process easily with the html file.

### Proposed Solution

parse jsonl to a html

### Alternatives Considered

_No response_

### Affected Area

Output / Formatting

### Additional Context

_No response_

chethanuk:
## Description

The only renderer for a review session is the live viewer HTTP server, so there's no way to archive a review as a CI artifact. `session.html` also pulls its two `/static/` assets by absolute path, so even lifting the rendered bytes out would not render offline.

Adds `viewer.ExportSession`, which drives the same `LoadSession` and the same template the HTTP handler uses, but renders into an `io.Writer` with `style.css` and `session.js` inlined as `template.CSS`/`template.JS` (an untyped string renders as literal `ZgotmplZ` for CSS and an escaped JSON literal for JS — silently broken either way).

`session.html` gains three gates on a new `Static` field, zero for every HTTP render, so the served page is byte-identical to before. The exported page inlines both assets and de-links the two site-absolute breadcrumb anchors, which are dead over `file://`.

Exposed as `ocr session export [session-id]` on the existing `session` command rather than a `viewer --export` flag, since `viewer` is `cobra.NoArgs` and `session_cmd.go` already owns session-id resolution and completion. The id is optional and defaults to the newest session — a successful `ocr review` never prints its own session id, so requiring one would force every CI archive step through `--format json | jq`.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

`make check`, `make test` (23 packages, `-race`) and `make coverage` all pass, coverage at 91.0% against the 90% threshold.

The new test asserts on real inlined asset content (`style.css`'s font stack, `session.js`'s inline-code replacement) rather than on the presence of a `<style>` tag, so it fails against a naive implementation that passes plain strings into `<style>`/`<script>`.

End to end against a synthetic session: the exported file contains no `ZgotmplZ`, no `/static/` references, and the only tag that could fetch anything is the inlined `<script>` (no `src`).

Also covered: no-id defaults to the newest session; an unknown id and an empty repo both error without creating the output file; the served HTTP page is unchanged (`TestHandleSession_ServedPageKeepsStaticRefs`); and both embedded assets are statically checked for no unescaped `</style` or `</script` terminators.

Documented in all five locales of `cli-reference.md`.

### Limitation

Opening the file in a real browser and confirming zero network requests was not automated — nothing in this repo drives a browser. What's proved is structural: no `<link>`, no ``, one `<script>` with no `src`, no `/static/` substring anywhere.

## Checklist

- [ ] I have signed the CLA

## Related Issues

Closes #1167


chethanuk:

lizhengfeng101:
@chethanuk Nice work — the offline-completeness story holds up well: the `{{icon}}` helper inlines the SVGs at render time and the only `url()`s in style.css are `data:` URIs, so nothing here reaches out to the network. That part I'm not worried about.

One thing before merge: this branch is based on a main that predates #1358 (the session-detail restyle), so it's currently built on the pre-restyle `session.html`/`style.css`. Both PRs touch `session.html`, `style.css` and `server_extra_test.go`, so please rebase onto latest main and re-run the tests. The good news is the export reuses the same template + CSS, so once rebased it inherits the new design for free — just double-check the hardcoded assertions (e.g. `<h3>Review Comments (` in TestExportSession_Variants) still match after the restyle.

Separately: the Static branch hand-copies the `app-header` markup. It happens to line up with the current header, but that block just changed in #1367 and #1358 — I'd pull the brand into a small sub-template that takes a 'linkable' flag rather than keeping two copies in sync, otherwise the export will silently drift the next time the header moves.

lizhengfeng101:
and rebase main

cmd/opencodereview/session_cmd_test.go
```diff
@@ -802,3 +803,129 @@ func TestRunSessionCompare_UnfinishedAfterRunIsNotResolved(t *testing.T) {
...
+// CI archive step through `--format json`.
+func TestRunSessionExport_DefaultsToNewest(t *testing.T) {
+	setTestHome(t, t.TempDir())
+	repoDir := t.TempDir()
+
+	older := session.New(repoDir, "main", "test-model", session.SessionOptions{ReviewMode: session.ReviewModeCommit, DiffCommit: "old"})
+	older.RecordReviewItemDone("a.go", "a.go", "a.go", "fp-a", nil)
+	older.Finalize()
+
+	time.Sleep(1100 * time.Millisecond) // session ids and StartTime are second-resolution
```
lizhengfeng101: `SessionID` comes from `generateUUID()` (internal/session/history.go:156), so two sessions can never share an id. Only `StartTime` is at risk here: `WriteSessionStart` persists it with `Format(time.RFC3339)` (internal/session/persist.go:143), and `ListSessions` orders by it through a non-stable `sort.Slice`. The sleep is doing real work — the comment just names the wrong reason.

```suggestion
	time.Sleep(1100 * time.Millisecond) // StartTime is persisted at second resolution
```

cmd/opencodereview/session_cmd_test.go
```diff
@@ -802,3 +803,129 @@ func TestRunSessionCompare_UnfinishedAfterRunIsNotResolved(t *testing.T) {
...
+	newer.Finalize()
+
+	body, err := exportTo(t, repoDir, "")
+	if err != nil {
+		t.Fatalf("runSessionExport: %v", err)
+	}
+	if !strings.Contains(body, newer.SessionID) {
+		t.Errorf("export defaulted to the wrong session; want %s", newer.SessionID)
+	}
+	if newer.SessionID != older.SessionID && strings.Contains(body, older.SessionID) {
```
lizhengfeng101: Follow-on from the comment above: because ids are UUIDs, `newer.SessionID != older.SessionID` is always true, so this half of the guard is dead.

```suggestion
	if strings.Contains(body, older.SessionID) {
```

lizhengfeng101 (commented):
Two small nits in the new `DefaultsToNewest` test: both the comment and the guard rest on session ids being timestamp-derived, but they are UUIDs. Behavior is unaffected.

lizhengfeng101 (approved):
LGTM

A5580:
https://app.tokenstorm.com/auth/login?redirect=/customer/earnhttps://cryptotabbrowser.com/landing/83/27779009https://cryptotabbrowser.com/landing/83/27779009

PR #1180, opened 2026-09-05
feat(viewer): number the lines inside the Existing Code block (#1172)

### LINKED ISSUE #1172: viewer: Improvement request: display line numbers inside the EXISTING CODE block
### Problem Statement

OCR displays line numbers above comment text, it would be better to display line numbers inside the `EXISTING CODE` block directly.

Attach 2 screenshots, one of them is the render of GitHub Copilot for reference.

Image
Image

### Proposed Solution

display line numbers inside the `EXISTING CODE` block directly.

### Alternatives Considered

_No response_

### Affected Area

WebUI Viewer

### Additional Context

_No response_

chethanuk:
## Description

The session page prints the Existing Code block with no line numbers and puts the
range in a separate `L120-L124` chip above the comment text. Mapping "the third line
of this snippet" back to a file means counting rows by hand, and the chip is the only
place the absolute numbers appear.

This numbers each line of the block in its own gutter. The gutter is
`user-select: none`, so copying the snippet still yields just the code, and it is
`aria-hidden` so a screen reader does not read a number before every line. The
chip is redundant once the gutter is there, so it is hidden visually with an
`.sr-only` class rather than dropped — otherwise a numbered block would have no
line location left in the accessibility tree at all.

Numbers are emitted only when `EndLine-StartLine+1` matches the line count of
`ExistingCode`. That guard is load-bearing rather than defensive: `internal/diff/resolver.go`
skips blank lines on both sides while matching (`splitAndNormalize`, and again in
`resolveFromFileContent`), so the stored span and the stored snippet can legitimately
disagree.

A zero `EndLine` is treated as a single-line finding. An inverted non-zero range is
left alone so the length check rejects it, matching the `hasRegion` test in
`cmd/opencodereview/sarif.go`.
When they do, the block renders exactly as it does today instead of printing numbers
that point at the wrong lines.

Nothing outside the viewer changes — the resolver, the store and the suggested-change
block are untouched.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

`TestNumberedCodeLines` covers the helper across 13 cases: empty input, a span/snippet
length mismatch, `endLine < startLine`, `startLine <= 0`, CRLF, and a single trailing
newline.

`TestParseTemplate_ExistingCodeLineNumbers` asserts the rendered DOM. Its
`resolved range is numbered` subtest fails on the parent commit — no `<span class="line-no"`
is present and the `comment-lines` chip still is — and passes with the change.
`TestParseTemplate_SessionWithComments` is unmodified and still passes.

Also run: `make check` (license-check, english-check, `go mod tidy`, `gofmt`, `go vet`)
and `go test -race ./internal/viewer/...`.

Manually: opened a session with both a resolver-matched comment and one where the model
supplied its own range, and confirmed the second still renders unnumbered with its chip
intact.

## Checklist

- [x] I have signed the CLA

## Related Issues

Closes #1172

### DIFF
--- a/internal/viewer/server.go
+++ b/internal/viewer/server.go
@@ -241,6 +241,49 @@ func severityCounts(comments []*ReviewComment) SeverityCount {
 	return counts
 }
 
+// codeLine is one rendered line of an Existing Code block. Num is the file
+// line number, or 0 when the number cannot be trusted.
+type codeLine struct {
+	Num  int
+	Text string
+}
+
+// numberedCodeLines pairs each line of existing_code with its file line number.
+//
+// Num is left at 0 on every line whenever the reported range and the snippet
+// cannot both be true. internal/diff/resolver.go matches existing_code against
+// the file with blank lines dropped on both sides (splitAndNormalize and
+// resolveFromFileContent), so endLine-startLine+1 is not guaranteed to equal
+// the number of lines in the snippet, and numbering it anyway would put line
+// numbers next to the wrong code. In a review tool no gutter beats a wrong one.
+func numberedCodeLines(code string, startLine, endLine int) []codeLine {
+	if code == "" {
+		return nil
+	}
+	raw := strings.Split(code, "\n")
+	// A trailing newline terminates the last line, it does not start a new one.
+	if len(raw) > 1 && raw[len(raw)-1] == "" {
+		raw = raw[:len(raw)-1]
+	}
+	lines := make([]codeLine, len(raw))
+	for i, text := range raw {
+		lines[i] = codeLine{Text: strings.TrimSuffix(text, "\r")}
+	}
+	if endLine == 0 {
+		// A record with only start_line set is a single-line finding. A
+		// non-zero inverted range is left alone so the guard below rejects
+		// it, matching the hasRegion test in cmd/opencodereview/sarif.go.
+		endLine = startLine
+	}
+	if startLine <= 0 || endLine-startLine+1 != len(lines) {
+		return lines
+	}
+	for i := range lines {
+		lines[i].Num = startLine + i
+	}
+	return lines
+}
+
 func parseTemplate(name string) (*template.Template, error) {
 	funcMap := template.FuncMap{
 		"formatDuration": formatDuration,
@@ -360,6 +403,7 @@ func parseTemplate(name string) (*template.Template, error) {
 				return "cat-default"
 			}
 		},
+		"numberedCodeLines": numberedCodeLines,
 	}
 	content, err := assets.ReadFile("templates/" + name)
 	if err != nil {
--- a/internal/viewer/server_startserver_test.go
+++ b/internal/viewer/server_startserver_test.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httptest"
+	"reflect"
 	"strings"
 	"testing"
 )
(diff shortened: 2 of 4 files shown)


Qiyuanqiii (changes requested):
The numbering logic and fallback behavior look solid overall, and the test coverage is good. I found one accessibility regression that I think should be fixed before merge.
When a block is successfully numbered, the existing comment-lines range is removed, while every gutter number is marked aria-hidden="true". That means screen-reader users lose the line-location information entirely: the old L10-L12 text is gone from the accessibility tree, and the replacement line numbers are hidden from it as well.
I think keeping the gutter hidden from assistive tech is reasonable to avoid making every code line noisy, but we should preserve the range somewhere accessible — for example with a visually hidden Lines 10 through 12 label.
One small robustness point as well: the endLine < startLine normalization seems broader than necessary. If the compatibility case is specifically end_line == 0 for single-line findings, checking endLine == 0 would avoid turning an actually invalid non-zero range into a seemingly trusted gutter.
Other than that, the implementation looks good to me.

Qiyuanqiii (approved):
Quick follow-up to my previous review: I think I was a bit too strong in marking this as Request changes.

The accessibility concern I mentioned is still worth addressing — keeping the line range available to screen readers would be a nice improvement — but I don't think it needs to block this PR on its own. The "endLine < startLine" point is also more of a small robustness suggestion than a blocker.

The numbering logic, fallback behavior, and test coverage all look good to me otherwise. Approving from my side.

chethanuk:
Both taken, pushed as `a5a2555`. Thanks @Qiyuanqiii

The accessibility one was a regression I introduced, not a pre-existing gap — `main:155`
renders that chip on a bare `{{if .StartLine}}`. I gated it on `(not $numbered)` and marked
every gutter number `aria-hidden`, so a numbered block lost the range entirely, and
`StartLine`/`EndLine` show up in only two places across `templates/` and `static/`, both on
that line.

Fixed by keeping the chip rather than adding a second label: it renders in every case now,
with `.sr-only` when the gutter is present. Sighted users keep the dedup, assistive tech
gets the same `L10-L12` string it had before this PR. The gutter stays `aria-hidden` —
you're right that a number before every line is noise.

On `endLine < startLine` I took the change, though the premise doesn't quite hold:

- `architecture.md:282-285` and `SKILL.md:96` both have the two fields going to zero
  together, and I couldn't find a record, fixture or code path producing
  `start_line > 0, end_line == 0`.
- `code_comment`'s schema is `category`, `content`, `existing_code`, `path`, `severity`,
  `suggestion_code` — no line numbers at all (only `file_read` has those), and everything
  downstream goes through `resolver.go` ascending by construction. The way in is a
  hand-edited JSONL, since `store.go:535-541` parses those fields unvalidated.

Took it because `sarif.go:241` already settles the policy with
`hasRegion := c.StartLine > 0 && c.EndLine >= c.StartLine`. An inverted range is distrusted
there, not repaired. Mine repaired it, and on a single-line snippet that meant printing a
gutter number derived from a range the rest of the codebase calls invalid. It checks
`endLine == 0` now, and the length guard rejects the rest.

Two more while I was in there:

- The doc comment said `ResolveLineNumbers` "skips resolution entirely when the model
  supplied a range itself". It can't — no such field in the schema; the guard at
  `resolver.go:37`/`:63` is two-pass idempotency. That line was also in the commit message
  and the PR body, so it's fixed in all three.
- The old `wantHasNot` on `class="comment-lines"` would have gone vacuous under the new
  markup, since `class="comment-lines sr-only"` doesn't contain it. Replaced with a
  positive assertion on the full span.

Both new assertions fail before the fix and pass after. `make check` clean.

chethanuk:
Both taken, pushed as `a5a2555`. Thanks @Qiyuanqiii

The accessibility one was a regression I introduced, not a pre-existing gap — `main:155`
renders that chip on a bare `{{if .StartLine}}`. I gated it on `(not $numbered)` and marked
every gutter number `aria-hidden`, so a numbered block lost the range entirely, and
`StartLine`/`EndLine` show up in only two places across `templates/` and `static/`, both on
that line.

Fixed by keeping the chip rather than adding a second label: it renders in every case now,
with `.sr-only` when the gutter is present. Sighted users keep the dedup, assistive tech
gets the same `L10-L12` string it had before this PR. The gutter stays `aria-hidden` —
you're right that a number before every line is noise.

On `endLine < startLine` I took the change, though the premise doesn't quite hold:

- `architecture.md:282-285` and `SKILL.md:96` both have the two fields going to zero
  together, and I couldn't find a record, fixture or code path producing
  `start_line > 0, end_line == 0`.
- `code_comment`'s schema is `category`, `content`, `existing_code`, `path`, `severity`,
  `suggestion_code` — no line numbers at all (only `file_read` has those), and everything
  downstream goes through `resolver.go` ascending by construction. The way in is a
  hand-edited JSONL, since `store.go:535-541` parses those fields unvalidated.

Took it because `sarif.go:241` already settles the policy with
`hasRegion := c.StartLine > 0 && c.EndLine >= c.StartLine`. An inverted range is distrusted
there, not repaired. Mine repaired it, and on a single-line snippet that meant printing a
gutter number derived from a range the rest of the codebase calls invalid. It checks
`endLine == 0` now, and the length guard rejects the rest.

lizhengfeng101 (approved):
nice work

PR #1185, opened 2026-09-05
fix(review): exit immediately on the second signal

HOWILLMAKEIT:
## Description

`ocr review` registered SIGINT/SIGTERM via `signal.NotifyContext`. Its watcher goroutine exits after the first signal without unregistering, so during the entire graceful-shutdown window every further signal lands in an unread channel and is dropped — a second Ctrl+C does nothing, however long the cleanup takes.

This change replaces `NotifyContext` with an owned signal channel in `cmd/opencodereview/interrupt.go`:

- the first signal cancels the context, running the existing graceful-shutdown defer chain;
- a second signal force-exits the process immediately, since the user has explicitly abandoned the graceful path;
- a late signal arriving after shutdown already completed is dropped rather than turning a successful run into exit 1;
- `stop()` is idempotent and mirrors `NotifyContext`'s stop semantics.

`scan_cmd.go` is untouched; it registers no signal handling today.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- `cmd/opencodereview/interrupt_test.go` covers first-signal graceful cancellation (SIGINT and SIGTERM), second-signal forced exit, stop/parent-cancel releasing the watcher, and two subprocess-level integration tests exercising the real `os.Exit` path end to end.
- The second-signal integration test asserts the child exits in under 100 ms with status 1; the single-signal test asserts the partial report lands on disk and the child exits 0.

- [x] Manual testing: verified the second-signal exit path and the single-signal graceful path via the subprocess integration tests

## Checklist

- [x] I have signed the CLA

## Related Issues

Related to #1141 (item 3).

### DIFF
--- a/cmd/opencodereview/interrupt.go
+++ b/cmd/opencodereview/interrupt.go
@@ -1,0 +1,100 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 alibaba/open-code-review Contributors
+
+package main
+
+import (
+	"context"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+)
+
+// forceExit terminates the process immediately when a second signal arrives.
+// It is a variable so tests can intercept the forced-exit path without
+// killing the test binary. No diagnostic is printed: an asynchronous write
+// would almost never complete before os.Exit kills the process, and a
+// synchronous write to a full stderr pipe would block the exit entirely.
+var forceExit = func(sig os.Signal) {
+	os.Exit(1)
+}
+
+// interruptContextWithForcedExit returns a child of parent that is canceled
+// by the first os.Interrupt or syscall.SIGTERM, together with a stop function
+// that restores default signal handling.
+//
+// The first signal cancels the context so executeReviewContext's defer chain
+// can run the graceful shutdown: flush the report, freeze the retry report,
+// persist the session manifest, close MCP clients. signal.NotifyContext
+// cannot serve this command's second requirement: its watcher goroutine
+// exits right after the first signal without unregistering, so for the whole
+// graceful-shutdown window every further signal lands in its unread buffered
+// channel and is dropped while the default kill behavior stays suppressed —
+// a second Ctrl+C does nothing. Here the watcher stays alive across the
+// whole shutdown window and a second signal force-exits the process at
+// once: the user has explicitly abandoned the graceful path, however long
+// the remaining cleanup would take.
+//
+// stop is idempotent. It unregisters the signal channel, releases the
+// watcher goroutine and cancels the context, mirroring the stop returned by
+// signal.NotifyContext.
+func interruptContextWithForcedExit(parent context.Context) (context.Context, context.CancelFunc) {
+	ctx, cancel := context.WithCancel(parent)
+	// Buffer 2: os/signal sends non-blockingly, so with a size-one buffer two
+	// back-to-back signals could drop the second before the watcher consumes
+	// the first — the very failure mode this function exists to eliminate.
+	ch := make(chan os.Signal, 2)
+	signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
+
+	done := make(chan struct{})
+	var once sync.Once
+	// mu + stopped close the race between stop() and the watcher's
+	// late-signal check: without it, a signal already in ch's buffer can be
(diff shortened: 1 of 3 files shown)


cmd/opencodereview/interrupt.go
```diff
@@ -0,0 +1,100 @@
...
+	"os/signal"
+	"sync"
+	"syscall"
+)
+
+// forceExit terminates the process immediately when a second signal arrives.
+// It is a variable so tests can intercept the forced-exit path without
+// killing the test binary.
+var forceExit = func(sig os.Signal) {
+	fmt.Fprintf(os.Stderr, "[ocr] second %v signal received, exiting immediately\n", sig)
```
Qiyuanqiii: This forced-exit path can still block on stderr.

`forceExit` writes synchronously to `os.Stderr` before calling `os.Exit(1)`. If stderr is a full pipe whose reader is still open but no longer consuming data, the second signal can block in `fmt.Fprintf`, so the process never reaches `os.Exit`. Further signals do not help because the watcher is already stuck in `forceExit`.

I reproduced this consistently with an unmodified copy of `interrupt.go`: after filling the child's stderr pipe, the first SIGINT starts graceful shutdown, but the second SIGINT leaves the process alive; even a third SIGINT has no effect. As soon as the stderr pipe is drained, the process exits with status 1. Removing the synchronous log write makes the same test exit immediately.

Could we make the forced-exit path independent of stderr availability and add a regression test for a blocked/full stderr pipe? Since the purpose of this PR is to guarantee that the second signal terminates a stuck shutdown, I think this should be fixed before merging.

A minimal fix would be to make this diagnostic best-effort so it can't block the actual exit:

```suggestion
	go fmt.Fprintf(os.Stderr, "[ocr] second %v signal received, exiting immediately\n", sig)
```

A regression test with a blocked/full stderr pipe would still be useful to make sure this behavior doesn't regress.

Qiyuanqiii (changes requested):
Blocking issue in the forced-exit path; details are in the inline comment.

cmd/opencodereview/interrupt.go
HOWILLMAKEIT: thanks, I will fix it soon :)

HOWILLMAKEIT:
thanks to @Qiyuanqiii , a problem is fixed in 85efaa3.

The diagnostic write now runs in a goroutine (`go fmt.Fprintf(...)`) so a full stderr pipe can no longer block the exit. The process reaches `os.Exit(1)` unconditionally; the message is best-effort.

I also added a regression test (`TestSecondSignalExitsWithBlockedStderr`) that reproduces this scenario: the child's stderr is redirected to a pipe that is filled to capacity before the signals are sent, and the test verifies the second SIGINT still terminates the process.

Qiyuanqiii (approved):

lizhengfeng101:
Solid fix — the second-signal analysis and the concurrency handling both hold up. Two smaller things I'd tweak before merge:

1. **The diagnostic line in `forceExit` will almost never actually print.** Writing it in a goroutine right before `os.Exit(1)` means the scheduler basically never runs it before the process is gone. I get the intent (don't let a blocked stderr hang the exit — nice regression test for that), but as written the "second signal received, exiting immediately" message is effectively dead code on a normal tty. Either drop it, or note in the comment that it's best-effort and usually won't show, so the next reader doesn't assume it's reliable.

2. **The 100ms bound in `TestSecondSignalExitsImmediately` looks tight enough to flake on a loaded CI runner.** Cross-process signal delivery + scheduling + `os.Exit` can drift close to that. Since the bug this guards against was "never exits until cleanup finishes," bumping the assertion to ~500ms–1s keeps the intent fully intact while removing the flake risk.

HOWILLMAKEIT:
Thank you for the review, @lizhengfeng101  — both suggestions are addressed in 6f7579e:

- The diagnostic write in `forceExit` is removed entirely. The comment now explains both reasons: an asynchronous write would almost never complete before `os.Exit` kills the process, and a synchronous write to a full stderr pipe would block the exit entirely.
- The assertion in `TestSecondSignalExitsImmediately` is relaxed from 100ms to 500ms, with the comment updated to explain the bound is loose enough for a loaded CI runner's cross-process signal latency while still far below the 30s cleanup the forced exit exists to skip.

lizhengfeng101 (approved):
LGTM

PR #1186, opened 2026-09-05
fix(rules): route Gradle scripts to Gradle rules

### LINKED ISSUE #1184: Gradle scripts other than build.gradle fall back to default rules
### OpenCodeReview Version

./dist/opencodereview version

### Operating System

macOS (Apple Silicon)

### Installation Method

Built from source

### LLM Provider

Anthropic (Claude)

### Bug Description

I noticed that `.gradle` files are supported, but only `build.gradle` gets the built-in Gradle rule. Other Gradle files, such as `settings.gradle` and `scripts/dependencies.gradle`, use the default rule instead.

### Steps to Reproduce

Run:

```bash
./dist/opencodereview rules check build.gradle
./dist/opencodereview rules check settings.gradle
./dist/opencodereview rules check scripts/dependencies.gradle
```
### Actual Behavior

`build.gradle` matches `**/build.gradle`. The other two files fall back to `default.md`.

### Expected Behavior

Gradle scripts should receive relevant Gradle-specific review guidance instead of the generic default rules.

### Logs / Error Output

```shell
build.gradle
Source: System built-in
Pattern: **/build.gradle

settings.gradle
Source: System built-in
Pattern: default

scripts/dependencies.gradle
Source: System built-in
Pattern: default
```

### Additional Context

`.gradle` is already included in the supported file types. This happens during local rule resolution and does not depend on the LLM provider.

Changing the mapping to `**/*.gradle` would cover these files, but the existing `build_gradle.md` rule only discusses snapshot dependencies. I would like guidance on whether that rule should be expanded or whether the current mapping is intentional.

Parikalp-Bhardwaj:
## Description

Only files named `build.gradle` currently receive the built-in Gradle rule. Other supported Gradle scripts, including `settings.gradle` and nested `.gradle` files, fall back to the default rules.

This PR changes the system mapping from `**/build.gradle` to `**/*.gradle`. It also adds coverage for root, nested, and uppercase Gradle paths and updates the rule tables in all five documentation languages.

`build.gradle.kts` remains mapped to the Kotlin rules.

The existing `build_gradle.md` content is unchanged. It currently focuses on snapshot dependencies, so I would appreciate confirmation that it should apply to all `.gradle` scripts.

## Type of Change

- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] Manual testing

Additional checks:

- `make check`
- `make coverage`
- `make build`
- `go test ./internal/config/rules/... -count=1`
- `ocr rules check build.gradle`
- `ocr rules check settings.gradle`
- `ocr rules check scripts/dependencies.gradle`
- `ocr rules check SETTINGS.GRADLE`

All tested `.gradle` paths resolve to `build_gradle.md`. Kotlin DSL files continue to use the Kotlin rules.

## Related Issues

Closes #1184

### DIFF
--- a/internal/config/rules/system_rules.json
+++ b/internal/config/rules/system_rules.json
@@ -4,7 +4,7 @@
     "**/*.properties": "properties.md",
     "**/*{mapper,dao}*.xml": "mapper_dao_xml.md",
     "**/pom.xml": "pom_xml.md",
-    "**/build.gradle": "build_gradle.md",
+    "**/*.gradle": "build_gradle.md",
     "**/package.json": "package_json.md",
     "**/Cargo.toml": "cargo_toml.md",
     "**/composer.json": "composer_json.md",
--- a/internal/config/rules/system_rules_test.go
+++ b/internal/config/rules/system_rules_test.go
@@ -82,6 +82,10 @@ func TestResolve_DefaultRules(t *testing.T) {
 		{"src/main/resources/dao/userdao.xml", "SQL Logic Error Detection"},
 		{"pom.xml", "snapshot"},
 		{"submodule/pom.xml", "snapshot"},
+		{"build.gradle", "snapshot version dependencies"},
+		{"settings.gradle", "snapshot version dependencies"},
+		{"scripts/dependencies.gradle", "snapshot version dependencies"},
+		{"SETTINGS.GRADLE", "snapshot version dependencies"},
 		{"src/main/resources/application.properties", "Configuration Error Detection"},
 		{"frontend/package.json", "latest"},
 		{"composer.json", "Composer Manifest Review Principles"},
@@ -182,6 +186,7 @@ func TestResolve_FallbackToDefault(t *testing.T) {
 		"readme.md",
 		"docs/architecture.txt",
 		"Makefile",
+		"gradlew",
 	}
 
 	for _, path := range paths {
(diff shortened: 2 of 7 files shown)


wu21-web:
The following types of .gradle files are generally considered to require no detailed review, they cost tokens and don't really value much:

* `build.gradle` / `build.gradle.kts`: The core build script of the project, defining dependencies, plugins, and tasks. Because it mainly declares dependencies and configurations, its logic is relatively straightforward, and it is often excluded from review by tools (such as GitHub Copilot) by default.

* `settings.gradle` / `settings.gradle.kts`: Defines the project name and module structure. It mainly serves a declarative purpose and contains relatively little logic, usually not worth reviewing line by line.

* `gradle.properties`: Used to set properties of the Gradle build itself (such as JVM parameters, build cache, etc.). It is usually configuration-based and does not contain business logic.

* `gradle-wrapper.properties`: Specifies the Gradle release version. This file is automatically generated or updated by the Gradle Wrapper and generally does not require review.

* Automatically generated or temporary files: Includes the entire .gradle directory, the build/ directory, and local.properties. These files are generated during the build process and should not be subject to version control, much less be subject to review.

- For `build.gradle.kts`, it might have 2 rule_docs that needs to be merged together (because it has kotlin properties, just guessing). Which is a headache.

And also note you haven't added any exclude patterns. (I would exclude them all actually)

**Therefore, I prefer closing this and the issue as not planned.**

Parikalp-Bhardwaj:
@wu21-web Thanks for the explanation. That makes sense, especially the limited review value and the kotlin dsl overlap, i am closing the PR and issue

PR #1187, opened 2026-09-05
feat(cmd): man doc support for ocr on UNIX-like systems

wu21-web:
## Description
feat(cmd): man doc support for ocr on UNIX-like systems

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [x] I have signed the CLA

## Related Issues

N/A


wu21-web:
May I request a review from @Qiyuanqiii , thanks!

Qiyuanqiii (changes requested):
The man-page generator itself looks solid, and CI is green. I found two installer issues that should be fixed before merge:

1. For custom install prefixes, the script tells users to add the section directory (for example "~/.local/share/man/man1") to "MANPATH". "MANPATH" should point to the manual hierarchy root ("~/.local/share/man"), with "man1/" beneath it.

2. In the no-sudo fallback, the script tells users that the generated pages are available under "$tmp/man" for manual installation, but "$tmp" is removed by the existing "EXIT" trap as soon as the installer exits. That makes the suggested fallback path unusable.

One small doc suggestion as well: it may be worth clarifying that automatic man-page installation currently applies to "install.sh"; with other installation methods, users can generate the pages manually with "ocr man <directory>".

Once the two installer issues are fixed, I’d be happy to take another look.

Qiyuanqiii (approved):
I don't see any remaining issues on my end.

PR #1192, opened 2026-09-06
fix(action): honor an explicit llm_protocol instead of forcing openai or anthropic

### LINKED ISSUE #1134: bug(action): Configure OCR step hardcodes llm.protocol to openai, breaking openai-responses protocol

po-et:
Fixes #1134.

The Configure OCR step derived `llm.protocol` from `llm_use_anthropic` alone, so it always wrote `openai` or `anthropic`, and users on the Responses API were sent to `/v1/chat/completions` no matter what they put in `OCR_LLM_PROTOCOL`.

This adds an optional `llm_protocol` input (`anthropic`, `openai`, `openai-responses`). When it is set, or when `OCR_LLM_PROTOCOL` is inherited from the job environment, that value is written and `llm.use_anthropic` is mirrored from it, the same way `ocr config set llm.protocol` mirrors the boolean. An unknown value fails the step with the accepted list. With neither set, nothing changes. The explicit protocol is added to the checkpoint fingerprint, appended only when set so existing checkpoints stay valid for everyone who never uses the input.

Contract tests cover the input, the inherited variable, precedence between the two, case folding, and the rejection of an unknown value. The GitHub Actions README documents the input.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -22,8 +22,18 @@ inputs:
       Selects the LLM protocol (mapped to env OCR_USE_ANTHROPIC). An explicitly supplied empty
       string, true, 1, or yes selects Anthropic
       case-insensitively; every other value selects the OpenAI-compatible
-      protocol, preserving the CLI environment contract.
+      protocol, preserving the CLI environment contract. Ignored when
+      llm_protocol names the protocol explicitly.
     required: true
+  llm_protocol:
+    description: >-
+      Explicit LLM protocol written to `llm.protocol`: anthropic, openai or
+      openai-responses (case-insensitive). When set it wins over
+      llm_use_anthropic, which is mirrored from it the way `ocr config set
+      llm.protocol` mirrors the boolean. Leave empty to keep deriving the
+      protocol from llm_use_anthropic; an OCR_LLM_PROTOCOL variable in the
+      job or step environment is honoured the same way when this is empty.
+    required: false
   llm_auth_header:
     description: Custom auth header name (mapped to env OCR_LLM_AUTH_HEADER).
     required: false
@@ -504,11 +514,16 @@ runs:
         OCR_LLM_URL: ${{ inputs.llm_url }}
         OCR_LLM_MODEL: ${{ inputs.llm_model }}
         OCR_USE_ANTHROPIC: ${{ inputs.llm_use_anthropic }}
+        OCR_LLM_PROTOCOL_INPUT: ${{ inputs.llm_protocol }}
         OCR_LLM_AUTH_HEADER: ${{ inputs.llm_auth_header }}
         OCR_EXTRA_BODY: ${{ inputs.llm_extra_body }}
         OCR_LANGUAGE: ${{ inputs.language }}
       shell: bash
       run: |
+        # An explicit protocol wins over llm_use_anthropic: the input first,
+        # then OCR_LLM_PROTOCOL inherited from the job or step environment.
+        # Read it before the derivation below reuses that variable name.
+        EXPLICIT_PROTOCOL="$(printf '%s' "${OCR_LLM_PROTOCOL_INPUT:-${OCR_LLM_PROTOCOL:-}}" | tr '[:upper:]' '[:lower:]')"
         NORMALIZED_USE_ANTHROPIC="$(printf '%s' "$OCR_USE_ANTHROPIC" | tr '[:upper:]' '[:lower:]')"
         case "$NORMALIZED_USE_ANTHROPIC" in
           ""|true|1|yes)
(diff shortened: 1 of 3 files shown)


Qiyuanqiii (changes requested):
Request changes — the runtime fix looks correct, but the inherited protocol is currently missing from the checkpoint fingerprint.

`Configure OCR` now correctly resolves the effective protocol from `llm_protocol` first and inherited `OCR_LLM_PROTOCOL` second. However, `Resolve review range` fingerprints only `${{ inputs.llm_protocol }}` via `OCR_FP_LLM_PROTOCOL`.

With `checkpoint_range: true`, this means two runs can have the same fingerprint even when the effective protocol changes solely through the job/step environment, for example:

```text
OCR_LLM_PROTOCOL=openai
→
OCR_LLM_PROTOCOL=openai-responses
```

If `llm_protocol` remains empty and the other inputs are unchanged, the previous checkpoint may therefore be treated as configuration-compatible and the next run can review only the delta, despite the effective LLM configuration having changed. This conflicts with the checkpoint contract that configuration changes must fail closed to a full review.

Could we fingerprint the same normalized effective protocol that `Configure OCR` resolves (input first, inherited env second), while retaining the current backward compatibility when neither source is set?

It would also be useful to add a contract test covering the env-only path: changing inherited `OCR_LLM_PROTOCOL` with identical action inputs should invalidate the previous checkpoint.

Other than this checkpoint interaction, the protocol precedence, `use_anthropic` mirroring, validation, documentation, and Configure OCR contract coverage look good.

po-et:
Good catch. Configure OCR now exports the normalized explicit protocol (input first, inherited OCR_LLM_PROTOCOL second) to GITHUB_ENV, and the resolve step fingerprints that value instead of the raw input, still appended only when set. Added the env-only contract case and rebased onto main after #1154.

Qiyuanqiii (approved):

harry-providend:
Hi team, will it be merged anytime soon?

lizhengfeng101:
> Hi team, will it be merged anytime soon?

hi, the merger will take place today.

lizhengfeng101 (approved):
LGTM

PR #1193, opened 2026-09-06
feat(action): expose --effort and --max-tokens-budget as inputs

### LINKED ISSUE #1147: feat(action): expose --effort and --max-tokens-budget as action inputs

po-et:
Fixes #1147.

The composite action builds the `ocr review` command itself and only exposes `--concurrency`, `--timeout`, `--background` and `--rule`. The two flags that shape token spend the most were unreachable from a workflow: `--effort` selects the review-round preset, and `--max-tokens-budget` is the only circuit breaker against a runaway review.

This adds `effort` (`low`, `medium`, `high`) and `max_tokens_budget` (non-negative integer) inputs. Each is forwarded only when set, so the CLI's own defaults still apply, and a malformed value fails the step with its own message before the review starts. `effort` is added to the checkpoint fingerprint, appended only when set so existing checkpoints stay valid for everyone who never uses it.

Contract tests cover forwarding, the empty case, and the two rejections. The GitHub Actions README gets a section for the inputs and lists `effort` under `config_changed`.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -67,6 +67,19 @@ inputs:
   rule:
     description: Path to a custom rules JSON file passed to `ocr review --rule`.
     required: false
+  effort:
+    description: >-
+      Review effort preset passed to `ocr review --effort`: low, medium or
+      high (one, two or three review rounds per file). Empty keeps the CLI's
+      configured or default preset. Changing it invalidates checkpoints.
+    required: false
+  max_tokens_budget:
+    description: >-
+      Cap on total token usage (input + output) for the whole review, passed
+      to `ocr review --max-tokens-budget`. Dispatch stops once it is
+      exceeded; partial results still publish and the run exits 0. A
+      non-negative integer; empty or 0 means unlimited.
+    required: false
   upload_artifacts:
     description: >-
       Upload raw JSON result and stderr as workflow artifacts. Must be the
@@ -461,6 +474,7 @@ runs:
         OCR_FP_ROUTE_SEVERITY_BELOW: ${{ inputs.route_severity_below }}
         OCR_FP_ROUTE_CATEGORIES: ${{ inputs.route_categories }}
         OCR_FP_BACKGROUND: ${{ inputs.background }}
+        OCR_FP_EFFORT: ${{ inputs.effort }}
         OCR_RULE_PATH: ${{ inputs.rule }}
         # `github.token` is always the "github-actions" app, so when the caller
         # did not override the token we know exactly which app wrote our summary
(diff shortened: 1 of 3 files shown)


Qiyuanqiii (changes requested):
One thing looks missing from the checkpoint contract.

#1147 explicitly calls out both `effort` and `max_tokens_budget` as inputs that should participate in the checkpoint config fingerprint. This PR adds `OCR_FP_EFFORT`, but `max_tokens_budget` is only forwarded to `ocr review` and never contributes to the fingerprint.

Could we include the budget there as well (ideally preserving the current empty-input backward compatibility), and add a contract test covering the fingerprint change? Since `0` and empty are both documented as unlimited, normalizing those to the same fingerprint value would also avoid unnecessary invalidation.

The forwarding/validation side otherwise looks good.

lizhengfeng101:
Superseded by #1154, which landed `effort` and `max_tokens_budget` and closed #1147. Closing as duplicate — thanks for the contribution!

PR #1194, opened 2026-09-07
fix(agent): budget file groups by review body size, not just diff size (#1171)

### LINKED ISSUE #1171: Grouping and size-guards measure diff size but review reads file bodies → over-budget groups fail with "context compression exceeded its threshold" instead of degrading to per-file review
**Version:** v1.11.2 (installed via `@alibaba-group/open-code-review@1.11.2`, GitHub Action)
**Model:** an OpenAI-compatible provider, thinking enabled, `MAX_TOKENS` at the template default (200000)
**Command:** `ocr review --from <merge-base> --to <head> --audience agent --format json` (concurrency 2)

## TL;DR

Every file-size guard in the review path measures **diff** size, never **file-body** size:

- `enforceGroupTokenBudget` splits a group only when the sum of `llm.CountTokens(d.Diff)` exceeds the limit — `internal/agent/grouping.go` (`total += int64(llm.CountTokens(d.Diff))`).
- The large-change pre-filter `filterLargeDiffs` likewise discards a file only when its **diff** exceeds the threshold — `internal/agent/agent.go` (`// Pre-filter: discard diffs whose diff content alone exceeds 80% of the token threshold`).

But a group's per-item review reads the **full file bodies** (via `file_read`, capped at 500 lines/call), and those reads accumulate in one conversation. So a group of files with **small diffs but large bodies** passes every guard, is bundled into a single review item, and then overflows the compression threshold at review time — returning `StopCompression` ("stopped because context compression exceeded its threshold", `internal/llmloop/loop.go`) and failing **every file in the group at once**.

Because `MAX_TOKENS=200000` with `tokenWarningThreshold=0.80`, the effective per-item ceiling is ~160k tokens (`internal/llmloop/compression.go`). Grouping (LLM strategy for `fileCount >= GroupingMinFiles`, default 4) will happily place up to `maxFilesPerGroup=10` files in one group as long as their *combined diff* is small.

## Why the existing per-file fallback doesn't save it

`groupDiffs` already falls back to one-file-per-group (`toSingleFileGroups`) on a grouping-LLM error, and `enforceGroupTokenBudget` degrades a group to per-file when its *diff* is too large. Per-file review of these same files works fine. But neither path triggers here, because the trigger condition is *diff* size — which is tiny — so the oversized bundle is dispatched as one item and there is no content-size-based degradation once review begins to overflow.

## Reproduction (observed)

A single semantic group of ~6 files, each with a **small diff** but a **large body** — each diff on the order of tens to ~150 changed lines, each body from several hundred up to a few thousand lines (largest ~170 KB). Combined diff: on the order of ~500 changed lines (well under the per-item budget); combined **body**: ~12,000 lines. The grouping step places them in one group and, because the *diff* is small, the group passes `enforceGroupTokenBudget` and is reviewed as a single item.

**Knife-edge behavior — the same group, opposite outcomes depending on the surrounding changeset:**

| overall changeset | this group | result |
|---|---|---|
| 33 changed files | same 6-file group | **completed** (whole run `status: complete`, ~28 min) |
| 40 changed files | **identical** 6-file group | **failed** — all 6 items `reason: "stopped because context compression exceeded its threshold"` (`status: partial`, ~50 min, ~5.5M tokens) |
| a small PR touching only 3–4 of these files | 3–4-file group | **completed** in 3–13 min |

So the group's own size is at the edge of the ~160k per-item ceiling; whether it fits is sensitive to run-level pressure (a marginally larger frozen "other changed files" list from `buildChangeFilesExceptGroup`, more sustained provider load causing the *async* memory-compression to fall behind, and model non-determinism). When it tips, that single item churns ~15 min to failure, and — combined with the larger run — pushes the whole job over its CI timeout.

The important, reproducible part is not the non-determinism but the **design gap**: a group whose members' **bodies** clearly cannot fit in one item's context is nonetheless assembled and dispatched as one item, because the guards only look at diffs.

## Impact

- On large review ranges (e.g. a release/promotion PR that accumulates many changes), any group of large-body/small-diff files (for example generated code, lockfiles, schema or data dumps, or large fixtures) can overflow and fail as a block.
- The failure is silent-ish: `status: partial` with the block marked failed, and — because the churn is slow — it can consume enough wall-clock to trip the CI job timeout, losing the *other* groups' findings that were computed but not yet posted.
- The same files review perfectly when they land in a smaller group on their own PR, so coverage is inconsistent between a change's own PR and a later aggregated range.

## Suggested fixes (in rough priority)

1. **Make the group budget content-aware.** In `enforceGroupTokenBudget` (and the dispatch look-ahead that sums `estimateDiffFileTokens`), estimate the tokens review will actually read — i.e. include a bounded estimate of each member's **file body** (e.g. `min(fileTokens, readableBudget)`), not just `CountTokens(d.Diff)`. A group whose estimated read cost exceeds the per-item ceiling should be split before dispatch.
2. **Degrade an over-budget group to per-file at review time.** When a grouped item returns `StopCompression`, retry its members individually (the per-file path already exists and works) rather than failing the whole block. This recovers coverage even when the pre-dispatch estimate is imperfect.
3. **Cap group assembly by body size, not just `maxFilesPerGroup`.** Add a per-group body-token budget so the grouping step never proposes a bundle that cannot fit.

## Secondary request: expose operator levers via the GitHub Action

The CLI has useful mitigations (`--max-tokens`, `--max-tools`, `--no-plan`, `--exclude`) but the Action (`action.yml`) passes only `--from/--to/--audience/--format/--timeout/--concurrency/--background/--rule`. Exposing `max_tokens` (per-item ceiling) and a grouping control (e.g. per-file mode or a group body-budget) as Action inputs would let operators mitigate this on large ranges without forking the Action. There are also no env overrides for the grouping thresholds (`GROUPING_MIN_FILES`, `maxFilesPerGroup`) or the token budget, so today there is no supported way to influence this behavior from a workflow.

## Environment notes

- `file_read` returns at most 500 lines/call (`internal/tool/file_read.go`), so reviewing a multi-thousand-line file body is many accumulating reads within one item.
- Compression is partly asynchronous (`triggerAsyncCompression`), so under provider latency the background summarizer can lag the tool-read cadence, which is consistent with the edge-case group flipping to overflow only under the heavier load of a large run.

beillahi:
Fixes #1171.

## Problem

`enforceGroupTokenBudget` splits a group only when `sum(CountTokens(d.Diff))` exceeds the limit — **diff text only**. But reviewing a patch reads the surrounding **file body** via `file_read`, so a grouped item's context is dominated by member bodies, not diffs. A group of **small-diff / large-body** files (generated config, dashboards, lockfiles, fixtures) therefore sails past the diff-only budget, is bundled into one review item, and then overflows at review time — returning `StopCompression` ("context compression exceeded its threshold") and failing **every member at once**. The same files review cleanly on their own PRs, where they land in smaller groups. (`filterLargeDiffs` has the same diff-only blind spot.)

See #1171 for the full write-up and the reproduction.

## Fix

- `groupReviewTokens` now counts each member's `Diff` **plus `NewFileContent`** (the body the review actually reads). `NewFileContent` is already populated at parse time, so this adds no I/O.
- `enforceGroupTokenBudget` splits an over-budget group to per-file review — the existing fallback shape — and never tries to split a single-file group (a lone oversized body is handled by read-chunking + compression, not here).
- `reviewGroupBudgetFraction = 0.55` reserves the rest of the per-item budget for the review's own working set (reasoning, tool-call framing, compression summaries that share the context window with the bodies being read).

## Calibration

Measured with the project's own `CountTokens`: a 6-file group totaling **~97k body tokens** overflowed a 160k per-item ceiling, while groups of **~77–80k** reviewed cleanly. `0.55` (≈88k of a 160k budget) splits the former with margin and keeps the latter — so it fixes the overflow without over-splitting healthy groups.

## Tests

Adds `grouping_body_budget_test.go` mirroring both real cases:
- the ~97k small-diff/large-body bundle → **splits to per-file** (and asserts the diffs alone stay under the limit, proving the body drives the split);
- the ~80k subset → **kept as one group**;
- a single oversized file → **never split**.

`go build ./...`, `go vet ./internal/agent/`, `gofmt`, and the full `internal/agent` suite pass; existing `enforceGroupTokenBudget` tests are unchanged and still green.

## Notes

An alternative/complementary fix is to degrade a group to per-file **after** a `StopCompression` at review time; this PR takes the pre-dispatch path because it also avoids the wasted churn (and inflated request count) of reviewing a bundle that cannot fit. Happy to adjust the fraction or approach per maintainer preference.

### DIFF
--- a/internal/agent/grouping.go
+++ b/internal/agent/grouping.go
@@ -21,6 +21,11 @@ import (
 
 const maxFilesPerGroup = 10
 
+// reviewGroupBudgetFraction is the share of the per-item token budget a group's
+// file bodies may occupy before it is split to per-file review; the remainder
+// covers the review's working set (reasoning, tool framing, compression).
+const reviewGroupBudgetFraction = 0.55
+
 // smallChangeSetLabel labels the single group a below-threshold change set is
 // bundled into. Unlike an LLM-produced label it carries no semantics, because
 // no partition was computed: every file simply went in together.
@@ -332,26 +337,39 @@ func enforceMaxFilesPerGroup(groups []FileGroup) []FileGroup {
 	return result
 }
 
-// enforceGroupTokenBudget splits groups whose combined diffs exceed the token limit.
+// groupReviewTokens estimates the context a group's review must hold: each
+// member's diff plus its file body, which is read via file_read and is the term
+// the old diff-only budget missed.
+func groupReviewTokens(g FileGroup) int64 {
+	var total int64
+	for _, d := range g.Diffs {
+		if d.IsDeleted {
+			continue
+		}
+		total += int64(llm.CountTokens(d.Diff)) + int64(llm.CountTokens(d.NewFileContent))
+	}
+	return total
+}
+
+// enforceGroupTokenBudget splits a group to per-file review when its estimated
+// review context (diffs + the bodies read to review them) exceeds the budget.
 func enforceGroupTokenBudget(groups []FileGroup, tokenLimit int) []FileGroup {
 	if tokenLimit <= 0 {
 		return groups
 	}
+	budget := int64(float64(tokenLimit) * reviewGroupBudgetFraction)
 	var result []FileGroup
 	for _, g := range groups {
-		total := int64(0)
-		for _, d := range g.Diffs {
-			total += int64(llm.CountTokens(d.Diff))
-		}
-		if total <= int64(tokenLimit) {
+		// A single-file group cannot be split further.
+		if len(g.Diffs) <= 1 || groupReviewTokens(g) <= budget {
 			result = append(result, g)
-		} else {
-			for _, d := range g.Diffs {
-				result = append(result, FileGroup{
-					Label: g.Label + " (split: " + d.NewPath + ")",
-					Diffs: []model.Diff{d},
-				})
-			}
+			continue
+		}
+		for _, d := range g.Diffs {
+			result = append(result, FileGroup{
+				Label: g.Label + " (split: " + d.NewPath + ")",
+				Diffs: []model.Diff{d},
+			})
 		}
 	}
 	return result
--- a/internal/agent/grouping_body_budget_test.go
+++ b/internal/agent/grouping_body_budget_test.go
@@ -1,0 +1,103 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 alibaba/open-code-review Contributors
+
+package agent
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/alibaba/open-code-review/internal/llm"
+	"github.com/alibaba/open-code-review/internal/model"
+)
+
+// bodyWithTokens returns text whose CountTokens is at least target. It measures
+// the repeat unit once and multiplies, avoiding an O(n^2) grow-and-recount loop.
+func bodyWithTokens(target int) string {
+	const unit = "alert rule threshold reduce noDataState execErrState summary description\n"
+	per := llm.CountTokens(unit)
+	if per < 1 {
+		per = 1
+	}
+	return strings.Repeat(unit, target/per+1)
+}
+
+// smallDiffLargeBody is the pathological shape: a tiny patch against a large
+// file. The old diff-only budget saw only the diff and never split.
+func smallDiffLargeBody(path string, bodyTokens int) model.Diff {
+	return model.Diff{
+		NewPath:        path,
+		Diff:           "@@ -1,1 +1,2 @@\n context\n+one changed line\n", // tiny
+		NewFileContent: bodyWithTokens(bodyTokens),
+	}
+}
+
+func TestEnforceGroupTokenBudget_SplitsLargeBodySmallDiffGroup(t *testing.T) {
+	const tokenLimit = 160000 // == PromptTokenLimit(200000)
+
+	// The real observability bundle that overflowed a 160k per-item ceiling:
+	// six small-diff files whose measured body tokens total ~97k. Combined diff
+	// is trivial, so the OLD diff-only budget kept them as one item.
+	bundle := []struct {
+		path       string
+		bodyTokens int
+	}{
+		{"grafana-alerts-dev.yaml", 35308},
+		{"grafana-alerts-production.yaml", 35992},
+		{"grafana-dashboard-journey-health.json", 11091},
+		{"grafana-dashboard-router-pay.json", 6214},
+		{"grafana-alert-contract.test.ts", 6056},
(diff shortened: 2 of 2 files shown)


Qiyuanqiii (approved):
This addresses the gap described in #1171 at the right layer. Accounting for the file bodies that may be pulled into the review context prevents small-diff/large-body groups from being dispatched as a single item only to fail later during compression.

The fallback to per-file review is consistent with the existing behavior, and the regression tests cover both the overflowing and healthy cases as well as the single-file boundary.

I don't see any blocking issues with the implementation. Looks good to me.

internal/agent/grouping.go
```diff
@@ -21,6 +21,20 @@ import (
...
+// remainder is reserved for the review's own working set — reasoning, tool-call
+// framing, and the compression summaries that share the context window with the
+// file bodies the agent reads. Reviewing a patch means reading the surrounding
+// file body via file_read, so a group's real cost is dominated by member bodies,
+// not diffs; a group whose bodies alone approach the budget cannot also hold that
+// working set and overflows ("context compression exceeded its threshold").
+//
+// Calibrated against observed runs: a 6-file group totaling ~97k body tokens
+// overflowed a 160k per-item ceiling, while ~77–80k groups reviewed cleanly. 0.55
+// (≈88k of a 160k budget) splits the former with margin and keeps the latter.
```
wu21-web: Can you clean these up? They seem long and pointless.

beillahi: Done — trimmed all three comments in `grouping.go` to the essentials and dropped the calibration narrative (that context lives in the PR description and #1171). Force-pushed.

lizhengfeng101:
Thanks for the detailed write-up and the PR — the diff-vs-body observation is a fair one. We've decided not to take this approach, and are closing both this PR and #1171.

The reason: a group's review context is built dynamically. File bodies are read on demand via `file_read` (the initial prompt carries only diffs), and the review can also pull in content via `code_search` and cross-file reads. So a pre-dispatch `diff + NewFileContent` estimate can't reliably predict review-time context — it's neither an upper bound (search / cross-file reads exceed it) nor a tight estimate (on-demand / partial reads + compression reduce it) — and the `0.55` fraction, fit to two observed data points, would over-split healthy multi-file groups and lose cross-file review context.

The runtime-degrade alternative doesn't really resolve it either: reading a file's body is necessary work to review it, not optional overhead, so splitting a group after it overflows just re-batches the same necessary tokens after paying for a failed attempt. Where the necessary cost genuinely exceeds a run's budget, that is better handled with operator-level controls (e.g. `--max-tokens`, per-file / concurrency settings) than with grouping heuristics. Appreciate the investigation.

PR #1199, opened 2026-09-08
feat(rules): add Rego policy review support

brianluby:
## Description

I am currently building a few tools that use OPA files that contain policy as code. I would like to enable default code reviews for Open Policy Agent (OPA) `.rego` files, which
are currently excluded by the extension allowlist.

- Register `.rego` and add Rego-specific review guidance covering authorization,
  undefined values, trust boundaries, iteration, and evaluation behavior.
- Add lowercase/uppercase allowlist and rule-resolution test cases.
- Update the review-rule mapping documentation in all five locales.

The guidance requires concrete evidence for security findings and recognizes
valid production `with` expressions, partial evaluation, and HTTP defaults.

## Type of Change

- [x] Documentation update

## How Has This Been Tested?

- [x] `make test` passes locally, including race detection.
- [x] Manual testing: `go run ./cmd/opencodereview rules check policies/authz.ReGo`
  selects `**/*.rego` and displays the embedded Rego guidance.
- [x] I have run it through code reviews with daybreak blue model

## Checklist

- [x] Coding style and validation checks pass (`make check`).
- [x] Self-review completed.
- [x] Tests added for the new behavior.
- [x] New and existing unit tests pass locally.
- [x] Relevant documentation updated.
- [x] Contributor License Agreement signed.

## Related Issues

No related issue linked.


brianluby:
I wanted to provide some examples of the rego policies being corrected. I am building a threat modeling tools that used IaC and source code to automatically create threat models.

```
─── crates/threatmitigator-rego/policies/rules/availability/kubernetes_dos_001.rego:27-28 ───
[maintainability · low] The supported workload types are now declared twice: once as the inline
membership set and once as the keys of `workload_kinds`. Adding a new alias (e.g.
`kubernetes_pod_v1` style additions) requires editing both, and a drift would either silently skip
the type or panic-free fail the map lookup. The membership check is redundant here —
`workload_kinds[resource.type]` is already undefined for any type outside the map, so the body fails
on its own.

-       resource.type in {"kubernetes_pod", "kubernetes_pod_v1", "kubernetes_deployment", "kubernetes_deployment_v1"}
        workload_kind := workload_kinds[resource.type]
```
       
```
─── crates/threatmitigator-rego/policies/lib/aws/iam.rego:25-27 ───
[bug · medium] The new shape normalization is only wired into `has_wildcard_actions`,
`has_wildcard_resources` and the two principal helpers. The other policy helpers in this same module
— `has_admin_access`, `requires_mfa_condition`, `allows_pass_role` and `allows_policy_modification`
— still index `policy_doc.Statement[_]`, `statement.Action[_]` and `statement.Condition.Bool[...]`
directly. They are called with exactly the same attribute (`resource.attributes.policy`) by
AWS-PRIV-004/-005/-007 and AWS-PRIV-012, so a `jsonencode()`-produced JSON string policy or the
single-string `Action`/`Condition` form that AWS explicitly accepts (and which appears in fixtures
such as `tests/fixtures/terraform/vulnerable/elevation_of_privilege/wildcard_role_policy.tf`,
`Action = "sts:AssumeRole"`) yields no match there, i.e. the scalar/string shapes this PR set out to
fix are still silently missed by those rules. Please route those helpers through
`policy_document`/`as_array` as well (or add a note about why they intentionally keep the strict
object/array form).

```

internal/config/rules/rule_docs/rego.md
```diff
@@ -0,0 +1,47 @@
+> Favor precision over recall: only raise an issue when you are confident it is a real defect, and stay silent when the surrounding context is unclear — a false alarm costs more reviewer trust than a missed minor issue. Treat authorization-bypass and data-exposure findings as blocking, and style or idiom suggestions as non-blocking. Ground findings in the Rego under review and observable caller, schema, and configuration evidence. Do not assume the runtime shape or provenance of `input` or `data`, the OPA/Regal version, or enforcement behavior (fail-open vs fail-closed) when that context is unavailable.
+
```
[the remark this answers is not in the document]
brianluby: I fixed the issue idenfied in [30573bb](https://github.com/alibaba/open-code-review/pull/1199/commits/30573bb8a5f93161d9c91cb8f8c4987c42ea0e2e)

internal/config/rules/rule_docs/rego.md
```diff
@@ -0,0 +1,48 @@
...
+- Unification (`=`) used where comparison (`==`) was intended, silently binding a variable instead of testing it — especially inside `with` overrides or function arguments
+- A rule body that is undefined (no value) on a reachable path where the caller expects `false` or an empty set, flipping the downstream default
+- Negation over a non-ground term (`not some_unbound_var == value`) whose scope is wider than intended
+
+#### Unsafe Variables and Iteration
+- Unbound or unsafe variables the compiler would reject being "fixed" by binding them to attacker-influenced `input` instead of a trusted value
+- `some x in input.collection` iterating a field that may be absent or a scalar rather than an array/set/object, making the rule undefined instead of denying
+- Partial iteration that checks only the first match (e.g. `input.roles[i] == "admin"` binding `i` once and succeeding) where every element needed checking; `every` or a comprehension with aggregation was required
+- Object key iteration (`some k, v in input.map`) that assumes keys present without `object.get` defaults or presence checks
```
Qiyuanqiii: `some k, v in input.map` enumerates key/value pairs that actually exist; it neither assumes absent keys nor requires `object.get`. `object.get` is for explicitly reading a key and providing a default when that lookup is missing. As written, this will teach the reviewer to flag valid object enumeration. Please remove this statement or rewrite it around a demonstrated optional-collection/schema problem. See the official [`some` object-iteration example](https://www.openpolicyagent.org/docs/policy-reference/keywords/some) and [`object.get`](https://www.openpolicyagent.org/docs/policy-reference/builtins/object).

internal/config/rules/rule_docs/rego.md
```diff
@@ -0,0 +1,48 @@
...
+
+#### Non-Determinism and Functions
+- Non-deterministic values or final decisions incorrectly reused across evaluations, with evidence that freshness or determinism requirements are violated; caching a compiled policy or using partial evaluation is not itself a defect, since calls such as `time.now_ns()` may remain in residual queries for runtime evaluation
+- `http.send` settings that demonstrably disable required safeguards (e.g. `tls_insecure_skip_verify: true` or a zero timeout on a path that requires a bound), or an error path that changes the authorization decision incorrectly; omitted options retain OPA's defaults of a five-second timeout and TLS certificate verification
+- Recommend HTTP response caching only for an established performance problem with an acceptable freshness contract; uncached requests are valid when decisions require fresh data
+- A custom function that is undefined for some inputs (no return value on a reachable path) used as though it always returns a value
+- `print()` or `trace()` calls left in committed policy, leaking evaluated values into logs
```
Qiyuanqiii: `print()` and `trace()` should not be grouped as unconditional log leakage. In server mode, `print()` can write to the output/log stream, while `trace()` emits a `Note` event in the query explanation. A trace can expose sensitive values when tracing/explanations are collected, but that condition needs to be established. Please split the guidance so ordinary logging is not attributed to every `trace()` call. See OPA's [`print`](https://www.openpolicyagent.org/docs/policy-reference/builtins/opa) and [`trace`](https://www.openpolicyagent.org/docs/policy-reference/builtins/tracing) documentation.

Qiyuanqiii (changes requested):
Thanks for adding first-class Rego support. The allowlist/rule-resolution wiring, regression tests, locale documentation, and the new principles heading all look good.

I am requesting changes on the current head because `rego.md` is executable review guidance: inaccurate statements here can repeatedly turn valid Rego into false-positive findings. The two inline comments cover the blocking cases.

Please also tighten two adjacent statements while updating the guide:

- The version guidance should recognize `import rego.v1` (available on supported pre-v1 toolchains) and that `in`, `every`, `if`, and `contains` require no import under OPA v1 syntax. The current wording mentions only matching `future.keywords` imports and omits `every`.
- `input.roles[i] == "admin"` is existential: Rego searches possible bindings for `i` and succeeds if any element matches. Please describe this as an existential check used where universal quantification was intended, rather than as checking only the first match.

These are precision issues rather than objections to the integration itself. Once the guidance is corrected, I think the PR will be ready from a code-review perspective. The current GitHub workflows are still `action_required`, so they will also need to run before merge.

internal/config/rules/rule_docs/rego.md
brianluby: Thanks for the feedback I have made the appropriate changes and added them to the PR [f1eda81](https://github.com/alibaba/open-code-review/pull/1199/commits/f1eda81851f938b84092d633e4c849510f3cb689). This has been a very encouraging process and I love the tool. It has been excellent to use to help assist it code reviews and especially from a security front. I have another app that I built as part of a hackathon which is a tool that takes text based security policies and converts them to OSCAL machine readable policies. I think my next PR will be adding first class support for OSCAL. Here is a link to that project. I am currently extending it to help with the process of creating policies. https://github.com/policy-forge/forge

internal/config/rules/rule_docs/rego.md
brianluby: Fixed in [f1eda81](https://github.com/alibaba/open-code-review/pull/1199/commits/f1eda81851f938b84092d633e4c849510f3cb689)

wu21-web (approved):
You are among a few people that remembered to update the documentation. Good job! 👍

lizhengfeng101 (approved):
LGTM

PR #1203, opened 2026-09-09
feat(action): add llm_protocol input to unlock openai-responses preset

psuckrow-routine:
## Problem

The CLI has supported the `openai-responses` protocol preset since v1.10.1 (`fix(llm): serve GPT-5.6 models via the OpenAI Responses API`, #938), but the composite action still derives `llm.protocol` from `llm_use_anthropic` only — a two-value selector that can produce `openai` or `anthropic` but never `openai-responses`.

As a result, callers running GPT-5.6 Sol/Terra/Luna through the composite action get an instant HTTP 400 from `/v1/chat/completions`:

> Function tools with reasoning_effort are not supported for gpt-5.6-terra in /v1/chat/completions.

Same class of failure documented in #559. Fix in the CLI exists; wrapper has no way to reach it.

## Change

Additive — no existing caller changes behavior.

- **New optional input `llm_protocol`** — accepts `openai`, `openai-responses`, or `anthropic` (case-insensitive). Empty (default) keeps today's derivation from `llm_use_anthropic`.
- **Validation** in the \`Validate inputs\` step, matching the fail-fast conventions already used for \`effort\` and \`llm_reasoning_effort\`.
- **Configure OCR** — when \`llm_protocol\` is non-empty it wins over \`llm_use_anthropic\`; \`OCR_USE_ANTHROPIC\` is derived from the resolved protocol so the CLI env var stays consistent. The existing \`reasoning_effort × anthropic-protocol\` guard still fires correctly because it reads the resolved \`OCR_LLM_PROTOCOL\`.
- **Checkpoint fingerprint** — \`llm_protocol\` added to the fingerprint so switching protocols invalidates prior checkpoints (a protocol change routes requests to a different endpoint and, for \`openai-responses\`, a different request shape).

## Usage after this lands

\`\`\`yaml
- uses: alibaba/open-code-review@main
  with:
    llm_url: https://api.openai.com/v1/responses
    llm_auth_token: \${{ secrets.OPENAI_API_KEY }}
    llm_model: gpt-5.6-terra
    llm_use_anthropic: 'false'
    llm_protocol: openai-responses
\`\`\`

## Test

Ran end-to-end against \`gpt-5.6-terra\` on a live PR from a workflow that wraps this action — direct \`curl\` probe to \`/v1/chat/completions\` returned 200 (model is available), \`ocr review\` still 400d before this patch (function-tool + reasoning_effort combination), and succeeded with inline + sticky summary posted after the patch with \`llm_protocol: openai-responses\` + the \`/v1/responses\` URL.

Legacy callers (no \`llm_protocol\` set) exercise the unchanged \`llm_use_anthropic\` branch.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -19,12 +19,14 @@ inputs:
     required: true
   llm_use_anthropic:
     description: >-
-      Selects the LLM protocol (mapped to env OCR_USE_ANTHROPIC). An explicitly supplied empty
-      string, true, 1, or yes selects Anthropic
+      Legacy protocol selector (mapped to env OCR_USE_ANTHROPIC). An
+      explicitly supplied empty string, true, 1, or yes selects Anthropic
       case-insensitively; every other value selects the OpenAI-compatible
       protocol, preserving the CLI environment contract. Ignored when
-      llm_protocol is set to a non-empty value.
-    required: true
+      llm_protocol is set to a non-empty value — callers using llm_protocol
+      can omit this entirely.
+    required: false
+    default: ''
   llm_protocol:
     description: >-
       Explicit LLM protocol override (case-insensitive): one of `openai`,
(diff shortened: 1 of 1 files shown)


Qiyuanqiii (commented):
Looks good to me overall. The two issues from the earlier OCR review have been addressed, and the explicit "llm_protocol" path now looks consistent with the existing CLI behavior.

One minor non-blocking note: when "llm_protocol" is explicitly set, "llm_use_anthropic" no longer affects the resolved configuration, but it is still included in the checkpoint fingerprint. That can cause an unnecessary full review if only the ignored legacy value changes. I don't think this needs to block the PR, though.

From the implementation side, this looks good to go once CI and CLA are clear.

wu21-web:
This is a duplicate.

psuckrow-routine:
@wu21-web duplicate of which PR?

wu21-web:
@psuckrow-routine https://github.com/alibaba/open-code-review/pull/1192

lizhengfeng101:
duplicate with https://github.com/alibaba/open-code-review/pull/1192

PR #1212, opened 2026-09-10
fix(frontend): generate provider presets from Go registry

### LINKED ISSUE #1211: refactor(llm): make Go provider registry the single source of truth for the VS Code preset
## Description

|---|---|---|

## Scope

## Proposed approach

## Acceptance Criteria

## Context

Qiyuanqiii:
### Summary

The extension's hand-maintained provider list can drift from the Go registry: a built-in CLI provider may appear as custom in the configuration UI, and the UI may offer outdated models or fallback defaults. This change generates the frontend presets from the Go registry and checks that the committed artifact stays synchronized.

The implementation follows the shared frontend layout introduced by #1031. The generated catalog, helpers, configuration form, and frontend regression tests now live under `extensions/frontend/`, which supplies both the VS Code and IDEA webviews. The generator directive, artifact test, and CI check all use the new path.

- Synchronize all 28 Go providers at base revision `85cecfe5f935da2b2aae8f91ce4fee8ed343a681`, including 13 missing frontend presets and model-list differences in 11 existing presets.
- Preserve Go model order and the first-model fallback. Keep the generated provider array name-sorted; reuse display-name-sorted options in the configuration dropdown.
- Preserve Responses and Bedrock metadata, including API-key field visibility based on the effective protocol and saved overrides.
- Align the IDEA host's built-in names and consistency test with the generated catalog so newly exposed providers are saved to the correct configuration section.
- Keep the subsequent option-reuse and regeneration-documentation improvements from #1298 in this PR.

### Generation and verification

```text
internal/llm/providers.go
  -> llm.ListProviders()
  -> go generate ./internal/llm
  -> extensions/frontend/src/shared/providers.generated.ts
  -> shared/providers.ts exports and frontend consumers
```

The generator under `internal/llm/gen/` imports the actual Go registry. It does not parse Go source, maintain a second provider allowlist, resolve user configuration, or contact provider APIs. The emitted fields are `name`, `displayName`, `protocol`, `baseUrl`, `authHeader`, `envVar`, `ambientAuth`, and `models`; `envVar` contains only the variable name.

`providers.ts` retains the existing named exports, lookup helpers, and model-list merging helper. The generated file imports `OcrProviderPreset` as a type only. The checked-in artifact lets extension builds consume the presets without installing Go.

The generator requires an explicit `-output` argument; the `go:generate` directive supplies it. An omitted or empty argument fails with a diagnostic explaining the supported regeneration command. Output is written to a temporary file in the destination directory, flushed, closed, and renamed over the destination. Repeated replacement was verified on Windows; this does not claim that `os.Rename` provides atomic replacement on every platform.

CI runs `go generate ./internal/llm` and checks the tracked generated file with `git diff --exit-code`. A stale file produces an error message containing the regeneration command and new frontend path. The Go artifact test independently compares the committed bytes with freshly rendered registry data without rewriting them.

### Catalog and fallback changes

The base frontend has 15 presets; Go has 28. The newly exposed frontend presets are `bedrock`, `edenai`, `gemini`, `iflytek`, `kimi-global`, `litellm`, `mistral`, `novita`, `ollama-cloud`, `openai-responses`, `siliconflow`, `siliconflow-cn`, and `xai`. These providers already exist in Go.

| Existing provider | Models, before -> after | First model before | First model after |
| --- | ---: | --- | --- |
| `anthropic` | 4 -> 6 | `claude-opus-4-8` | `claude-opus-5` |
| `baidu-qianfan` | 6 -> 10 | `ernie-5.1` | `ernie-5.1` |
| `dashscope` | 4 -> 10 | `qwen3.7-max` | `qwen3.8-max` |
| `dashscope-tokenplan` | 12 -> 14 | `qwen3.7-max` | `qwen3.8-max` |
| `hy-tokenplan` | 1 -> 1 | `hy3-preview` | `hy3` |
| `kimi` | 3 -> 5 | `kimi-k2.7-code` | `kimi-k3` |
| `mimo` | 5 -> 2 | `mimo-v2.5-pro` | `mimo-v2.5-pro` |
| `tencent-tokenhub` | 1 -> 11 | `hy3-preview` | `deepseek-v4-pro` |
| `volcengine` | 3 -> 6 | `doubao-seed-2-0-lite-260428` | `doubao-seed-evolving` |
| `z-ai` | 4 -> 6 | `glm-5.2` | `glm-5.3` |
| `z-ai-coding` | 4 -> 5 | `glm-5.2` | `glm-5.3` |

The nine changed first entries affect fallback selection when there is no saved per-provider model. Saved models retain precedence, including the custom-model input for a saved model absent from the available options. Generation does not migrate or rewrite saved configuration. The generated provider array remains name-sorted with `anthropic` first; sorting the dropdown labels does not change that initial-provider behavior. The generator preserves model order and duplicates.

### Scope and compatibility

- Go registry values, provider transports, resolver behavior, and model validation are unchanged. `providers.go` gains only the generation directive and regeneration instructions.
- Existing frontend exports and configuration keys are preserved. Newly recognized built-ins use `providers.<name>.model`.
- The preset type represents all four protocols already supported by Go. The custom-provider editor's protocol choices are unchanged.
- The official form hides the API Key field for effective `anthropic-bedrock` authentication, including overrides from another preset. A Bedrock preset overridden to `openai` shows the field. AWS credential, profile, and region resolution remain the CLI's responsibility.
- The shared frontend change reaches both VS Code and IDEA through their existing build targets. The IDEA host's names-only set is aligned with the generated catalog, and its existing consistency test parses the generated JSON payload instead of the removed hand-written TypeScript array. A configuration regression checks that newly exposed presets are saved under `providers`, and the IDEA workflow now also runs for shared frontend changes. No host bridge or configuration-writing algorithm is changed.
- This PR keeps a read-only regeneration check. Workflow write permissions, automated commits/pushes, and an autofix bot remain separate work.

## Type of Change

- [x] Bug fix
- [x] CI / Build / Tooling
- [ ] Breaking change
- [ ] Refactoring without functional changes

The catalog and fallback-default corrections are user-visible.

## How Has This Been Tested?

This revision is rebased onto main `85cecfe5`. The three IDEA conflict resolutions preserve the upstream English comments and existing test names alongside the generated-catalog checks and built-in configuration-routing regression. Go checks, coverage, both frontend builds, and the frontend, VS Code, and IDEA test suites were rerun after rebasing.

Base: `85cecfe5f935da2b2aae8f91ce4fee8ed343a681`. Local environment: Windows amd64, Go 1.26.5, Node.js 24.14.0, npm 11.9.0, and Yarn 1.22.22. Make targets were invoked through GNU Make with Git Bash; the race detector used the installed GCC toolchain.

| Check | Result |
| --- | --- |
| `make license-add` / `make check` | Passed; no missing license headers, English check, formatting, tidy, and vet passed |
| `make test` | All 24 Go packages passed, including race detection |
| `make coverage` | Passed: 91.2% total against the 90% threshold |
| `go generate ./internal/llm`, twice | Passed; byte-identical output and no diff from the staged artifact |
| Frontend `npm run typecheck` | Passed |
| Frontend `npm test -- --runInBand` | 51 tests passed across both suites, including 34 provider regressions |
| Frontend production build, `OCR_TARGET=vscode` | Passed for webview and configuration panel |
| Frontend production build, `OCR_TARGET=idea` | Passed for webview and configuration panel |
| VS Code `yarn lint` / `yarn compile` | Passed |
| VS Code `yarn test --runInBand` | All 88 tests passed on Windows; the unmodified current base also passes all 88 |
| IDEA `gradlew test -PskipFrontend=true` | All 221 tests passed across 19 suites, with no skipped tests |
| IDEA `gradlew verifyPluginStructure -PskipFrontend=true` | Passed |
| Pre-commit OCR review | Previous feature reviews completed; the resolved IDEA source was reviewed again after rebasing, with 0 findings |
| `git diff --check` and LF verification | Passed |

Go regression coverage includes the artifact at its new frontend path, exact metadata mapping, escaping, model order and duplicate preservation, empty model arrays, LF output, repeatability, missing-output diagnostics, and contextual write failures. Frontend regressions cover name ordering, normalized lookups for every generated preset, built-in configuration paths, model merging, Bedrock/Responses metadata, protocol overrides, and unknown providers.

The full VS Code suite passes on both the PR revision and an unmodified archive of base `85cecfe5`, using the same dependencies. Main now sets the POSIX cancellation test platform explicitly. No files under `extensions/vscode/` are changed by this PR.

IDEA checks used the repository's Gradle 8.14.3 wrapper, JBR 21.0.8, and the locally installed PyCharm 2025.1.3.1 platform through `-PlocalIdePath`. That machine-specific override is not committed; CI retains its configured IntelliJ Community platform. The IDEA regression suites include catalog/name parity, normalized provider lookup, and the new built-in configuration-routing case.

The initial OCR feature reviews completed without failed files. The IDEA follow-up had one non-fatal invalid-range `file_read` tool diagnostic before completing. The subsequent rebase review completed with 0 findings in the resolved production file. Its default filters exclude generated artifacts and test files; those five files were inspected locally and covered by the checks above. The remaining low-priority tracked-file-only CI observation is documented below.

## Known Limitations

- Presets describe the source revision, not a live provider catalog. No provider endpoint or AWS credential chain was exercised. An older installed CLI may not recognize newer built-ins or models until updated.
- The CI diff check detects drift in the tracked artifact. If a later change deletes the tracked file, regeneration can create an untracked file that `git diff` alone does not detect; this PR does not broaden that guard.
- IDEA retains its existing names-only host set, now checked against the generated artifact in CI. Adding a built-in Go provider requires updating that set as well as regenerating the frontend catalog; model/transport metadata is not copied into the host.
- Builds and automated tests do not replace interactive verification in the VS Code Extension Host or IntelliJ IDEA. Those interactive checks and native Linux/macOS execution were not performed locally.

## Checklist

- [x] Existing maintainer-requested display-name sorting and independent name-order assertion are retained.
- [x] Generator, artifact tests, frontend code, and CI use `extensions/frontend/`.
- [x] Mainline registry values are preserved and the artifact was regenerated from them.
- [x] Automated tests and local diff inspection were completed.
- [ ] Contributor has reviewed the latest AI-assisted revision before requesting further maintainer review.

### AI/LLM Disclosure

AI was used when resubmitting and resolving the conflict with the main branch.
Codex (GPT-6 Astra) assisted with implementation, the shared frontend migration, validation, and this description. Automated pre-commit review used `ocr` 1.11.9 with the configured `deepseek-v4-flash` model; the provider reported `deepseek-flash` in the completed review responses.
I reviewed the changes made by AI, and I personally take responsibility for the code submitted.

## Related Issues

Closes #1211.


Qiyuanqiii (commented):
I did another pass focused specifically on readability and robustness. The overall structure looks good: the handwritten TypeScript surface is much smaller now, the generator is straightforward, and the generated-file flow is easy to follow.

I found one robustness issue that I think is worth fixing before merge:

- Ambient-auth UI behavior does not follow the effective protocol. "OfficialForm" currently hides the API Key field based only on "preset.ambientAuth". The CLI resolver, however, gives "providers.<name>.protocol" precedence: for example, "bedrock" overridden to "openai" is no longer ambient-auth and needs a token, while a provider explicitly overridden to "anthropic-bedrock" becomes ambient-auth. The VS Code UI should derive this from the saved entry protocol first, then fall back to the preset metadata, matching the CLI semantics. A small helper plus regression coverage for both directions would make this robust.

A few smaller, non-blocking maintainability notes:

- The CI guard runs "go generate ./..." even though it only verifies "providers.generated.ts". Scoping this to "go generate ./internal/llm" would avoid unrelated future generators changing the CI workspace.
- The generator writes the artifact directly with "os.WriteFile"; an atomic temp-file + rename would avoid leaving a truncated generated file if generation is interrupted. This is low risk because the file is reproducible.
- The TypeScript regression test assumes every preset has at least one model via "preset.models[0]", while the Go generator tests explicitly support empty model arrays. Handling the empty-list case explicitly would make the test contract clearer and avoid a future "undefined.trim()" failure.

None of the latter three are blockers on their own. The main thing I would address is keeping the VS Code ambient-auth behavior aligned with the CLI's effective-protocol rules.

AI review disclosure: This focused code-quality evaluation was performed with GPT-5.6 Sol (Extra High reasoning). The findings above were reviewed before posting.

lizhengfeng101:
Two small things from reading through this — the generation approach itself looks right, and the ambient-auth logic lines up with the resolver's `protocol == bedrock || (isPreset && preset.AmbientAuth && entry.Protocol == "")` rule, override precedence included.

1. `OfficialForm`'s provider `Select` sorts by `name` but renders `displayName`, which makes the dropdown look shuffled now that it is 28 entries — "Anthropic Claude API, Baidu Qianfan API, AWS Bedrock (Anthropic models), Alibaba DashScope API…" reads as unsorted. Could we sort by `displayName` at render time and leave the generated order alone?

2. `providers.ts` re-exports the generated array as-is, so `expect(PROVIDER_PRESETS).toEqual(generatedPresets)` in `src/shared/__tests__/providers.test.ts` compares the same object with itself and can never fail. Worth dropping, or replacing with an assertion that the list is name-sorted — that is the contract the generated header actually promises, and nothing checks it today.

lizhengfeng101:
and rebase main

Qiyuanqiii:
> and rebase main

Conflicts have been resolved, and the above‑mentioned issues have also been addressed.

lizhengfeng101:

Qiyuanqiii:
> Heads up — #1031 just landed on main and moved the VS Code webview and shared types into a new `extensions/frontend/` package, so please rebase onto the latest main and relocate any changes under `extensions/vscode/src/webview/` or `extensions/vscode/src/shared/` to `extensions/frontend/src/`.

@lizhengfeng101 A new review is needed.

Issue #674, opened 2026-08-01
希望支持ACP


Pheobe-Southwood:
### Problem Statement

在使用Paseo等工具时，需要通过ACP将open-code-review注册为一个单独的Provider，以便于多Agent协同工作

### Proposed Solution

无

### Affected Area

Other

stay-foolish-forever:
@Pheobe-Southwood Thanks for the suggestion!

We agree that ACP support would be a valuable addition, making OpenCodeReview easier to use with Paseo and other ACP-compatible clients. We have added it to our development roadmap and are currently evaluating the best implementation approach.

We don’t have a confirmed timeline yet, but we’ll share updates here as the work progresses. Thanks again for bringing this to our attention!

chaojixinren:
@stay-foolish-forever Hi! I'd like to work on this issue. Could you please assign it to me?

chaojixinren:
# [Design Proposal] ACP server (stdio) + ocr subprocess harness

## Motivation

OCR can only be invoked as a one-shot CLI process today. Every host that wants to integrate has to write its own plugin, there is no progress anyone can subscribe to at runtime (under `--format json` progress goes straight to `io.Discard`, so it exists on no stream at all), and cancelling mid-run leaves orphaned git processes behind.

Following what ACP v1 is for: **put a protocol adapter in front of the review engine**, so any compatible client (Paseo / Zed / JetBrains) uses one entry point to start a review, receive live progress, and cancel at any time, while the review logic under `internal/` stays untouched. Non-goal: turning OCR into a general-purpose conversational agent.

## Overall structure

```mermaid
flowchart TB
    C["ACP client<br/>Paseo / Zed / JetBrains"]

    subgraph acp["ocr-acp process (standalone binary)"]
        M["cmd/ocr-acp<br/>wire up and serve"]
        P["internal/acp/protocol<br/>SDK wrapper, generated types"]
        S["internal/acp/server<br/>session lifecycle · capabilities"]
        CMD["internal/acp/command<br/>slash commands + NL parsing"]
        H["internal/acp/ocrproc<br/>subprocess harness"]
    end

    C <==>|"stdio · newline-delimited JSON-RPC 2.0"| M
    M --> P --> S
    S --> CMD
    S --> H
    H ==>|"os/exec process boundary"| R["ocr review --format json --audience human"]
    R --> ENG["internal/ review engine<br/>no new import edge"]
```

**Key trade-offs**:

[...]

PR #679, opened 2026-08-02
[WIP] feat: add ACP support

stay-foolish-forever:
## Description

Add Agent Client Protocol (ACP) support to the H2 2026 roadmap.

This draft PR serves as a planning placeholder while the implementation approach is evaluated. The current direction is to expose OpenCodeReview as a specialized review agent for Paseo and other ACP-compatible clients through a decoupled ACP server adapter.

This PR does not implement ACP support yet.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring (no functional changes)
- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Documentation-only change; verified with `git diff --check`.

- [ ] `make test` passes locally
- [x] Manual testing (describe below)

## Checklist

- [x] My code follows the project's coding style (`go fmt`, `go vet`)
- [x] I have performed a self-review of my code
- [ ] I have added tests that prove my fix is effective or my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] I have updated the documentation accordingly (if applicable)
- [x] I have signed the CLA

## Related Issues

Tracks #674.

PR #1218, opened 2026-09-11
feat(acp): support ACP service for OpenCodeReview

chaojixinren:
## Current Stage

This PR adds an independent ACP stdio adapter around the existing OCR CLI. **Phases 1-8 are complete for the agreed scope; phase 9 delivery review and release handoff remain in progress.** The current head is `210f52b`.

The final implementation uses existing CLI logs for live activity and elapsed time, and existing JSON output for findings, coverage, usage and failure details. It does not change the root `cmd/opencodereview/` or `internal/` code relative to the PR baseline. Custom structured group-stage progress, exact phase/model-wait timers and the associated CLI changes have been removed. Windows Job Objects, cancellable stdio and path handling are retained.

Current-head ACP cross-compilation checks passed for Linux arm64, macOS amd64/arm64 and Windows amd64/arm64. ACP tests, repository CI and Pages checks were still running or queued when this description was updated; this is not a claim that all current-head checks have passed. The ACP workflow currently has no native Windows runtime job. Final review, target-platform/client verification and installation/upgrade checks against published release artifacts remain handoff work.

## ACP Module Architecture

The module keeps the ACP protocol boundary independent from OCR's core review engine. The adapter owns client-facing sessions and orchestration, while the existing `ocr` binary remains responsible for review execution and its review LLM.

```mermaid
flowchart LR
    C["ACP clients<br/>Paseo / Zed / JetBrains"]
    T["JSON-RPC 2.0 transport<br/>stdio"]
    S["ocr-acp<br/>ACP Server"]
    SE["Session and command layer<br/>available_commands_update<br/>session/prompt / session/cancel"]
    IP["Intent parser<br/>slash commands / natural language<br/>clarification state"]
    O["OCR process orchestrator<br/>timeouts / cancellation / cleanup<br/>stdout-stderr isolation"]
    CLI["Existing ocr CLI<br/>review / scan"]
    R["Existing CLI output<br/>stderr logs / stdout JSON"]

    C <--> T
    T <--> S
    S --> SE
    SE --> IP
    SE --> O
    O --> CLI
    CLI --> R
    R --> O
    O --> SE
    SE --> S
    S --> T
```

The client sends prompts and commands through the transport. `ocr-acp` parses them into a review or scan intent, asks for clarification when required, launches the existing OCR CLI as a child process, and streams progress and structured results back through the same ACP session. Cancellation and timeout signals flow into the process orchestrator, which is responsible for terminating and cleaning up the managed process tree before delivering the final outcome.

The implemented transport is stdio and the exposed OCR operations are review and scan. Streamable HTTP and OCR session-command integration are outside this PR's scope.

## Why ACP Support

OpenCodeReview (OCR) is Alibaba's open-source AI code review CLI, implemented in Go under the Apache-2.0 license. It has been used by tens of thousands of developers inside Alibaba and has found millions of code defects. OCR uses a hybrid architecture of deterministic engineering and agents: file selection, file-level context isolation, rule matching, comment locations, and reflection are protected by deterministic engineering logic, while agents handle dynamic decisions and context retrieval. This improves accuracy and reduces token consumption under the same model.

Today, OCR maintains a separate plugin or skill for each host, such as Claude Code, Codex, Cursor, and OpenCode. This makes integration work grow with every new host and makes behavior and documentation harder to keep consistent. OCR also cannot currently be discovered or consumed by ACP clients such as Paseo, Zed, and JetBrains ACP Agent Registry, so it has no standard entry point for multi-agent workflows where a general coding agent writes code and OCR acts as a dedicated review agent.

Agent Client Protocol (ACP) is an open protocol for connecting editors/clients with AI agents, with a role similar to LSP. A client implements ACP once and can then use any compatible agent. The client communicates with the agent through a JSON-RPC 2.0 session, sends prompts and commands, receives streaming progress and tool-call updates, gets structured results with file/line locations, and can cancel an in-flight session. The agent only needs to provide an ACP server. This lets OCR replace “one host, one integration” with “one server, every compatible client” and enables standard multi-agent collaboration.

## Existing Work and Motivation

OCR already provides:

- CLI commands: `ocr review`, `ocr scan`, `ocr session`, and `ocr delegate`;
- a VS Code extension;
- CI/CD integrations such as GitHub Actions and GitLab CI;
- integrations with multiple LLM providers; and
- host-specific plugins or skills for Claude Code, Codex, Cursor, and OpenCode.

Issue #674 requested registering OCR as an independent provider through ACP for tools such as Paseo. PR #679 added an “ACP server adapter” to the H2 2026 roadmap, with the explicit direction to expose OCR as a dedicated review agent while keeping it decoupled from the core review engine. That roadmap PR is a planning placeholder and contains no implementation.

## Design and Scope of This PR

This work follows the roadmap boundary: an independent ACP adapter process wraps the existing `ocr` binary instead of embedding protocol code into OCR's internal review engine.

- The adapter now provides a standalone ACP Server executable and communicates with clients using JSON-RPC 2.0 over stdio. Streamable HTTP remains future work.
- The adapter invokes `ocr review` and `ocr scan` as child processes and consumes their structured JSON output; OCR session-command integration remains outside the implemented scope.
- The adapter owns ACP sessions, prompt parsing, clarification, progress/display updates, cancellation, timeout handling, and protocol-channel hygiene.
- OCR continues to own review execution and the review LLM; the adapter does not import OCR internal packages, duplicate OCR's provider system, or change its source or prompts.
- Progress uses one collapsible execution entry with the latest CLI activity, total elapsed time, command and bounded logs. No custom CLI progress protocol or inferred group-stage timers are required.
- Windows uses Job Objects and cancellable stdio. Graceful CTRL_BREAK delivery requires a shared console; without one, cancellation relies on Job Object termination and may not preserve a partial JSON result.
- Release builds cover Linux, macOS and Windows on amd64/arm64. Documentation is synchronized across English, Chinese, Japanese, Korean and Russian.
- `available_commands_update` now exposes `/review`, `/scan`, and parameter hints so clients can discover the supported entry points.

## Staged Plan

| Phase | Scope | Status |
| --- | --- | --- |
| 1 | CLI investigation | Complete |
| 2 | Architecture and ownership boundaries | Complete |
| 3 | CLI contract and mock OCR | Implemented in this PR |
| 4 | Prompt parsing and clarification | Implemented in this PR |
| 5 | OCR process orchestration | Implemented in this PR |
| 6 | ACP server, session lifecycle, and real-client smoke test | Complete; official SDK smoke and manual Zed/VSCode ACP client acceptance completed |
| 7 | End-to-end, race, coverage, static, vulnerability, and platform regression | Prior macOS/Linux validation complete; latest local regressions pass; current-head CI still pending at update time |
| 8 | Real OCR client integration, packaging, and release preparation | Complete for the agreed scope; prior Zed/VSCode acceptance and Docker evidence recorded; log-based display, five-language documentation and Linux/macOS/Windows release builds implemented |
| 9 | Delivery review and release handoff | In progress; final review and published-artifact installation/upgrade evidence remain |

## Final Project Goal

Deliver a usable, tested, and documented standalone ACP Server for OCR. A user should be able to configure one command in Paseo or another compatible ACP client, use natural language or slash commands to start a review or scan, see streaming progress and structured line-level findings in the client UI, and cancel the operation at any time. The completed work will land the ACP roadmap item in the upstream `alibaba/open-code-review` repository and close issue #674.

## Type of Change

## How Has This Been Tested?

- [x] Local root-module `make check`, `make test` and `make build` passed after restoring CLI source to the PR baseline.
- [x] Local ACP `make check`, `make test` and `make build` passed, including the two stderr regression fixes.
- [x] ACP race-enabled coverage passed at **91.1%** after removing the custom progress protocol; the subsequent stderr fixes passed the full race-enabled test suite.
- [x] Regression tests reproduce and fix false truncation warnings for lines larger than the 32 KiB read buffer but below the 64 KiB line limit, and preserve the pending log event before a non-EOF read error. Over-limit truncation remains tested.
- [x] Official Python ACP SDK smoke passed using ordinary CLI logs and final JSON: handshake, command discovery, review/scan, findings and locations, no-findings completion, partial coverage/budget failure, ResourceLink, cancellation, repeated cancellation and EOF cleanup.
- [x] Opt-in real CLI binary contract tests passed against the rebuilt CLI without the removed source changes.
- [x] Local Windows amd64/arm64 adapter builds passed; current-head ACP CI cross-compilation also passed for Windows amd64/arm64, Linux arm64 and macOS amd64/arm64.
- [x] Documentation site production build passed with two asset/entrypoint size warnings; `git diff --check` and LF checks passed.
- [ ] Tag-triggered release execution and published-artifact installation/upgrade verification remain pending.

Earlier evidence is revision-specific: Linux ACP CI passed at `c9279e7` ([run](https://github.com/alibaba/open-code-review/actions/runs/35177361930)); Docker validation at that revision covered Linux arm64 and QEMU-emulated amd64, plus riscv64 build/start and core no-race tests. RISC-V race execution was blocked by QEMU ThreadSanitizer VMA support. Descendant-cleanup tests intermittently failed their deadlines in containers and passed reruns; their cause was not conclusively established. These earlier runs do not establish that the final revision has passed the same runtime matrix.

### RISC-V support and validation

Linux riscv64 has experimental source-build compatibility through the existing Unix implementation. Build both OCR CLI and ACP from source (`make build` and `make -C acp build`); the release workflow currently packages amd64/arm64 only and ACP CI has no riscv64 job.

Historical Docker/QEMU validation on 2026-09-17 used Go 1.26.5 at `c9279e7`: both binaries built and started, and all five ACP internal packages passed without race (93.6% coverage). Command-package tests passed except the excluded `TestRealBinaryUnreadStdoutExits`, which builds a race-enabled child. Go could compile race binaries, but QEMU could not start them (`ThreadSanitizer: unsupported VMA range`, `Found 49 - Supported 39 and 48`). Native RISC-V hardware, Python SDK smoke, graphical clients and the full OCR CLI test suite remain unverified on this architecture. These results do not establish runtime validation of the latest revision.

The author previously confirmed manual Zed and VSCode ACP client acceptance for the tested installations. The final simplified display still needs its own graphical-client confirmation. Mock protocol tests do not establish real OCR/LLM review quality. Parser tests use deterministic fakes; real-endpoint coverage is opt-in via `OCR_ACP_REAL_LLM=1`. Python smoke provisioning may download the pinned Python 3.12.13 runtime and SDK dependencies.

The final documentation update did not run OCR review, as requested by the author. Detailed local review/acceptance records remain under `temp/`, not in an `acp/ACCEPTANCE.md` file. This is a nested Go module, so its gates are `make -C acp ...` and its workflow is `.github/workflows/acp-ci.yml`.

## Checklist

- [x] I have added tests that prove my feature works
- [x] I have signed the CLA

## Related Issues

- Related to #674 (ACP provider support request)
- Related roadmap: #679

## AI / LLM Disclosure

This PR was developed with assistance from OpenAI Codex gpt-6-astra. The implementation and tests were reviewed locally and with the repository's OpenCodeReview automation. No external LLM API or API key is required for the test suite; parser LLM integration is covered by deterministic fakes, with real-endpoint coverage remaining opt-in.

### DIFF
--- a/acp/internal/intent/natural.go
+++ b/acp/internal/intent/natural.go
@@ -1,0 +1,162 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 alibaba/open-code-review Contributors
+
+package intent
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"io"
+	"strings"
+)
+
+// parseNatural turns free text into an intent through one LLM tool call. There
+// is no deterministic fallback: a failed call becomes a clarification or a
+// rejection, never a keyword match.
+func (p *Parser) parseNatural(ctx context.Context, text string, st *State) (Result, error) {
+	if p.llm == nil {
+		return RejectResult("natural-language parsing is unavailable because no parsing LLM is configured",
+			"Use /review or /scan."), nil
+	}
+
+	req := LLMRequest{
+		System: systemPrompt,
+		User:   buildUserPrompt(text, st),
+		Tool:   SubmitIntentTool(),
+	}
+
+	callCtx := ctx
+	if p.timeout > 0 {
+		var cancel context.CancelFunc
+		callCtx, cancel = context.WithTimeout(ctx, p.timeout)
+		defer cancel()
+	}
+
+	call, err := p.llm.CallTool(callCtx, req)
+	if err != nil {
+		return RejectResult("could not parse the request with the LLM: "+err.Error(),
+			"Use /review or /scan, or try again."), nil
+	}
+	if call.Name != submitIntentToolName {
+		return RejectResult("the LLM returned an unexpected tool: "+call.Name,
+			"Try again, or use /review or /scan."), nil
+	}
+
+	var raw rawIntent
+	dec := json.NewDecoder(bytes.NewReader(call.Arguments))
+	dec.DisallowUnknownFields()
+	if err := dec.Decode(&raw); err != nil {
+		return RejectResult("the LLM returned a malformed intent: "+err.Error(),
+			"Try again, or use /review or /scan."), nil
+	}
+	if err := ensureEOF(dec); err != nil {
+		return RejectResult("the LLM returned trailing data after the intent: "+err.Error(),
+			"Try again, or use /review or /scan."), nil
+	}
+
+	return p.applyRaw(ctx, st, &raw)
+}
+
+func ensureEOF(dec *json.Decoder) error {
+	var extra json.RawMessage
+	err := dec.Decode(&extra)
+	if err == nil {
+		return errors.New("extra JSON value")
+	}
+	if errors.Is(err, io.EOF) {
+		return nil
+	}
+	return err
+}
+
+func (p *Parser) applyRaw(ctx context.Context, st *State, raw *rawIntent) (Result, error) {
+	switch raw.Action {
+	case actionReview:
+		if raw.Review == nil {
+			p.clearPending(st)
+			return RejectResult("the LLM returned action review without a review object",
+				"Try again, or use /review."), nil
+		}
+		s := &pendingSlots{action: actionReview, extra: append([]string(nil), raw.Extra...)}
+		switch raw.Review.Type {
+		case reviewWorkspace:
+			s.reviewType = reviewWorkspace
+		case reviewRange:
+			s.reviewType = reviewRange
+			s.from, s.to = raw.Review.From, raw.Review.To
+		case reviewCommit:
+			s.reviewType = reviewCommit
+			s.commit = raw.Review.Commit
+		default:
+			p.clearPending(st)
+			return RejectResult("the LLM returned an unknown review type: "+raw.Review.Type,
+				"Try again, or use /review."), nil
+		}
+		return p.finalizeReview(ctx, st, s)
+
+	case actionScan:
+		if raw.Scan == nil {
+			p.clearPending(st)
+			return RejectResult("the LLM returned action scan without a scan object",
+				"Try again, or use /scan."), nil
+		}
+		s := &pendingSlots{
+			action: actionScan,
+			paths:  append([]string(nil), raw.Scan.Paths...),
+			extra:  append([]string(nil), raw.Extra...),
+		}
+		return p.finalizeScan(ctx, st, s)
+
+	case actionClarify:
+		if strings.TrimSpace(raw.Question) == "" {
+			p.clearPending(st)
+			return RejectResult("the LLM returned clarify without a question",
+				"Try again, or use /review or /scan."), nil
+		}
+		if partial := slotsFromRaw(raw); partial != nil {
+			p.storePending(st, partial, raw.Missing)
+		}
+		return ClarifyResult(raw.Question, raw.Missing...), nil
+
+	case actionReject:
+		if strings.TrimSpace(raw.Reason) == "" {
+			p.clearPending(st)
+			return RejectResult("the LLM returned reject without a reason",
+				"Try again, or use /review or /scan."), nil
+		}
+		p.clearPending(st)
+		return RejectResult(raw.Reason, raw.Hint), nil
+
+	default:
+		p.clearPending(st)
+		return RejectResult("the LLM returned an unknown action: "+raw.Action,
+			"Try again, or use /review or /scan."), nil
+	}
+}
+
+// slotsFromRaw preserves whatever partial slots a clarify call carried so the
+// next turn can merge into them.
+func slotsFromRaw(raw *rawIntent) *pendingSlots {
+	if raw.Review != nil {
+		s := &pendingSlots{action: actionReview, extra: append([]string(nil), raw.Extra...)}
+		switch raw.Review.Type {
+		case reviewRange:
+			s.reviewType, s.from, s.to = reviewRange, raw.Review.From, raw.Review.To
+		case reviewCommit:
+			s.reviewType, s.commit = reviewCommit, raw.Review.Commit
+		default:
+			s.reviewType = reviewWorkspace
+		}
+		return s
+	}
+	if raw.Scan != nil {
+		return &pendingSlots{
+			action: actionScan,
+			paths:  append([]string(nil), raw.Scan.Paths...),
+			extra:  append([]string(nil), raw.Extra...),
+		}
+	}
+	return nil
+}
--- a/acp/internal/intent/natural.go
+++ b/acp/internal/intent/natural.go
@@ -79,16 +79,9 @@ func (p *Parser) applyRaw(ctx context.Context, st *State, raw *rawIntent) (Resul
 			return RejectResult("the LLM returned action review without a review object",
 				"Try again, or use /review."), nil
 		}
-		s := &pendingSlots{action: actionReview, extra: append([]string(nil), raw.Extra...)}
+		s := mergeReviewSlots(slotsFromState(st), raw)
 		switch raw.Review.Type {
-		case reviewWorkspace:
-			s.reviewType = reviewWorkspace
-		case reviewRange:
-			s.reviewType = reviewRange
-			s.from, s.to = raw.Review.From, raw.Review.To
-		case reviewCommit:
-			s.reviewType = reviewCommit
-			s.commit = raw.Review.Commit
+		case reviewWorkspace, reviewRange, reviewCommit:
 		default:
 			p.clearPending(st)
 			return RejectResult("the LLM returned an unknown review type: "+raw.Review.Type,
@@ -117,6 +110,8 @@ func (p *Parser) applyRaw(ctx context.Context, st *State, raw *rawIntent) (Resul
 		}
 		if partial := slotsFromRaw(raw); partial != nil {
 			p.storePending(st, partial, raw.Missing)
+		} else {
+			p.clearPending(st)
 		}
 		return ClarifyResult(raw.Question, raw.Missing...), nil
 
@@ -134,6 +129,35 @@ func (p *Parser) applyRaw(ctx context.Context, st *State, raw *rawIntent) (Resul
 		return RejectResult("the LLM returned an unknown action: "+raw.Action,
 			"Try again, or use /review or /scan."), nil
 	}
+}
+
+func mergeReviewSlots(prev *pendingSlots, raw *rawIntent) *pendingSlots {
+	s := &pendingSlots{action: actionReview, extra: append([]string(nil), raw.Extra...)}
+	if raw.Review.Type == reviewRange {
+		s.reviewType = reviewRange
+		s.from, s.to = raw.Review.From, raw.Review.To
+	} else if raw.Review.Type == reviewCommit {
+		s.reviewType = raw.Review.Type
+		s.commit = raw.Review.Commit
+	} else {
+		s.reviewType = raw.Review.Type
+	}
+	if prev == nil || prev.action != actionReview || prev.reviewType != s.reviewType || s.reviewType == reviewWorkspace {
+		return s
+	}
+	if s.from == "" {
+		s.from = prev.from
+	}
+	if s.to == "" {
+		s.to = prev.to
+	}
+	if s.commit == "" {
+		s.commit = prev.commit
+	}
+	if raw.Extra == nil {
+		s.extra = append([]string(nil), prev.extra...)
+	}
+	return s
 }
 
 // slotsFromRaw preserves whatever partial slots a clarify call carried so the
--- a/.github/workflows/acp-ci.yml
+++ b/.github/workflows/acp-ci.yml
@@ -1,0 +1,87 @@
+name: ACP CI
+
+# acp/ is a nested Go module. The root CI resolves its package list with
+# `go list ./...`, which does not descend into a nested module, so nothing in
+# acp/ is built or tested by ci.yml. This workflow is the counterpart to
+# pages-ci.yml, which exists for the same reason.
+#
+# No `branches:` filter on pull_request on purpose: development lands on an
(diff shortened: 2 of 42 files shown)


wu21-web:
Please use the new pull request template. Thanks!

acp/internal/intent/natural.go
```diff
@@ -0,0 +1,162 @@
...
+
+func (p *Parser) applyRaw(ctx context.Context, st *State, raw *rawIntent) (Result, error) {
+	switch raw.Action {
+	case actionReview:
+		if raw.Review == nil {
+			p.clearPending(st)
+			return RejectResult("the LLM returned action review without a review object",
+				"Try again, or use /review."), nil
+		}
+		s := &pendingSlots{action: actionReview, extra: append([]string(nil), raw.Extra...)}
```
Qiyuanqiii: There is a clarification-state edge case here that I think should be handled deterministically rather than delegated back to the model.

The previous pending slots are included in `buildUserPrompt`, but this branch constructs a fresh `pendingSlots` only from the current LLM response. For example:

1. Turn 1 establishes a range review with `from=main` and `to` missing.
2. Turn 2 supplies `feature`.
3. The model validly returns `{"type":"range","to":"feature"}` without repeating `from`.

At that point the already-known `from=main` is lost, and `finalizeReview` asks for it again.

Since the pending clarification state is owned by the adapter, I think the parser should merge the current response into the compatible pending state itself instead of requiring the LLM to reproduce every previously known slot. A two-turn regression test where the second tool call returns only the missing field would be useful here.

Qiyuanqiii (changes requested):
The overall Phase 4 direction looks solid, and the amount of validation and test coverage here is good. CI, ACP CI, and CodeQL are green.

I’m requesting changes for two clarification-state issues before merge:

1. I confirmed the existing finding in `finalizeReview`: a commit review with a missing commit is stored back as `reviewRange`, and `reviewClarifyQuestion(["commit"])` consequently asks for `--to` instead of `--commit`. This should preserve `reviewCommit`, use a commit-specific question, and have regression coverage for both the pending state and the question text.

2. In the natural-language path, the pending slots are included in the next model prompt but are not deterministically merged with the next tool-call result. A valid follow-up response containing only the newly supplied field can therefore discard fields already known from the previous turn. I left an inline comment on the fresh `pendingSlots` construction with a concrete example.

I also agree that protecting structural `ExtraBody` fields and the managed `anthropic-version` header would be worthwhile hardening, but I would not make the lower-severity URL/error-snippet/Makefile findings blockers by themselves.

Once the two state-machine issues above are addressed and covered by multi-turn tests, I think this will be much closer to approval.

Qiyuanqiii:
Please explicitly complete the AI/LLM disclosure required by the current PR template: either state that no AI/LLM was used, or, if it was used, disclose the tool(s) and specific model(s).

chaojixinren:
@Qiyuanqiii Thanks for the detailed review. I addressed the two clarification-state issues you identified.

1. Commit reviews now preserve `reviewCommit` when `--commit` is missing. The parser emits a commit-specific clarification question and keeps the pending state typed as a commit review. Regression coverage now verifies both the pending state and the question text.

2. Natural-language follow-ups now merge the new tool-call fields with the adapter-owned compatible pending slots. A follow-up response may provide only the newly supplied field without discarding values collected in the previous turn. A two-turn regression test covers the `from=main` followed by `to=feature` case.

I also added coverage for clearing clarification state when the model returns no partial intent, and for preventing stale slash-command selectors from leaking across review types.

The fixes are included in commits `39cd254` and `c094012`. Verification passes with:

- `make -C acp test`
- `make -C acp check`
- `make -C acp coverage` (approximately 90.1% total coverage)

I agree that protecting structural `ExtraBody` fields and the managed Anthropic headers is worthwhile hardening. I kept those items separate from the clarification-state fix so the current change remains focused. The lower-severity URL, error-snippet, and Makefile suggestions are also reasonable follow-up improvements.

I have also updated the AI/LLM disclosure to explicitly identify the tools and models used during development.

PrayWithYou (commented):
Thanks for the follow-up. I rechecked the latest changes, and there has been good progress here.

The two clarification-state issues from my original review are addressed:
- commit clarification now preserves the commit review type and asks for `--commit`;
- natural-language review follow-ups now merge compatible pending slots instead of losing previously known values.

The recent process-lifecycle fixes also look useful. In particular, the stream/process cleanup changes and the once-only event-channel closure address the lifecycle regressions introduced while Phase 5 was being added. ACP CI is green again as well.

I still see a few items before I can approve:

1. `actionClarify` still stores `slotsFromRaw(raw)` directly. A partial `clarify -> clarify` follow-up can therefore replace, rather than merge with, compatible adapter-owned pending state. I think this should use the same deterministic merge principle as the `actionReview` path.

2. A context that is already cancelled, or a request deadline that has already expired, still reaches `cmd.Start()` before cancellation/deadline handling begins. I would prefer these to terminate before spawning OCR. The completion-vs-cancellation race in `await()` is related to the same lifecycle contract.

3. The PR disclosure currently names OpenAI Codex, but the repository policy asks for the specific model as well, so that still needs to be added.

I would keep the ExtraBody/header/URL/UTF-8 hardening items non-blocking as before.

Overall, this is much closer now. Once the remaining clarification-state and pre-start cancellation semantics are tightened up, I think the implementation will be in a much better position for approval.

PrayWithYou:
My main GitHub account was suspended, so I’m replying using this one.

chaojixinren:
@PrayWithYou Thanks for the follow-up review. Commit 42d84af addresses the remaining clarification and cancellation issues.
- Clarification state: actionClarify now merges compatible pending state instead of replacing it. Omitted fields retain previously collected values, explicit lists replace previous lists, and switching intent types does not carry incompatible fields forward. Regression tests cover consecutive clarification turns followed by a completed request.
- Cancellation and deadlines: The runner now checks termination before setup and immediately before cmd.Start(). It also rechecks observable cancellation/deadline state when selecting the final outcome, including when process completion is already available. This avoids relying on watcher scheduling or select ordering. Tests cover requests cancelled or expired before startup, the completion/cancellation race, and preserving an already-delivered outcome.
The startup check and OS process creation are not atomic; cancellation occurring between them is handled by the running-process cleanup path.
The PR disclosure now explicitly names OpenAI Codex with gpt-6-astra. ACP CI is green on the latest commit. The non-blocking hardening items remain separate from this fix.

chaojixinren:
截屏2026-09-15 23 43 48
截屏2026-09-15 23 45 52
截屏2026-09-15 23 49 04
截屏2026-09-15 23 49 55

chaojixinren:
**Follow-up: the tool-call details added by `c9279e7` cannot be seen in the VS Code / Cursor ACP client**

While re-verifying the structured progress work on Windows, I found that the client used for the VS Code side of the manual acceptance does not render tool-call `content` at all. This is a client-side limitation, not an adapter or OCR defect, but it makes a few statements in this PR and its docs over-claimed, so I am recording it here.

**Client behaviour**

`formulahendry.acp-client@0.2.0` (the "ACP Client" extension; latest release) renders a tool call from `toolCallId`, `title` and `status` only. In `src/ui/ChatWebviewProvider.ts`:

- `case 'tool_call'` calls `addToolCall(tc.toolCallId, tc.title, tc.status)`; `addToolCallInline()` builds only `<span class="tc-icon">` and `<span class="tc-title">`.
- `case 'tool_call_update'` calls `updateToolCall(update.toolCallId, update.status, update.title)`; it rewrites the icon and the title, nothing else.

The `content` array does reach the webview — the extension host forwards the whole notification in `handleSessionUpdate` — but it is never read in the render path, and the only `<details>` element in the webview is the "Thinking..." block. So in this client there is no expandable body: `Command:`, the working directory, the bounded `[ocr]` log tail and the new `**Review stages**` markdown are all dropped.

I filed the gap upstream with the code references, a generic payload example and a minimal repro:
https://github.com/formulahendry/vscode-acp/issues/69

**What this means for the adapter**

Nothing needs to change here. With an `ocr` built from this branch, the CLI emits the structured `[ocr-progress] {...}` stderr snapshots (same one-file review: 0 lines with the v1.12.4 release binary, 17 with the branch build), and the adapter forwards them as `tool_call_update.content` together with the command, working directory and diagnostics. Clients that render tool-call content — Zed — keep working, and this is the protocol-correct representation regardless of any single client.

chaojixinren:
One thing I want to clarify before treating the current ACP interface as stable is that both the command coverage and the natural-language intent contract are still expected to evolve.

The current ACP adapter starts with `/review` and `/scan`, but the overall command coverage is not finalized yet. OCR already exposes a broader set of CLI commands, modes, and flags, and it is still unclear which of those should also be exposed through ACP, which should remain CLI-only, and whether some capabilities should be represented differently at the ACP layer instead of mirroring the CLI one-to-one.

This also applies to future OCR features. For example, #1510 proposes a new `ocr review --staged` mode for reviewing a frozen Git index snapshot. If that lands, ACP will need to decide whether and how staged review should be exposed, rather than automatically mapping every new OCR CLI flag into ACP. That decision would affect the slash-command surface, structured intent schema, validation rules, clarification behavior, command discovery, documentation, and tests.

The natural-language parsing layer is also still an initial version. It currently maps free-form requests into the structured review/scan intent schema, but the conventions around intent selection, clarification, rejection, argument inheritance, unsupported capabilities, and how newly added OCR features should be described to the parser LLM still need further refinement and testing. The current prompt and tool schema should therefore not be treated as a finalized long-term contract yet.

My intention for this PR is primarily to establish the ACP foundation: the protocol boundary, session lifecycle, OCR process orchestration, cancellation and timeout handling, progress/result mapping, and an initial working interface for review and scan.

Once the ACP integration is in use and the expected command coverage becomes clearer, I plan to follow up on the higher-level command surface and natural-language parser contract separately, including how new OCR capabilities such as #1510 should be represented.

PR #1210, opened 2026-09-10
fix(llm): add deepseek-flash to DeepSeek provider

Wsp030914:
## Description

Add the official `deepseek-flash` model to the built-in DeepSeek provider.

The DeepSeek API accepts `deepseek-flash`, but the current provider validation only allows legacy model names, causing valid requests to be rejected locally.

This change:

- Adds `deepseek-flash` to the Go provider registry.
- Synchronizes the VS Code provider preset.
- Preserves the existing model order and default selection.
- Adds a regression test for provider model lookup.
- Retains legacy model names for backward compatibility.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring (no functional changes)
- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] `make test` passes locally
- [x] Manual testing (describe below)

Additional verification:

- `make check`
- `make build`
- DeepSeek provider regression test
- Successful live review using `ocr review --provider deepseek --model deepseek-flash`

## Checklist

- [x] My code follows the project's coding style (`go fmt`, `go vet`)
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix is effective or my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have updated the documentation accordingly (not applicable: provider model lists are not documented)
- [x] I have signed the CLA

## Related Issues

Closes #1207

PR #1219, opened 2026-09-11
chore: remove stale alias pointing to `deepseek-flash`

wu21-web:
## Description

There is a stale pointer named `deepseek-v4-flash` pointing to the new `deepseek-flash`.
Per the docs:
Screenshot 2026-09-11 at 10 04 34 PM

Therefore the `deepseek-v4-flash` should be removed, because it will always points towards to the new model.
The last PR updating the models list is incomplete.

`deepseek-v4-pro` is not effected because per the docs deepseek said it still supports v4 pro model

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [x] I have signed the CLA
**No AI used in any way**

## Related Issues

N/A

### DIFF
--- a/extensions/vscode/src/shared/providers.ts
+++ b/extensions/vscode/src/shared/providers.ts
@@ -64,7 +64,7 @@ export const PROVIDER_PRESETS: OcrProviderPreset[] = [
     protocol: 'openai',
     baseUrl: 'https://api.deepseek.com',
     envVar: 'DEEPSEEK_API_KEY',
-    models: ['deepseek-v4-pro', 'deepseek-v4-flash', 'deepseek-flash'],
+    models: ['deepseek-v4-pro', 'deepseek-flash'],
   },
   {
     name: 'tencent-tokenhub',
--- a/internal/llm/providers.go
+++ b/internal/llm/providers.go
@@ -203,7 +203,6 @@ var registry = []Provider{
 		EnvVar:      "DEEPSEEK_API_KEY",
 		Models: []string{
 			"deepseek-v4-pro",
-			"deepseek-v4-flash",
 			"deepseek-flash",
 		},
 	},


Qiyuanqiii:
Would you like me to take a look？XD

Qiyuanqiii:
Thanks for cleaning this up. I understand the motivation here: `deepseek-v4-flash` now points to the same underlying V4.1-Flash model as `deepseek-flash`, so keeping both names in the visible model list can look redundant.

I think there is one compatibility detail worth considering before removing it from the registry, though.

According to DeepSeek's official V4.1-Flash announcement, `deepseek-flash` is now the canonical model name, while `deepseek-v4-flash` has been retired as a model version but is still temporarily kept as a compatibility route to V4.1-Flash:

https://www.deepseek.com/en/news/deepseek-v4-1-flash/

So while the old name is deprecated, it is still currently an accepted API identifier.

This matters a little more in OCR because the provider `Models` list is not only used for the model picker. The resolver also uses that list to validate explicit `--model` overrides. Removing the alias here would therefore make:

`ocr review --provider deepseek --model deepseek-v4-flash`

fail locally, even though DeepSeek still accepts that identifier upstream.

This also seems consistent with the intent of #1210, which explicitly kept legacy model names for backward compatibility. The author there also mentioned that `deepseek-v4-flash` was intentionally retained while DeepSeek continues redirecting it to V4.1-Flash.

Would it make sense to either:

1. keep the alias until DeepSeek actually removes the compatibility route; or
2. hide deprecated aliases from the picker while still accepting them in resolver validation?

The second option would avoid showing duplicate model names without breaking existing configurations that still use the old identifier.

One small documentation detail as well: the same DeepSeek announcement says `deepseek-v4-pro` is also being phased out, with requests scheduled to route to V4.1-Flash starting September 14, 2026 at 04:00 UTC until V4.1-Pro is released.

I think the cleanup direction makes sense; I’d just prefer to preserve the upstream-supported compatibility alias while DeepSeek itself still guarantees it.

wu21-web:
@Qiyuanqiii I have considered all of the points you made upon creation, but I still created this. But we will see and get this reopened when that route is removed.

I created this despite knowing the compatibility route. And I belief DeepSeek will not serve `deepseek-v4-flash` model again.

LLM goes forwards, not backwards. This is a part of the OCR concept.

The timeout is 15 days.

Qiyuanqiii:
ok

wu21-web:
@Qiyuanqiii I think you have a misunderstanding of the compatibility route. Or is it me?
We have 2 understandings here maybe:
1. The route is for programs that are still using the old model name so they won't fail.
2. The route is a substitute for the new model and will be removed later.

Which one is yours?

Qiyuanqiii:
> @Qiyuanqiii I think you have a misunderstanding of the compatibility route. Or is it me? We have 2 understandings here maybe:
> 
> 1. The route is for programs that are still using the old model name so they won't fail.
> 2. The route is a substitute for the new model and will be removed later.
> 
> Which one is yours?

I think the two interpretations are actually describing two parts of the same compatibility lifecycle rather than being mutually exclusive.

My understanding is primarily your first one: the compatibility route exists so programs and existing configurations that still use `deepseek-v4-flash` do not suddenly fail.

The second part is also true in the sense that this is temporary. DeepSeek's wording is:

> “For compatibility, `deepseek-v4-flash` and `deepseek-v4-flash-vision-exp` temporarily route to V4.1-Flash.”

So I read that as:

`old identifier works during migration` → `users migrate to deepseek-flash` → `DeepSeek may eventually remove the old route`

I definitely don't mean that `deepseek-v4-flash` should remain a recommended alternative to `deepseek-flash`. New configurations should use `deepseek-flash`, and I would be completely fine with hiding the deprecated alias from the model picker.

My concern is only that OCR's current `Models` list also acts as the resolver allowlist. Removing the old ID there means existing OCR configs using it stop working *before* DeepSeek removes the upstream compatibility route, which defeats the purpose of that transition period.

So I think we're mostly aligned on the intended direction; the difference is just where the compatibility window should end. My preference would be to stop advertising the old ID now, but continue accepting it until DeepSeek actually removes the route. After that, removing it entirely makes perfect sense.

Official announcement for reference:
https://www.deepseek.com/en/news/deepseek-v4-1-flash/

wu21-web:
What models list? I had problem understanding.

Qiyuanqiii:
> What models list? I had problem understanding.

Ah, sure — by “models list” I mean the `Models` field in the built-in DeepSeek provider entry in `internal/llm/providers.go`.

The slightly confusing part is that this list is doing two jobs in OCR:

1. it’s used for showing/selecting models;
2. it’s also used by the resolver to validate an explicit `--model` value.

The second one is what I was talking about.

So if we remove `deepseek-v4-flash` from that list, OCR will reject something like:

`ocr review --provider deepseek --model deepseek-v4-flash`

before the request even reaches DeepSeek, even though DeepSeek still accepts that old identifier through the temporary compatibility route.

That’s basically my only concern here. I’m totally fine with not showing/recommending the deprecated alias anymore — I just think “don’t show it” and “don’t accept it anymore” are slightly different things while the upstream compatibility route still exists.

Hope that explains what I meant a bit better 😄

wu21-web:
@Qiyuanqiii Please 👍 if you don't have any doubts or questions. Let me know, thanks.

stay-foolish-forever (approved):

PR #1222, opened 2026-09-12
fix(tool): swap inverted file_read line range instead of erroring

seamnniel:
## Summary

`file_read` hard-errors when the model passes `end_line < start_line`:

```
file_read failed: invalid line range: start_line 430 is greater than end_line 120
```

This is a common model slip — the model derives `start_line` from the old side of a diff hunk while `end_line` comes from the new file's length. One bad call fails the whole file's review item (observed: 6/6 items in a review failed this way).

Fix: swap the inverted range and serve it. The response still reports `Total lines`, so the model can recover if the swap was not its intent. Updated `TestExecute_InvalidLineRange` to `TestExecute_SwappedLineRange` asserting the swapped range returns content.

## Validation

- `go test ./internal/tool/` — pass
- `make check` — pass
- Patched binary built and run locally: `ocr llm test` and `ocr review --commit <sha> -p` work.

## Disclosure

This change was developed with AI assistance (Devin / SWE-2) per the project's agent guidelines; the change was reviewed and tested before submission.

### DIFF
--- a/internal/tool/file_read.go
+++ b/internal/tool/file_read.go
@@ -37,11 +37,14 @@ func (p *FileReadProvider) Execute(ctx context.Context, args map[string]any) (st
 
 	maxLines := fileReadMaxLines
 	if endLine > 0 {
-		requested := int(endLine) - int(startLine) + 1
-		if requested <= 0 {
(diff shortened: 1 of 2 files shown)


PrayWithYou (commented):
need CLA

Issue #1197, opened 2026-09-08
Preview omits diff-level vendor directory drops from file totals and exclusion reasons; no tracked-vendor override in v1.11.6


acoliver:
## Observed

With OCR v1.11.6, installed build `04284b5dc` matching tag commit `04284b5dc19e0f52528ca466a767e4a2c100e5f5`, an explicit Git range containing24 changed tracked files produces a successful preview listing17 files with `excluded_count: 0`. The other7 are tracked Rust source/tests under `vendor/`, not binaries or Git-ignored files. Other tested ranges reconcile34/34,13/13 and33/33.

Command shape:

```sh
ocr review --preview --from <exact-base> --to <exact-head> \
  --rule /absolute/full-scope.json --repo /absolute/repository \
  --provider zai-anthropic --model glm-5.3 --format json
```

The rule has a `rules` entry matching `**/*`, top-level `include: ["**/*"]` and `exclude: []`. These are supported schema fields, and the include list is active in preview. No LLM review was run.

This arose while preparing review of intentionally modified vendored code in `vybestack/llxprt-code-rs`, with24 files in the exact local range `2f0af9c4098c3672eb323363a6f0df62e689aa00` to `28e2399eb21b9c4820c31f635ddc04deece47fdc`. Local feature commits may not be remotely available. The seven omitted files are:

- `vendor/serdes-ai-models/src/anthropic/model.rs`
- `vendor/serdes-ai-models/src/anthropic/stream.rs`
- `vendor/serdes-ai-models/src/anthropic/types.rs`
- `vendor/serdes-ai-responses/src/client/assembler.rs`
- `vendor/serdes-ai-responses/src/client/mod.rs`
- `vendor/serdes-ai-responses/src/types.rs`
- `vendor/serdes-ai-responses/tests/client_ws.rs`

## Source-qualified cause and distinction

At exact v1.11.6, `internal/diff/git.go:27–40` defines `providerDirIgnoreDirs`, including `vendor/`; `filterDiffs` at387–400 removes these paths before agent file selection. `internal/agent/preview.go` therefore never classifies them or includes them in excluded counts. `ocr delegate` uses the same selection counts.

[...]

PR #1223, opened 2026-09-12
fix(preview): show provider-excluded diffs

CowToTheSky:
## Summary
- Preserve provider-directory exclusions for Preview reporting without retaining their full diff payload during normal reviews.
- Report those files as `provider_directory` exclusions, distinct from include-overridable `default_path` exclusions.
- Keep normal review behavior unchanged: files under `vendor/` remain unreviewed.

Refs #1197

## Validation
- `make check`
- `make test`
- Manually verified `review --preview --format json` with a tracked modified file under `vendor/`.

## AI disclosure
This PR was developed with assistance from OpenAI Codex (GPT-6). I reviewed the implementation and validation results and take responsibility for the changes.


internal/agent/agent.go
```diff
@@ -576,12 +577,13 @@ func (a *Agent) loadDiffs(ctx context.Context) error {
...
 
-	parsed, err := provider.GetDiff(ctx)
+	set, err := provider.GetDiffSet(ctx)
 	if err != nil {
 		return fmt.Errorf("get diffs: %w", err)
 	}
 
-	a.diffs = parsed
+	a.diffs = set.Included
+	a.providerExcluded = set.Excluded
```
PrayWithYou: `loadDiffs` is also used by normal review runs, so keeping `set.Excluded`
on the Agent retains the full excluded `model.Diff` values for the lifetime
of the review. Those values include both the unified diff and
`NewFileContent`, which can be substantial for exactly the noisy directories
being filtered here (`vendor/`, `node_modules/`, etc.).

Could we keep normal reviews on `GetDiff()` and only request the excluded
set on the Preview path, or retain lightweight metadata for excluded files
instead of the full diff payload?

internal/agent/preview.go
```diff
@@ -51,9 +51,25 @@ func (a *Agent) preview(ctx context.Context) (*DiffPreview, error) {
...
+	for _, d := range a.providerExcluded {
+		result.TotalInsertions += d.Insertions
+		result.TotalDeletions += d.Deletions
+		result.ExcludedCount++
+		result.Entries = append(result.Entries, DiffPreviewEntry{
+			Path:          effectivePath(d),
+			Insertions:    d.Insertions,
+			Deletions:     d.Deletions,
+			Status:        diffStatus(d),
+			ExcludeReason: ExcludeDefaultPath,
```
PrayWithYou: I don't think `default_path` is quite the same exclusion class here.

Today `default_path` refers to the per-file built-in path gate, which can be
overridden by a user `include`. Provider directory exclusions are unconditional
and happen before that filter, so `include` cannot re-admit `vendor/` or
`node_modules/`.

Using the same reason makes the existing documented remediation for
`default_path` incorrect for these entries. Could we distinguish provider-level
directory exclusions with a separate reason (or otherwise make the
non-overridable semantics explicit)?

PrayWithYou (changes requested):
The overall direction looks right, and the change does address the preview visibility problem from #1197: provider-level directory exclusions are now surfaced instead of silently disappearing from the reported totals.

I left two inline comments that I consider blocking before approval:
- avoid retaining the full provider-excluded diff payload for the lifetime of a normal review;
- distinguish these unconditional provider-directory exclusions from the existing `default_path` class, whose documented remediation is different.

One scope note as well: I don't think this PR should currently use `Fixes #1197`. That issue also asks for the documentation/remediation semantics to be clarified and for a supported way to review intentionally tracked vendored code. This PR deliberately keeps `vendor/` unreviewable, so it only addresses part of the issue. `Refs #1197` or a follow-up issue for the remaining work would be more accurate.

Other than those points, the implementation and tests are moving in the right direction, and I don't see a need to broaden the scope further.

lizhengfeng101:
Nit: `filterDiffs` no longer has any callers now that `GetDiff` goes through `GetDiffSet`/`partitionDiffs` — worth deleting it as dead code before merge.

internal/diff/git.go
```diff
@@ -386,15 +413,26 @@ func matchGitignoreDirectory(relPath, pattern string) bool {
 
 // filterDiffs removes diffs whose file paths are excluded.
 func (p *Provider) filterDiffs(diffs []model.Diff) []model.Diff {
+	return p.partitionDiffs(diffs).Included
+}
+
```
lizhengfeng101: Nit: `filterDiffs` no longer has any callers now that `GetDiff` goes through `GetDiffSet`/`partitionDiffs`, and as an unexported method nothing outside the package can reach it either. Suggest dropping the dead wrapper:

```suggestion
```

lizhengfeng101 (approved):
nice job

PR #1226, opened 2026-09-13
docs(delegate): require git --no-pager in agent skill examples

### LINKED ISSUE #1059: docs(delegate): explain why git --no-pager is required in PTY-based agent environments
## Issue Type

Incorrect information

## Document Location

- `skills/open-code-review-delegate/SKILL.md`, Step 3: Get Diffs
- `plugins/open-code-review/skills/open-code-review-delegate/SKILL.md`, Step 3: Get Diffs

## Description

Step 3 currently instructs agents to run `git diff` or `git show` directly. In an Agent Bash environment that provides a pseudo-terminal (PTY), Git can detect stdout as interactive and launch its configured pager, usually `less`.

Because the agent has no interactive input channel, `less` waits indefinitely. If the host tool enforces a command timeout and terminates the process group with `SIGKILL`, the observed result is exit code 137.

This is not caused by diff size or stdout pipe buffering. Large diffs may make the issue appear more frequent because they exceed one screen, but the triggering conditions are:

1. The command runs with a PTY.
2. Git selects an interactive pager such as `less`.
3. The pager receives no user input.
4. The host eventually terminates the hung command.

A deterministic local reproduction uses `script` to allocate a PTY:

```sh
TERM=xterm GIT_PAGER=less script -q /dev/null \
  git diff HEAD -- path/to/changed-file
```

With a sufficiently long diff, the command starts `less` and does not exit until input is provided or the process is terminated.

The same command completes immediately when the pager is disabled:

```sh
TERM=xterm GIT_PAGER=less script -q /dev/null \
  git --no-pager diff HEAD -- path/to/changed-file
```

I also reproduced this in Codex by forcing the same PTY and pager conditions: `git --no-pager diff` completed with exit 0, while the default-pager command remained blocked for four seconds and then exited 137 when the timeout harness sent `SIGKILL`.

## Suggested Change

Add the following subsection before the Step 3 command examples:

### Pager and large-output behavior in Agent and CI environments

Always pass `--no-pager` when obtaining diffs or file content through Git:

```bash
git --no-pager diff <ref> -- <path>
git --no-pager show <commit> -- <path>
```

Some Agent Bash tools allocate a pseudo-terminal. Without `--no-pager`, Git may start an interactive pager such as `less`, which waits for input that the agent cannot provide. A host-side timeout can then terminate the command and report exit code 137.

This is a pager and PTY interaction, not a Git diff size limitation.

For a potentially large diff, `--no-pager` alone does not bound the amount of stdout captured by the host tool. Redirect the diff to a temporary file, then read that file in bounded chunks:

```bash
git --no-pager diff --stat <ref> -- <path>
diff_file=$(mktemp /tmp/ocr-diff.XXXXXX)
git --no-pager diff <ref> -- <path> > "$diff_file"
# Read "$diff_file" with a file-reading tool in bounded chunks.
```

This avoids both an interactive pager waiting for input and an Agent or CI runner buffering the entire diff in memory.

Update the existing examples accordingly:

```bash
git --no-pager diff <merge_base>..<to> -- <path>
git --no-pager show <commit> -- <path>
git --no-pager diff HEAD -- <path>
```

The recommendation is intentionally local to each command. Environment-level alternatives such as `GIT_PAGER=cat` or `core.pager=cat` can be overridden by user configuration, while `git --no-pager` is explicit and reliable. For large diffs, use file redirection instead of streaming the full content to the host tool.

jimpablo:
## Description

Delegation-mode skills told host agents to run `git diff` / `git show` to collect diffs. In Agent Bash environments that allocate a PTY, Git can start an interactive pager (`less`), which waits for input the agent cannot provide. A host timeout then kills the process group and reports exit 137.

This PR prefixes the Step 3 git examples with `--no-pager` and explains why that flag must be on the git command itself (`GIT_PAGER` / `core.pager` can be overridden by user config). The plugin skill copy stays synchronized with the canonical skill. The Claude Code delegate command is updated so it does not contradict the skills.

Out of scope: the `mktemp` / redirect-large-diff recipe from the issue.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] `node scripts/github-actions/check-plugin-contract.test.js` passes locally (includes a new assertion that reads both shipped `SKILL.md` files)
- [x] `node scripts/github-actions/check-plugin-contract.js` passes locally
- [ ] `make test` passes locally (docs-only; Go tests not required)

## Checklist

- [x] I have signed the CLA

AI/LLM disclosure: this PR was prepared with assistance from Grok (xAI). I reviewed every line of the diff and the test output; commits have no AI authorship trailers.

## Related Issues

Fixes #1059


PrayWithYou (approved):
Looks good to me. Using git --no-pager directly in the delegation skill is the right fix for PTY-based agent environments where Git may otherwise launch an interactive pager and block waiting for input.

As long as the canonical delegation skill and its plugin mirror stay synchronized, I don't see a blocking issue here.

The user-facing delegation docs still use the plain git diff / git show forms, but I would treat aligning those as a non-blocking follow-up since this PR is specifically scoped to agent skill instructions.

LGTM.

skills/open-code-review-delegate/SKILL.md
```diff
@@ -54,22 +54,29 @@ Pass the reviewable file paths from Step 1. Output is grouped by rule content 
 
 ### Step 3: Get Diffs
 
-Use git directly based on the mode/ref info from Step 1:
```
wu21-web, line 57: There is no need to speak that much. Agents only need what they need to do, not the reason behind it. They are usually smart enough to figure that out.

I would keep this block cleaner.

scripts/github-actions/check-plugin-contract.test.js
```diff
@@ -525,6 +525,93 @@ function testDeclarationTargetsAreCoveredByTheTreesWeValidate() {
   }
 }
 
+// --- shipped delegate skill: Step 3 git examples ---------------------------
+//
+// Issue #1059: Agent Bash PTYs can make Git start `less`, which hangs until
+// the host kills the command (exit 137). The shipped skills must tell the
+// agent to pass `--no-pager` on the git command itself. This reads the real
+// SKILL.md files rather than a fixture so a reverted example fails CI.
```
wu21-web: Please clean these comments up.
```suggestion

```

plugins/open-code-review/claude-code/commands/delegate-review.md
```diff
@@ -30,10 +30,10 @@ ocr delegate rule <path1> <path2> ...
 
 ### Step 3: Get Diffs and Review
 
-For each reviewable file, get its diff using git (based on mode/ref from Step 1):
-- Range: `git diff <merge_base>..<to> -- <path>`
-- Commit: `git show <commit> -- <path>`
-- Workspace: `git diff HEAD -- <path>` (or read directly for untracked files)
+For each reviewable file, get its diff using git (based on mode/ref from Step 1). Always pass `--no-pager` so a PTY does not start `less` and hang until the host kills the command (exit 137):
```
wu21-web: I wouldn't spef a `137` here. There is no OOM and a user won't know the agent is stuck just because of missing `--no-pager` in the subprocess.

plugins/open-code-review/skills/open-code-review-delegate/SKILL.md
```diff
@@ -59,22 +59,29 @@ Pass the reviewable file paths from Step 1. Output is grouped by rule content 
...
 
-Use git directly based on the mode/ref info from Step 1:
+Use git directly based on the mode/ref info from Step 1. Always pass
+`--no-pager` on the git command itself. Some Agent Bash environments
+allocate a pseudo-terminal (PTY); without `--no-pager`, Git may start an
+interactive pager such as `less`, which waits for input the agent cannot
+provide. A host-side timeout can then terminate the command and report
+exit code 137. This is a pager and PTY interaction, not a Git diff size
+limitation. Environment-level alternatives such as `GIT_PAGER=cat` or
+`core.pager=cat` can be overridden by user configuration.
```
wu21-web: [Ditto](https://github.com/alibaba/open-code-review/pull/1226/changes#r3999883852)

wu21-web (commented):
Thank you for diving into this. I might be strict on these documentation but it is worth taking a look at.

skills/open-code-review-delegate/SKILL.md
jimpablo: Thanks, that's fair. The extra rationale was noise for the agent. I shortened Step 3 to the `--no-pager` instruction and the three examples.

plugins/open-code-review/skills/open-code-review-delegate/SKILL.md
jimpablo: Same cleanup on the plugin copy so the two files stay in sync.

plugins/open-code-review/claude-code/commands/delegate-review.md
jimpablo: Agreed — I dropped the `137` mention. The command now just says to always pass `--no-pager`.

scripts/github-actions/check-plugin-contract.test.js
jimpablo: Removed those comments. Thanks for catching it.

jimpablo:
@wu21-web thank you for the review — the stricter pass on the docs was helpful.

I shortened the skill and command text to the action only (`git --no-pager` on the examples, no PTY/`less`/137 rationale), applied the same cleanup to the plugin copy, and removed the extra test commentary. The follow-up is in bdbb948.

Please take another look when you have a moment.

PR #1228, opened 2026-09-13
fix(llm): install the embedded BPE loader from package init

### LINKED ISSUE #1012: TestFilterLargeDiffs_Boundary fails because exactNTokens fixture is not token-exact
### Problem

`TestFilterLargeDiffs_Boundary` currently fails on the latest `main` branch because its `exactNTokens` helper assumes that repeating `"a "` `n` times produces exactly `n` `cl100k_base` tokens. The tokenizer can merge that repeated pattern, so the assumption does not hold.

### Reproduction

At commit `9c5e90d0b1b64f84119754b8517b6d9b0978d15f`:

```sh
go test ./internal/agent -run '^TestFilterLargeDiffs_Boundary$' -count=1
```

Output on macOS arm64 with the repository's Go 1.25.5 toolchain:

```text
--- FAIL: TestFilterLargeDiffs_Boundary (0.40s)
    agent_test.go:630: fixture drift: llm.CountTokens(<80-token string>) = 39, want 80
FAIL
```

The same failure also appears during `go test ./...`.

### Cause

The helper constructs its fixture with:

```go
s := strings.TrimSpace(strings.Repeat("a ", n))
```

BPE tokenization is content-dependent, so repeated source words are not guaranteed to map one-to-one to tokens.

### Expected behavior

The boundary test should construct text based on the actual tokenizer result (or use a verified non-merging fixture) so a clean checkout's tests pass while still testing exactly 80 versus 81 tokens.

One robust option is to append text incrementally and stop when `llm.CountTokens` reaches the requested count, failing only if it skips past it.

basil-k-aji-dev:
## Description

`TestSelectFilesTooLarge_Boundary` (the test #1012 reports under its former name `TestFilterLargeDiffs_Boundary`) fails on a clean checkout with `fixture drift: llm.CountTokens(<80-token string>) = 39, want 80`.

The issue attributes this to BPE merging in the `exactNTokens` fixture, but that is not the cause. The fixture is fine — `strings.Repeat("a ", 80)` really is 80 `cl100k_base` tokens. The failure comes from token counting silently degrading:

- `InitEmbeddedLoader()` was only called from `cmd/opencodereview/main.go`.
- Every other consumer of `internal/llm` — the whole test suite included — therefore left tiktoken on its **default loader, which downloads the encoding file over the network**.
- When that download is unavailable, `countTokensWithEncoding` falls back to `len(text)/4`.

For the fixture, `len("a a a ... a") / 4` = `159 / 4` = **39**, which is exactly the number in the issue report.

So the test does not fail because of the tokenizer; it fails on any machine where the encoding download is blocked or unavailable, and passes everywhere else. That also explains why it is not failing for everyone.

This affects **two** tests, not just the one in the issue — `internal/scan`'s `TestSelectScanItems_LargeBoundary` shares the same helper and fails identically.

### Fix

Install the embedded BPE loader from the `internal/llm` package `init()` rather than relying on an explicit startup call. The embedded data already ships in the package; it was just not wired up unless `main()` asked for it.

`InitEmbeddedLoader()` stays exported and is now redundant in `main()`, so that call is removed.

This is more than a test fix: the silent `len(text)/4` fallback affects real token budgeting for any consumer that does not go through the CLI entry point.

### Why not adjust the fixture instead

#1012 suggests building the fixture from the tokenizer's own output. That would make the test pass under the degraded byte estimate too, hiding the real defect. Keeping the strict `exactNTokens` check leaves it as a canary, which is what caught this.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Reproduced the failure deterministically by pointing `TIKTOKEN_CACHE_DIR` at an empty directory and blackholing the proxy env, which makes the encoding download fail:

```
$ TIKTOKEN_CACHE_DIR=<empty> HTTPS_PROXY=http://127.0.0.1:9 go test ./internal/agent ./internal/scan -run Boundary -count=1
--- FAIL: TestSelectFilesTooLarge_Boundary
    agent_test.go:650: fixture drift: llm.CountTokens(<80-token string>) = 39, want 80
--- FAIL: TestSelectScanItems_LargeBoundary
    agent_test.go:285: fixture drift: llm.CountTokens(<80-token string>) = 39, want 80
```

With this change, the full suite passes under those same offline conditions, and `make check` and `make build` pass.

Added `TestCountTokensUsesEmbeddedBpeData`, which never calls `InitEmbeddedLoader` and asserts the count is not the `len(text)/4` estimate. I verified it fails (`= 39`) when the new `init()` is removed.

One environment note, for accuracy: `make test` uses `-race`, and the race detector cannot start on my arm64 box (`ThreadSanitizer: unsupported VMA range, Found 47 - Supported 48`) for any package, on `main` as well as on this branch. I ran the identical package list with `LC_ALL=C go test -count=1` instead; all 23 packages pass. CI will exercise the `-race` path.

## Checklist

- [ ] I have signed the CLA

AI disclosure: this change was developed with AI assistance (Claude Opus 5, via Claude Code). I reviewed the diff and the reasoning myself, no commits are attributed to AI, and I will answer maintainer questions in my own words.

## Related Issues

Closes #1012


lizhengfeng101:
@basil-k-aji-dev Great job! I actually ran into this exact problem locally yesterday!

lizhengfeng101 (approved):
LGTM

PR #1122, opened 2026-08-31
fix(rules): route JavaScript module files to JS rules

teddyli18000:
## Description

OpenCodeReview already accepts `.mjs` and `.cjs` files through `supported_file_types.json`, but the built-in JavaScript/TypeScript rule mapping only covers `.ts`, `.js`, `.tsx`, and `.jsx`.

As a result, reviewable `.mjs` and `.cjs` files fall back to the generic `default.md` rule instead of receiving the existing JavaScript/TypeScript review guidance.

This PR fixes that routing gap by:

* extending the existing JavaScript/TypeScript rule pattern to include `.mjs` and `.cjs`;
* adding `.mjs` and `.cjs` coverage to the existing system rule resolver test matrix;
* synchronizing the documented built-in rule mapping across all supported locales.

`.mjs` and `.cjs` intentionally use the same existing JavaScript/TypeScript review rules as `.js` and `.ts`; no separate rule semantics are introduced.

## Type of Change

* [x] Bug fix (non-breaking change that fixes an issue)
* [ ] New feature (non-breaking change that adds functionality)
* [ ] Breaking change (fix or feature would cause existing functionality to change)
* [ ] Refactoring (no functional changes)
* [x] Documentation update
* [ ] CI / Build / Tooling

## How Has This Been Tested?

* [x] Added `.mjs` and `.cjs` cases to the existing system rule resolver tests.
* [x] Verified the updated `system_rules.json` is valid JSON.
* [ ] Full repository tests are pending upstream CI execution.

## Checklist

* [x] My code follows the project's existing style.
* [x] I have performed a self-review of the change.
* [x] I have added tests that prove the fix is effective.
* [ ] New and existing unit tests pass locally.
* [x] Documentation has been updated for the built-in rule mapping.
* [x] I have signed the CLA.

## Related Issues

No existing issue found for this routing gap.

PR #1232, opened 2026-09-14
fix(rules): route Python stub files to the Python review rules

### LINKED ISSUE #1127: Python stub (.pyi) files fall back to generic review rules
## OpenCodeReview Version

Current `main` at `24edcbd3b7d0e10adf7f7277aaeaa6374869235a`.

## Operating System

Windows

## Installation Method

Built from source

## LLM Provider

Not applicable. This happens during built-in rule resolution before any LLM request.

## Bug Description

OpenCodeReview accepts `.pyi` files through `supported_file_types.json`, but the built-in Python rule mapping only covers `.py` and `.ipynb`:

```json
"**/*.{py,ipynb}": "python.md"
```

As a result, Python stub files are reviewable but silently fall back to `default.md` instead of receiving language-specific guidance.

## Steps to Reproduce

With no custom, project, or global review rules configured, run:

```shell
ocr rules check stubs/example.pyi
```

The result is:

```text
File: stubs/example.pyi
Source: System built-in
Pattern: default
```

For comparison, a `.py` file resolves through the Python mapping and loads `python.md`.

## Expected Behavior

The built-in behavior for `.pyi` files should be explicit and appropriate for Python type stubs.

Two possible directions are:

1. Include `.pyi` in the existing Python mapping:

   ```json
   "**/*.{py,pyi,ipynb}": "python.md"
   ```

2. Add a dedicated rule for Python stub files if their review requirements should differ from regular Python source.

Would the maintainers prefer `.pyi` to reuse `python.md`, or to have a separate stub-specific rule?

## Additional Context

`.pyi` is already part of the supported extension allowlist, so this does not require changing file eligibility.

A similar routing gap for `.mjs` and `.cjs` was fixed in #1122. This report is limited to `.pyi`; other extensions that currently use the generic fallback may require separate language rules or have ambiguous semantics.

basil-k-aji-dev:
## Description

`.pyi` is already in `supported_file_types.json`, so Python stub files are eligible for review, but the built-in path rule map only covered `.py` and `.ipynb`:

```json
"**/*.{py,ipynb}": "python.md",
```

Stub files therefore fell through to `default.md` and were reviewed with generic guidance instead of the Python rules.

This adds `pyi` to the existing Python pattern.

### Why reuse `python.md` rather than add a stub-specific rule

#1127 asks which of the two directions maintainers prefer. This PR takes the first one, for three reasons:

- It matches the precedent set by #1122, which fixed the same class of routing gap for `.mjs` and `.cjs` by extending the existing pattern rather than adding a rule file.
- `.pyi` is Python source and the rules in `python.md` are about Python semantics, so nothing in it is wrong for a stub.
- It is a one-line change with no new rule content to maintain or translate.

If maintainers would rather have stub-specific guidance (a rule that leans on signature and typing concerns and drops the runtime-behaviour items), I am happy to redo this as a separate `python_stub.md`. This change does not block that — it just stops stubs falling back to the generic rules in the meantime.

No change to file eligibility: `.pyi` was already allowlisted.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Added two cases to `TestResolve_DefaultRules`. I confirmed they fail without the JSON change, reproducing the issue:

```
--- FAIL: TestResolve_DefaultRules/stubs/example.pyi
    Resolve("stubs/example.pyi"): expected rule containing "Mutable Default Arguments",
    got "#### Correctness\nIs the logic correct? ..."
```

and pass with it.

Also reproduced the issue's own CLI check against the built binary:

```
$ ocr rules check stubs/example.pyi
File: stubs/example.pyi
Source: System built-in
Pattern: **/*.{py,pyi,ipynb}
```

Previously this reported `Pattern: default`.

`make check` and `make build` pass, and the full package list passes.

One environment note, for accuracy: `make test` runs with `-race`, and the race detector cannot start on my arm64 machine (`ThreadSanitizer: unsupported VMA range, Found 47 - Supported 48`) for any package, on `main` as well as on this branch. I ran the same package list with `LC_ALL=C go test -count=1` instead; all packages pass. CI covers the `-race` path.

## Checklist

- [x] I have signed the CLA

AI disclosure: developed with AI assistance (Claude Opus 5, via Claude Code). I reviewed the change and the reasoning myself, no commits are attributed to AI, and I will answer maintainer questions in my own words.

## Related Issues

Closes #1127


lizhengfeng101 (changes requested):
Looks good — please also update the pattern in the five `pages/src/content/docs/{en,zh,ja,ko,ru}/review-rules.md` tables (same files as #1123 touched), and optionally add a do-not-report note in `python.md`'s Dead Code section for unused imports/parameters in `.pyi` stubs; then this is ready to merge.

basil-k-aji-dev:
Thanks — both done in f213cf3.

- Updated the rule table in all five `pages/src/content/docs/{en,zh,ja,ko,ru}/review-rules.md` files to `**/*.{py,pyi,ipynb}`, leaving each locale's description text untouched.
- Took the optional suggestion too. The Dead Code section already lists "variables, imports, or function parameters that are declared but never read", which is precisely what a stub looks like, so without a carve-out `.pyi` review would flag re-exported imports and unreferenced parameter names as dead. Added a do-not-report bullet in that section saying a stub declares an interface it never executes.

`make check` (license + english-check included) and the full test suite pass.

internal/config/rules/rule_docs/python.md
```diff
@@ -8,6 +8,7 @@
 - Code blocks that can never be reached (e.g., branches where the condition is always false, code after a `return`, `raise`, `break`, or `continue`)
 - Variables, imports, or function parameters that are declared but never read or referenced
 - Large blocks of commented-out code with no apparent intent to preserve
+- Do not report unused imports, parameters, or assignments in `.pyi` stub files; a stub declares an interface it never executes, so re-exported imports and unreferenced parameter names are expected rather than dead
```
lizhengfeng101: `assignments` here is narrower than the `variables` used in the Dead Code item just above, and the gap is not only cosmetic: in a stub, the standard way to declare a module-level variable is a bare annotation — `version: str`, `DEFAULT_TIMEOUT: int` — with no assignment at all. Only type aliases and `_T = TypeVar("_T")` are actual assignments. Read literally, the most common form of "declared but never read" in a `.pyi` file falls outside this carve-out, which is exactly the false positive it is meant to suppress.

Suggest aligning the vocabulary with the item above and naming annotation-only declarations explicitly:

```suggestion
- Do not report unused imports, variables, or parameters in `.pyi` stub files; a stub declares an interface it never executes, so re-exported imports, annotation-only declarations, and unreferenced parameter names are expected rather than dead
```

Non-blocking — the routing fix itself is correct and I would not hold the PR for this.

lizhengfeng101 (approved):
LGTM

Issue #1008, opened 2026-08-19
OpenCode review tool fails with long background context


cristeaadrian:
### OpenCodeReview Version

```text
open-code-review v1.9.7 (f269d0ce0) darwin/arm64
```

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

Other OpenAI-compatible endpoint

### Bug Description

The official OpenCode plugin exposes only a `background` string and always passes it to OCR through `--background`.

With multi-paragraph review context:

- `ocr_review` preview succeeds.
- The full `ocr_review` call exits immediately without creating an OCR session.
- OpenCode reports only `OpenCodeReview exited with code 1.`
- Running the same review and context directly with `--background-file` creates the OCR session and proceeds normally.

The adapter also discards the child process termination signal. Its `close` handler converts a null exit code to `1`, so the caller cannot distinguish a signal termination from a normal exit and receives no useful diagnostic when stdout and stderr are empty.

### Steps to Reproduce

1. Install the official OpenCode plugin from this repository.
2. Invoke `ocr_review` in preview mode for a branch range and confirm the file selection succeeds.
3. Invoke the full review with a multi-paragraph `background` value.
4. Observe the generic exit-code error and confirm that `ocr session list` contains no new session.
5. Save the same background text to a Markdown file and run:

```shell
ocr review --from main --to feature --background-file context.md
```

6. Observe that OCR creates a session and starts reviewing normally.

### Expected Behavior

The OpenCode adapter should support large review context without silently losing the OCR process.

Suggested options:

- Expose a `backgroundFile` tool argument.
- Transparently write long background text to a private temporary file and pass `--background-file`.
- Remove the temporary file on success, failure, or cancellation.
- Preserve and report the child process exit signal alongside stdout and stderr.

### Logs / Error Output

```shell
OpenCodeReview exited with code 1.
```

No OCR session is created for the failed tool call and no stdout or stderr is returned.

### Additional Context

The failure was reproduced twice through the OpenCode tool. LLM connectivity and OCR preview both passed immediately before the full calls. The direct `--background-file` run used the same provider, refs, and review context.

OpenCode version: `1.18.18`.

preston-56:
I'd like to take this.

#1234 fixed the plugin-side gaps: `backgroundFile` now maps to the CLI's `--background-file`, and the `close` handler reports the terminating signal instead of coercing a null exit code to 1, which is what produced the bare "exited with code 1" here.

[...]

PR #206, opened 2026-06-24
feat(background-file) Add the background-file CLI option

Victor-D:
feat(background-file) Add the background-file CLI option to read a local business context file

## Description

Following discussion in https://github.com/alibaba/open-code-review/discussions/176 add the new -B flag : 
```
  -b, --background string       optional requirement/business context for the review
  -B, --background-file string  path to a Markdown file used as review background (combined with --background; inline value appears first when both are set)
  ```

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring (no functional changes)
- [x] Documentation update
- [x] CI / Build / Tooling

## How Has This Been Tested?

- [x] `make test` passes locally
- [x] Manual testing (describe below)

soft limit
```
% ./dist/opencodereview review --repo /home/USER/bitbucket/REPO_NAME --commit <commit-hash> -B /tmp/context.md
[ocr] --background-file content is 4623 characters, exceeding the recommended 2000 (continuing but review quality might be impacted)
[ocr] 2 file(s) changed, reviewing 1 in /home/USER/bitbucket/REPO_NAME
[ocr] Skipping hieradata/roles/web.eyaml — filtered by path/extension rules
[ocr] Filtered 1 file(s) by include/exclude rules
```

hard limit
```
user@HOSTNAME open-code-review % ./dist/opencodereview review --repo /home/USER/bitbucket/REPO_NAME --commit <commit-hash> -B /tmp/context.md
Error: background content is 14631 characters, exceeding the hard limit of 8000 (aborting)
```

I also checked I see the two prompts in the session file when using -b and -B flags.

## Checklist

- [x] My code follows the project's coding style (`go fmt`, `go vet`)
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix is effective or my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have updated the documentation accordingly (if applicable)
- [x] I have signed the CLA

PR #1234, opened 2026-09-14
feat(opencode): add backgroundFile input and report termination signals

preston-56:
## Description

The OpenCode plugin only ever passed `--background`, so there was no way to give OCR a long review context from OpenCode. The CLI has supported `--background-file` since #206. This adds a `backgroundFile` input that maps to it.

It also fixes error reporting in `runOcr`. The `close` handler used `exitCode ?? 1`, so a child killed by a signal was reported as "exited with code 1". When OCR wrote no output, that was the only message the caller received, and it named the wrong cause. The handler now reads the signal and reports it.

Passing `background` and `backgroundFile` together is rejected. #1016 settled this as precedence plus a stderr warning, but `formatReviewResult` returns only stdout on exit 0, so that warning never reaches the caller. Rejecting in the plugin is what keeps the inline text from being dropped silently.

I left inline `--background` unchanged instead of writing long text to a temporary file. `loadBackgroundFile` applies sanitization, the `<ocr_user_background>` wrapper, and an 8000-character limit. None of these apply to inline `--background`, so rerouting inline text would subject it to a limit it does not have today. That is a decision for the maintainers.

I could not reproduce the original exit-1 report. Inline `--background` does not appear to be length-limited anywhere. Both problems fixed here are visible in the code regardless.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

`npm run check` in `plugins/open-code-review/opencode` passes: build, typecheck, and 32 tests.

New tests cover:

- `backgroundFile` is forwarded as `--background-file` in the argv.
- `background` combined with `backgroundFile` is rejected before OCR starts.
- A child killed by a signal reports that signal and a null exit code, rather than a fabricated exit code 1.

`make test` and `make check` both pass locally.

## Checklist

- [x] I have signed the CLA

## AI/LLM disclosure

Developed with AI assistance (Claude Opus 5, via Claude Code). I reviewed the full diff. No AI attribution in commits.

## Related Issues

Refs #1008


plugins/open-code-review/opencode/open-code-review.ts
```diff
@@ -283,6 +295,10 @@ const reviewArgs = {
   to: optionalString("Target ref for a branch/range comparison. Must be paired with 'from'."),
   resume: optionalString("Resume a previous OCR review session by ID."),
   background: optionalString("Business or requirement context that the implementation should satisfy."),
+  backgroundFile: optionalString(
+    "Path to a Markdown file holding the review background, for context too long to pass inline. " +
+      "A relative path resolves against the repository root. Cannot be combined with 'background'.",
+  ),
```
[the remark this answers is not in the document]
preston-56: Description fixed. Path resolution left to the CLI, which takes absolute paths by design.

lizhengfeng101 (commented):
Solid fix, and the signal test is nicely thorough. One blocker:

**`backgroundFile` is invisible under OpenCode 2.x.** You updated `ReviewInput` and `reviewArgs` (V1), but `reviewInputSchema` at line 428 — the JSON Schema `setupV2` registers as the tool's input — doesn't have it, and that schema is `additionalProperties: false`. So 2.x models can't see or pass the parameter. `buildReviewArgs` is shared so the runtime is fine; it's just the declaration. V2 support landed one commit ago in #1213, easy to miss. Add the property and a schema assertion test (`v2 numeric inputs require positive integers`, test line 629, is the pattern) — behavioral tests stay green here regardless, so only a schema assertion catches this.

Two description nits:

- #1016 landed *precedence with a stderr warning*, not rejection (see `selectBackground`). I agree with rejecting in the plugin — `formatReviewResult` drops stderr on exit 0, so the warning never reaches the caller — but that's the justification, not #1016. Same for the code comment: the silence is the plugin's, not OCR's.
- `Refs #1008` rather than `Closes`? You're upfront that the exit-1 symptom wasn't reproduced. This gives a working path and better diagnostics, but closing it means nobody revisits the actual bug.

Minor: `exited with code ${exitCode}` can print `code null` when both are null — unreachable in practice, but the old `?? 1` guaranteed a number.

Also, per `AGENTS.md`, please note in the description whether AI/LLM tooling was used and which.

plugins/open-code-review/opencode/open-code-review.ts
```diff
@@ -79,6 +83,13 @@ function buildReviewArgs(input: ReviewInput, repo: string): string[] {
   if (input.preview && input.resume) {
     throw new Error("'preview' and 'resume' cannot be used together.")
   }
+  // OCR warns on stderr and lets the file win (selectBackground). On exit 0
+  // formatReviewResult returns stdout only, so that warning never reaches the
```
lizhengfeng101:
```suggestion
  // OCR warns on stderr and lets the file win, but formatReviewResult drops
  // stderr on exit 0, so the caller would never see it.
```

preston-56: Took your wording. Applied it across the whole four-line block rather than committing the suggestion directly, since it was anchored to 86-87 and the last two lines would have been left behind as a fragment.

lizhengfeng101 (approved):
LGTM

PR #1238, opened 2026-09-14
docs(architecture): remove incorrect provider_directory entry from wh…

### LINKED ISSUE #1237: docs: provider_directory is documented as a whyExcluded reason, and is missing from four locales
## Description

#1223 documented the new `provider_directory` exclusion reason in `pages/src/content/docs/en/architecture.md`, but not in the four localized copies. Readers of `zh`, `ja`, `ko` and `ru` see an exclusion-reason list that is missing a reason `--preview` can actually print. The FAQ pages were updated in all five locales; only `architecture.md` was missed.

Before syncing, though, the English page needs a correction — otherwise the sync copies a factual error into four more files.

## Part 1: fix the English page

The new line was added to this list:

> Once diffs are loaded, every file passes through [`whyExcluded`](...). The function returns one of:
> ```
> binary          — file is binary
> user_exclude    — matched a pattern in your `exclude` list
> unsupported_ext — extension is not in supported_file_types.json
> default_path    — matched a built-in test-file exclude pattern
> provider_directory — matched an unconditional provider directory exclusion
> ```

`whyExcluded` does not return `provider_directory`. The constant appears nowhere in `internal/agent/selection.go`; the reason is attached in `internal/agent/preview.go`, outside `selectFiles` entirely. The same section says so a few lines further down — "they never reach the per-file filter" — so the page currently contradicts itself.

The sentence right after the list is where this belongs:

> `deleted` and `too_large` are **not** returned by `whyExcluded`; `selectFiles` applies them after the gates

`provider_directory` sits even further out than those two: `deleted` and `too_large` at least pass through `selectFiles`, while a provider-directory file is dropped by the diff provider before selection runs at all. Either move it out of the list into that paragraph, or drop it from the list and rely on the `providerDirIgnoreDirs` paragraph below, which already explains it correctly.

Also minor: the list aligns its `—` at column 16, and `provider_directory` is 18 characters, so that line breaks the alignment of the block.

## Part 2: sync all four locales

The English page has two relevant places:

1. The exclusion-reason area (line 67 today) — carry over whatever Part 1 settles on, not the current wording.
2. The `providerDirIgnoreDirs` paragraph (lines 87-91) gained a sentence explaining that preview reports these files as `provider_directory`, that they never reach the per-file filter, and that an `include` rule cannot make them reviewable. This part is correct as written and can be translated as-is.

The corresponding reason lists in the localized pages:

| locale | file | reason list |
| --- | --- | --- |
| zh | `pages/src/content/docs/zh/architecture.md` | line 60 |
| ja | `pages/src/content/docs/ja/architecture.md` | line 48 |
| ko | `pages/src/content/docs/ko/architecture.md` | line 64 |
| ru | `pages/src/content/docs/ru/architecture.md` | line 75 |

Each page already carries the `providerDirIgnoreDirs` paragraph, so it only needs the new sentence.

One thing to match rather than decide: the reason codes themselves (`binary`, `default_path`, `provider_directory`, ...) are literal CLI output and stay in English — only the description after the dash is translated. The `ru` page translates those descriptions, while `zh`, `ja` and `ko` currently keep this particular block in English. Follow whatever each page already does.

## Scope

- File(s): `pages/src/content/docs/{en,zh,ja,ko,ru}/architecture.md`

## Acceptance Criteria

- [ ] The English page no longer lists `provider_directory` among `whyExcluded`'s return values
- [ ] All five pages state that an `include` rule cannot override a provider-directory exclusion, and that such files never reach the per-file filter
- [ ] Descriptions are in the same language as the surrounding block on each page
- [ ] No source-code changes, so `make check` is sufficient

## Context

Found while reviewing #1223 after it merged (3bce7ad).

---

*AI/LLM disclosure: the analysis above is my own. Claude Code (Claude Opus 5) was used to polish the English wording and to file this issue via `gh`.*

Fayupable:
whyExcluded never returns provider_directory - the constant appears nowhere in internal/agent/selection.go, and is attached in internal/agent/preview.go, outside selectFiles entirely. The English page contradicted itself: the providerDirIgnoreDirs paragraph a few lines down already correctly said these files never reach the per-file filter.

- en: remove the provider_directory line from the whyExcluded list
- zh/ja/ko/ru: add the missing sentence to the providerDirIgnoreDirs paragraph (preview reports these as provider_directory, and an include rule cannot override the exclusion), matching each page's existing language convention
- ko: also fix a wrong function name (filterDiffs -> isProviderDirExcluded) in the same paragraph

Fixes #1237

## Description

Fixes a self-contradiction in the English architecture.md (provider_directory was listed as a whyExcluded return value, but the function never returns it - verified against internal/agent/selection.go and internal/agent/preview.go), and syncs the corrected explanation to the zh/ja/ko/ru copies, which were missing it since #1223. Also fixes an unrelated wrong function name in the ko page found while editing the same paragraph.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Ran `make check` (license-check, english-check, go vet) - passed, and confirmed `git status` shows no files touched outside the 5 architecture.md pages. Verified every technical claim in the issue directly against the source (grepped internal/agent/selection.go and internal/agent/preview.go, confirmed the exact line numbers cited for en/zh/ja/ko/ru, confirmed the ko page's filterDiffs reference doesn't exist anywhere in the codebase).

## Checklist

- [ ] I have signed the CLA
- [x] I disclosed the tool/model below and reviewed its output; I did not attribute commits to AI and will answer maintainer questions and review comments myself without AI/LLM.

AI/LLM disclosure: I used Gemini and Claude to help translate the zh/ja/ko/ru sentences. I also used Claude to verify against the source that this closes #1237 correctly and completely (checked for anything left unaddressed). The scope of the fix itself was already clear from the issue. I reviewed and understand every change before pushing.

## Related Issues

Closes #1237


lizhengfeng101 (approved):
LGTM

PR #1241, opened 2026-09-14
fix(preview): scope preview totals to the tracked changeset

### LINKED ISSUE #1235: fix(preview): untracked provider-directory files are missing from workspace preview totals
## Description

#1223 made `--preview` list provider-directory exclusions (`vendor/`, `node_modules/`, `target/`, ...) so that preview totals account for the whole changeset instead of silently dropping those files. It fixes the tracked path, but **workspace mode still drops untracked files under those directories before any diff is built**, so they never reach `partitionDiffs` and never land in `DiffSet.Excluded`.

In `internal/diff/git.go`, `untrackedFilesList` filters with `isPathExcluded`, which combines two independent rules — the provider-directory blocklist and `.gitignore` matching:

```go
if !p.isPathExcluded(line, patterns) {   // isProviderDirExcluded(line) || gitignore match
	files = append(files, line)
}
```

Only the provider-directory half should be deferred to `partitionDiffs`; the `.gitignore` half must keep dropping files. (That half is nearly redundant here anyway, since `git ls-files --others --exclude-standard` has already dropped ignored files.)

Because of this, the comment added in #1223 to `internal/agent/preview.go` promises more than the code delivers:

> Preview lists them separately to make its file and line totals match the Git changeset without implying that include rules can make them reviewable.

That holds for tracked changes only.

## Reproduction

This repository gitignores `vendor/`, so `vendor/` cannot demonstrate the behaviour. `target/` is in `providerDirIgnoreDirs` and is not gitignored, so use that. Avoid path segments such as `dist/` or `node_modules`, which this repository's `.gitignore` matches at any depth.

Untracked — the file is silently dropped:

```console
$ mkdir -p target && printf 'package main\n\nfunc demo() {}\n' > target/demo.go
$ git status --short target/
?? target/
$ ocr review --preview

Preview: 1 file(s) changed  |  +2  -0

Excluded from review (1):
  [M]  internal/diff/git_test.go (default_path)
```

Staged — the same file is counted and reported:

```console
$ git add target/demo.go && git status --short target/
A  target/demo.go
$ ocr review --preview

Preview: 2 file(s) changed  |  +5  -0

Excluded from review (2):
  [A]  target/demo.go            (provider_directory)
  [M]  internal/diff/git_test.go (default_path)
```

Clean up with `git reset -- target/demo.go && rm -rf target/`.

(The `internal/diff/git_test.go` line is unrelated local noise. The relevant part is `target/demo.go` and the `1 -> 2` file / `+2 -> +5` line difference.)

## Possible fixes

The mechanical change is small, but there is a real cost to weigh, so this wants a maintainer decision.

1. **Defer the provider-directory check.** Split `isPathExcluded` into its two halves — #1223 already extracted `isProviderDirExcluded` — and have `untrackedFilesList` apply only the `.gitignore` half. The synthesized diffs then flow into `partitionDiffs`, which sorts them into `Excluded` with no further plumbing. Cost: `untrackedFileDiffs` reads every such file in full and builds a synthetic `+`-prefixed diff for it. In a repository where `node_modules/` is untracked and not gitignored that is tens of thousands of file reads, and this path runs for normal reviews too, not just `--preview` — which is exactly the cost the provider-level drop exists to avoid.

2. **Count lines without synthesizing a body.** Return the excluded paths separately, skip `combined`, and append lightweight `model.Diff{IsNew: true, NewPath: f, Insertions: <line count>}` records to `DiffSet.Excluded`. Still reads each file, but avoids building and retaining the diff text. Roughly 20 lines plus one merge point.

3. **Narrow the promise instead.** Reword the `preview.go` comment (and `Preview`'s doc comment) to say "tracked changes", and leave the untracked gap documented. Zero risk.

Option 3 may well be enough on its own. The scenario in #1197 — Rust or `go mod vendor` projects that commit their vendored tree — is *tracked*, and #1223 already covers it. The untracked gap needs a provider directory that is neither tracked nor gitignored, e.g. a fresh `npm install` before `.gitignore` is set up, which is a much narrower case than paying the read cost on every review.

## Scope

- File(s): `internal/diff/git.go` (`untrackedFilesList`, `isPathExcluded`), `internal/agent/preview.go` (comments)
- Also: `cmd/opencodereview/output_helpers_test.go`

## Acceptance Criteria

- [ ] Either untracked provider-directory files are counted in `total_files` / `total_insertions` and reported as `provider_directory`, or the "match the Git changeset" wording is narrowed to tracked changes
- [ ] `cmd/opencodereview/output_helpers_test.go` no longer pairs `vendor/lib.go` with `ExcludeDefaultPath` — after #1223 that combination cannot occur, and it should be `ExcludeProviderDirectory`
- [ ] Coverage threshold holds (`make coverage`)

A note for whoever writes the test: `vendor/` is gitignored in this repository, so use `target/` or a temporary repo. `initPreviewRepo` in `internal/agent/preview_run_test.go` is the existing harness.

## Context

Found while reviewing #1223 after it merged (3bce7ad). Related: #1197.

Worth a separate look: the FAQ wording added in #1223 — "never reviewable, even when `include`d" — is accurate for the current implementation, but it also closes the door on #1197's third request, a supported way to review a vendored tree that is genuinely part of the project. If that use case should stay open, the FAQ may want softer wording across all five locales.

---

*AI/LLM disclosure: the analysis and the reproduction above are my own. Claude Code (Claude Opus 5) was used to polish the English wording and to file this issue via `gh`.*

anmolxlight:
Fixes #1235.

## What
Narrows the preview totals promise to tracked changes (option 3 from the issue). Untracked provider-directory files are dropped in `untrackedFilesList` before any diff exists, so they are neither listed nor counted. Counting them (options 1/2) would mean reading every file under an untracked `node_modules/`/`target/` on every review, not just `--preview`, since that list feeds the normal workspace path too. The gap needs a provider dir that is neither tracked nor gitignored, which is narrower than the read cost.

## Changes
- `internal/agent/preview.go`: `Preview` doc + inline comment now say totals match the tracked Git changeset, with the reason recorded.
- `cmd/opencodereview/output_helpers_test.go`: `vendor/lib.go` pairs with `ExcludeProviderDirectory` (that combination is the only one possible after #1223).
- `internal/agent/preview_run_test.go`: new `TestPreviewOmitsUntrackedProviderDirFile` (untracked `main.go` + untracked `target/demo.go`; asserts no `target/` entry, totals exactly 1 file / +1).

FAQ wording left alone: provider_directory "never reviewable, even when included" stays accurate, nothing about reviewability moved.

## Gates
- `make test` pass, `make check` pass, `make coverage` 91.5% (threshold 90%).

## AI/LLM disclosure
Per AGENTS.md: this change was implemented with Claude Code (Claude Opus 5) from the issue spec, then reviewed, verified (all gates re-run locally), and submitted by me. I understand every line and can explain it on review.


lizhengfeng101:
@anmolxlight anmolxlight seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

lizhengfeng101 (approved):
LGTM

PR #1242, opened 2026-09-14
fix(preview): collapse provider-directory rows and keep changeset order

### LINKED ISSUE #1236: fix(cli): provider-directory entries flood the preview and break Will review alignment
## Description

#1223 made `--preview` list provider-directory exclusions. Three presentation problems only became reachable with that change, because those paths could not enter `Preview.Entries` before it.

### 1. The column width is computed globally and shared by both sections

In `outputPreviewText` (`cmd/opencodereview/output.go`):

```go
maxPathLen := 0
for _, e := range p.Entries {          // every entry, including excluded ones
	if n := len(sanitizeTerminal(e.Path)); n > maxPathLen {
		maxPathLen = n
	}
}
pathFmt := fmt.Sprintf("%%-%ds", maxPathLen)
```

`pathFmt` is then used for both `Will review` and `Excluded from review`, so a single deep vendored path inflates the padding in the section users actually read:

```console
Preview: 4 file(s) changed  |  +10  -0

Will review (1):
  [A]  demo_real.go                                                                       +3    -0

Excluded from review (3):
  [A]  target/.pnpm/@scope+some-package@1.2.3/packages/some-package/esm/internal/index.js (provider_directory)
  [A]  target/demo.go                                                                     (provider_directory)
  [M]  internal/diff/git_test.go                                                          (default_path)
```

Without that one long path, the same run reads:

```console
Will review (1):
  [A]  demo_real.go              +3    -0
```

On a normal terminal width the insertion/deletion counts are pushed off-screen or wrap. One long path is enough — it does not take many files.

### 2. There is no bound on the excluded list

A repository that commits its vendored tree now gets thousands of lines of paths it cannot act on. `cargo vendor` output is routinely that large, and that is exactly #1197's reporter. The `Excluded from review` section has never had a row cap; before #1223 those files simply never reached it.

### 3. Provider entries are prepended, so the listing no longer follows Git's order

In `internal/agent/preview.go`, `preview` appends the provider-excluded diffs before iterating `selectFiles(a.diffs)`:

```go
for _, d := range providerExcluded { ... }          // all provider entries first
for _, dec := range a.selectFiles(a.diffs) { ... }  // then everything else
```

So every provider-directory file sorts ahead of the rest regardless of where it sits in the changeset, both in the terminal and in `--output json`'s `files` array. Interleaving them back into Git's order would keep the listing scannable against `git status` / `git diff --stat`. If the fix for problem 2 collapses provider entries into a single line, this still matters for the JSON payload.

## Suggested fix

For problems 1 and 2, aggregating `provider_directory` entries into a single line addresses both, and reads better besides:

```
Excluded from review (3423):
  [M]  internal/diff/git_test.go   (default_path)
  [B]  assets/logo.png             (binary)
  3421 file(s) in provider directories (vendor/, target/) — not reviewable
```

It also conveys something the current output loses. `provider_directory` and `default_path` look identical in the terminal, but an `include` rule can override `default_path` and cannot override `provider_directory` — see the comment on `ExcludeProviderDirectory` in `internal/model/preview.go`. Collapsing the non-overridable class into one line says "you cannot act on these" without needing an extra symbol.

Two constraints:

- **Terminal rendering only.** `--output json` must keep every entry in `files`, so script consumers are unaffected.
- The aggregate line should name the directories actually matched, not the whole `providerDirIgnoreDirs` list.

Whatever shape the fix takes, the width calculation has to skip entries that are no longer printed per row, or problem 1 survives.

## Scope

- File(s): `cmd/opencodereview/output.go` (`outputPreviewText`), `internal/agent/preview.go` (`preview`, entry ordering)
- Tests: `cmd/opencodereview/output_helpers_test.go`, `internal/agent/preview_run_test.go`

## Acceptance Criteria

- [ ] A long excluded path no longer affects `Will review` column alignment
- [ ] The excluded section stays bounded when a provider directory holds thousands of files
- [ ] `--output json` output stays complete, and its `files` array follows the changeset order rather than listing provider entries first
- [ ] Coverage threshold holds (`make coverage`)

## Context

Found while reviewing #1223 after it merged (3bce7ad). The sample output above is from a local build of that commit. `target/` stands in for `vendor/` because `vendor/` is gitignored in this repository; note also that a path containing `dist/` or `node_modules` will not reproduce it, since this repository's `.gitignore` matches those at any depth.

---

*AI/LLM disclosure: the analysis and the reproduction above are my own. Claude Code (Claude Opus 5) was used to polish the English wording and to file this issue via `gh`.*

jimpablo:
## Description

Fixes the three `--preview` presentation problems that became reachable once #1223 started listing provider-directory exclusions (#1236):

1. **Column width** — the path column is sized per section, from the rows that section actually prints. A long excluded path no longer pads the `Will review` rows.
2. **Unbounded excluded list** — in text output, `provider_directory` entries collapse into one line that names only the directories actually matched (`diff.ProviderDirPrefix`). They are skipped when sizing the excluded column. `--format json` is unchanged and still lists every entry in `files`.
3. **Ordering** — `DiffSet` records each provider exclusion's changeset position and `ForEachInOrder` walks Included/Excluded in Git's order instead of prepending provider entries. The terminal listing and the JSON `files` array follow that order.

Scope note: #1236 lists `output.go`, `preview.go` and their two test files. This also touches `internal/diff`, for two reasons. Once `partitionDiffs` splits the changeset into `Included` and `Excluded`, `preview.go` can no longer recover the original order, so `DiffSet` records where each exclusion sat. And `diff.ProviderDirPrefix` lets the summary line name the matched directories from the existing `providerDirIgnoreDirs` blocklist instead of a second copy of it.

The five FAQ locales now describe the text/JSON split: terminal preview collapses provider-directory files into one summary line; `--format json` still lists every entry.

Untracked provider-directory files stay with #1235 / #1241. This does not add a vendor-include override (#1197).

Example on a scratch repo with a vendored `target/` tree, a `vendor/` file, and a test file:

```
Will review (1):
  [M]  a.go                 +2    -0

Excluded from review (4):
  [M]  foo_test.go          (default_path)
  3 file(s) in provider directories (target/, vendor/) — not reviewable
```

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] `make check` passes locally
- [x] `make coverage` reports 91.5% (threshold 90%)

New tests, covering the renderer, `Preview`, `runPreview` and `internal/diff`:

- `TestOutputPreviewText_ExcludedPathDoesNotWidenWillReview`
- `TestOutputPreviewText_AggregatesProviderDirectories`
- `TestOutputPreviewJSON_KeepsEveryFileInEntryOrder`
- `TestPreviewKeepsChangesetOrder`
- `TestRunPreviewJSON_ProviderDirectoryKeepsChangesetOrder`
- `TestGetDiffSetWalksChangesetOrder`
- `TestProviderDirPrefix`

Updated: `TestOutputPreviewText_WithExcludedFiles`. Since #1241 its `vendor/lib.go` entry is `provider_directory`, which this PR renders as the summary line, so the test now expects that line instead of a per-row path.

Manual: built `ocr` and ran `ocr review --preview` and `ocr review --preview --format json` against a scratch repo containing `a.go`, `foo_test.go`, a deep `target/.pnpm/...` path, `target/demo.go`, and `vendor/lib.go`. Text output is shown above. The JSON `files` array lists all 5 paths in Git order.

Also checked against a 3,421-file provider tree (the excluded section stays at one summary line; JSON keeps all 3,423 entries in `git diff` order), commit and range modes, and a gitignored tracked file sitting between entries.

Pre-commit `ocr review --audience agent` could not run locally (no LLM endpoint configured).

## Checklist

- [x] I have signed the CLA

AI/LLM disclosure: Grok 4.6 (Grok Build / xAI) was used to implement the fix and tests and to draft this description. Claude Code (Claude Opus 5) was used afterwards to check the PR against #1236 and to correct two code comments and the FAQ wording. I reviewed the diff. Commits carry no AI attribution.

## Related Issues

Closes #1236


Parikalp-Bhardwaj:
@jimpablo I was working on the same issue as well, but you pushed the PR first.

One thing i noticed is that the FAQ still says `ocr review --preview` lists every candidate file.

With this change, provider-directory entries are summarized in text output, while JSON still keeps every entry.

Should we update the preview example/wording in `pages/src/content/docs/*/faq.md` so the docs match the new behavior? @wu21-web

jimpablo:
Thanks @Parikalp-Bhardwaj — sorry for the overlap, and thanks for reading the FAQ against this change.

I double-checked against #1236 (and the same author's follow-ups) before deciding not to touch the docs here:

- #1236's scope is `output.go` / `preview.go` plus the two test files. The constraints are **terminal rendering only** and **JSON `files` stays complete**. FAQ is not in the file list or the acceptance criteria.
- The FAQ sample was already a schematic (`src/foo.go modified`, no `Will review` / `Excluded from review` layout). It never matched the live CLI, including before this PR. The useful part for "my file isn't being reviewed" is the reason table, not a full dump of every path.
- #1237 (now closed via #1238) treated the five FAQ locales as already updated by #1223, and only asked to fix `architecture.md`. The remaining FAQ note from the same author, in #1235, is about softening "`never reviewable`, even when `include`d" for #1197's vendor-include request — which this PR is deliberately not doing.

So the table row for `provider_directory` is still the documented contract. The one sentence that is slightly stale is "lists every candidate file": text now collapses that class, while `--format json` still lists every entry, as #1236 required.

Happy to add a short wording note in all five `faq.md` locales in this PR if @wu21-web wants that here. I would keep the sample schematic rather than replacing it with a full CLI dump, unless you'd rather take the docs in a follow-up.

wu21-web:
@jimpablo @Parikalp-Bhardwaj You should always document these changes.

jimpablo:
Thanks @wu21-web — will do. Updating the five FAQ locales in this PR so the preview wording and sample match the new text/JSON split.

Parikalp-Bhardwaj:
> @jimpablo @Parikalp-Bhardwaj You should always document these changes.

I agree, since this changes user-visible preview behavior, the FAQ should be updated in the same PR so the documented behavior stays accurate.

jimpablo:
Agreed — that's now in this PR. `487274f` updates all five FAQ locales: the wording and sample show the terminal summary line, and note that `--format json` still lists every entry.

lizhengfeng101:
@jimpablo merge conflicts

wu21-web:
Can I review this?

jimpablo:
@lizhengfeng101 Rebased on main and fixed the conflicts. Had to update one existing test after #1241 landed, see the description.

lizhengfeng101 (approved):
LGTM

kefu84510-cmd (approved):

kefu84510-cmd (approved):
Hello

PR #1221, opened 2026-09-12
fix(vscode): generate provider presets from Go registry

PrayWithYou:
## Description

This PR restores #1212 from its saved local commits and description after an account change. It includes the display-name sorting and provider-order regression test from the previous review, and has been rebased onto current main. The recovered source was the last locally saved head, `c4019f8`. The restored commits use my current account's Git author identity so the CLA check can identify the contributor; their file trees are unchanged by that metadata update.

Human attribution: @Qiyuanqiii is my previous account. The co-author trailer preserves that account's contribution; it refers to the same contributor as @PrayWithYou.

### Summary

- Generate the VS Code provider presets from the existing Go registry, replacing the extension's hand-maintained provider/model list.
- Synchronize all 28 built-in providers at the base revision, including 13 providers absent from the extension and model-list drift in 11 existing presets.
- Preserve each Go model list verbatim, including its order and first-model default. Keep the generated registry sorted by internal provider name, and sort the configuration dropdown by display name, computing its static options once when the webview module loads.
- Carry through the existing Responses protocol and Bedrock ambient-authentication metadata. Show or hide the API Key field according to the effective protocol, including saved overrides.
- Add deterministic generation, a checked-in generated TypeScript file, Go/TypeScript regression tests, and a CI regeneration check.

### Root Cause and Fix

The CLI resolved built-in providers from `internal/llm/providers.go`, while the extension maintained a separate array in `extensions/vscode/src/shared/providers.ts`. Nothing enforced the comment asking contributors to keep those files aligned. As a result, the extension could offer an outdated model list or treat a CLI built-in provider as custom.

The data path is now:

```text
internal/llm/providers.go
  -> llm.ListProviders()
  -> go generate ./internal/llm
  -> extensions/vscode/src/shared/providers.generated.ts
  -> the existing shared/providers.ts exports and consumers
```

The generator imports the Go package and serializes its actual registry values. It does not parse Go source text, maintain a second provider allowlist, resolve user configuration, or contact provider APIs. The emitted fields are `name`, `displayName`, `protocol`, `baseUrl`, `authHeader`, `envVar`, `ambientAuth`, and `models`. `envVar` contains the variable name, not its value.

`providers.generated.ts` carries the generated-file warning and regeneration command. `providers.ts` retains the existing interface, lookup helpers, and model-list merging helper. Its import from the generated file is a runtime import; the generated file's reference back to `OcrProviderPreset` is type-only and is erased during compilation.

[...]

PR #1253, opened 2026-09-14
ci: autofix provider presets for same-repository PRs

### LINKED ISSUE #1211: refactor(llm): make Go provider registry the single source of truth for the VS Code preset

PrayWithYou:
## Description

Automatically regenerate and commit stale provider presets for pull requests from branches in this repository. Instead of only reporting that the generated catalog is out of date, the workflow commits the corrected artifact back to the PR branch. The independent CI check from #1221 remains in place to validate the committed result.

**Draft follow-up to #1221; merge #1221 first.** This PR contains only the optional autofix workflow, helper, tests, and documentation. The generator and VS Code changes remain in #1221 so the automation can be reviewed separately.

### Behavior

The workflow runs on matching `pull_request` events targeting `main`, including changes to the Go LLM package, generated catalog, module files, or autofix implementation.

| Situation | Result |
| --- | --- |
| Same-repository PR with stale presets | Regenerate and commit only `extensions/vscode/src/shared/providers.generated.ts` to the PR branch. |
| Catalog already matches | Leave the branch unchanged. |
| PR deleted the artifact, including when its replacement is ignored by Git | Restore the generated artifact and include it in the fix. |
| PR branch predates the autofix scripts | Use the helper from the workflow revision; the PR branch does not need to contain those scripts. |
| Generator is missing, fails, or produces no regular artifact | Fail before the commit and push step. |
| Generation introduces other tracked or non-ignored untracked changes | Fail before committing. |
| Fork or Dependabot PR | Skip the write job; the independent CI check and manual regeneration instructions still apply. |
| PR head advanced before the helper starts | Skip the superseded run. |
| Branch advances, is deleted, or is rewound during the update | Reject the outdated push. |
| Permissions, branch rules, or connectivity prevent confirmation | Fail with manual recovery guidance rather than retrying a potentially stale result. |

### Follow-up fixes in `25ad96d`

The follow-up addresses three cases found during review:

- **PR branches without the helper:** The workflow first checks out `github.workflow_sha`, tests that revision's helper, and preserves it outside the Git worktree. It then checks out the exact PR head for generation and committing. No files from the workflow revision are included in the generated fix.
- **Deleted and ignored artifacts:** The helper checks whether the generated artifact is tracked before returning `unchanged`. It can force-add that specific allowlisted path without staging other ignored files.
- **Successful commands that perform no generation:** The CI-only `--generate` mode removes the old artifact before running `go generate ./internal/llm` and requires a newly produced regular file. An old committed catalog can no longer hide a missing generator.

The `--generate` mode is intended for disposable CI checkouts. Local regeneration continues to use `go generate ./internal/llm`.

### Permissions and concurrent updates

Default workflow permissions remain `contents: read`; only the eligible autofix job requests `contents: write`. Both checkouts disable persisted credentials. The regeneration step does not receive the push token, and the commit helper supplies authentication only to the push subprocess.

The helper validates the repository, target ref, expected PR head, staged paths, and generated file before committing. The fix must be a direct child of the expected head.

Pushes use `--force-with-lease=<ref>:<expected-sha>`. The explicit expected SHA protects against concurrent branch updates, deletion, and rewind. Together with the direct-parent check, an accepted update is a fast-forward. The helper does not pull, rebase, or retry an outdated generated result.

Commits use the standard `github-actions[bot]` identity. Existing branch rules, required checks, and CLA policy still apply; this proposal introduces no CLA exemption or additional credential requirement.

This implementation targets the current public GitHub.com repository. Private-repository authentication and GitHub Enterprise Server support are outside its scope.

### Checks after a bot commit

The bot's new commit must be validated independently; passing checks on the previous head do not validate the updated branch.

A live bot push/check cycle still needs to be exercised, including any approval required for checks on the new commit. This PR itself comes from a fork, so its write job is intentionally skipped and cannot demonstrate the same-repository push path.

## Type of Change

- [x] New feature
- [x] CI / Build / Tooling
- [x] Documentation update
- [ ] Breaking change

### Latest revision: `25ad96d`

Focused validation was performed on Linux with Node 22.16.0 and Git 2.47.3.

- [x] All **30 helper tests** passed, including **11 new regression tests**.
- [x] **Six additional local workflow/CLI checks** passed using real Git repositories, a local bare remote, the workflow's shell commands, and a small Go 1.23.2 generator fixture.
- [x] Node syntax, YAML/shell structure, LF/SPDX/ASCII checks, and `git diff --check` passed. Uploaded file blobs matched the locally tested files.

The added checks cover PR heads without the helper, restoration of ignored artifacts, unchanged output, a successful command that runs no generator, and generation failure. Existing tests continue to cover exact-file commits, bot identity, invalid metadata, unrelated changes, missing credentials, file permissions, concurrent branch changes, push rejection, and uncertain push outcomes.

This was focused validation, **not a full repository test run**. Full `make check`, `make test`, coverage, `actionlint`, OCR CLI review, Node 14/24 compatibility, and integration with the actual #1221 generator were **not rerun for this revision**.

### Earlier validation: `fc345674`

The following results were recorded for the original implementation and are retained here for context. They do not establish that the latest revision passes those same checks.

| Check | Recorded result |
| --- | --- |
| `make license-add` and `make check` | Passed. |
| Full `make test`, including race detection | Passed. |
| `make coverage` | Passed; reported coverage was **91.2%**. |
| `actionlint` v1.7.12 | Passed with optional shellcheck/pyflakes integrations disabled. |
| Original 19 helper integration tests | Passed on Node 14.21.3 and Node 24.14.0. |
| Integration with #1221's actual Go generator | Passed in an isolated checkout; the fix contained only the expected catalog, verified byte-for-byte. |
| OCR CLI self-review | Completed for the original proposal and helper. |

The original validation environment was Windows amd64 with Git 2.54.0, Go 1.26.5, GNU Make 4.4.1, GCC 15.2.0, and Node 24.14.0.

The full Actions contract suite on Linux and a live GitHub bot push/check cycle remain unverified. Earlier Windows aggregate runs encountered `/bin/bash`-dependent failures, and the separate plugin-contract suite encountered symlink permission restrictions. The post-review and translation-sync suites passed.

## Checklist

- [x] Focused regression tests completed for the latest revision.
- [x] Workflow scope, permissions, concurrency behavior, and manual recovery are documented.
- [x] Human contributions and AI assistance are disclosed below.
- [ ] Revalidate the latest revision with the actual generator after #1221 is available.
- [ ] Complete live same-repository bot push/check validation.
- [ ] Complete contributor review of the latest AI-assisted changes before marking this PR ready.

### Human contributions

@wu21-web proposed automatically committing the regenerated catalog back to the PR branch in [the review discussion on #1221](https://github.com/alibaba/open-code-review/pull/1221#discussion_r3999972570), rather than only failing CI after regeneration. That proposal shaped the core behavior of this PR, and @wu21-web is credited as a co-author for this design contribution.

The implementation, tests, and follow-up fixes are submitted by @PrayWithYou with the AI assistance described below.

@Qiyuanqiii is my previous account. Its co-author attribution preserves that contribution history; @Qiyuanqiii and @PrayWithYou refer to the same contributor.

### AI assistance

AI tools were used for implementation, tests, review, and documentation:

- **Initial implementation:** Codex, based on GPT-6.
- **Initial self-review:** The repository's OCR CLI using `deepseek-v4-flash`.
- **Follow-up review and fixes in `25ad96d`:** ChatGPT, GPT-6 Astra Pro, assisted with review, implementation, regression tests, and documentation.

I have reviewed and verified all AI-generated content.

The validation scope and outstanding checks are listed above. The latest changes remain pending contributor review; this PR stays in draft until that review and the remaining validation are complete.

## Related Issues

Depends on #1221, which addresses #1211.

This PR implements the optional autofix workflow proposed during that review. It does not independently close #1211.


wu21-web:
Can I review this?

PrayWithYou:
> Can I review this?

Sure, I'm in class right now. I cannot find the button to mark this PR ready for review on the mobile app.

.github/workflows/provider-presets-autofix.yml
wu21-web, line 1: Why do we need another workflow for this? I think `ci.yml` is already perfect. We only need a new step.

scripts/github-actions/autofix-provider-presets.js
wu21-web, line 1: If you already have the go command that can generate the providers prsts for VSC extensions, and already a go env in the `ci.yml`, personally I think there is no need to write a duplicate `autofix-provider-presets.js`

package.json
wu21-web, line 1: Ditto. Codex might have been working on the behalf of you on a branch or context that does not have your prior PR's context. If I was you (and if I use Codex), I probably merge that branch into this branch directy. When your last pull request gets merged, you can remerge the base into this one, so Codex knows that this PR is dependent on another branch (feature).

.github/workflows/provider-presets-autofix.yml
wu21-web, line 1: I will write a `.patch` for you to view and then you will understand what I mean.

wu21-web: A new step at most actually.

PrayWithYou: ok

wu21-web: @PrayWithYou 
[ci.patch](https://github.com/user-attachments/files/32223074/ci.patch)

PrayWithYou: Thanks for putting the patch together, @wu21-web! It makes the smaller change you had in mind much clearer. I agree that reusing `ci.yml` is a good direction.

I'd like to make two adjustments while adapting it: create the autofix commit from the PR head, since the default checkout uses the test merge commit; and stop an outdated run if the branch advances, so we regenerate against the latest registry instead of rebasing an older generated result. I'll also limit the commit to `providers.generated.ts`.

I'll keep this work in #1253 so #1221 stays focused on generation and validation. Thanks again for taking the time to help simplify this.

wu21-web: OK, reference other review comments are you ready to rewrite.
My .patch is just the idea, you'll still write this yourself.

.github/workflows/ci.yml
```diff
@@ -22,6 +22,102 @@ concurrency:
...
+          else
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Commit provider presets to the PR branch
+        id: provider-presets-commit
+        if: steps.provider-presets-generate.outputs.changed == 'true'
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
```
wu21-web, line 83: Why a `GH_TOKEN`? Like don't need that for authoring commits.

.github/workflows/ci.yml
```diff
@@ -22,6 +22,102 @@ concurrency:
...
+          git check-ref-format "$ref"
+          remote="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}.git"
+          current=$(git ls-remote --heads "$remote" "$ref" | cut -f1)
+          if [[ "$current" != "$HEAD_SHA" ]]; then
+            echo "::notice::The PR branch changed; skipping this outdated autofix."
+            exit 0
+          fi
+          git -c core.hooksPath=/dev/null \
+            -c user.name='github-actions[bot]' \
+            -c user.email='41898282+github-actions[bot]@users.noreply.github.com' \
```
wu21-web, line 103: I'd use a non-existing user, so CLA does not trigger (on bots?). If you insist that is fine.

.github/workflows/ci.yml
```diff
@@ -22,6 +22,102 @@ concurrency:
...
+    timeout-minutes: 10
+    permissions:
+      contents: write
+    container:
+      image: golang:1.26.6
+    env:
+      HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+      HEAD_REF: ${{ github.event.pull_request.head.ref }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
```
wu21-web, line 42: You don't have to pin this for now, it is a change for another job.

```suggestion
      - uses: actions/checkout@7
```

.github/workflows/ci.yml
```diff
@@ -22,6 +22,102 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  provider-presets-autofix:
```
wu21-web, line 25: I'd prefer keeping steps instead of heavy bash syntax and command. Then users know what goes wrong and better for debugging too.

.github/workflows/ci.yml
```diff
@@ -22,6 +22,102 @@ concurrency:
...
+    # Keep the test job on the merge ref with read-only permissions. Autofix
+    # must create its commit from the contributor's exact PR head instead.
+    if: >-
+      github.event_name == 'pull_request' &&
+      github.event.pull_request.head.repo.full_name == github.repository &&
```
wu21-web, line 30: Why not trigger this on a fork that doesn't have its name as `$github.repository`?

wu21-web (commented):
Changes look better and understandable now. Thank you. Can we push this into a step? It still can be simplified, since the job still gets triggered when you modify something unrelated to the `providers`

PrayWithYou:
起这么早

wu21-web:
出来玩了

wu21-web:
你要累就睡觉吧，没有想打扰你
晚会处理也没问题

PrayWithYou:
> 你要累就睡觉吧，没有想打扰你 晚会处理也没问题

我其实是一直没睡着，一直在看GitHub相关封禁贴，我现在review的都少了 ，天天想着我的号

PrayWithYou:
> 你要累就睡觉吧，没有想打扰你 晚会处理也没问题

在处理了

PR #1254, opened 2026-09-14
fix(review): exit non-zero when a partial run has real item failures

### LINKED ISSUE #1027: ocr timed out and failed for some files, but status code=0
### OpenCodeReview Version

v1.9.8 (9c5e90d0b) linux/amd64

### Operating System

Linux (x86_64)

### Installation Method

GitHub Release binary

### LLM Provider

### Bug Description

When ocr is run with --timeout set and timeout happens I expect that ocr would finish with non-zero return code.

Having non-zero return code is critical for detecting ocr and CI harness misconfiguration on CI.

### Steps to Reproduce

Set timeout and run on slow LLM:
```
$ ocr review --commit HEAD --concurrency 1 --timeout 10 --audience agent
[ocr] Summary: 2 file(s) reviewed, 0 comment(s), ~251232 token(s) used (input: ~232407, output: ~18825), cache(read: ~200118, write: ~0), 10m48s elapsed
[ocr] Session: 496b007a-1dc5-4a77-8b71-aaeb928f9e13
Review partially complete: 0 finding(s); 1 of 2 selected item(s) failed.

LLM retry report: 0/16 requests retried, 0 retries, 0 recovered, 1 failed, 0 cancelled
- myfile.c / main_task #13: timeout -> failed


$ echo $?
0
```

### Expected Behavior

```
$ echo $?
1
```

### Logs / Error Output

```shell

```

### Additional Context

_No response_

HoneyTyagii:
## Summary

Fixes #1027.

`ocr review` exits 0 on a partial run (some items completed, some failed with timeout/provider errors), so CI cannot detect an incomplete review or a misconfigured harness from the exit status alone. The GitHub Action then skips posting (`Post review comments` gates on `OCR_EXIT_CODE == '0'`) while the job stays green.

## Change

- `reviewResultError` (cmd/opencodereview/review_cmd.go): a `partial` manifest whose failed set contains any non-budget item now returns an error, so the run exits non-zero. Budget-only partials (the documented `--max-tokens-budget` controlled-truncation contract) keep exiting 0. Partial results are still published before the non-zero exit, so JSON consumers keep the full coverage diagnosis.
- Updated the unit and e2e tests that pinned the old exit-0 contract, including a new mixed budget+provider partial case that counts only real failures.
- Docs synced: `--max-tokens-budget` flag help, action.yml input description, and the exit-codes tables / FAQ entries across en/zh/ja/ko/ru pages.

## Verification

- `go build ./...`, `go vet ./cmd/...`, `gofmt -s -l` clean
- `go test ./cmd/... ./internal/session/...` passes (LC_ALL=C)
- New pinned cases: partial with a timeout failure exits non-zero; budget-only partial still exits 0

Fixes #1027


rajpratham1 (changes requested):
The new exit-code behavior is directionally correct, but the process status now depends on nonBudgetFailures() treating every non-FailureBudget classification as a real failure. Since this changes the CLI/CI contract, please add coverage for the other possible failure classifications and verify that only controlled budget truncation returns 0 for a partial run. After those cases are covered, this should be ready to merge.

PR #1255, opened 2026-09-14
fix(tool): normalize Windows paths in search and comments

zzz-yu:
## Description\n\nFixes #1088.\n\nGit pathspecs and diff paths are repository-relative and use `/` regardless of the host OS. On Windows, the code_search tool previously validated only `/` separators, so patterns such as `..\pkg` or `pkg\..\internal` bypassed the traversal check and were passed to Git unchanged. Windows-style comment paths (for example `pkg\util.go`) also failed strict path matching and were silently dropped from review results.\n\nThis change:\n- normalizes backslashes in `file_patterns` before validation and Git invocation;\n- rejects traversal components after normalization;\n- normalizes comment paths with forward slashes and cleans `./` prefixes and duplicate separators;\n- adds regression coverage for Windows traversal attempts and path matching.\n\n## How Has This Been Tested?\n\n- `go test ./...` (all packages pass on Windows with Go 1.26.5)\n- `gofmt` and `git diff --check`\n\n## Checklist\n\n- [x] My code follows the project coding style (`go fmt`, `go vet`)\n- [x] I have performed a self-review of my code\n- [x] I have added tests that prove my fix is effective or my feature works\n- [x] New and existing unit tests pass locally with my changes\n- [ ] I have updated the documentation accordingly (not applicable)\n- [ ] I have signed the CLA\n- [x] I used AI/LLM assistance and disclosed it below; I reviewed every line and did not attribute commits to AI.\n\n### AI/LLM disclosure\n\nOpenAI Codex (GPT-5) assisted with issue investigation, patch drafting, and test design. Prompt used: “Identify a reproducible path-handling defect in this Go/Agent tool, implement a minimal fix with regression tests, and run the repository tests.” I reviewed the complete diff and verified the behavior with `go test ./...`.\n\n## Related Issues\n\nCloses #1088


rajpratham1 (approved):
The Windows path normalization changes look correct and are well covered by regression tests, including traversal rejection and valid Windows-style path handling. The additional normalization in hasTraversalPathComponent() is slightly redundant but harmless and does not block the fix. I don't see any blocking issues in this PR.

wu21-web:
This is a duplicate of #1088 .

brahmanandmathpati:
The fix itself is fine, honestly. The bug was real — Git pathspecs are always /-based, but the old traversal check only looked for /-style .., so on Windows you could sneak ..\pkg past it and it'd go straight to Git unmodified. That's a genuine bypass. Your fix normalizes backslashes to forward slashes before checking for traversal, which is the right order — you always want to validate the normalized form, not the raw input. Good call adding the regression tests for both that and the comment-path matching bug too.

lizhengfeng101:
This is a duplicate of https://github.com/alibaba/open-code-review/issues/1089 .

PR #1256, opened 2026-09-14
fix(suggestdiff): preserve case and indentation changes

### LINKED ISSUE #1249: bug: suggestion diff hides indentation-only and case-only changes (ComputeLineDiff compares with EqualFold + TrimSpace)
### OpenCodeReview Version

open-code-review dev windows/amd64 (built from source at main 6e5cd8c; `ocr version` prints no release tag for source builds)

### Operating System

Windows

### Installation Method

Built from source

### LLM Provider

### Bug Description

`internal/suggestdiff.ComputeLineDiff` decides whether two lines are "the same" with:

```go
strings.EqualFold(strings.TrimSpace(oldLines[i-1]), strings.TrimSpace(newLines[j-1]))
```

at `internal/suggestdiff/diff.go:40` (LCS table) and `:53` (backtrack).

Two independent problems fall out of that single comparison:

1. `TrimSpace` makes a pure-indentation change invisible.
2. `EqualFold` makes a pure-case change invisible, even though every language OCR reviews is case-sensitive.

When the comparison succeeds the line is emitted as `DiffContext` **carrying the old content** (`diff.go:54`, `Content: oldLines[i-1]`). So the rendered suggestion both (a) shows no +/- markers at all and (b) echoes the *original* text where the suggestion should be.

Net effect: any finding whose fix is "re-indent this block" or "fix the casing of this identifier" renders as an unmodified grey context block. The user sees no change, and if they copy what is displayed they copy the old code.

### Steps to Reproduce

Deterministic unit-level repro. Add this file to `internal/suggestdiff/` and run `go test ./internal/suggestdiff/ -run TestRepro -v`:

```go
package suggestdiff

import "testing"

func TestReproIndentationOnlyChange(t *testing.T) {
	old := []string{"func f() {", "\treturn nil", "}"}
	new := []string{"func f() {", "\t\treturn nil", "}"}
	got := ComputeLineDiff(old, new)
	for _, l := range got {
		t.Logf("type=%d content=%q", l.Type, l.Content)
	}
	added, deleted := 0, 0
	for _, l := range got {
		if l.Type == DiffAdded {
			added++
		}
		if l.Type == DiffDeleted {
			deleted++
		}
	}
	if added == 0 || deleted == 0 {
		t.Errorf("indentation-only change not rendered as a change: added=%d deleted=%d", added, deleted)
	}
}

func TestReproCaseOnlyChange(t *testing.T) {
	old := []string{"const maxRetries = 3"}
	new := []string{"const MaxRetries = 3"}
	got := ComputeLineDiff(old, new)
	added, deleted := 0, 0
	for _, l := range got {
		t.Logf("type=%d content=%q", l.Type, l.Content)
		if l.Type == DiffAdded {
			added++
		}
		if l.Type == DiffDeleted {
			deleted++
		}
	}
	if added == 0 || deleted == 0 {
		t.Errorf("case-only change not rendered as a change: added=%d deleted=%d", added, deleted)
	}
}
```

End-to-end: any review whose `SuggestionCode` differs from `ExistingCode` only by leading whitespace (e.g. "move this statement into the `if` block") or only by letter case (e.g. `const maxRetries` -> `const MaxRetries`). No LLM call is needed to observe it — the defect is entirely in the renderer.

### Expected Behavior

- A leading-whitespace difference is a real change: emit DiffDeleted + DiffAdded (or at least mark the line changed).
- A letter-case difference is a real change: emit DiffDeleted + DiffAdded.
- When a line is treated as equal for alignment purposes, it must not be displayed with the old content as though it were the suggestion.

### Logs / Error Output

```shell
=== RUN   TestReproIndentationOnlyChange
    zz_repro_test.go:11: type=0 content="func f() {"
    zz_repro_test.go:11: type=0 content="\treturn nil"
    zz_repro_test.go:11: type=0 content="}"
    zz_repro_test.go:24: indentation-only change not rendered as a change: added=0 deleted=0
--- FAIL: TestReproIndentationOnlyChange (0.00s)
=== RUN   TestReproCaseOnlyChange
    zz_repro_test.go:36: type=0 content="const maxRetries = 3"
    zz_repro_test.go:45: case-only change not rendered as a change: added=0 deleted=0
--- FAIL: TestReproCaseOnlyChange (0.00s)
FAIL
FAIL	github.com/alibaba/open-code-review/internal/suggestdiff	0.801s

(type=0 is DiffContext; note the second line is echoed as the OLD "\treturn nil", not the suggested "\t\treturn nil", and no DiffAdded/DiffDeleted was produced at all.)
```

### Additional Context

Call path: `cmd/opencodereview/output.go:259 buildDiffLines` (splits `ExistingCode`/`SuggestionCode`) -> `suggestdiff.ComputeLineDiff` -> rendered at `output.go:119-130`.

`TrimSpace` may well have been deliberate (absorbing indentation drift in LLM output). `EqualFold` almost certainly was not: it silently discards semantically meaningful changes in case-sensitive languages. A conservative fix would keep fuzzy matching for LCS *alignment* but emit DiffDeleted/DiffAdded whenever the raw lines differ (e.g. compare `a == b` after normalizing only a trailing `\r`), so a matched-but-different pair is still shown as a change.

Happy to open a PR with the fix plus the two tests above if you agree with the direction.

---

**AI/LLM Disclosure:** AI/LLM was used to help find and write this report (code reading, drafting the wording). I verified every claim above by reading the source at the cited lines and by actually building the project and running the tests quoted in the Logs section; the reproduction output is real output from my machine, not generated text.

mo-tunn:
## Description

`ComputeLineDiff` uses case-insensitive, trimmed line equality to align the LCS. The same fuzzy equality was also treated as unchanged context during backtracking, which could hide real case-only or indentation-only suggestion changes.

This change keeps fuzzy equality for LCS alignment, but separates rendering semantics:

- exact raw line matches remain `DiffContext`;
- fuzzy-but-not-exact matches are emitted as a deletion followed by an addition;
- regression tests cover case-only and indentation-only changes and assert the rendered line contents/order.

## Type of Change

- [x] Documentation update
- [x] CI / Build / Tooling

## How Has This Been Tested?

- `go test ./internal/suggestdiff` — pass
- Regression test verified to fail against the pre-fix `main` behavior before applying the production change
- `go test ./...` — pass
- `go vet ./...` — pass
- `go build ./...` — pass
- `git diff --check` — pass
- `make build` — pass
- `make coverage` — pass, 90.9% total coverage (90% threshold)
- `make check` — license check and `go mod tidy` passed, then the Windows GNU Make environment could not launch extensionless `gofmt`; the remaining equivalent checks were run directly and passed (`gofmt -s -l .`, license check, English-only check, and `go vet ./...`)
- `make test` — cannot run the repository's `-race` gate on this Windows/amd64 environment because the available C compiler targets `i686-w64-mingw32`; `runtime/cgo` fails with `cc1.exe: sorry, unimplemented: 64-bit mode not compiled in`. The non-race full test suite passes as noted above.

## Checklist

- [x] My code follows the project coding style (`go fmt`, `go vet`)
- [x] I have signed the CLA

OpenAI ChatGPT (GPT-5.6 Sol) was used to investigate the issue, draft the change, and design the regression tests.

## Related Issues

Closes #1249


lizhengfeng101 (approved):
LGTM

PR #1259, opened 2026-09-15
fix(cli): flush telemetry before error exit

### LINKED ISSUE #1250: bug: telemetry is never flushed on a failing run - os.Exit(1) skips the deferred Shutdown
### OpenCodeReview Version

open-code-review dev windows/amd64 (built from source at main 6e5cd8c)

### Operating System

Windows

### Installation Method

Built from source

### LLM Provider

### Bug Description

`cmd/opencodereview/main.go:20-27`:

```go
if telemetry.Init(ctx) {
	defer telemetry.ShutdownWithTimeout(ctx, 5*time.Second)
}

if err := rootCmd.Execute(); err != nil {
	fmt.Fprintf(os.Stderr, "Error: %v\n", err)
	os.Exit(1)
}
```

`os.Exit` does not run deferred functions, so the moment `rootCmd.Execute()` returns an error the process exits without ever calling `telemetry.Shutdown`. Everything still sitting in the OTel buffers (spans and metrics for the whole run) is dropped, and no warning is printed either, because the `[ocr] WARNING: telemetry shutdown error: ...` line lives inside `Shutdown` (`internal/telemetry/shutdown.go:40`), which never runs.

`internal/telemetry/shutdown.go:13-14` states the contract that is being broken: "It should be called before process exit to ensure buffered data is exported."

The runs that lose their telemetry are exactly the ones you most want telemetry for - the failing ones.

### Steps to Reproduce

1. Enable telemetry (OTLP endpoint or the console exporter) so that `telemetry.Init` returns true.
2. Run any command that exits non-zero, e.g. `ocr review --from <bad-ref> --to <ref>` in a repository, or any command with an invalid/missing configuration.
3. Compare with a successful run of the same command.

Observed: the successful run exports its spans/metrics; the failing run exports nothing (and prints no shutdown warning).

### Expected Behavior

Telemetry should be flushed on every exit path, including the error path - i.e. `ShutdownWithTimeout` should run before the process terminates with a non-zero status.

### Logs / Error Output

```shell
No log line is emitted at all on the failing path, which is part of the problem: the only diagnostic (`[ocr] WARNING: telemetry shutdown error: ...`) is inside the function that is skipped.
```

### Additional Context

The usual fix is to keep the `defer` in a helper that returns an exit code so the deferred call still runs:

```go
func main() {
	llm.AppVersion = Version
	os.Exit(run())
}

func run() int {
	ctx := context.Background()
	if telemetry.Init(ctx) {
		defer telemetry.ShutdownWithTimeout(ctx, 5*time.Second)
	}
	if err := rootCmd.Execute(); err != nil {
		fmt.Fprintf(os.Stderr, "Error: %v\n", err)
		return 1
	}
	return 0
}
```

Note there is a second, smaller exit path worth checking while you are in here: any `os.Exit` reached from inside a Cobra `RunE` (rather than by returning an error) would have the same effect.

I did not open a PR for this one because I wanted to confirm first whether you consider the error path in scope for telemetry.

---

**AI/LLM Disclosure:** AI/LLM was used to help find and draft this report. I verified the code path by reading `cmd/opencodereview/main.go` and `internal/telemetry/shutdown.go` at the cited lines; no part of the report is inferred without that reading.

iluv7:
## Summary

- move CLI execution into a `run` helper that returns an exit code
- allow deferred telemetry shutdown to run before non-zero process exit
- add a subprocess regression test that verifies telemetry is flushed on the error path

Fixes #1250

## Testing

- `make check`
- `make test`
- `make coverage` (91.8%)
- `ocr review --audience agent --background "Fix issue #1250 by ensuring deferred telemetry shutdown runs before the CLI exits with a non-zero status, with a regression test for the failing command path."` (0 findings)

## AI/LLM Disclosure

OpenAI Codex (GPT-5) was used to analyze the issue, implement the fix, and draft the regression test and PR description.


cmd/opencodereview/main_test.go
```diff
@@ -0,0 +1,31 @@
...
+	if os.Getenv("OCR_TEST_RUN_ERROR") == "1" {
+		rootCmd.SetArgs([]string{"--invalid-flag"})
+		os.Exit(run())
+	}
+
+	cmd := exec.Command(os.Args[0], "-test.run=^TestRunFlushesTelemetryOnError$")
+	cmd.Env = append(os.Environ(),
+		"OCR_TEST_RUN_ERROR=1",
+		"OCR_ENABLE_TELEMETRY=1",
+	)
```
wu21-web, line 27:
```suggestion
	home := t.TempDir()
	cmd.Env = append(os.Environ(),
		"OCR_TEST_RUN_ERROR=1",
		"OCR_ENABLE_TELEMETRY=1",
		"OTEL_EXPORTER_OTLP_ENDPOINT=",
		"HOME="+home,
		"USERPROFILE="+home,
	)
```

I am not sure if creating a test for main.go is suitable, but here is a suggestion.

iluv7: Agreed, I'll isolate the test from the user's local configuration by using a temporary HOME/USERPROFILE and clearing OTEL_EXPORTER_OTLP_ENDPOINT. Thanks for the suggestion!

wu21-web: @iluv7 
> I am not sure if creating a test for main.go is suitable

iluv7: @wu21-web 
I think a subprocess test makes sense here since the bug is about process exit behavior, but I’m happy to adjust it if you prefer another approach.

wu21-web: It is the first time someone write a `main_test.go`, the test ifself look fine.

wu21-web:
I would like to ask an attribution to me in the HEAD commit message because you applied exactly my suggestion.

cmd/opencodereview/main_test.go
iluv7: Added you as a co-author in the updated HEAD commit. Thanks again for the suggestion!

lizhengfeng101 (approved):
LGTM

PR #1262, opened 2026-09-15
feat(allowlist): exclude secret paths from review

### LINKED ISSUE #1240: feat(allowlist): exclude extensionless and dotfile secrets (.env, id_rsa, .netrc) from review by default

hyy321:
## Description

This PR adds built-in path-based protection for common secret files such as `.env`, SSH private keys, `.netrc`, `.npmrc`, `.pypirc`, and `.dockercfg`.

The secret-path patterns are kept separate from the existing default exclude patterns because they have different precedence: default excludes can be overridden by an include rule, while secret paths cannot.

Both review and scan apply the secret-path check before user include/exclude rules and report matches as `secret_exclude`. For review diffs, both `OldPath` and `NewPath` are checked so a rename from a protected path to an allowed path cannot expose content from the protected source path.

Agent runtime tools are intentionally left unchanged, following the scope discussed in #1240.

**Known limitation:** This PR keeps the secret patterns scoped to #1240, so files such as `.env.production`, `.env.staging`, and `.env.development` are not covered yet. Broader .env.* coverage is tracked in #1265.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Additional validation:

- `make check`
- `make build`
- `git diff --check upstream/main...HEAD`
- Local OCR self-review with `deepseek-v4-pro`: 0 findings across 7 selected items

Tests cover secret-path matching, include/exclude precedence, rename handling, logging, and review/scan selection behavior.

## Checklist

- [x] I have signed the CLA

## AI/LLM Disclosure

AI/LLM tools were used during implementation and review.

Tools: Codex and OpenCodeReview.\
OCR self-review model: `deepseek-v4-pro`.

I reviewed the final changes and test results before submission.

## Related Issues

Closes #1240

Follow-up: #1265


internal/config/allowlist/default_secret_patterns.json
wu21-web, line 1: Why do we have all these workarounds here, because in my solution these are just no allowed (not even supported), so they are excluded at first. 
This is my way of solving this problem. Just asking, thanks!

hyy321: Thanks for asking. I followed the design in #1240 here. I kept the secret-path check separate from the extension allowlist so files like Dockerfile and Makefile keep their current behavior.
Secret paths are different because include rules shouldn't be able to make them reviewable again, and they need their own secret_exclude reason. #1240 also discussed handling dotfiles through the extension path, but that would change the existing extensionless-file behavior.

wu21-web: Sorry, I didn't read the issue. My bad. I will speak to the maintainer.

lizhengfeng101:
@hyy321 Thanks — the implementation is clean and the rename/`OldPath` handling and dedicated log wording are nice touches. Two non-blocking points before merge:

**1. `.env.production` / `.env.staging` blind spot (most important)**

The patterns cover `.env`, `.env.local`, and `.env.*.local`, but not per-environment files like `.env.production` or `.env.development` — which are just as common and just as likely to hold real secrets. This is a faithful, conservative reading of the issue (avoiding false positives on `.env.example`), so it need not block this PR, but it is the largest real-world gap. Suggested:
- Call out the limitation explicitly in the PR / merge notes.
- Open a follow-up that switches to a denylist approach: match `.env.*` while excluding `.example` / `.sample` / `.template`, so `.env.<env>` is covered.

**2. Missing documentation**

`pages/src/content/docs/*/review-rules.md` does not mention `secret_exclude` at all — users who see that reason in `--preview` have no way to know what it means. Please add at least a line explaining it is a built-in protection that `include` cannot override.
- Note: this PR touches no `.md`. If you add docs, put them in `review-rules.md` for each locale — **do not touch README**, or you will trigger the multi-language README sync requirement in CLAUDE.md. `docs/i18n/README.*` stays out of scope.

hyy321:
> @hyy321 Thanks — the implementation is clean and the rename/`OldPath` handling and dedicated log wording are nice touches. Two non-blocking points before merge:
> 
> **1. `.env.production` / `.env.staging` blind spot (most important)**
> 
> The patterns cover `.env`, `.env.local`, and `.env.*.local`, but not per-environment files like `.env.production` or `.env.development` — which are just as common and just as likely to hold real secrets. This is a faithful, conservative reading of the issue (avoiding false positives on `.env.example`), so it need not block this PR, but it is the largest real-world gap. Suggested:
> 
> * Call out the limitation explicitly in the PR / merge notes.
> * Open a follow-up that switches to a denylist approach: match `.env.*` while excluding `.example` / `.sample` / `.template`, so `.env.<env>` is covered.
> 
> **2. Missing documentation**
> 
> `pages/src/content/docs/*/review-rules.md` does not mention `secret_exclude` at all — users who see that reason in `--preview` have no way to know what it means. Please add at least a line explaining it is a built-in protection that `include` cannot override.
> 
> * Note: this PR touches no `.md`. If you add docs, put them in `review-rules.md` for each locale — **do not touch README**, or you will trigger the multi-language README sync requirement in CLAUDE.md. `docs/i18n/README.*` stays out of scope.

Thanks, that makes sense. I’ll keep the current secret patterns scoped to #1240, note the .env.<environment> limitation in the PR, and open a follow-up issue for broader .env.* coverage. I’ll also add an explanation of secret_exclude to the localized review-rules.md files.

lizhengfeng101:
> Thanks, that makes sense. I’ll keep the current secret patterns scoped to #1240, note the .env. limitation in the PR, and open a follow-up issue for broader .env.* coverage. I’ll also add an explanation of secret_exclude to the localized review-rules.md files.

@hyy321 Will you add an explanation regarding `secret_exclude` to the current PR?

hyy321:
> > Thanks, that makes sense. I’ll keep the current secret patterns scoped to #1240, note the .env. limitation in the PR, and open a follow-up issue for broader .env.* coverage. I’ll also add an explanation of secret_exclude to the localized review-rules.md files.
> 
> @hyy321 Will you add an explanation regarding `secret_exclude` to the current PR?

Yes, I’ll include that in the current PR. I’m updating the localized review-rules.md files to explain secret_exclude and that include cannot override it.

hyy321:
I’ve added it to the current PR. The localized review-rules.md files now explain secret_exclude, including that it is a built-in protection that include cannot override.

lizhengfeng101 (approved):
LGTM

PR #1278, opened 2026-09-15
fix: recover tool-call JSON with trailing content and back off repeated failures

### LINKED ISSUE #1276: code_comment tool-call fails on trailing content after valid JSON, then retries the same finding indefinitely with no backoff
## Bug description

When a model's `code_comment` tool-call arguments contain a well-formed JSON value followed by trailing non-JSON content, `parseToolArgs` (`internal/llmloop/loop.go`) rejects the whole string because it requires the entire raw string to be valid JSON:

```go
func parseToolArgs(raw string) (map[string]any, error) {
	var args map[string]any
	if err := json.Unmarshal([]byte(raw), &args); err != nil {
		return nil, err
	}
	...
}
```

`executeToolCall` then returns the parse error as the tool result and lets the model retry next round — but there is no backoff or dedup for a repeated failure of the *same* tool call within a task. In practice the model reattempts the identical finding, reworded slightly each time, and keeps failing the same way.

## Reproduction

Observed running `openai.gpt-oss-20b-1:0` via Bedrock on a ~21-file diff, `--max-tokens-budget 2000000`, default concurrency. gpt-oss-20b's harmony chat format leaves trailing control tokens after its `code_comment` JSON (e.g. `...}]}/>\n<|start|><|start|><|channel|>final<|message|>...`), so `parseToolArgs` fails even though the JSON payload itself is well-formed. One specific finding was reattempted **6 times** in a single review run (tool calls #103, 106, 109, 114, 115, 116 in the retained result), each attempt a full LLM round-trip resending the accumulated conversation, before the run exhausted its token budget with 3 of 21 files never reviewed (`status: "partial"`).

Three runs of the identical diff/model/config varied 3.1M / 3.3M / 9.2M total tokens — consistent with this retry pattern landing at different points in different conversations (a retry late in a long conversation costs far more than an early one, since the full history resends each round).

This isn't specific to gpt-oss — any model that appends non-JSON content after its tool-call arguments, or rewords-and-retries after a tool error, hits the same failure mode.

## Expected behavior

1. A tool-call argument string containing a complete, well-formed JSON value followed by trailing content should still parse successfully (extract the first balanced top-level JSON value instead of requiring the entire string to be valid JSON).
2. A tool call that keeps failing identically within the same task should back off — e.g. a stronger corrective message after 2 consecutive failures, and give up on that one call (drop it, don't fail the task) after a small bounded number of attempts, rather than retrying indefinitely against the round budget.

## Related

Possibly related to #1247 (`--max-tokens-budget` enforcement gap) — both result in actual token spend exceeding the configured budget, but via distinct mechanisms (dispatch-time gate timing vs. an unbounded same-finding retry loop within an already-dispatched group). Filing separately since the root causes and fixes are different.

I'm preparing a PR with a fix for both parts above and will link it here.

liRONCO11:
## Summary

`code_comment` tool-call arguments containing a well-formed JSON value followed by trailing non-JSON content fail the whole parse in `parseToolArgs`, and a tool call that keeps failing the same way for the same task retries indefinitely with no backoff — burning tokens re-attempting an identical finding rather than either succeeding or giving up.

Fixes #1276

## Changes

1. `internal/llmloop/tool_args_json.go` (new) — `extractTopLevelJSON` scans for the first balanced top-level `{}`/`[]` JSON value, correctly skipping over string contents (including escaped quotes). `parseToolArgs` tries a whole-string parse first, and only on failure retries against the extracted span; if no balanced value exists at all it falls back to the original error unchanged (no regression for genuinely non-JSON input).

2. `internal/llmloop/tool_failure_streak.go` (new) — tracks a per-`(taskKey, toolName)` failure streak. 1st failure: unchanged error text. 2nd: same error plus an explicit instruction to fix the arguments or call `task_done`. 3rd and later: a non-error "skipped, won't be retried" result, so the loop stops burning rounds on a call that's already failed twice in a row. Any success resets the streak. `task_done` itself is excluded (telling it to "call task_done" would be circular).

Both changes are generic — no model-specific string matching (e.g. nothing gpt-oss/harmony-format specific) — since any provider can produce trailing content after valid JSON or retry a failed call without making progress.

## Tests

- `tool_args_json_test.go` — reproduces the original bug (valid JSON followed by trailing control-token-style content, confirms it now parses), confirms genuinely invalid input still errors, and unit-tests the extraction helper directly (trailing garbage, quoted braces, escaped quotes, unbalanced input).
- `tool_failure_streak_test.go` — asserts the exact 1st/2nd/3rd/4th-failure wording, confirms the streak is scoped per `(taskKey, toolName)` (a different task or tool starts fresh), and confirms a success resets the streak.

## Verification

- `go build ./...`, `go vet ./...` clean
- `go test ./...`: all packages passing
- `go test -race -count=1 -coverprofile=coverage.out ./...`: all green under the race detector, 91.6% coverage (CI gate 90%)
- `gofmt -s -l .` empty, license/English-only checker scripts clean, `go mod tidy` produces no diff
- One pre-existing, unrelated skip (`TestInterruptChildHarness`), untouched by this change


lizhengfeng101 (approved):
LGTM

PR #1279, opened 2026-09-15
feat(allowlist): protect per-environment .env files

### LINKED ISSUE #1265: feat(allowlist): cover per-environment .env.* files in secret-path protection
### Problem Statement

#1240 adds built-in secret-path protection for `.env`, `.env.local`, and `.env.*.local`, but common per-environment files such as `.env.production`, `.env.staging`, and `.env.development` are still not covered.

These files are commonly used for environment-specific configuration and may contain real secrets, so they should receive the same protection as the existing `.env` patterns.

This was identified as a non-blocking follow-up during review of #1262.

### Proposed Solution

Extend the built-in secret-path matching to cover `.env.*`, while keeping common committed template files reviewable.

For example:

- Exclude `.env.production`
- Exclude `.env.staging`
- Exclude `.env.development`
- Keep `.env.example` reviewable
- Keep `.env.sample` reviewable
- Keep `.env.template` reviewable

The existing secret-path behavior should stay the same:

- Matches are reported as `secret_exclude`.
- User `include` rules cannot override the protection.
- The change remains limited to review/scan selection.
- Agent runtime tools and content-based secret detection remain out of scope.

### Alternatives Considered

Matching `.env.*` without exceptions would also exclude committed template files such as `.env.example`, `.env.sample`, and `.env.template`. Keeping the current patterns unchanged avoids those false positives but leaves common `.env.<environment>` files unprotected.

### Affected Area

Review Rules

### Additional Context

_No response_

hyy321:
## Description

This PR extends the built-in secret-path protection added in #1262 to cover common per-environment files such as `.env.production`, `.env.staging`, and `.env.development`.

The existing `.env.local` and `.env.*.local` patterns are replaced by the broader `.env.*` pattern. `.env.example`, `.env.sample`, and `.env.template` remain exempt from that rule so committed template files can still be reviewed normally.

The template exception applies only to the `.env.*` rule and does not bypass other secret-path protections. For example, `.ssh/.env.example` is still excluded by the existing `.ssh/**` rule.

The existing `secret_exclude` precedence and review/scan selection behavior remain unchanged. The localized review-rules documentation has also been updated to describe the template exceptions.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Additional validation:

- `go test -run TestIsSecretPath -count=1 ./internal/config/allowlist`
- `go test ./internal/config/allowlist`
- `make check`
- `make build`
- `git diff --cached --check`
- Local OCR self-review with `deepseek-v4-pro`: 0 findings across 2 selected items

Tests cover per-environment `.env.*` paths, template exceptions, case-insensitive matching, and ensure template exceptions do not bypass other secret-path rules.

## Checklist

- [x] I have signed the CLA

## AI/LLM Disclosure

AI/LLM tools were used during implementation and review.

Tools: Codex and OpenCodeReview.  
OCR self-review model: `deepseek-v4-pro`.

I reviewed the final implementation, diff, and test results before submission.

## Related Issues

Closes #1265

Follow-up to #1262


internal/config/allowlist/secret_path.go
```diff
@@ -42,13 +43,27 @@ func initSecret() {
...
 // the extension allowlist and the default exclude patterns.
-func IsSecretPath(path string) bool {
+func IsSecretPath(filePath string) bool {
 	secretOnce.Do(initSecret)
-	lowerPath := strings.ToLower(path)
+	lowerPath := strings.ToLower(filePath)
+	envTemplate := isEnvTemplate(lowerPath)
 	for _, pattern := range secretPatterns {
+		// Template names are exempt only from the broad .env.* rule.
+		if envTemplate && pattern == "**/.env.*" {
```
lizhengfeng101: @hyy321 There is an implicit string-based coupling between the Go code and the JSON; this implementation is prone to errors.

internal/config/allowlist/secret_path.go
```diff
@@ -42,13 +43,27 @@ func initSecret() {
...
 		if matched, _ := doublestar.Match(pattern, lowerPath); matched {
 			return true
 		}
 	}
 	return false
 }
+
+func isEnvTemplate(filePath string) bool {
+	switch path.Base(filePath) {
+	case ".env.example", ".env.sample", ".env.template":
```
wu21-web: What's the point of reviewing a template? This doesn't add too much value.

internal/config/allowlist/secret_path.go
hyy321: I’ll remove the dependency on the literal JSON pattern and handle per-environment .env.* paths directly in the matcher. Existing secret patterns will still use the JSON-based matching path, so template exceptions won’t bypass rules such as .ssh/**.

internal/config/allowlist/default_secret_patterns.json
```diff
@@ -1,7 +1,6 @@
 [
   "**/.env",
-  "**/.env.local",
-  "**/.env.*.local",
+  "**/.env.*",
```
wu21-web:
```suggestion
  "**/*.env.*",
```

Use this so that doesn't omit `config.env.example`, which is also common.

internal/config/allowlist/secret_path.go
hyy321: Updated. The .env family is now handled directly by the matcher, and the .env patterns have been removed from default_secret_patterns.json. The JSON list now only contains unconditional secret-path globs, so there is no longer a literal dependency between the Go logic and the JSON patterns. Template paths still fall through to the remaining rules, so cases such as .ssh/.env.example remain protected.

internal/config/allowlist/default_secret_patterns.json
hyy321: This is addressed by the latest change. The .env family is no longer matched through a JSON glob and is handled directly by the matcher instead, so this pattern replacement no longer applies.

lizhengfeng101:
@hyy321 [CI / test (pull_request)](https://github.com/alibaba/open-code-review/actions/runs/34979646957/job/104416301038?pr=1279)Failing after 4m

hyy321:
> @hyy321 [CI / test (pull_request)](https://github.com/alibaba/open-code-review/actions/runs/34979646957/job/104416301038?pr=1279)Failing after 4m

I traced the failure to TestCloseAll_TimeoutPreservesCollectedErrors in internal/mcp. This PR does not touch the MCP package. I also ran the same CI command locally successfully, and the failing test passes 3 consecutive runs with -race. The surrounding stress tests already account for -race timing differences, so this looks like a timing-sensitive flake. I don't appear to have permission to rerun the failed job; could you rerun it?

lizhengfeng101 (approved):
LGTM

internal/config/allowlist/secret_path.go
verdrellw60-max: > I’ll remove the dependency on the literal JSON pattern and handle per-environment .env.* paths directly in the matcher. Existing secret patterns will still use the JSON-based matching path, so template exceptions won’t bypass rules such as .ssh/**.

>

PR #1285, opened 2026-09-15
fix(rules): route all .gradle build scripts to the gradle rule

xiaodu55:
## What

`build_gradle.md` currently applies only to `**/build.gradle`. Every other Gradle build script — `settings.gradle`, standalone `.gradle` scripts included via `apply from:` — falls through to `default.md`, even though they can pin dependency versions the same way (`pluginManagement`, `dependencyResolutionManagement`, `dependencies` blocks). The snapshot-version rule therefore never reaches them.

This widens the system path rule to `**/*.gradle` and updates the routing tables in the en/ja/ko/ru/zh review-rules docs. Same pattern as #1232 (`.pyi` → python rules).

## Checks

- `go test -count=1 ./...`: all packages pass, incl. new resolve cases for `build.gradle` / `settings.gradle` / included scripts
- `gofmt -s`, `go vet`, license check: clean


wu21-web:
Why? The rule doc is only for `build.gradle` itself, not for glob `*.gradle`.

xiaodu55:
Thanks for the context — that matches a deliberate scoping: the doc was written for `build.gradle` itself. Closing this and leaving the routing as-is; if version pinning in `settings.gradle` (`pluginManagement` / `dependencyResolutionManagement`) ever deserves the snapshot rule, that can be revisited as its own change.

PR #1291, opened 2026-09-15
security: validate credential commands and warn on plaintext secrets

Kamakhya-s:
- Add validateKeyCmd() in internal/llm/keycmd.go to reject suspicious shell patterns (command separators, backticks, redirections) in api_key_cmd and auth_token_cmd before the shell executes them. Provides a clear diagnostic for tampered or accidentally malformed config rather than silently running destructive expressions.

- Add plaintext credential warnings in internal/llm/resolver.go when api_key or auth_token are stored as static strings in the config file. Credential is still accepted; the warning nudges users toward api_key_cmd / auth_token_cmd so secrets are not at rest in JSON.

- Extend .gitattributes with explicit binary markers for web fonts (woff, woff2, ttf, otf, eot), WebAssembly (.wasm), compressed archives (.zip, .tar, .gz, .tgz), and other image formats (.webp, .bmp, .tiff) to prevent CRLF corruption on Windows clones.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [ ] I have signed the CLA


lizhengfeng101:
Thanks for the security hardening — the direction is sound and the code comments are thoughtful. A few things to address before this can move forward:

**1. Missing tests.** No tests were added, but `AGENTS.md` requires tests for new code and CI enforces a 90% coverage threshold. The `validateKeyCmd` error path (backtick / NUL rejection) is entirely uncovered. Please add cases for: a backtick command being rejected, a NUL byte being rejected, and the new plaintext warnings firing.

**2. Warning noise.** `ResolveEndpoint` runs on every `ocr review` / `ocr scan`, so a static `api_key` in the config (a common, legitimate setup) prints a `[ocr] WARNING` to stderr on *every* run. This risks warning fatigue. Could we gate it (print once, or make it suppressible) rather than emitting it unconditionally each run? Also note that when both `api_key` and `api_key_cmd` are set, the user now gets two warnings for the same field.

**3. Comment length.** The block comments around `suspiciousPatterns` / `validateKeyCmd` are quite long relative to the logic. Please trim to the essentials — the rationale for excluded patterns can be condensed.

**4. CLA.** The CLA checkbox is unchecked — please sign it so this can be merged.

Happy to re-review once these are addressed.

lizhengfeng101 (approved):
LGTM

PR #1292, opened 2026-09-15
fix(install): add download timeout to asset fetch in install.sh

SameerAlSahab:
fix(install): add timeout to asset download in install.sh

The release binary download had no --connect-timeout or --max-time, while the sha256sum.txt download right after it already had both. On a stalled or unreachable connection this leaves the installer hanging indefinitely instead of failing with a clear error, which is especially disruptive for `curl | sh` usage in CI or automated setups.

Add the same timeout flags used for the checksum fetch, with a higher --max-time (120s vs 15s) since the binary is larger than the checksum file.

## Description

The main binary download in `install.sh` had no `--connect-timeout`/`--max-time` flags, unlike the `sha256sum.txt` download right below it. On a stalled connection (common in the regions this script's mirror support targets), the install hangs forever instead of failing cleanly. This adds matching timeout flags to the asset download, using a longer `--max-time` (120s) than the checksum fetch (15s) to account for the binary being much larger.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Ran `install.sh` locally against a normal release download to confirm the timeout flags don't affect the happy path (binary downloads and installs as before). Simulated a slow/unreachable endpoint to confirm the script now fails with a clear `download failed` error after ~120s instead of hanging indefinitely.

## Checklist

- [ ] I have signed the CLA

## Related Issues

N/A


install.sh
```diff
@@ -58,7 +58,8 @@ main() {
   trap 'rm -rf "$tmp"' INT TERM EXIT
 
   printf 'downloading %s %s (%s/%s)...\n' "$BIN" "$VERSION" "$os" "$arch"
-  curl -fsSL -o "$tmp/$asset" "$base/$asset" || err "download failed: $base/$asset"
+  curl -fsSL --connect-timeout 5 --max-time 120 -o "$tmp/$asset" "$base/$asset" ||
+    err "download failed: $base/$asset"
```
wu21-web: This might be duplicate, consider removing it because curl output is not empty. Curl outputs first and then you gave a general error message. I'd keep the curl one because it gives more details.

```suggestion
```

wu21-web (changes requested):

install.sh
wu21-web: This might be duplicate, consider removing it because curl output is not empty. Curl outputs first and then you gave a general error message. I'd keep the curl one because it gives more details.

```suggestion
  curl -fsSL --connect-timeout 5 --max-time 120 -o "$tmp/$asset" "$base/$asset"
```

wu21-web (changes requested):

install.sh
```diff
@@ -58,7 +58,8 @@ main() {
   trap 'rm -rf "$tmp"' INT TERM EXIT
 
   printf 'downloading %s %s (%s/%s)...\n' "$BIN" "$VERSION" "$os" "$arch"
-  curl -fsSL -o "$tmp/$asset" "$base/$asset" || err "download failed: $base/$asset"
+  curl -fsSL --connect-timeout 5 --max-time 120 -o "$tmp/$asset" "$base/$asset" ||
```
lizhengfeng101: --max-time recommended value 1800

wu21-web: Can we enable the progress bar for curl? Since it is good to know how much data curl have received.
Here is a compacted one:
```suggestion
  curl -fL --progress-bar --connect-timeout 5 --max-time 1800 -o "$tmp/$asset" "$base/$asset" ||
```

wu21-web (approved):

SameerAlSahab:
Updated the code based on the suggestions from @lizhengfeng101 and @wu21-web. Please have a look when you get a chance!

wu21-web:
@SameerAlSahab Does the install.ps1 behaves the same as the shell script? It is good to keep them in sync.

SameerAlSahab:
> @SameerAlSahab Does the install.ps1 behaves the same as the shell script? It is good to keep them in sync.

I have updated that too sir! Take a look.

wu21-web (approved):
Looks great. Thank you!

lizhengfeng101:
Thanks for the fix — the hang on a stalled connection is real and worth addressing.

One suggestion on the choice of flag: `--max-time` caps *total* transfer time, which conflates two different things. Too low (120s) can kill a slow-but-healthy mirror download of the binary; too high just makes CI wait longer before failing on a genuine stall — which undercuts the goal here.

A stall-detection approach sidesteps the "how many seconds" question entirely:

```sh
curl -fsSL --connect-timeout 5 --speed-limit 1024 --speed-time 30 \
  -o "$tmp/$asset" "$base/$asset" || err "download failed: $base/$asset"
```

This aborts only if throughput stays under 1KB/s for 30s — fast to fail on a dead connection, but won't penalize a slow mirror that's still making progress, and needs no `--max-time` ceiling.

wu21-web:
can we add that status bar for curl

SameerAlSahab:
> Thanks for the fix — the hang on a stalled connection is real and worth addressing.
> 
> One suggestion on the choice of flag: `--max-time` caps _total_ transfer time, which conflates two different things. Too low (120s) can kill a slow-but-healthy mirror download of the binary; too high just makes CI wait longer before failing on a genuine stall — which undercuts the goal here.
> 
> A stall-detection approach sidesteps the "how many seconds" question entirely:
> 
> ```shell
> curl -fsSL --connect-timeout 5 --speed-limit 1024 --speed-time 30 \
>   -o "$tmp/$asset" "$base/$asset" || err "download failed: $base/$asset"
> ```
> 
> This aborts only if throughput stays under 1KB/s for 30s — fast to fail on a dead connection, but won't penalize a slow mirror that's still making progress, and needs no `--max-time` ceiling.

Yeah nice idea!! I also preserved the status progress bar too beside that @wu21-web . Thanks!

SameerAlSahab:
I also checked install.ps1 as well. Since PowerShell's Invoke-WebRequest natively handles connection and request timeouts via -TimeoutSec, I kept it as is for Windows environments unless you would like any adjustments there too. But ig Timeout is enough there.

lizhengfeng101:
@SameerAlSahab Are there any updates that haven't been pushed yet?

SameerAlSahab:
> @SameerAlSahab Are there any updates that haven't been pushed yet?

No sir. I guess the current ones are fine.

lizhengfeng101 (approved):
LGTM

PR #1293, opened 2026-09-15
docs(ci): standardize review-language configuration

### LINKED ISSUE #1266: docs(ci): review-language configuration differs per CI example and is missing from four of them
### Issue Type

Missing documentation

### Document Location

- `examples/README.md` (index)
- `examples/github_actions/README.md`
- `examples/gitlab_ci/README.md`
- `examples/{bitbucket_pipelines,codeup_ci,gerrit_ci,gitflic_ci}/`
- `action.yml` — the `language` input, lines 56-61
- `pages/src/content/docs/*/configuration.md` — "Configuring the review language"

### Description

Only two of the six CI examples under `examples/` can set the review output
language, and they do it through two different mechanisms. Neither mechanism is
described in the docs site. For the other four examples, the obvious thing to
try (`OCR_LANGUAGE`) is silently ignored — the review still comes out in
English, with no error and no warning.

This came up in #1020. A user on CODING — which has no example under
`examples/` — asked whether review output can be Chinese, and was told:

> In CI you don't need the extra step: the pipeline examples already read an
> `OCR_LANGUAGE` variable and run `ocr config set language` for you, so setting
> `OCR_LANGUAGE=Chinese` in your CI variables is enough.

That holds for `gitlab_ci`, but not for the four examples that have no language
handling at all — including `codeup_ci`, which is the example that same reply
recommends starting from. A CODING user following it sets `OCR_LANGUAGE`, gets no
error, and receives English output.

To be clear, this is not #263. That asked whether Chinese output exists at all,
and it does (`ocr config set language`). The gap here is narrower: the CI path to
it is implemented in only one example, and named inconsistently across the two
that support it.

**Current coverage:**

| Example | Mechanism | User-facing knob |
|---|---|---|
| `github_actions` | `action.yml` `language` input, wired to an internal `OCR_LANGUAGE` env, which runs `ocr config set language` | `with: language: Chinese` |
| `gitlab_ci` | `.gitlab-ci.yml` reads `OCR_LANGUAGE` from the environment | CI variable `OCR_LANGUAGE=Chinese` |
| `bitbucket_pipelines` | none | — |
| `gitflic_ci` | none | — |
| `codeup_ci` | none (no `ocr config set` at all; LLM settings come from `OCR_LLM_*`) | — |
| `gerrit_ci` | none | — |

`OCR_LANGUAGE` is not a product-level variable. The binary resolves `OCR_LLM_*`,
`OCR_CONFIG_PATH`, `OCR_ENABLE_TELEMETRY`, `OCR_VIEWER_ALLOWED_HOSTS` and others
from the environment (`internal/llm/resolver.go:51-66` and friends), but there
is no `OCR_LANGUAGE` there. It exists only as per-example shell glue.

**Two statements in the repo contradict each other.**

`action.yml:56-59` documents the input as:

> Review output language, written via `ocr config set language`
> (e.g. English, Chinese). No env var exists for this.

while `examples/gitlab_ci/README.md:44` presents `OCR_LANGUAGE` as a
first-class variable:

> | `OCR_LANGUAGE` | No | No | Review output language, written via `ocr config set language` (e.g. `English`, `Chinese`). Defaults to OCR's built-in default. |

Each is locally true, but read together they are contradictory — and a reader
who trusts the gitlab_ci wording will set `OCR_LANGUAGE` in a Bitbucket, Gerrit,
GitFlic or CODING pipeline and get English output with no diagnostic.

Two smaller gaps in the same area:

- `examples/github_actions/README.md` never mentions the `language` input at all.
- `examples/README.md` (13 lines) lists five examples but omits `codeup_ci/`,
  even though the directory exists.

**Reproduction:**

```bash
grep -rn "OCR_LANGUAGE" examples/                        # only examples/gitlab_ci/
grep -rn '"OCR_[A-Z_]*"' internal/ cmd/ | grep -v _test  # no OCR_LANGUAGE
grep -c '`language`' examples/github_actions/README.md   # 0
```

### Suggested Change

Three possible directions — happy to work on whichever you prefer:

**A. Resolve `OCR_LANGUAGE` in the binary**, next to the existing `OCR_LLM_*`
variables. Every example would then work with no per-example glue, and
`gitlab_ci`'s `if [ -n "$OCR_LANGUAGE" ]; then ...; fi` line could be dropped.
The open question is whether this conflicts with the intent behind
`action.yml`'s "No env var exists for this" — i.e. whether keeping the language
setting in the config file rather than the environment is deliberate.

**B. Add the `ocr config set language` step to the four examples that lack it**,
leaving the binary unchanged.

**C. Documentation only** — state in `configuration.md` and in each example's
README which mechanism applies there, and note explicitly that the remaining
examples have no language knob.

A and B are small changes; C is markdown-only.

katepallewarprathmesh-sketch:
## Summary

Closes #1266 with a backward-compatible A+B+C solution

- Adds product-level `OCR_LANGUAGE` resolution through the existing `language` configuration path.
- Preserves explicit config precedence, existing GitHub Actions `language` input, and GitLab compatibility behavior.
- Documents and wires `OCR_LANGUAGE` across the CI examples, including CodeUp.
- Updates the examples index and localized configuration documentation.
- Adds focused coverage for config precedence and empty/unset environment behavior.

## Validation

- `make test`
- `make check`
- `make build`
- `git diff --check`

## AI disclosure

This change was developed with AI assistance using GitHub Copilot in VS Code. The implementation was reviewed and validated.


wu21-web:
@katepallewarprathmesh-sketch Please link this https://github.com/alibaba/open-code-review/issues/1266 for development.

katepallewarprathmesh-sketch:
> @katepallewarprathmesh-sketch Please link this #1266 for development.

Done

action.yml
```diff
@@ -56,7 +56,8 @@ inputs:
   language:
     description: >-
       Review output language, written via `ocr config set language`
-      (e.g. English, Chinese). No env var exists for this.
+      (e.g. English, Chinese). The GitHub Actions input is `language`; the OCR
+      CLI also supports the `OCR_LANGUAGE` environment variable.
```
wu21-web: Why talk about CLI usage in an action? I know you mean it can be set via commands in other platforms, but we should only talk about GitHub Actions here.
```suggestion
```

examples/github_actions/README.md
```diff
@@ -24,6 +24,19 @@ The core of the demo is a single action step:
 
 See [`action.yml`](../../action.yml) for the full list of inputs, outputs, security guidance, and the four comment-posting modes (sticky summary + incremental).
 
+### Configure the review language
+
+Use the Action-specific `language` input to choose the review output language:
+
+```yaml
+- uses: alibaba/open-code-review@main
```
wu21-web, line 32: Be aware that an action SHA might be enforced later on.

pages/src/content/docs/en/configuration.md
```diff
@@ -391,8 +391,13 @@ it defaults to English when unset:
 ```bash
 ocr config set language 中文
 ocr config set language English
+OCR_LANGUAGE=Chinese ocr review
 ```
 
+Both commands configure the same underlying `language` setting. For GitHub
```
wu21-web: This is conflicting with the other translations. Chinese "方式" should match "ways". And the original said 2 commands, but they are not commands and even if they are there are more than 2. 

```suggestion
Both ways configure the same underlying `language` setting. For GitHub
```

wu21-web (commented):
Looks good to me. Can you also format the markdown tables? That is optional.

examples/github_actions/README.md
```diff
@@ -14,7 +14,7 @@ cp ocr-review.yml .github/workflows/ocr-review.yml
 The core of the demo is a single action step:
 
 ```yaml
-- uses: alibaba/open-code-review@main
+- uses: alibaba/open-code-review@<full-commit-sha> # vX.Y.Z
```
wu21-web: I am not asking you to change this, it is just a reminder. This is a job of another open PR. You should revert.

wu21-web: Can you format using npx prettier?

katepallewarprathmesh-sketch: @wu21-web I have updated this now

lizhengfeng101:
Please route all three `ApplyLanguage` call sites through `configuredLanguage` so `OCR_LANGUAGE` behaves consistently. Right now only `loadLLMRuntime` (`shared.go:259`) uses the new helper, while `scan_cmd.go:188` and `llm_cmd.go:75` still read `appCfg.Language` directly — so `ocr scan` silently ignores `OCR_LANGUAGE`, which contradicts the docs stating both configure the same `language` setting.

Also add a test covering the `ocr scan` path with `OCR_LANGUAGE` set (and no config file); the missing coverage there is why this gap wasn't caught.

PR #1299, opened 2026-09-16
ci: autofix provider presets in the existing test job

### LINKED ISSUE #1211: refactor(llm): make Go provider registry the single source of truth for the VS Code preset

NanaseInori:
## Description

Automatically regenerate and commit a stale VS Code provider catalog when an eligible pull request changes its inputs. The implementation reuses the existing Linux CI test job and leaves normal validation on the simulated merge result.

**Follow-up to #1298 (the restoration of #1221): merge #1298 first.** That prerequisite supplies the generator, checked-in catalog, regression tests and read-only regeneration check. This PR adds the optional automatic commit behavior as a separate change, keeping the provider synchronization fix within its reviewed scope.

This restores and completes #1253 on top of `71ed3a288df6435f240b7194f19041b4fcc4ab7d`, with the latest review feedback incorporated into one commit. Thanks to @wu21-web for the original suggestion, proposed patch and subsequent review.

### Behavior and implementation

The existing `test` job in `.github/workflows/ci.yml` now contains separate, named steps to detect relevant changes, prepare a temporary worktree, generate, validate, commit, push and clean up. There is no additional job, workflow or runtime helper. The existing Go container and checkout action are reused.

Only same-repository pull requests are eligible; Dependabot is excluded. The path check compares the simulated merge commit with its base parent, so an unrelated PR does not run autofix just because `main` changed provider data. Relevant paths are:

- `internal/llm/providers.go` and `internal/llm/protocol.go`;
- `internal/llm/gen/`;
- `go.mod` and `go.sum`;
- `extensions/vscode/src/shared/providers.generated.ts`.

Checkout retains the normal merge ref, fetches both parents with `fetch-depth: 2`, and disables persisted credentials. After verifying that the merge commit's second parent matches the event PR head, generation takes place in a temporary Git worktree at that exact head. This prevents the bot from committing base-only changes to the contributor's branch. The original merge checkout remains unchanged, and an `always()` cleanup step removes the temporary worktree.

Generation removes the previous catalog before running `go generate ./internal/llm`. It requires a newly produced regular file with ordinary permissions, rejects unrelated tracked changes and non-ignored untracked files, and stages only the catalog. This handles missing or ignored catalogs and rejects successful no-op generation when the directive is absent.

The workflow's Bash steps execute directly from CI. Eligible PR branches do not need to contain a separate autofix script. Executable regression tests live in `scripts/github-actions/autofix-provider-presets.test.js`, are included in the existing `npm run test:github-actions` command, and exercise the actual workflow shell blocks. Changes to `ci.yml` now trigger that contract workflow.

### Changes requested during the previous review

| Review point | Result in this revision |
| --- | --- |
| Reuse the existing CI job and avoid extra runs | Uses the existing Linux `test` job and skips generation for unrelated PR changes. |
| Make the Bash flow easier to review | Separates path detection, worktree preparation, generation, validation, commit, push and cleanup into named steps. |
| Follow the repository's checkout convention | Uses the existing `actions/checkout@v7`; no separate pinned checkout step is added. |
| Clarify where a token is needed | Local commit creation requires no token. Only the push step explicitly receives `GITHUB_TOKEN` in its environment. |
| Use the normal bot identity and clarify CLA handling | Commits use `github-actions[bot]`; repository CLA policy still applies and no exemption is assumed. |
| Account for fork PRs | Forks skip autofix because the upstream repository token cannot write to contributor repositories; the read-only check and manual instructions remain. |

### Permissions and branch updates

GitHub scopes permissions to jobs rather than individual steps. Reusing the test job therefore changes that job to **`contents: write`**. The workflow default and other jobs remain read-only, and GitHub's normal fork-token restrictions still apply. Keeping the token out of the generation and commit step environments reduces incidental exposure; it is not a security sandbox between steps in the same job. Eligible same-repository generator code must be trusted: it shares the container and Git metadata with the later push step. This permission change is an explicit part of this follow-up for maintainers to assess.

The push uses the repository's standard token, not a PAT or additional secret. The encoded authentication header is masked, Git HTTP tracing is disabled, and the credential is supplied through the push subprocess environment.

| Situation | Result |
| --- | --- |
| Fork, Dependabot or unrelated PR | Skip autofix. |
| Generated catalog already matches | No commit or push. |
| Eligible PR has a stale, deleted or ignored catalog | Commit only the freshly generated catalog on the event PR head. |
| Generation fails, produces no regular catalog, or changes other files | Fail before committing and clean up the worktree. |
| PR branch changes before commit creation | Skip the outdated run. |
| Branch advances, is deleted or is rewound during push | Reject the outdated update using an explicit expected-SHA lease. |
| Push fails or its acceptance cannot be confirmed | Stop without retrying and provide instructions to inspect the current head before regenerating manually. |

The bot commit must be a direct child of the expected PR head. Combined with `--force-with-lease=<ref>:<expected-head>`, an accepted update is a fast-forward without overwriting a concurrently changed branch. The workflow does not pull, rebase or retry an old generated result.

### Validation after an automatic commit

The bot's new commit needs checks on its own SHA. The original run continues to describe the earlier merge result, and an automatic push does not guarantee that the rest of that run will pass.

For GitHub.com, a `pull_request` synchronization triggered by `GITHUB_TOKEN` creates workflow runs that require approval by someone with write access. See [GitHub's token documentation](https://docs.github.com/en/actions/concepts/security/github_token#when-github_token-triggers-workflow-runs). Branch protection, workflow approval and CLA requirements remain in effect.

## Type of Change

- [x] CI / Build / Tooling
- [x] Tests and documentation
- [ ] Breaking change

## How Has This Been Tested?

Local environment: Windows amd64, Go 1.26.5, Git Bash, Node 24.14.0 and Node 14.21.3.

- [x] All **29** workflow regression tests passed on Node 24 and Node 14. Tests execute the actual Bash steps using shallow merge checkouts, a generator fixture and local bare Git remotes.
- [x] Integration with #1298's actual Go generator at `eac569f284e172d63f39e3545f67cca9824a6372` passed. Starting from an intentionally stale catalog, the workflow generated, validated, committed and pushed to a local bare remote. The fix changed only the catalog, was a direct child of the expected head, and restored the expected bytes exactly. The original merge checkout remained unchanged and the temporary worktree was removed.
- [x] `make license-add` and `make check` passed.
- [x] `actionlint` passed for both modified workflows, with the optional shellcheck/pyflakes integrations disabled.
- [x] Full `make test` passed, including race detection.
- [x] OCR completed its review of all three selected production/configuration files and reported five findings. This revision adds explicit diagnostics for an unavailable remote head or mismatched merge parent, and cleans up failed worktree preparation without publishing an invalid output path; the new failure paths have executable tests. The shared-job write-permission concern is retained as the documented trust tradeoff above. The suggestion that no PR checks can be triggered by `GITHUB_TOKEN` was checked against the current GitHub documentation: `pull_request:synchronize` is an approval-required exception. Codex independently reviewed the final changes and the regression tests.
- [x] LF line endings, `git add --renormalize .`, and `git diff --cached --check` passed for all five changed files.

The full Linux Actions contract suite and a live GitHub bot push/approval/check cycle were not run locally. The PR is submitted from a fork, so its own autofix steps are intentionally skipped. The local integration establishes generator and Git-update behavior, not hosted-runner permissions or repository-policy behavior. No Go runtime implementation is changed in this follow-up.

## Contributors and AI/LLM Disclosure

@wu21-web (Tao Xin) proposed the automatic commit behavior and provided the patch and review feedback that shaped this integration. His contribution is credited with a Git `Co-authored-by` trailer.

A second `Co-authored-by` trailer preserves the contribution under @Qiyuanqiii, my previous account. The earlier submission was under @PrayWithYou; this restoration is under @NanaseInori.

AI/LLM assistance was used for implementation, tests, review and documentation: Codex (GPT-6 Astra) and the repository's OCR CLI (`deepseek-v4-flash`). Earlier iterations also used ChatGPT (GPT-6 Astra). Validation results and remaining limitations are listed above.

## Related Issues

Depends on #1298, which addresses #1211. Restores and completes #1253. This follow-up does not independently close #1211.


lizhengfeng101:
Thanks for the thorough work on this — the implementation and test coverage are genuinely careful.

That said, I'd like to close this in favor of keeping the read-only check from #1298 only, without the automatic commit/push.

Reasoning:

- **It breaks the repo's established CI convention.** The existing checks (`gofmt`, line-ending renormalize, `go mod tidy`) all *detect and instruct the contributor to run a command* — none of them write back to the branch. #1298's read-only regeneration check fits that pattern; auto-committing to the PR branch is a new and more invasive behavior.
- **The cost outweighs the benefit.** To push, the shared `test` job has to be elevated to `contents: write` while running on a self-hosted runner and executing PR-authored generator code (`internal/llm/gen/` is in the trigger set). That's a real trust/permission expansion, and the only thing it saves a contributor is running `go generate ./internal/llm` locally.
- **Operational noise.** The autofix push triggers a `pull_request: synchronize` that, combined with `cancel-in-progress: true`, cancels the current run and leaves an approval-gated run on the new head.

A failing CI check with a clear "run `go generate ./internal/llm` and commit" message is sufficient here. Closing this; #1298 remains the path forward.

Appreciate the effort, and thanks for the clear AI/LLM disclosure.

NanaseInori:
ok

PR #1300, opened 2026-09-16
test(diff): cover gitignore-drop interleaved with provider-exclude

### LINKED ISSUE #1297: test(diff): cover gitignore-drop interleaved with provider-exclude in ForEachInOrder
## Description

`DiffSet.ForEachInOrder` (internal/diff/git.go) walks `Included` and `Excluded`
in the original changeset order using the arithmetic `upTo := excludedAt[i] - i`,
where `excludedAt[i]` is the exclusion's index in the *mixed stream* of included
and provider-excluded files.

That mixed-stream index deliberately does **not** count files dropped by
`.gitignore`, because `partitionDiffs` adds a gitignore-dropped file to neither
`Included` nor `Excluded`. The ordering is therefore correct only as long as a
gitignore-dropped file contributes to neither counter — an invariant the code
relies on but that no test exercises directly.

The existing test `TestGetDiffSetWalksChangesetOrder` covers an
included → provider-excluded → included interleaving (`a.go`, `target/mid.go`,
`z.go`), but never mixes in a gitignore-ignored file. If a future change to
`partitionDiffs` ever miscounted dropped files, the ordering would silently skew
and no test would catch it.

This is a test-completeness gap, not a known defect: the current logic is
correct. We want a regression test that pins the invariant.

## Scope

- File(s): `internal/diff/git_test.go`
- Function/area: `DiffSet.ForEachInOrder` / `partitionDiffs` in `internal/diff/git.go`
- Add a test case where the changeset interleaves all three kinds of file:
  included, provider-excluded (e.g. under `target/`), and gitignore-ignored,
  arranged so the gitignore-dropped file sits *between* an included and an
  excluded file. Assert that `ForEachInOrder` visits only the included and
  provider-excluded files, in the original changeset order, with correct
  `providerExcluded` flags, and never visits the gitignore-dropped file.

## Acceptance Criteria

- [ ] A new test in `internal/diff/git_test.go` covers the gitignore-drop +
      provider-exclude interleaving described above
- [ ] The test fails if `excludedAt` accounting is perturbed to count dropped
      files (verify by temporarily breaking it locally)

## Context

Surfaced while reviewing the ordering logic introduced in #1242. The arithmetic
was verified correct by hand across the empty / all-excluded / interleaved /
gitignore-mixed cases; this issue only fills the missing regression coverage for
the last case.

jimpablo:
## Description

Adds the missing `ForEachInOrder` regression coverage from #1297: a tracked `.gitignore`-dropped file sitting between an included file and a provider-directory exclusion must not shift changeset order or be visited.

This is test-only. `partitionDiffs` / `ForEachInOrder` production logic is unchanged.

The gitignore-dropped file is committed before `.gitignore` is added so `git diff HEAD` still emits it and `partitionDiffs` is the layer that drops it — the case `excludedAt` must not count.

## Type of Change

- [ ] Breaking change (fix or feature that would change existing functionality)
- [ ] Documentation update
- [x] CI / Build / Tooling

## How Has This Been Tested?

- [x] `make check` passes locally

The new test fails if `excludedAt` is perturbed to count gitignore-dropped files (verified locally, not committed).

## Checklist

- [x] I have signed the CLA

AI/LLM disclosure: Grok 4.6 (Grok Build / xAI) was used to implement the test. I reviewed the diff. Commits carry no AI attribution.

## Related Issues

Closes #1297


lizhengfeng101 (approved):
LGTM

PR #1301, opened 2026-09-16
fix(diff): stop reading gitlink entries as file content

LiewSDK:
## Problem

A submodule change appears in a diff as a gitlink (mode `160000`) whose body is a `Subproject commit <sha>` line. `finalizeDiff` treats that path like a regular file and tries to read it:

- **workspace mode** — stats a directory:
  `[ocr] WARNING: cannot read file sub for review: file path "sub" is a directory`
- **at a ref** (`--commit`, `--from/--to`) — `git show <ref>:<gitlink>` exits 128:
  `[ocr] WARNING: cannot read file <path> at ref <sha>: exit status 128`

On their own these are noise. The practical problem is that they fire on **every** review of a commit or branch that bumps a submodule, and callers that wrap `ocr` and treat stderr as failure cannot tell them apart from genuine read errors. On a repository with many submodules that makes branch review unusable for such tooling — which is how I hit this (a repo with 55 submodules, behind a wrapper that is fail-closed on stderr).

## Reproduce

```bash
git init main && cd main && echo x > a.py && git add -A && git commit -m init
git submodule add ../sub sub && git commit -m "add submodule"
# advance the submodule so the gitlink is modified
git -C sub checkout <newer-sha>

ocr delegate preview
```

Before:
```
# Files (1 reviewable / 1 total)
[ocr] WARNING: cannot read file sub for review: file path "sub" is a directory
```

After:
```
# Files (0 reviewable / 1 total)
```
(stderr empty)

Note the `1 reviewable` before — the gitlink was being counted as a reviewable file in workspace mode.

## Change

Mark gitlink entries as submodules while parsing (both the `Subproject commit` body lines and the workspace directory case), flag them non-reviewable, and return from `finalizeDiff` before either read path.

Adds `Diff.IsSubmodule`. Kept as a separate field rather than reusing `IsBinary` alone so the reason a path is unreviewable stays visible to consumers; both are set for compatibility with existing filtering.

## Compatibility

Gitlinks were already excluded from review at a ref, so file selection and counts are unchanged there. Only the spurious warnings and the pointless `git show`/`os.Stat` calls go away. Workspace mode changes a gitlink from wrongly counted as reviewable to correctly excluded.

## Tests

- `TestParseDiffText_MarksGitlinkAsSubmodule` — parse sets the flags
- `TestParseDiffText_GitlinkAtRefEmitsNoWarning` — captures `os.Stderr` and asserts nothing is written for a gitlink at a ref

Both were written before the fix and observed failing. Verified locally against `main` (`a694be5`):

- `gofmt -s -l` clean, `go vet` clean
- `scripts/verify-license.sh` — all headers valid
- `go run scripts/verify-english-only.go` — clean
- `go test -race -count=1 ./internal/diff/... ./internal/model/...` — pass

Happy to adjust naming or split the parse/read halves if you'd prefer them separate.


LiewSDK:
Closing this for now — I'm not in a position to complete the CLA on this side, so rather than leave an unmergeable PR sitting in your queue I'll take it off the board.

The analysis stands if it's useful to anyone: gitlink entries (mode 160000) are read as file content in `finalizeDiff`, producing `cannot read file ... is a directory` in workspace mode and `... exit status 128` at a ref. Repro and a two-test fix are in the diff above, and the branch remains at LiewSDK/open-code-review:fix/gitlink-submodule-read if a maintainer wants to pick it up or if I can revisit the CLA later.

Thanks for the tool — the deterministic pre-LLM filtering is genuinely well built.

PR #934, opened 2026-08-15
fix(diff,tool): fix CRLF parsing and cross-platform path traversal handling

MeiSiristhebest:
## Description

When unified diff text contains Windows-style CRLF (`\r\n`) line endings (e.g. from Windows environments, CI runners, or repositories with `core.autocrlf=true`), `ParseDiffText` and `ParseHunks` split lines on `\n` without stripping trailing `\r`.

This caused several severe issues:
1. `d.NewPath` retained a trailing `\r` (e.g. `"foo.go\r"`), causing subsequent file reading in `finalizeDiff` to fail with `stat file ...: The filename, directory name, or volume label syntax is incorrect` / `file not found`, leaving `d.NewFileContent` completely empty during review.
2. Strict equality checks for `--- /dev/null` and `+++ /dev/null` failed, resulting in lost `IsNew` and `IsDeleted` metadata.
3. Hunk line content retained trailing `\r`.

Additionally, in `internal/tool/code_search.go`, `hasTraversalPathComponent` only checked `/` separators, allowing Windows-style `..\` path traversal patterns to bypass the check.

### Changes Made:
- In `internal/diff/parser.go` and `internal/diff/hunk.go`: strip trailing `\r` when iterating lines.
- In `internal/tool/code_search.go`: normalize path separators using `filepath.ToSlash` in `hasTraversalPathComponent`.
- Added regression unit tests covering CRLF diff parsing, CRLF hunk parsing, and backslash path traversal in code search.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring (no functional changes)
- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] `make test` passes locally
- [x] Added `TestParseDiffText_CRLFLineEndings` in `internal/diff/parser_test.go`
- [x] Added `TestParseHunks_CRLFLineEndings` in `internal/diff/hunk_test.go`
- [x] Added `TestCodeSearch_RejectsBackslashPathTraversal` in `internal/tool/code_search_test.go`

## Checklist

- [x] My code follows the project's coding style (`go fmt`, `go vet`)
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix is effective or my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have signed the CLA

## Related Issues

Closes #933

PR #1302, opened 2026-09-16
fix(diff): strip trailing CR from CRLF diff text

### LINKED ISSUE #933: bug(diff): CRLF line endings cause corrupted NewPath and missing IsNew/IsDeleted metadata
### OpenCodeReview Version
main (latest)

### Operating System
Windows / Cross-platform with CRLF

### Installation Method
go build / npm

### LLM Provider
Any

### Bug Description
When unified diff text contains Windows-style CRLF (`\r\n`) line endings (common with `core.autocrlf=true` or repositories configured with autocrlf / CRLF checkout), `ParseDiffText` and `ParseHunks` in `internal/diff` split lines on `\n` without stripping trailing `\r`.

This causes several severe regressions:
1. `diffHeaderRe` captures `b/(.+)`, polluting `d.NewPath` with a trailing `\r` (e.g. `"file.go\r"`).
2. Subsequent calls to read `d.NewPath` in `finalizeDiff` fail with `stat file "file.go\r": CreateFile ... The filename, directory name, or volume label syntax is incorrect` / `file not found`, leaving `d.NewFileContent` completely empty during review.
3. Metadata checks `line == "--- /dev/null"` and `line == "+++ /dev/null"` fail because the string contains `\r`, losing `IsNew` and `IsDeleted` flags.
4. `ParseHunks` leaves trailing `\r` in `HunkLine.Content`.

### Steps to Reproduce
Pass a unified diff containing CRLF line endings to `ParseDiffText` (or run `ocr review` in an environment where diffs output CRLF):

```go
diffText := "diff --git a/fresh.go b/fresh.go\r\n" +
    "new file mode 100644\r\n" +
    "--- /dev/null\r\n" +
    "+++ b/fresh.go\r\n" +
    "@@ -0,0 +1,1 @@\r\n" +
    "+line1\r\n"

diffs, err := ParseDiffText(context.Background(), diffText, dir, "", nil)
// Output:
// [ocr] WARNING: cannot read file fresh.go for review: stat file "fresh.go\r": CreateFile ...: The filename, directory name, or volume label syntax is incorrect.
// d.NewPath == "fresh.go\r" (corrupted)
// d.IsNew == false (lost)
```

### Expected Behavior
Trailing `\r` should be stripped when splitting diff lines, so file paths are clean and metadata markers match accurately.

### Additional Context
Additionally, in `internal/tool/code_search.go`, `hasTraversalPathComponent` only checks `/` separators and misses Windows backslashes (`..\secret`), which can be addressed together.

SurefireStudios:
## Description

`ParseDiffText` and `ParseHunks` split diff text on `"\n"` and keep the carriage return that CRLF-terminated text leaves on every line.

Git writes its own structural lines with `\n`, but the diff text can arrive already converted — a `.patch` checked out under `core.autocrlf=true`, or output captured through a Windows shell. The stray `\r` is not cosmetic:

- It rides into the `^diff --git a/(.+?) b/(.+)$` capture, so `NewPath` becomes `"file.go\r"`. `finalizeDiff` then cannot open the file and logs `cannot read file ...: The filename, directory name, or volume label syntax is incorrect`, leaving `NewFileContent` empty — the review runs with no file content.
- It defeats the equality checks `line == "--- /dev/null"` and `line == "+++ /dev/null"`, so the diff silently loses `IsNew` / `IsDeleted`.
- `ParseHunks` leaves it on `HunkLine.Content`.

The fix is one helper, `splitDiffLines`, used by both parsers. It strips a single trailing `\r` per line, which is how the rest of the package already treats one: `resolver.go:255` trims it before matching lines, and `filereader.go` and `viewer/server.go` do the same when reading file content. `TestResolveLineNumbers_FallbackToFileContent_CRLF` already encodes that position for file content; this extends it to the diff text itself.

### Two corrections to the issue

**The path-traversal half is already fixed.** #933 closes by saying `hasTraversalPathComponent` in `internal/tool/code_search.go` "only checks `/` separators and misses Windows backslashes (`..\secret`)". On current `main` it opens with `strings.ReplaceAll(pathspec, "\\", "/")`. I tested `..\secret`, `a\..\b` and `..\..\etc` against it — all three are detected, and `a\b`, `..foo`, `foo..` are not false positives. Nothing to do there, so this PR leaves that file alone.

**`IsNew` is not always lost.** The issue's own reproducer includes a `new file mode 100644` line, and that header is matched with `HasPrefix`, so a trailing `\r` does not affect it — `IsNew` survives in that exact case. The flags are only lost when the `/dev/null` marker is the sole signal, because that comparison is `==`. Both shapes are covered by the tests.

## Type of Change

## How Has This Been Tested?

Windows, Go 1.25.5, `CGO_ENABLED=0`.

Reproduced first, on `main` before changing anything — the failing assertions matched the issue exactly, including the warning text:

```
NewPath="fresh.go\r" OldPath="fresh.go" IsNew=true IsDeleted=false
[ocr] WARNING: cannot read file fresh.go for review: stat file "fresh.go\r": ...
```

and with the `/dev/null` marker as the only signal:

```
add -> IsNew=false IsDeleted=false NewPath="a.go\r"
del -> IsNew=false IsDeleted=false NewPath="b.go\r"
```

After the fix, `NewPath="fresh.go"`, `IsNew=true`, `IsDeleted=true`, no CR in hunk content.

Verified the four new tests are load-bearing by reverting the strip to a no-op (`lines[i] = line`): all four fail, while `TestSplitDiffLines_LeavesLFTextAlone` still passes, since LF behaviour is unchanged.

- Full suite (`go list ./... | grep -v /extensions/`): **23 packages ok, 0 failures**. `-race` needs cgo, which this machine lacks, so that pass is left to CI.
- `internal/diff` coverage **94.0%**, above the 90% threshold; `splitDiffLines` and `ParseDiffText` both 100%.
- `scripts/verify-license.sh`, `scripts/verify-english-only.go` (488 files), `gofmt -s -l`, `go vet` — all clean.
- `git ls-files --eol` reports `i/lf w/lf` for all three files.

One test expectation I got wrong and corrected rather than papered over: `ParseHunks` yields a trailing empty context line from the final newline. That is existing behaviour, identical for LF input and unrelated to CR, so the test asserts it rather than hiding it.

## Checklist

- [ ] I have updated the documentation accordingly (if applicable) — no user-facing behaviour or flag changed
- [ ] I have signed the CLA — pending; will sign when the bot prompts

**AI/LLM disclosure, per `AGENTS.md` rule 1:** this change was written with Claude Code (model: claude-opus-5). Per rule 6 the commit carries no `Assisted-by`/`Co-developed-by` trailer.

## Related Issues

Fixes #933


wu21-web:
This is a duplicate of #934

SurefireStudios:
@wu21-web is right — thanks for catching it. #934 covers the same bug and was opened by @MeiSiristhebest, who also reported #933, on the same day. Mine is the duplicate and the later one; I'm happy to close this in their favour.

For transparency about how I missed it: I did check the open-PR queue for anything referencing #933, but I truncated PR bodies when scanning. `#933` appears at character 2274 of #934's 2280-character body, so the reference fell outside what I looked at. My check, my mistake.

One finding from this that may be useful to #934 either way, since it is currently conflicting:

Its `internal/tool/code_search.go` hunk changes `hasTraversalPathComponent` to normalise backslashes before splitting. That change is **already on `main`** — the function now opens with `norm := strings.ReplaceAll(pathspec, "\\", "/")`. I tested the current implementation against `..\secret`, `a\..\b` and `..\..\etc` (all detected) and `a\b`, `..foo`, `foo..` (no false positives). So that half of #934 has landed by another route, which is very likely the source of the conflict — dropping that hunk on a rebase should clear it, leaving only the CRLF fix, which is still unfixed on `main` today.

Two small details I verified while reproducing, in case they are worth a test in #934:

- `IsNew` is not lost in the reproducer from #933 as written, because it also contains a `new file mode 100644` line and that header is matched with `HasPrefix`, which a trailing `\r` does not affect. The flags are only lost when a `/dev/null` marker is the sole signal, since that comparison is `==`.
- `ParseHunks` emits a trailing empty context line from the final newline. That is pre-existing and identical for LF input, not a CR artefact.

Maintainers: say the word and I'll close this, or rebase it if #934 has stalled. No preference beyond not adding review noise.

**AI disclosure, per AGENTS.md rule 1:** this comment and the PR were written with Claude Code (model: claude-opus-5). The verification above is from commands actually run against a checkout of current `main`.

lizhengfeng101 (approved):
LGTM

lizhengfeng101:
Thanks for the thorough fix and writeup — we're going to merge this.

Heads-up: #934 by @MeiSiristhebest independently fixed the same issue and diagnosed #933 accurately. Since they identified the bug first, we'd like to credit them as a co-author on this merge. Are you OK with adding a `Co-authored-by:` trailer for them?

One process item: the CLA still shows as pending on your side — please sign when the bot prompts so we can merge.

SurefireStudios:
Yes, absolutely — and thank you for doing it. @MeiSiristhebest reported #933 and fixed it in #934 before I opened this, so the credit is theirs by right; I'd have been uncomfortable with the merge landing without it. I see the trailer is already on the merge commit, so nothing further needed from me there.

On the CLA: I can't find anything to action from my side. There's no CLA check among the 12 on this PR and no CLA bot has commented on it, so there's nothing prompting me. If it's tracked somewhere outside the PR — an internal system, or an email I should be watching for — point me at it and I'll sign straight away.

MeiSiristhebest:
Thank you @lizhengfeng101 and @SurefireStudios! Glad to see this landed and appreciate the co-author credit!

PR #1248, opened 2026-09-14
fix(agent): enforce --max-tokens-budget inside a running group

mgoulart:
## Description

`--max-tokens-budget` was only checked in `dispatchSubtasks` before starting the *next* group, so with the default `--concurrency 8` any review that splits into 8 or fewer groups never hit it. In a real run a 2M budget finished at 12.15M tokens with all groups marked `completed` (details in #1247).

This PR enforces the budget inside a running group:

- `internal/llmloop/loop.go`: `Deps.MaxTokensBudget`; `RunMainTask` checks aggregate usage before every round and stops with a new `StopTokenBudget` reason. Like `StopMaxRounds`, the group gets one grace round to submit what it already has; the grace prompt now says the budget is exhausted rather than assuming rounds ran out.
- `internal/agent/agent.go`: passes the budget into the runner, classifies `StopTokenBudget` as `FailureBudget` (so the group is reported `failed(budget)` like skipped groups), records a `token_budget_reached` warning, and sets `budgetExceeded` so the dispatch loop also stops starting new groups.
- `internal/scan/agent.go`, `shared_flags.go`: plumbing and flag help text.
- Docs (`cli-reference.md`, both SKILL.md copies, `action.yml`, GitHub Actions README): describe the budget as checked before every round.

`--max-tools` being advisory only is a separate gap; #1118 addresses that, and this change is independent of it.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring (no functional changes)
- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] `make test` passes locally (`go test -race -count=1 ./...`, plus `make check`)
- [x] Manual testing (describe below)

New tests:
- `TestRunMainTask_TokenBudgetStopsBeforeNextRound`: a runner over budget stops after the current round, runs exactly one grace round, and reports `StopTokenBudget`.
- `TestRunMainTask_ZeroTokenBudgetNeverStops`: budget 0 keeps the old unlimited behaviour.
- `TestDispatchSubtasks_TokenBudgetStopsInFlightGroup`: two groups under budget; the first completes, the second is stopped in flight, partial manifest is written with 1 completed / 1 `failed(budget)`.
- `TestMainLoopStopUnknownValue` updated for the new enum value.

[...]

Issue #1247, opened 2026-09-14
--max-tokens-budget is never enforced when groups <= --concurrency (only checked at dispatch)


mgoulart:
### OpenCodeReview Version
1.12.1 (reproduced on `main` at 6e5cd8c)

### Bug Description
`--max-tokens-budget` is only checked in `dispatchSubtasks`, right before the *next* group starts. Nothing inside a running group looks at it. With the default `--concurrency 8`, a diff that splits into 8 or fewer groups starts every group immediately, so the budget is never applied.

Measured on a 4-file diff with `--max-tokens-budget 2000000` (Claude Opus via an OpenAI-compatible gateway, effort medium):

- 4 groups dispatched at once, gate never fired
- finished normally after 14m54s at 12.15M tokens (~6x the budget), 246 tool calls, one group ran 94 rounds
- manifest shows all groups `completed`, no budget warning

`--max-tools` did not bound it either since it only sets `MAX_TOOL_REQUEST_TIMES` in the prompt; that is #935 / #1118 and not addressed here. #409 / #508 designed the budget as a dispatch-time gate, but the docs describe it as a cap on total usage, and for most PR-sized reviews it does nothing.

### Steps to Reproduce
1. Take a diff that splits into fewer groups than `--concurrency`.
2. `ocr review --max-tokens-budget 200000 ...` with a tool-happy model.
3. Compare `total_tokens` in the output to the budget.

### Expected Behavior
Once aggregate usage crosses the budget, running groups stop looping (one final round to submit what they have), are reported `failed(budget)`, and partial results still publish.

### Logs / Error Output
```
duration: 14m54s   total_tokens: 12.15M (cache_read 11.55M, output 99k)
groups: 4 completed, 0 failed(budget)   tool_calls: 246
```

### AI/LLM Disclosure
Claude (claude-fable-5-1) helped investigate and draft this; the measurements are ours and I reviewed the text.

### Additional Context
Fix in #1248.

PR #1303, opened 2026-09-16
fix(agent): enforce token budget during active loops

### LINKED ISSUE #1295: token budget not enforced
### OpenCodeReview Version

open-code-review v1.12.1 (5589990) linux/amd64 built at: 2026-09-15T21:56:40+08:00

### Operating System

Linux (x86_64)

### Installation Method

Built from source

### LLM Provider

### Bug Description

Supplementary write-up for #1247. --max-tokens-budget is only checked
inside dispatchSubtasks, right before starting the *next* group
(agent.go ~685-698), and that's the only place the budgetExceeded flag
gets set:

    if a.args.MaxTokensBudget > 0 {
        ...
        if projected > a.args.MaxTokensBudget {
            ...
            a.budgetExceeded.Store(true)
        }
    }

The only place that flag is read is inside a group's own round loop
(agent.go:1449):

    if round > 1 && a.args.MaxTokensBudget > 0 && a.budgetExceeded.Load() {
        ...
        break
    }

With the default --concurrency 8, a diff producing 8 or fewer groups
dispatches all of them in the first pass, so there's never a "next
group" to trigger the check, budgetExceeded never flips true, and no
other code path polls real aggregate usage mid-run.

### Steps to Reproduce

1. Read internal/agent/agent.go ~685-698 (the only site that sets
   budgetExceeded) and :1449 (the only site that reads it).
2. Confirm there is no other call site for budgetExceeded.Store(true) and
   no per-round live-usage check outside this flag — grep for
   "MaxTokensBudget" and "budgetExceeded" across internal/agent/*.go.
3. Confirm main has no StopTokenBudget symbol (the fix PR #1248
   introduces this; it isn't merged).
4. Cross-reference against #1247's live repro: 4-file diff,
   --max-tokens-budget 2000000, default --concurrency 8 → finished at
   12.15M tokens (~6x budget), all groups reported completed, no budget
   warning.

### Expected Behavior

Aggregate token usage should be checked before every round inside a
running group, not only when a next group is about to dispatch, so runs
that never dispatch a "next" group still respect --max-tokens-budget.

### Logs / Error Output

```shell
go build ./... / go vet ./... / go test ./... all clean at commit
5589990.
```

### AI/LLM Disclosure

### Additional Context

Supplements #1247 (PR #1248 open, not yet merged, adds the missing
per-round check).
AI disclosure: this report was investigated and drafted with AI
assistance (Claude), verified against the source at commit 5589990 via
direct code reading plus go build/vet/test — not via a live LLM-driven
review run.

iluv7:
## Description

`--max-tokens-budget` was only checked in `dispatchSubtasks`, immediately before dispatching another group. The flag that stops later review rounds, `budgetExceeded`, was set only when that dispatch check rejected a group.

With the default concurrency of 8, a review containing eight or fewer groups could dispatch every group before any LLM request reported token usage. There was then no next group to trigger another dispatch check, so `budgetExceeded` remained false and the active groups continued running additional rounds beyond the configured aggregate budget. Checks for larger reviews were also timing-dependent because the budget was read before waiting for a concurrency slot.

This change moves the live aggregate check into the shared main-task loop while retaining the existing dispatch look-ahead:

- `llmloop.Runner` checks aggregate usage before every main-task LLM request and returns a dedicated `StopTokenBudget` reason when the budget has been exceeded.
- An over-budget conversation receives one restricted grace round so it can submit findings it has already identified, but it cannot continue analysis with the normal tools.
- Review groups stopped this way are classified as `failed(budget)`, set `BudgetExceeded`, and emit one `token_budget_reached` warning.
- Scan uses the same loop-level check and reports the same budget state and warning. Its flag is atomic because scan subtasks run concurrently.
- `0` remains unlimited, and the dispatch-time estimate still avoids starting work that is already projected to exceed the budget.

The budget remains a cooperative aggregate limit rather than an exact hard cap: requests already in flight and the restricted grace round may finish after the threshold is crossed. The fix closes the gap where active groups never observed the aggregate budget at all.

## Type of Change

## How Has This Been Tested?

- `make check`
- `make test`
- `go test -race -count=1 ./internal/llmloop ./internal/agent ./internal/scan`
- `ocr review --audience agent`: 5 files reviewed, 0 findings

The new tests cover:

- stopping a running main-task conversation before its next normal LLM round once aggregate usage exceeds the budget;
- preserving the restricted grace round;
- propagating `failed(budget)`, `BudgetExceeded`, and `token_budget_reached` through review;
- propagating the same budget state and warning through scan;
- preserving the existing unlimited-budget behavior.

After rebasing onto the latest `main`, `make check` and the affected packages' race tests passed again.

## Checklist

- [x] My code follows the project coding style (`gofmt`, `go vet`, license, and English-only checks).
- [x] I added tests that prove the fix is effective.
- [x] New and existing unit tests pass locally.
- [x] I updated the CLI help and relevant documentation.
- [x] The commit message is in English and contains no AI attribution trailer.
- [x] I used AI/LLM assistance, disclosed it below, reviewed the output, and will answer maintainer questions from my own understanding.

**AI/LLM disclosure, per `AGENTS.md` rule 1:** this change was developed with OpenAI Codex (model: GPT-5) for implementation, tests, documentation, and wording. The required pre-commit review was run with `ocr review --audience agent` using `deepseek-v4-pro`. Per rule 6, the commit contains no `Assisted-by`, `Co-developed-by`, or similar trailer.

## Related Issues

Fixes #1295

Related: #1247, #1248


wu21-web:
This is likely a duplicate of https://github.com/alibaba/open-code-review/pull/1248 and the upstream issue is likely a duplicate of its related issue

PR #1306, opened 2026-09-16
fix(tool): enforce the global code search result limit

wxggzz:
## Description

Enforce the documented global 100-result limit in `code_search`.

`git grep --max-count` limits matches **per file**, not across the repository. Previously, two files with 60 matches each produced 120 results even though the response claimed that only the first 100 were shown. Exactly 100 matches also incorrectly triggered the truncation notice.

Request one extra match per file to detect overflow, then count only successfully parsed text matches and stop on the 101st. Binary-file diagnostics do not consume the result budget. This keeps the existing output format and file order. It bounds the rendered tool response; it does not introduce a global byte limit or streaming cancellation for the Git subprocess.

## Type of Change

## How Has This Been Tested?

On macOS arm64 with Go 1.26.8:

- Added 27 real-Git regression cases: workspace, commit (shared Git runner), and plain-directory modes; each covers 99 matches, exactly 100 across files, exactly 100 in one file, 120 across files, 101 in one file, binary diagnostics before text, binary diagnostics plus exactly 100/101 text matches, and binary-only results.
- Before the fix, the new cases reproduced 120 returned matches and false truncation notices. After the fix they pass.
- [x] `make test` passes locally, including race detection.
- `make check` and `make build` pass.
- `make coverage` passes: total statement coverage 91.6% (threshold 90%).
- `git diff --check` passes; both changed files use LF.
- The required `ocr review --audience agent --background ...` was attempted using the local build, but could not run because no LLM endpoint was configured (`resolve LLM endpoint: no valid LLM endpoint configured`). No live-model review is claimed.

## Checklist

- [x] Documentation already specifies the 100-result contract; no documentation change is needed
- [ ] I have signed the CLA
- [x] AI-assisted output has been personally reviewed and the contributor can answer maintainer questions independently

## AI disclosure

AI assistance: OpenAI Codex was used for investigation, implementation, regression tests, and code review. The exact runtime model identifier has not been independently verified and is not claimed here. No AI attribution trailers are included in the commit.

I have personally reviewed and understood the code, tests, and PR description, and can answer maintainer questions independently. The CLA remains pending.

## Related Issues

No existing issue is claimed as resolved. Existing open issues/PRs mentioning `code_search`, truncation, and result limits were checked to avoid duplicating an active fix.


lizhengfeng101:
Pushed a small follow-up commit (`cca1aade`) as maintainer.

**Motivation:** with the global limit enforced, a truncated file (or a file omitted entirely beyond the 100-match budget) is indistinguishable from a file that genuinely has that many matches — the caller cannot tell its view is incomplete.

**Change:** instead of breaking at the 101st match, the parse loop now scans all `git grep` output to count the distinct files that have text matches, while still rendering only the first 100. The truncation notice becomes:

> `Showing the first 100 matches across N matching files. Some files are partially shown or omitted entirely. Narrow file_patterns to see the rest.`

`N` is exact (unaffected by the per-file `--max-count` cap); per-file `Match lines` still reflects the rendered count by design. Tests updated accordingly (`make check` / `make test` / `make coverage` at 91.9% all pass).

internal/tool/code_search.go
```diff
@@ -209,18 +208,31 @@ func (p *CodeSearchProvider) gitGrep(ctx context.Context, searchText string, cas
 			continue
 		}
 		fname := parts[offset]
-		m := match{}
 		ln, parseErr := strconv.Atoi(parts[offset+1])
 		if parseErr != nil {
+			// Skip binary-file diagnostics and other non-match lines.
```
lizhengfeng101:
```suggestion
			// Skip lines whose line-number field is not numeric.
```

lizhengfeng101 (approved):
LGTM

PR #1288, opened 2026-09-15
fix(diff): flag untracked binary files as binary in workspace mode

xiaodu55:
## Description

In workspace mode, untracked binary files were fabricated as all-added text diffs and entered review selection as reviewable (a 55 MB executable reported ~270k insertions; `.DS_Store` counted as reviewable). Committed files are excluded as `binary` because `git diff` itself emits the `Binary files ... differ` marker, which the parser maps to `IsBinary`.

`untrackedFileDiffs` now sniffs the content for a NUL byte within git's 8000-byte window (the same heuristic as git and the scan provider's `binarySniffWindow`) and emits git's standard binary marker instead of fabricated `+` lines, so untracked binaries are excluded for the same reason as committed ones.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## How Has This Been Tested?

- [x] `make test` passes locally — full suite via `go test -count=1 ./...` (23 packages; `-race` skipped on this Windows machine for lack of a C toolchain)
- [x] Manual testing — the reproduction script from #1287: the untracked run now reports `0 reviewable / 1 total` with `[binary] (excluded: binary)`, identical to the committed run; `gofmt -s` and `go vet` clean

## Checklist

- [x] My code follows the project's coding style (`go fmt`, `go vet`)
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix is effective or my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] I have updated the documentation accordingly (if applicable) — N/A: behavior fix, no doc-facing surface change
- [x] I have signed the CLA
- [x] I did not use AI/LLM to create this PR, or I disclosed the tool/model below and reviewed its output; I did not attribute commits to AI and will answer maintainer questions and review comments myself without AI/LLM.
  - Disclosed: AI-assisted via ZCode (GLM model). Root-cause analysis, patch draft, and test draft were produced with AI assistance; I reviewed the diff, verified the fix locally with the issue's reproduction script, and will answer maintainer questions and review comments myself.

## Related Issues

Fixes #1287

PR #1310, opened 2026-09-16
perf(diff): skip oversized untracked files before reading

### LINKED ISSUE #1307: perf(diff): skip reading oversized untracked files into memory
## Description

In workspace mode, every untracked file is read fully into memory before any size check runs. `readWorkspaceFileForDiff` uses `os.ReadFile` with no cap (`internal/diff/workspace_file.go:58`), called per file by `untrackedFileDiffs` (`internal/diff/git.go:644`). The token/size gate that would exclude an oversized file runs only *after* the whole file is in memory and fabricated into a diff string (roughly doubling peak memory).

The token gate already excludes any untracked file whose diff exceeds ~160k tokens (~600 KB of text) as `too_large`, so review correctness is not affected. The remaining risk is purely memory: a multi-gigabyte untracked text file with no NUL byte in its first 8000 bytes (a large CSV, SQL dump, or log) is read in full just to be discarded, and can exhaust memory before the gate sees it. `Lstat` already runs in `readWorkspaceFileForDiff`, so the file size is known without an extra syscall.

**Expected behavior:** an untracked file larger than 64 MiB is flagged as binary (emitting git's standard `Binary files ... differ` marker, the path added in #1288) and excluded, **without** its contents ever being read into memory.

## Scope

- File(s): `internal/diff/git.go` (`untrackedFileDiffs`), `internal/diff/workspace_file.go` (`readWorkspaceFileForDiff`)
- Approach: stat the file before reading; when its size exceeds 64 MiB, emit the binary marker and skip the read, reusing the marker emission from #1288. No source file over 64 MiB is a realistic review target, so this never skips something that would otherwise be reviewed.
- Only the untracked path is affected; tracked/committed files go through git, which applies its own big-file handling.

## Acceptance Criteria

- [ ] An untracked file above 64 MiB is excluded as `binary` without being read into memory
- [ ] Files at or below 64 MiB are unaffected (existing NUL-byte sniffing still applies)
- [ ] A unit test covers the oversized-untracked-file case

## Context

Discovered while reviewing #1288 (fixes #1287, untracked binaries reviewed as text). #1288 flags untracked binaries via NUL-byte sniffing, but the sniff still happens after the full read. This issue tracks the remaining memory concern, which predates #1288 and is out of that PR's scope.

iluv7:
## Description

In workspace mode, every untracked file is fully read into memory before the token and size filters run. A multi-gigabyte text file can therefore exhaust memory even though it would later be excluded from review. The file size is already available from `Lstat` before the read.

## Change

- Limit reads on the untracked-file path to 64 MiB. Files above that threshold emit Git's binary diff marker without loading their contents.
- Skip the follow-up content read for binary diffs in the parser.
- Add regression coverage for an oversized untracked file and for the exact read-limit boundary.

## Testing

- `make check` — passed
- `make test` — passed
- `git diff --check` — passed

## AI disclosure

I used OpenAI Codex (GPT-5) to implement and test the change.

Closes #1307.


lizhengfeng101 (changes requested):
@iluv7 Nice work — the approach is sound and the tests cover both the workspace path and the read-limit boundary. One thing worth calling out as done right: the `if d.IsBinary { return }` guard in `finalizeDiff` is essential, not just an optimization — without it, the oversized file would be re-read with no limit during parsing (`ref == ""` → `readWorkspaceFileForDiff`), defeating the whole point. Good catch.

Two items before merge:

1. **Rebase onto current `main`.** #1288 landed after this branch was cut and also touches `untrackedFileDiffs`. GitHub reports no textual conflict, but please rebase and re-run `make test` so the combined behavior is verified: oversized → too-large branch, small binary → `looksBinary` branch, text → fabricated diff.

2. **Consolidate the binary-marker emission.** After the rebase there are two sites building the `Binary files ... differ` header with different formats — #1288 emits `Binary files a/f and b/f differ`, this PR emits `new file mode 100644\nBinary files /dev/null and b/f differ`. Extract a small helper both call, standardizing on this PR's more complete form (it sets `IsNew` too):

   ```go
   func untrackedBinaryDiff(path string) string {
       return fmt.Sprintf("diff --git a/%s b/%s\nnew file mode 100644\nBinary files /dev/null and b/%s differ\n", path, path, path)
   }
   ```

Otherwise LGTM.

iluv7:
> @iluv7 Nice work — the approach is sound and the tests cover both the workspace path and the read-limit boundary. One thing worth calling out as done right: the `if d.IsBinary { return }` guard in `finalizeDiff` is essential, not just an optimization — without it, the oversized file would be re-read with no limit during parsing (`ref == ""` → `readWorkspaceFileForDiff`), defeating the whole point. Good catch.
> 
> Two items before merge:
> 
> 1. **Rebase onto current `main`.** #1288 landed after this branch was cut and also touches `untrackedFileDiffs`. GitHub reports no textual conflict, but please rebase and re-run `make test` so the combined behavior is verified: oversized → too-large branch, small binary → `looksBinary` branch, text → fabricated diff.
> 2. **Consolidate the binary-marker emission.** After the rebase there are two sites building the `Binary files ... differ` header with different formats — #1288 emits `Binary files a/f and b/f differ`, this PR emits `new file mode 100644\nBinary files /dev/null and b/f differ`. Extract a small helper both call, standardizing on this PR's more complete form (it sets `IsNew` too):
>    ```go
>    func untrackedBinaryDiff(path string) string {
>        return fmt.Sprintf("diff --git a/%s b/%s\nnew file mode 100644\nBinary files /dev/null and b/%s differ\n", path, path, path)
>    }
>    ```
> 
> Otherwise LGTM.

Thanks for the review. I rebased the branch onto `main` after #1288 and extracted `untrackedBinaryDiff` so both oversized files and smaller NUL-containing files use the same marker.

lizhengfeng101 (approved):
LGTM

PR #1312, opened 2026-09-16
fix(session): tolerate truncated resume tail

### LINKED ISSUE #1308: Crash with "unexpected end of JSON input" when try to resume a scan
### OpenCodeReview Version

open-code-review v1.12.3 (71ed3a288d) darwin/arm64 built at: 2026-09-16T00:20:30Z

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

### Bug Description

When try to resume a scan, it returns an error. I set it up with the DeepSeek API.

### Steps to Reproduce

1. Start a scan
2. Abort with control +c
3. Resume with ocr scan --resume

### Expected Behavior

Continue scanning where from the last valid point.

### Logs / Error Output

```shell
Error: load resume session: parse resume session "4493c9f6-7eee-4b0f-8e0e-9b278ecabb53": unexpected end of JSON input (run 'ocr session list' to see available sessions)
```

### AI/LLM Disclosure

### Additional Context

_No response_

hyy321:
Fixes #1308.

Resume loading now ignores a single unterminated final JSONL record, which can be left behind when a session is interrupted during a buffered write.

Newline-terminated records are still parsed strictly, so malformed completed records continue to fail as before.

Added regression tests for:
- recovering completed checkpoints when the final record is truncated;
- keeping malformed newline-terminated records fatal.

## Validation

- `go test ./internal/session`
- `make check`
- `make test`
- `git diff --check`

## AI assistance

- Codex (gpt-6) was used to assist with code analysis.


internal/session/resume.go
```diff
@@ -127,14 +127,18 @@ func loadResumeState(repoDir, sessionID string, skipUnparseable bool) (*ResumeSt
...
 				return nil, err
 			}
 		}
-		if readErr == io.EOF {
-			break
-		}
+
 		if readErr != nil {
 			return nil, fmt.Errorf("read resume session %q: %w", sessionID, readErr)
 		}
```
wu21-web, line 144: @hyy321 “no trailing newline” does not necessarily mean the JSON bytes themselves are truncated.

```suggestion
		line, readErr := reader.ReadBytes('\n')

		if len(line) > 0 {
			if err := state.applyResumeLine(line); err != nil &&
				readErr != io.EOF && !skipUnparseable {
				return nil, err
			}
		}

		if readErr == io.EOF {
			break
		}
		if readErr != nil {
			return nil, fmt.Errorf("read resume session %q: %w", sessionID, readErr)
		}
```

hyy321: You're right that a final line without `\n` can still contain syntactically complete JSON.

The current behavior follows the fix direction in #1308, where a final record not terminated by `\n` is treated as not fully written and as carrying no committed state, so I intentionally avoid parsing the EOF tail here.

I agree the current comment makes the stronger assumption that the JSON bytes themselves are necessarily truncated. I can update the comment to describe the record as uncommitted instead.

wu21-web: Wait for maintainer's resp.

lizhengfeng101:
@hyy321 Nice fix — the EOF-before-parse ordering is correct and the two tests guard both branches well.

One small suggestion: consider adding a boundary case where the file contains *only* a truncated record with no preceding complete records. The current tests always have a valid record before the torn tail, so this "pure truncated tail" path (should yield an empty state without error) isn't directly covered.

hyy321:
> @hyy321 Nice fix — the EOF-before-parse ordering is correct and the two tests guard both branches well.
> 
> One small suggestion: consider adding a boundary case where the file contains _only_ a truncated record with no preceding complete records. The current tests always have a valid record before the torn tail, so this "pure truncated tail" path (should yield an empty state without error) isn't directly covered.

Good point. I'll add a test for the truncated-only case to make that behavior explicit.

hyy321:
Added a regression test for the truncated-only case. It verifies that a session file containing only an unterminated truncated record resumes with an empty state without error. Thanks for the suggestion.

wu21-web (approved):
Looks good to me. Thank you!

lizhengfeng101 (approved):
LGTM

PR #1314, opened 2026-09-16
feat(rules): add a C# rule doc

Menzi-Thami:
## Description

`.cs` is already in `supported_file_types.json`, but no rule doc covered it, so every C# file resolved to `default.md`. This adds `rule_docs/csharp.md` and maps `**/*.{cs,csx}` to it.

The doc follows the house format used by `go.md` and `java.md`: a precision-over-recall preamble telling the agent to establish call sites, DI lifetimes and disposal ownership with `file_read`/`code_search` before making a non-local claim, and not to duplicate what the compiler, nullable reference type analysis or Roslyn analyzers already report.

Sections cover the failure modes those tools do *not* reliably catch:

- **Nullability** — null-forgiving on values the code cannot guarantee; `null` returned for "not found" against a contract that promises otherwise.
- **Async and tasks** — `async void` outside an event handler; `.Result`/`.Wait()` deadlocks; an accepted `CancellationToken` that is never passed on; unobserved tasks; sharing a `DbContext` across parallel iteration.
- **Disposal and resource lifetime** — ownership of injected disposables; `using` on a type whose real cleanup is `DisposeAsync`; `HttpClient` per call.
- **Exceptions** — swallowing into `null`/`default`/empty; `throw ex;` resetting the stack trace; `finally` discarding the in-flight exception.
- **LINQ, collections, equality** — deferred execution enumerated after its scope ended; multiple enumeration; mutable dictionary keys; `Equals` without `GetHashCode`.
- **EF Core** — a `DbContext` used concurrently or captured in a singleton; N+1 navigation access; raw SQL handed an interpolated string.
- **ASP.NET Core** — captive dependencies; over-posting; middleware ordering; the options interfaces chosen so a reload is never observed.
- **Time and determinism** — static clock reads in logic that *branches* on time, with an explicit carve-out so plain audit stamping is not flagged.
- **Concurrency** — locking on publicly reachable objects; a synchronous critical section held across an `await`.
- **Testing** — assertions that cannot fail; shared mutable state between tests.

Docs: the `**/*.{cs,csx}` row is added to the built-in rules table in `review-rules.md` for all five locales, so the documented table matches `system_rules.json`.

## Type of Change

- [x] Documentation update

## How Has This Been Tested?

Not run: Go is not installed on the machine this was written on, so I could not execute the suite locally. I am relying on CI, and will fix anything it reports.

What I did verify by reading the tests that govern this change, in `internal/config/rules/system_rules_test.go`: every filename referenced by `system_rules.json` must be embedded, and every file under `rule_docs/` must be referenced by `system_rules.json` (no orphans). This change satisfies both — one new doc, one new mapping. `system_rules.json` was re-validated as JSON, and `**/*.{cs,csx}` uses the same brace-expansion form as existing entries. No Go code is touched.

## Checklist

- [ ] I have added tests — no new code paths; the existing `system_rules_test.go` invariants cover the mapping
- [ ] New and existing unit tests pass locally — see above
- [ ] I have signed the CLA — will sign when the bot prompts
- [ ] **AI/LLM disclosure:** this was drafted with assistance from Claude (Anthropic) via Claude Code. No commits are attributed to AI. The content has been reviewed before submission. Flagging it here rather than ticking the "did not use AI" box, since the template asks for disclosure either way.


Menzi-Thami:
Withdrawing — keeping these rules in my own configuration instead. Thanks for the tool.

PR #1317, opened 2026-09-16
fix: recover from truncated resume session

hyphen0009:
## Description

This PR adds graceful interrupt handling for scan sessions and a small
resume-session enhancement on top of the truncated-tail handling already
merged in #1312.

### Changes

- Gracefully handle the first Ctrl+C during scans.
- Persist checkpoints and close the session cleanly before exiting.
- Force exit on a second Ctrl+C.
- Thread the cancellation context through the scan flow.
- Add recovery diagnostics for resume sessions.
- Preserve a valid final JSONL record when it is complete but has no
  trailing newline.
- Keep incomplete/torn final records recoverable without changing the
  strict handling of malformed records in the middle of the session.

The truncated-final-record handling from #1312 is intentionally not
duplicated; the implementation in `main` remains the source of truth.

## Type of Change

## Testing

Added regression tests covering:

- Graceful abort → resume flow
- Valid newline-less final records
- Resume recovery diagnostics
- Review-session recovery
- Concurrent checkpoint writes

### Local verification

The code was statically reviewed and `git diff --check` is clean.

The full Go test/build/vet/check suite could not be run in the development
environment because the Go toolchain was unavailable.

CI should perform the project's standard test and verification suites.

## AI Assistance Disclosure

AI/LLM assistance was used during development.

Tools used:
- AI coding agent for investigation, implementation, conflict resolution,
  and test creation/review.


lizhengfeng101:
Thanks for the fix! Heads-up: the core of this — tolerating a truncated final JSONL record in `loadResumeState` — was just landed by #1312 (merged), which fixes the same `unexpected end of JSON input` on `--resume`. Both branch from the same `resume.go`, so this will now conflict with `main`.

Could you rebase onto `main` and drop the parts already covered by #1312? A couple of things here still look like net-new value worth keeping:

- **Graceful first Ctrl+C** in `scan_cmd.go` / `review_cmd.go` / `output.go` so checkpoints persist and the session closes cleanly — out of scope for #1312.
- **Recovering a valid-but-newline-less final record.** #1312 unconditionally drops the EOF residue; your `applyTrailingFragment` parses it first and keeps it if valid. If that's intended, it'd be great as a small enhancement on top of #1312's logic (plus the `Recovered` / `RecoveredFragment` diagnostics), rather than a full rewrite of the loop.

So: rebase, drop the duplicated tail-truncation handling, and keep the Ctrl+C work (and the newline-less-record recovery if you want it) layered on top. Happy to re-review once it's rebased.

hyphen0009:
Thanks for the review and for pointing me to #1312.

I've rebased the branch onto upstream/main and removed the duplicated truncated-tail implementation from #1312.

The PR now focuses on the graceful first Ctrl+C / clean session shutdown behavior, along with the small valid-newline-less-final-record enhancement layered on top of the existing #1312 implementation.

I've also updated the tests and pushed the rebased branch. The existing PR has been updated accordingly.

PR #1342, opened 2026-09-16
fix(vscode): add files allowlist to fix vsce packaging

### LINKED ISSUE #1318: vsce package excludes out/ from the VS Code extension .vsix (gitignore fallback), breaking the documented build steps
**OpenCodeReview Version**: n/a, this is a build-tooling issue in extensions/vscode, not the ocr CLI itself. Repo commit at time of testing: main branch, `@vscode/vsce` 3.9.2.

**Operating System**: macOS (Apple Silicon)

**Installation Method**: Built from source

**LLM Provider**: n/a (build-time issue, not a review-time issue)

## Bug Description

Following the README's own "Build a release package" steps in `extensions/vscode/README.md` fails at the packaging step:

```
$ cd extensions/vscode
$ yarn install
$ yarn package
...
extension (webpack 5.110.1) compiled successfully
webview (webpack 5.110.1) compiled successfully
configPanel (webpack 5.110.1) compiled successfully
 ERROR  Extension entrypoint(s) missing. Make sure these files exist and aren't ignored by '.vscodeignore':
  extension/out/extension.js
```

The webpack build genuinely produces `out/extension.js`, `out/webview.js`, and `out/configPanel.js` beforehand (visible in the log right above the error), and `extensions/vscode/.vscodeignore` does not exclude `out/`. So on its face this looks like a contradiction.

## Root Cause

`extensions/vscode/.gitignore` has `out/` on line 5, correctly keeping the build output out of git. But `@vscode/vsce`'s file-discovery walks up the directory tree and also honors `.gitignore` files (not just `.vscodeignore`) when the extension isn't declaring a `files` allowlist in `package.json` and there's no explicit flag telling it to ignore the git-based exclusion. Since this extension lives inside the larger `open-code-review` monorepo (which is a git repository), `vsce package` picks up `extensions/vscode/.gitignore`'s `out/` rule and silently drops the entire `out/` directory from the packaged `.vsix`, even though `.vscodeignore` correctly says it should be included.

I confirmed this precisely: building the exact same output with the exact same command outside of any git repository (copied `extensions/vscode` to a plain, non-git directory and ran `vsce package` there) packages correctly and includes `out/`. Rebuilding inside the repo reproduces the failure every time.

## Steps to Reproduce

1. Clone the repo, `cd extensions/vscode`
2. `yarn install`
3. `yarn package` (or `npx vsce package --no-yarn` directly)
4. Observe: webpack compiles `out/extension.js` etc. successfully, then `vsce` fails with "Extension entrypoint(s) missing"

## Expected Behavior

`yarn package` should produce a working `.vsix` per the README's own documented steps, without needing to work around git-based exclusion manually.

## Suggested Fix

A few options, roughly in order of how surgical they are:

- Add a `"files"` allowlist to `extensions/vscode/package.json` (e.g. `["out/**", "resources/**", "LICENSE.txt", "readme.md", "package.nls*.json"]`). When `vsce` sees a `files` field it switches to npm-style packing and stops consulting `.gitignore` for inclusion/exclusion, which resolves this directly and is the conventional fix vsce's own docs recommend for exactly this situation.
- Alternatively, pass `vsce package`'s ignore-file override so it only consults `.vscodeignore` (I did not find a bare `--no-git-ignore`-style flag on `vsce package` in 3.9.2, `--ignoreFile <path>` only lets you swap which vscodeignore-style file is read, it does not disable the gitignore fallback), so the `files`-field approach is likely the cleanest fix available without a vsce version bump.
- Or move `extensions/vscode`'s `.gitignore` `out/` rule out of the ignore chain vsce walks (not recommended, since git still legitimately should not track build output).

I'm happy to open a PR with the `files`-field fix if that's the preferred direction, just didn't want to submit one blind without confirming which approach the maintainers want first.

## Additional Context

This affects anyone trying to build the extension from source per the README, not just CI — no CI job currently packages/verifies the `.vsix` (issue #441 only added a CI job for the extension's own build/lint/test, not `vsce package` itself), which is likely why this has gone unnoticed.

I used Claude Code (AI assistance) to help investigate and diagnose the root cause of this issue, including confirming it by reproducing the build both inside and outside the git tree.

HoneyTyagii:
## Description

Fixes #1318

`vsce package` walks up the directory tree and picks up `.gitignore` rules when `package.json` lacks an explicit `files` allowlist. Since the extension lives inside the `open-code-review` monorepo, the root `.gitignore` rule for `out/` was silently dropping build output from the `.vsix`, even though `.vscodeignore` explicitly includes it.

## Fix

Add an explicit `files` field to `extensions/vscode/package.json`. This switches `vsce` to npm-style packing, which only consults `.vscodeignore` and ignores parent `.gitignore` rules.

## Changes

- `extensions/vscode/package.json`: added `files` allowlist with `out/**`, `resources/**`, `LICENSE`, `README.md`, `README.zh-CN.md`, `package.nls.json`


austinuni60-netizen (approved):

lizhengfeng101:

PR #1356, opened 2026-09-17
docs(skill): skip ocr install pre-check on the common path

lizhengfeng101:
## What

Add an explicit instruction to the open-code-review skill telling the agent **not** to probe whether the `ocr` CLI is installed (e.g. `command -v ocr` or `ocr --version`) before running a review. The agent should assume the CLI is available and run the review directly, falling back to the Troubleshooting install path only when the command fails with `command not found`.

Applied to both files that must stay in sync:
- `skills/open-code-review/SKILL.md` (canonical)
- `plugins/open-code-review/skills/open-code-review/SKILL.md` (plugin mirror)

## Why

In practice the agent was conservatively spending a tool call on a `command -v ocr` probe before every review, even though the skill already handles the not-installed case in its Troubleshooting section. That probe is wasted work on the common path where `ocr` is present. Making "don't pre-check" explicit removes the redundant round-trip while keeping the existing `command not found` fallback fully intact.

## Notes

No behavioral code change — documentation/instructions only.

AI disclosure: model used — Claude Opus 4.8. All content was reviewed by the author.


MuoDoo (approved):
LGTM

PR #1359, opened 2026-09-17
fix: dedupe overlapping incremental review comments

### LINKED ISSUE #1344: Duplicate inline comments on same file+line within a single run (incremental doesn't dedupe intra-run)
## Summary

With `incremental: true`, a single `ocr review` run still posts multiple duplicate inline comments on the **same file + line** (and sometimes the same file + similar line ranges). The `incremental` overlap filter only dedupes against comments from *previous runs* — it does not dedupe comments generated within the current run.

## Repro (real-world)

A PR with ~10 changed files produced 90 bot comments. These duplicate groups were posted in one run:

- 3 separate comments on the same `file:134` all flagging the same bug (version hardcoded to 0), with different severities (`bug · high`, `bug · medium`)
- 2 comments on `file:140` covering the same issue
- 2 comments on `file:39` and 2 on `file:41` in a shell hook, both about fetch error suppression
- 2 near-identical `actions/checkout@v7` comments on the same line with different severities

## Root cause hypothesis

Parallel per-file/batch dispatch: each batch independently reviews its slice of the diff and generates comments with no cross-batch dedup. The post-time `incremental` overlap check (IoU vs existing comments) only sees comments already on GitHub, not comments generated in sibling batches of the same run. When the same diff hunk (or overlapping line ranges) appears in two batches — e.g., a renamed file, a file split across batches, or overlapping context lines — the LLM flags the same issue twice.

## Expected behavior

Within one run, comments that overlap another comment from the same run on the same (path, line range) should be suppressed, or at least collapsed/deduped. The `incremental_overlap_threshold` IoU check should also apply to same-run comments, not just historical ones.

## Workaround

Setting `review_concurrency: 1` forces sequential file review and eliminates the cross-batch duplicate comments, at the cost of review speed.

## Environment

- GitHub Action `alibaba/open-code-review@f1101fd7f51304c82e4a4f292bbee88aea0823cf`
- `@alibaba-group/open-code-review` latest (>= v1.9.6)
- `effort: high`, `incremental: true`, `sticky_summary: true`

devnp2007:
## What changed

* Deduplicate overlapping inline comments produced within the same review run when `incremental: true`.
* Reuse the existing incremental overlap/IoU semantics for same-run comments.
* Preserve the existing historical-comment deduplication behavior.
* Add regression coverage for both exact same-line duplicates and overlapping multiline comments.

## Why

When review comments are produced in parallel batches, multiple findings can target the same file and line within a single run.

The existing incremental filtering only compared current findings against comments from previous runs. As a result, overlapping findings generated during the same run could all be posted.

This change performs the same overlap check against already-accepted comments from the current run before comments are sorted and batched for publishing.

## Testing

* `node scripts/github-actions/post-review-comments.test.js`
* Added regression coverage for same-line same-run duplicates.
* Added regression coverage for overlapping multiline same-run comments using the existing IoU threshold semantics.

Fixes #1344


scripts/github-actions/post-review-comments.test.js
```diff
@@ -598,6 +598,69 @@ async function testIncrementalSkipsOverlapping() {
...
+  assert.strictEqual(sent[0].path, "src/a.js");
+  assert.strictEqual(sent[0].line, 10);
+  assert.strictEqual(sent[1].path, "src/b.js");
+  assert.strictEqual(sent[1].line, 5);
+
+  assert.strictEqual(outputs.comments_skipped, "1");
+  assert.strictEqual(outputs.comments_inline, "2");
+}
+
+
```
lizhengfeng101:
```suggestion

```

lizhengfeng101 (approved):
LGTM

PR #1363, opened 2026-09-17
fix(llm): replay tool_call extra_content for OpenAI-compatible providers

### LINKED ISSUE #1357: Review fails with HTTP 400 after tool call when using Gemini 3 on Vertex AI OpenAI-compatible endpoint
### OpenCodeReview Version

open-code-review v1.12.4 (f1101fd7f) darwin/arm64, built at: 2026-09-16T09:58:39Z

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

## Describe the bug

When using a Gemini 3 model through Vertex AI's OpenAI-compatible Chat Completions endpoint, `ocr review` consistently fails with `HTTP 400 Bad Request` after the first tool call.

The initial LLM request succeeds and returns a tool call such as `file_read`. `ocr` executes the tool successfully, but the following LLM request — which contains the assistant tool call and tool result — is rejected by the provider with `400 Bad Request`.

The response body is empty.

In contrast, `ocr llm test` succeeds:

```text
✓ Connection test successful
```

This appears to be because the connection test only performs a single request and does not exercise a tool-call round trip.

## Steps to reproduce

1. Configure a custom OpenAI-compatible provider using:

```text
https://aiplatform.googleapis.com/v1/projects/<PROJECT_ID>/locations/global/endpoints/openapi/chat/completions
```

with:

```text
google/gemini-3.8-flash
```

2. Run:

```bash
ocr llm test
```

The connection test succeeds:

```text
✓ Connection test successful
```

3. Run a review against any small diff:

```bash
ocr review --from <base-branch> --to <feature-branch>
```

4. The model successfully requests a tool such as `file_read`, but the next LLM request fails with HTTP 400.

---

## Actual behavior

Example CLI output:

```text
[ocr] 1 file(s) changed, reviewing 1 in <repo>
[ocr] Skipping plan phase for group "<file>" (1 file(s), max 7 lines, total 7 lines)

[ocr]   ▶ file_read file_path=<file> start_line=450 end_line=520
[ocr]   ✔ file_read (28ms)

[ocr] Subtask error for group "<file>":
LLM completion error: POST "https://aiplatform.googleapis.com/v1/projects/<PROJECT_ID>/locations/global/endpoints/openapi/chat/completions": 400 Bad Request

[ocr] Summary: 1 file(s) reviewed, 0 comment(s), ~3336 token(s) used
Review failed: 0 finding(s); 1 of 1 selected item(s) failed.

LLM retry report summary: 1 of 2 requests affected -- 1 request failed

Core review (1 request):
- <file>: rejected by provider (HTTP 400) -> failed

Error: review failed: all 1 file review(s) failed — check your LLM configuration and API key
```

The failure is reproducible across repeated runs.

---

## Session flow

The session log at:

```text
~/.opencodereview/sessions/<repo>/<session>.jsonl
```

shows the following sequence.

### Request 1 succeeds

The model returns an assistant response with no normal content and one tool call:

```json
{
  "content": "",
  "tool_calls": [
    {
      "id": "call_<REDACTED>",
      "name": "file_read",
      "arguments": "{\"end_line\":520,\"file_path\":\"<file>\",\"start_line\":450}"
    }
  ]
}
```

### Tool execution succeeds

`file_read` executes successfully and returns the requested file contents.

### Request 2 fails

The next LLM request fails with:

```text
400 Bad Request
```

The response body recorded by `ocr` is empty.

The session log does not contain fields such as:

```text
thought_signature
extra_content
```

## OR simply

Steps to Reproduce

1. Configure a custom provider using the OpenAI protocol:
   - URL: `https://aiplatform.googleapis.com/v1/projects/<PROJECT_ID>/locations/global/endpoints/openapi/chat/completions`
   - Model: `google/gemini-3.8-flash`
2. Run `ocr llm test` → `✓ Connection test successful`
3. Run a review on any small diff (mine was 1 file, 7 changed lines):
   `ocr review --from <base-branch> --to <feature-branch>`
4. The model requests `file_read`, the tool succeeds, and the next LLM request fails with HTTP 400.

It happens on every run (3 of 3).

## Expected behavior

`ocr` should be able to complete tool-calling conversations with Gemini 3 models through Vertex AI's OpenAI-compatible endpoint.

Provider-specific metadata attached to assistant tool calls should be preserved when the conversation history is sent back to the provider on subsequent turns.

If the provider rejects a request, the response body should also be surfaced or logged when available so the underlying validation error can be diagnosed.

### Logs / Error Output

```shell
Session log (`~/.opencodereview/sessions/<repo>/<session>.jsonl`):
1. `llm_response` (request 1) succeeds: `content: ""`, one tool call
   `{"id":"call_<REDACTED>","name":"file_read","arguments":"{\"end_line\":520,\"file_path\":\"<file>\",\"start_line\":450}"}`
2. The tool result is `ok: true`, with the file contents returned.
3. `llm_error` (request 2): `400 Bad Request`, empty body, after 584ms.
```

### AI/LLM Disclosure

### Additional Context

Additional Context

**Suspected cause (unconfirmed — the 400 body is empty):**
Gemini 3 tool calls include a thought signature, which the OpenAI-compatible endpoint returns under `tool_calls[].extra_content.google.thought_signature`. Vertex requires it to be sent back unchanged with the assistant tool call on the next request. If `ocr` keeps only the standard fields (`id`, `name`, `arguments`) when it rebuilds the history, the signature is lost and the provider rejects the request.

Evidence:
- The first request succeeds, and the tool itself runs fine.
- The failure is always on the first request after a tool call.
- `ocr llm test` (no tool-call round trip) succeeds.
- Neither `thought_signature` nor `extra_content` appears in the session log. Caveat: the log stores ocr's parsed version, not raw HTTP payloads, so this suggests the cause but doesn't prove it.

**Possible fixes:**
1. Keep `extra_content` (or all unknown fields) on `tool_calls` and send them back unchanged on later turns.
2. Log the provider's response body on 4xx/5xx errors.
3. Add a tool-call round trip to `ocr llm test` so this kind of incompatibility shows up during setup.

xiaodu55:
## What

Gemini 3 models reached through Vertex AI's OpenAI-compatible chat completions endpoint attach a thought signature to each tool call as `tool_calls[].extra_content`, and require it to be sent back unchanged with the assistant tool call on the next request. `ocr` rebuilt conversation history from the standard fields only (`id`, `type`, `function`), so the signature was dropped and every request after the first tool call was rejected with HTTP 400 (#1357).

This change preserves the field verbatim through the whole round trip:

- `ToolCall` gains an `ExtraContent json.RawMessage` field with a `json:"extra_content,omitempty"` tag, so session persistence round-trips it unchanged.
- Non-streaming: `mapOpenAIResponse` extracts `extra_content` from each tool call's raw response JSON (new `toolCallExtraContent` helper).
- Streaming: the SDK accumulator ignores unmodeled fields, so the field is captured from the delta fragments' `ExtraFields` and re-attached to the mapped tool call, with negative delta indices (sent by some gateways such as Bedrock or liteLLM for a single call) clamped to 0 to match the accumulator's own handling.
- Replay: `buildOpenAIParams` echoes `extra_content` back on the assistant tool call via `SetExtraFields` — the same mechanism already used for `reasoning_content` (#805).

The extraction is provider-agnostic: whatever sits under `extra_content` is replayed byte for byte, and providers that never send the field produce exactly the same wire format as before.

## Why

- Fixes #1357: `ocr review` with Gemini 3 on Vertex AI failed on every run at the first post-tool-call request (`LLM completion error: ... 400 Bad Request`) while `ocr llm test` passed, because the test never exercises a tool-call round trip.
- The report's follow-up expectation about surfacing the response body turned out to be already satisfied: `apierror.Error()` in openai-go appends the response body to its message. The empty body in the report is what the provider returned, not something `ocr` drops, so no `ocr`-side change is needed there.

## How checked

- New deterministic, no-live-model regression tests in `internal/llm/native_turn_test.go`, following the issue-#805 test pattern:
  - `TestOpenAIChatCompletions_ReplaysToolCallExtraContentAcrossTurns` — the signature survives `mapOpenAIResponse` → `NewToolCallMessage` → `buildOpenAIParams` byte for byte.
  - `TestOpenAIChatCompletions_ToolCallExtraContentOmittedWhenAbsent` — no `extra_content` field appears on the wire for providers that don't send it.
  - `TestOpenAIChatCompletions_StreamingCapturesToolCallExtraContent` — SSE fixture covering the streaming capture and replay path.
  - `TestOpenAIChatCompletions_StreamingClampsNegativeToolCallIndex` — a delta tool call with `index: -1` still lands its signature on slice position 0.
- `go test` passes across the repo's Go packages (`internal/llm`, `internal/session`, `internal/agent` re-run explicitly), plus `gofmt -s`, `go vet ./...`, the license check and the English-only check. No new dependencies.

A follow-up commit applies review feedback on this PR: the streaming index clamping above, a whitespace-tolerant null check in `toolCallExtraContent`, and fixtures aligned to Vertex AI's actual `extra_content` object shape.

## AI/LLM Disclosure

I used an AI coding assistant (ZCode CLI, powered by GLM) to help analyze the codebase and draft this change; I reviewed and tested every change myself before submitting, and I can explain each part of the diff.


AyushDubey23:
Hi @xiaodu55, thanks for putting this up! The approach of keeping `extra_content` on `ToolCall` makes sense, but I noticed a few edge cases and an omission from #1357 you might want to address:

1. **Streaming index clamping (`deltaTool.Index = -1`)**:
   In `completionsStreamingInner`:
   ```go
   extraContentByChoice[choice.Index][deltaTool.Index] = json.RawMessage(ec.Raw())
   ```
   Some OpenAI-compatible providers/gateways (like AWS Bedrock or liteLLM) return `deltaTool.Index = -1` when there's only one tool call. The SDK accumulator handles this by clamping negative indices to `0` (`clampToZero(deltaTool.Index)`). Storing this under `-1` means the re-attachment loop (`int64(j)` starting at `0`) will miss it, and `ExtraContent` will be dropped in streaming mode. Clamping the index to zero before storing would prevent that.

2. **Missing 4xx response body surfacing**:
   Part of the problem reported in #1357 was that provider validation failures only show `400 Bad Request` with an empty error body, which made this issue tricky to diagnose in the first place. This PR doesn't address surfacing the response body on 4xx errors yet.

3. **Vertex AI test fixture format**:
   In `native_turn_test.go`, the test fixture uses an array:
   ```json
   "extra_content": [{"google": {"thought_signature": "sig-abc-123"}}]
   ```
   On Vertex AI's actual Gemini endpoint, `extra_content` is an object:
   ```json
   "extra_content": {"google": {"thought_signature": "sig-abc-123"}}
   ```
   While `json.RawMessage` accepts both, aligning the test fixture with Vertex AI's actual schema would be more accurate.

4. **Null check with whitespace**:
   In `toolCallExtraContent`, `string(ec) == "null"` won't match if there's leading/trailing whitespace (e.g. `" null "`). `bytes.Equal(bytes.TrimSpace(ec), []byte("null"))` is safer.

xiaodu55:
Thanks for the careful review — all fair points. Follow-up commit pushed on this branch:

1. **Streaming index clamping** — fixed: negative `deltaTool.Index` is now clamped to 0 before storing, matching the accumulator's `clampToZero`, with a regression test that feeds `index: -1` through the SSE fixture (`TestOpenAIChatCompletions_StreamingClampsNegativeToolCallIndex`).
2. **4xx response body** — no code change here: `apierror.Error()` in openai-go already appends the response body to its message (`POST "...": 400 Bad Request <body>`), so the empty body in the report is what the provider actually returned, not something `ocr` drops. Keeping this PR focused on the metadata round trip; if maintainers want extra error formatting on top, happy to discuss in a follow-up.
3. **Fixture format** — aligned both fixtures to the object form (`"extra_content": {"google": {...}}`). The replay is verbatim either way, so both shapes work.
4. **Null check with whitespace** — adopted `bytes.TrimSpace` in `toolCallExtraContent`.

xiaodu55:
One precision on point 2, since #1042 came up on the issue: openai-go extracts only the top-level `error` field of the response body into `apierror` (`gjson.GetBytes(contents, "error")` in `requestconfig.go`), and `Error()` appends that object's raw JSON. So a standard `{"error": {...}}` body does surface its message and details — on the current SDK that part of #1042 should no longer reproduce — while an empty or non-standard body, like the empty 400 Vertex returned in this issue, stays opaque.

That residual gap is a pre-existing concern independent of the metadata round trip, so I kept this PR scoped to #1357; happy to address #1042 separately if maintainers want it.

xiaodu55:
Correction to my earlier comments and the PR description, for the record: I claimed the empty error body was what the provider returned. That was wrong — as @Moon-Wrecker documented in #1393, Vertex did send an explanation, array-wrapped as `[{"error": {...}}]`, which openai-go's `gjson` path `"error"` doesn't match, so `RawJSON()` came back empty and only the status line survived. The opaque error and the dropped thought signature are two faces of the same provider-shape mismatch.

#1393 covers everything this PR set out to do — the same `ExtraContent` round trip including the negative streaming index case, plus the error-body surfacing — so the cleanest path is to consolidate there. I'll close this once its checks are green. Thanks @AyushDubey23 for the review that flushed out the index clamping, and @Moon-Wrecker for the definitive diagnosis.

xiaodu55:
Closing this now rather than waiting: #1393's main workflows are still gated on the first-time-contributor approval, so "green checks" isn't something this thread can schedule — and the consolidation rationale doesn't depend on them anyway. It covers the full scope of #1357 (and #1042) with the reporter's own reproduction to validate against. If anything from this branch turns out to be missing there, happy to help.

PR #1368, opened 2026-09-17
fix(diff): decline ambiguous same-file existing_code matches

### LINKED ISSUE #991: Review comment attaches to the first same-file existing_code match when duplicates exist
### OpenCodeReview Version

open-code-review v1.9.5 (ea1de41a) darwin/arm64 built at: 2026-08-17T03:52:37Z https://github.com/alibaba/open-code-review

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

### Bug Description

When the same `existing_code` appears multiple times in the same file, OCR may attach the review comment to the first matching location instead of the location the model actually commented on.

I noticed that PR #921 fixed a related duplicate matching issue for cross-file relocation, where the same code snippet may appear in multiple files. This issue is slightly different: the duplicate candidates are in the same file.

In this case, the file contains two identical lines:

```go
ev.ErrorClass = ArgosRuntimeStatusTimeout
```

The model commented on the second occurrence in the `Analyze` branch, where a non-timeout error was incorrectly marked as timeout. However, OCR attached the final review comment to the first identical occurrence at line 286.

Image

### Steps to Reproduce

1. Run `ocr review` on a diff where the same `existing_code` appears multiple times in the same file.
2. Make the model produce a `code_comment` whose `existing_code` matches those repeated snippets.
3. Open the generated review result.
4. Observe that the comment is attached to the first matching location instead of the later location that the comment content refers to.

### Expected Behavior

When `existing_code` matches multiple candidate locations within the same file, OCR should not directly choose the first match.

If the target code block can be determined from the comment content, surrounding context, or candidate locations, OCR should attach the comment to the corresponding location.

If OCR cannot determine the target location reliably, it should avoid outputting an incorrect precise line number and keep the comment unlocated instead.

### Additional Context

I have a #992 ready to address the ambiguous location resolution when `existing_code` has multiple candidate matches.

alter:
## Description

Fixes #991. When a review comment's `existing_code` matches more than one identical location within the same file, `matchConsecutive` (and `resolveFromFileContent`'s own separate copy of the same scan) returned the first occurrence, so the comment could be anchored to the wrong line. For example, the model's comment about the second occurrence of a duplicated line would land on the first one instead.

This is the minimal fix the issue reporter proposed themselves: when a same-file match is ambiguous, decline instead of guessing, the same way `RelocateAcrossFiles` already declines when the same snippet shows up in more than one file. Callers already treat a decline as "still unresolved" and fall through to cross-file relocation and then LLM-based re-location (`internal/llmloop/loop.go`), so an ambiguous match now gets a real shot at disambiguation instead of landing on a silently wrong line number.

`resolveFromFileContent` had its own separate first-match-wins loop instead of calling `matchConsecutive`. It now builds an `[]indexedLine` and reuses `matchConsecutive`, so the hunk-based and whole-file-fallback paths share one matching rule.

## Type of Change

## How Has This Been Tested?
- Renamed/rewrote the two existing tests that asserted the old "first match wins" behavior (`TestMatchConsecutive_AmbiguousMatchDeclines`, `TestResolveLineNumbers_FallbackToFileContent_AmbiguousMatchKeepsZero`) to assert the corrected decline instead, and added `TestResolveFromHunk_AmbiguousDuplicateDeclines` for the hunk-level path and `TestMatchConsecutive_OverlappingDuplicateDeclines` for overlapping duplicate windows. All were confirmed failing against the pre-fix code before landing.
- Manual mutation testing (reverse control): dropping the ambiguity guard, an off-by-one on the guard's boundary, and reverting only the `resolveFromFileContent` refactor were each introduced on purpose and confirmed caught by the suite, then reverted.
- `make test`: no new failures vs. main (one pre-existing, environment-specific failure in `internal/diff.TestCommitDiffMergeCommitReviewsFirstParentDiff`, unrelated to this change, root-caused to a local `merge.ff=only` git config, not a code defect).
- `make vet`, `make build`, `gofmt -s -l .`, `go run scripts/verify-english-only.go` all pass. Coverage floor independently confirmed above 90% (91.6% total / 94.3% for `internal/diff`) with that one pre-existing failing test skipped.

## AI/LLM Disclosure
I used an AI coding assistant (Claude Code, Claude Sonnet 5) to research open issues, identify this one as a well-scoped, uncontested fix, draft the fix and its tests, and run the verification commands above. I reviewed and understand every line of the change before opening this PR.


Bangkah (approved):
Clean refactoring, reuses matchConsecutive properly, drops blind first-match guessing for ambiguous duplicates. Solid tests. LGTM, pending CLA clearance

PR #1372, opened 2026-09-17
Create a user level skill and use local copilot only

ZoltanDobrovolni:


ZoltanDobrovolni:
Unintentional PR -> removed

PR #1389, opened 2026-09-17
fix(sarif): encode artifact paths as URI references

Adi202001:
## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [ ] I have signed the CLA


chaojixinren (commented):
I've completed the review and found no blocking issues. The implementation and tests look good. Please fill in the PR description with the problem being fixed, a brief example, and the testing performed before merging.

lizhengfeng101:
@Adi202001 Good catch on the `#` case, and nice call leaving `fingerprintSource` on the raw path — that keeps existing findings from being re-reported.

Four things before this goes in:

1. **The docs change is en-only.** `integrations/ci.md` also exists under `zh`, `ja`, `ko` and `ru`. Please mirror the new paragraph into all four — see ab741e63 for how the previous SARIF doc change handled it.

2. **`./generated:main.go` deserves a comment.** It's the one case in the table where the output gains something the input didn't have, and the reason (RFC 3986 §4.2: a first segment containing `:` would be mistaken for a scheme) isn't written down anywhere. Without a note, the next reader will take the `./` for a bug and "fix" it.

3. **Consider a named helper**, e.g. `sarifArtifactURI(path string) string`. It reads better than the inline `(&url.URL{Path: ...}).String()`, and it can be unit-tested directly instead of through `outputSARIF` plus a JSON round-trip.

4. **Pin the fingerprint.** Nothing currently locks in the property from my first paragraph. Please assert the concrete `partialFingerprints` value for a path like `src/my file.go`. Otherwise someone swaps `c.Path` for `artifactURI` later, the suite stays green, and users get a wave of duplicate alerts in CI.

lizhengfeng101:
@Adi202001 Looking forward to your update

Issue #811, opened 2026-08-09
Anthropic signed thinking blocks are lost (replayed as plain text, signature dropped) across tool-use turns


Fanzzzd:
## OpenCodeReview version

- Reproduced on current `main`: `71d29810be1e7f7b183c021fe9b06a2e850f1037`
- Installation: built from source
- Affected protocol: Anthropic Messages (`protocol: anthropic`)
- Follow-up to #805; the OpenAI-compatible Chat Completions slice is tracked in #806

## Bug description

The Anthropic adapter parses `thinking` blocks into the normalized `ResponseMessage.ReasoningContent`, but assistant history for the next tool-result request is rebuilt from only visible text and `tool_use` blocks. Two things go wrong for extended-thinking models:

1. The signed `thinking` block (and any `redacted_thinking` block) is not replayed. Anthropic requires thinking blocks around tool use to be passed back unmodified and in their original order; a strict endpoint can reject the continuation, and a tolerant one returns HTTP 200 while silently losing reasoning continuity.
2. It is worse than a plain drop: a thinking + tool_use turn has no visible text, so `ChatResponse.Content()` falls back to the reasoning text, and the private thinking is replayed as an ordinary **visible `text` block**. Hidden reasoning leaks into visible assistant content, and its `signature` disappears entirely.

## Source trace

All references are `main` @ `71d2981`:

1. [`mapAnthropicResponse`](https://github.com/alibaba/open-code-review/blob/71d29810be1e7f7b183c021fe9b06a2e850f1037/internal/llm/client.go#L833-L868) flattens `thinking` blocks into a joined `ReasoningContent` string — the block `signature` is already gone at this point.
2. `RunPerFile` rebuilds history via `resp.Content()` + `resp.ToolCalls()` (the lossy projection described in #805), and [`Content()`](https://github.com/alibaba/open-code-review/blob/71d29810be1e7f7b183c021fe9b06a2e850f1037/internal/llm/client.go#L154-L165) substitutes reasoning text when visible content is empty.
3. [`buildAnthropicParams`](https://github.com/alibaba/open-code-review/blob/71d29810be1e7f7b183c021fe9b06a2e850f1037/internal/llm/client.go#L724-L749) emits only `text` and `tool_use` blocks for assistant history; no code path emits a `thinking` block param.

## Deterministic reproduction

No live endpoint or API key needed — this exercises the adapter's own response→history→request chain:

```go
[...]

PR #1371, opened 2026-09-17
fix(llm): preserve tool_call extra_content and enrich error body for OpenAI providers

AyushDubey23:
## Description

When using Gemini models (such as `google/gemini-3.8-flash`) via Vertex AI's OpenAI-compatible Chat Completions endpoint, tool calls return provider metadata carrying a thought signature under `tool_calls[].extra_content` (e.g. `extra_content.google.thought_signature`). Vertex AI requires this metadata to be echoed back unchanged with the assistant tool call in the next request.

Previously, `internal/llm/client.go` discarded unmodeled tool-call fields when mapping completions (`mapOpenAIResponse`) and building parameters (`buildOpenAIParams`), causing subsequent turns to be rejected with `HTTP 400 Bad Request`. Additionally, when the provider rejected requests with 4xx errors, the raw response body was dropped, leaving users with an opaque `400 Bad Request` and an empty error message.

This change:
1. Adds `ExtraContent json.RawMessage` to `llm.ToolCall` to preserve unmodeled provider metadata across conversation turns.
2. Extracts `extra_content` from raw tool call JSON in `mapOpenAIResponse` with whitespace-trimmed null validation.
3. Captures `extra_content` from streaming chunks in `completionsStreamingInner`, properly clamping negative indices (`deltaTool.Index = -1` used by certain gateways like Bedrock or custom proxies) to slice index `0`.
4. Replays `extra_content` in `buildOpenAIParams` on `ChatCompletionMessageFunctionToolCallParam` via `SetExtraFields`.
5. Adds `formatOpenAIError` to unpack and surface provider response bodies on 4xx/5xx API errors when available, making validation failures transparent and easy to diagnose.
6. Adds unit tests covering non-streaming replay (using Vertex AI's object schema), absent metadata omission, streaming capture with index clamping, and 4xx response body enrichment.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring (no functional changes)
- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] Unit tests pass locally: `go test -v -run "TestOpenAIChatCompletions" ./internal/llm`
- [x] Verified english-only check: `go run scripts/verify-english-only.go`
- [x] Verified static analysis: `gofmt -s -w internal/llm && go vet ./internal/llm/...`
[...]

PR #1393, opened 2026-09-18
fix(llm): replay tool_call extra_content and surface provider error bodies

### LINKED ISSUE #1042: Opaque "400 Bad Request" error with missing details when using Gemini / OpenAI-compatible endpoints

### LINKED ISSUE #1357: Review fails with HTTP 400 after tool call when using Gemini 3 on Vertex AI OpenAI-compatible endpoint

Moon-Wrecker:
## Description

Two reported bugs, one provider behaviour.

Gemini 3 on an OpenAI-compatible endpoint attaches a thought signature to each tool call as `tool_calls[].extra_content` and rejects the next request unless it is echoed back unchanged. `llm.ToolCall` modelled only `id`/`type`/`function`, so `mapOpenAIResponse` dropped it and `buildOpenAIParams` could not replay it — every request after the first tool call failed with `400 Bad Request` (#1357).

The same provider array-wraps its error documents as `[{"error": {...}}]`. openai-go extracts the payload with the gjson path `"error"` (`internal/requestconfig/requestconfig.go:656`), which does not match at an array root, so `RawJSON()` came back empty and the user saw a bare `400 Bad Request` with nothing after it (#1042). That is why #1357 was so hard to diagnose: the provider *did* explain the failure, and we threw the explanation away.

They are fixed together because the second bug is what hides the first.

### Where the metadata is kept

`ToolCall` gains `ExtraContent json.RawMessage` tagged `json:"-"`, and `internal/session` persists it deliberately in both projections.

That combination is deliberate and follows the rule the codebase already states for `Message.Native` at `internal/llm/client.go:105-110`:

> `json:"-"` because Payload is an SDK struct unsuitable for **incidental marshaling**; internal/session persists it **deliberately** via `ChatResponse.Native()`.

A bare `json:"-"` keeps the wire format honest but loses the value from the session record. A plain struct tag persists it, but by the incidental marshaling that comment rejects. Doing both is what `Native` does, so `extra_content` does the same.

This also closes a gap in the response record. `internal/session/history_test.go:281` (`TestSetResponse_PersistsNativePayload`) guards a *"training-reflow gap … otherwise a session exported for training cannot reconstruct a request that replays byte for byte"*, but the `llm_response` writer emitted only `id`/`name`/`arguments`, so the metadata that produced the turn was dropped. `toolCallsForJSON` and the `SetResponse` projection now carry it in both `llm_request` and `llm_response`. To be precise about scope: resume does not need this — `ResumeState` is a file-fingerprint checkpoint index — it is session export and diagnostics that do, exactly as for `native_payload`.

### What changed

- `mapOpenAIResponse` reads `extra_content` from each tool call's raw JSON. The SDK's tool-call union exposes no `ExtraFields` map, so the raw payload is the only place it survives.
- Streaming captures it from `choice.Delta.ToolCalls[i].JSON.ExtraFields`, keyed by choice and clamped tool index. It has to be taken from the deltas because `ChatCompletionAccumulator` synthesises its result field by field and never sets raw JSON. It also has to be captured *before* the `reasoning_content` lookup's early `continue` — Gemini sends tool calls without `reasoning_content`, so a capture placed after it would never run. There is a test that fails if that ordering is reversed.
- `buildOpenAIParams` replays it via `SetExtraFields` on the per-tool-call param, the mechanism `reasoning_content` already uses (#805).
- `withProviderErrorBody` appends the provider body when the SDK left the error without one. The trigger is "an `*openai.Error` whose `RawJSON()` is empty", not "is this Gemini", so no provider-specific transport behaviour is introduced. The response body is restored after reading so later consumers still see it whole, only the diagnostic text is bounded (64 KiB), and control bytes are stripped so a body cannot rewrite the terminal. Applied to the chat-completions, streaming and Responses API exits.
- `copyMessages` deep-copies `ExtraContent`. Its contract is a deep copy; every `ToolCall` field used to be a string, so an element copy was enough. Adding a byte slice broke that quietly, and the new test catches it.

Providers that never send `extra_content` produce a byte-identical request to before.

## Lineage

This is the same defect #1070 fixed for reasoning and thinking state, showing up one level lower. That change gave assistant turns an opaque `Native` replay payload after `reasoning_content` (#805) and Anthropic thinking signatures (#811) were found to be dropped when a tool-call turn was replayed. `extra_content` sits on the tool call itself rather than the message, so it falls outside what the SDK's own response-to-request converters carry, and survived that pass.

It was first reported as #947 by @vincekent in August, with the root cause already correct and `SetExtraFields` already named as the mechanism.

One deliberate divergence from #947, which asked that opaque fields stay out of session logs: this PR persists `extra_content` in both JSONL projections. #947 was filed on 2026-08-16 and #1070 merged on 2026-08-28, introducing `nativeTurnForJSON` and `TestSetResponse_PersistsNativePayload` — which require the analogous Anthropic thinking signature to be written to the `llm_response` record so a session exported for training can reconstruct a byte-identical request. The request was reasonable when written; the project's position on opaque replay state moved after it. The wire format is unchanged either way, and `ToolCall.ExtraContent` stays `json:"-"` so nothing is written incidentally.

A second divergence: #947 asks for every unknown tool-call field to be preserved, and this PR keeps only the named `extra_content` key. So it references #947 rather than closing it, pending the capture discussion below.

## Credit

There is good prior work on these two issues from @basil-k-aji-dev, @xiaodu55, @AyushDubey23 and @fkdosilovic, and this PR builds on it rather than around it. The `json:"-"` choice and the diagnosis that the streaming accumulator drops raw JSON, the named `extra_content` key and the negative-index clamping, the recognition that #1357 and #1042 are one problem, and the provider-neutral shape of the error enrichment (including trimming `RawJSON()` before deciding to enrich) all come from their PRs. @subaru-ye's streaming reproduction and @pdrivom's AI Studio verification were directly useful too. Thanks to all of them.

## How Has This Been Tested?

- [x] Manual testing (below)
- [x] `make test` — Passed

`go test` passes on `internal/llm`, `internal/session`, `internal/llmloop`, `internal/agent`, `internal/scan`, `internal/mcp`, `internal/viewer` and `cmd/opencodereview`, and `go test -race` passes on `internal/llm` and `internal/session`. I have not run the full suite to completion on this machine; `TestRangeDiffDetectsRename` in `internal/diff` fails here for me both with and without this change, because of a stray directory in my working tree, so I have left it out rather than report a pass I did not see.

**Live, against Vertex AI** (`aiplatform.googleapis.com/.../endpoints/openapi/chat/completions`, `google/gemini-3.8-flash`), same repository and diff both times — a small Python file with a planted SQL injection and a `ZeroDivisionError`:

| | v1.12.5 release binary | this branch |
|---|---|---|
| tool round trip | `400 Bad Request` with an empty body | completed, 9 requests |
| result | `all 1 file review(s) failed` | both planted bugs found, with fixes |
| `thought_signature` failures | 1, fatal | 0 |

Verification note: previous reports on this bug were against AI Studio (`generativelanguage.googleapis.com`). This is the Vertex surface.

The error change also proved itself unintentionally in that run. The endpoint rate-limited me partway through, and instead of a bare `429 Too Many Requests` the reason came through:

```
429 Too Many Requests : [{
  "error": {
    "code": 429,
    "message": "Resource exhausted. Please try again later. ...",
    "status": "RESOURCE_EXHAUSTED"
  }
}]
```

That is the array-wrapped shape from #1042, surfaced from a real response rather than a fixture.

New tests, all deterministic and offline: non-streaming replay; streaming with the metadata on the opening chunk and on a later chunk; a streamed tool call carrying no `reasoning_content` (the ordering trap); negative delta index clamping; interleaved tool calls keeping their own metadata; a later chunk not erasing an earlier value; absent field leaving the wire format unchanged; persistence in `llm_request` and `llm_response`; the deep-copy contract; array-wrapped error surfacing on both the non-streaming and streaming exits; no duplication when the SDK already extracted a payload; the 64 KiB bound; the response body still readable afterwards; and terminal control bytes stripped.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## Checklist

- [x] New and existing unit tests pass locally with my changes (see the note above on the full suite)
- [ ] I have updated the documentation accordingly (if applicable) — no user-facing surface changed
- [x] I have signed the CLA

AI/LLM disclosure: I used Claude (Opus 5) to help refine this fix, with OpenAI Codex for an independent review pass. I reviewed and understand every line, ran the live Vertex verification myself, and will answer review comments myself.

One thing I deliberately did not change: surfacing a provider error body puts that body into terminal output and the session JSONL. That is the point of #1042, and it is not new behaviour — `apierror.Error()` already formats the raw payload into every error for providers that use the standard `{"error": ...}` shape. This only brings array-wrapping providers to the same footing. Worth knowing when sharing session files.

## Related Issues

Fixes #1357
Fixes #1042
Refs #947


internal/llm/client.go
```diff
@@ -862,6 +901,77 @@ func (c *OpenAIClient) buildOpenAIParams(model string, req ChatRequest) openai.C
...
+	var apiErr *openai.Error
+	if !errors.As(err, &apiErr) || apiErr.RawJSON() != "" || apiErr.Response == nil || apiErr.Response.Body == nil {
+		return err
+	}
+	body, readErr := io.ReadAll(apiErr.Response.Body)
+	_ = apiErr.Response.Body.Close()
+	apiErr.Response.Body = io.NopCloser(bytes.NewReader(body))
+	if readErr != nil || len(bytes.TrimSpace(body)) == 0 {
+		return err
+	}
```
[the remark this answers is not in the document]
Moon-Wrecker: Revisiting this properly, because chasing the concern down found three real things — none of them the one flagged, but the flag is what led to them.

**On the suggestion itself.** It is behaviourally identical to the current code: both branches assign the same `io.NopCloser(bytes.NewReader(body))` from the same partial `body`, so moving the assignment inside the `if` changes nothing observable.

**On the concern underneath it.** I tried the stronger form — partial bytes followed by a reader replaying the failure — and backed it out. `DumpResponse` discards its own error (`apierror.go:53`), so a failing reader makes it return nil: a dump that would have shown the partial bytes shows nothing instead, which is worse in exactly the diagnostic path this code serves. It also makes a transient failure permanent. And the branch is unreachable: `requestconfig.go:643-652` reads the non-2xx body, returns immediately if that read fails, and only then buffers it and builds the error, so what arrives here is always an in-memory reader.

**What the concern did surface**, though, is that mutating a shared `*openai.Error` has a consequence the flag did not name: **the function was not idempotent.** Restoring the body — which is the right thing to do — is exactly what lets a second pass read it again and append the payload twice. Reproduced as 1 → 2 → 3 copies. Not reachable today (`CompletionsWithCtx` returns at `client.go:663` for streaming, so each of the three call sites fires once), but a latent trap, and a merge or a moved call site would hit it silently. Enrichment now returns a private `enrichedError` carrying `Unwrap()`, and the helper returns early when it finds one, so it survives arbitrary later `%w` wrapping. `TestWithProviderErrorBody_IsIdempotent` applies it four times and asserts one copy.

Two more, both found while tracing the same code and both fixed:

- **C1 controls were not filtered.** The old predicate kept every rune `>= 0x20` except DEL, which lets U+0080–U+009F through — including U+009B (CSI) and U+0090 (DCS), which a terminal decoding C1 obeys just as it obeys the ESC-prefixed forms. That contradicted the comment above it. Now `unicode.IsControl`, with `\n` and `\t` kept so pretty-printed JSON still reads well.
- **`"null"` counted as a payload.** For a body like `{"error":null,"message":"quota exhausted..."}` gjson yields the literal `null`, which is non-empty, so the guard concluded the SDK had a payload and the real message was dropped — the user saw `400 Bad Request null`. Empty and `null` are both treated as absent now.

Also cut the oversized-body path to slice before building the string, so a huge body is no longer copied whole just to be truncated.

Tests added for each: `TestWithProviderErrorBody_IsIdempotent`, `TestOpenAIClient_SurfacesBodyWhenSDKExtractedNull`, and the terminal-control test extended to assert U+009B and U+0090 alongside ESC and BEL.

Moon-Wrecker:
Hi @chaojixinren , @wu21-web , @austinuni60-netizen , @Qiyuanqiii , @basil-k-aji-dev ,  can you kindly review this PR.

Qiyuanqiii:
No big deal. I’m planning to go through all open PRs starting this afternoon

wu21-web:
Please do not tag individual users for reviews. Somebody will review it for you. I will review for you soon, but please use ocr first.

wu21-web:
@Moon-Wrecker Before reviewing, I suggest splitting changes up into different PRs that are small and focused. Big chunks make it hard to review. Be aware that this closes 3 issues on merge, I don't really recommend that ..

Qiyuanqiii (commented):
One scope question: how do you see #1393 relative to #1233?
#1393 fixes the concrete extra_content / thought-signature failure, while #1233 generalizes the same area to arbitrary opaque tool-call fields. Do you intend #1393 to supersede #1233 for #947, or do you still see #1233 as a useful follow-up for the generic case?
Just trying to understand the intended ownership between the two approaches.

Qiyuanqiii (commented):
I think there are really two separate questions here.
#1233 and #1393 substantially overlap on the concrete tool-call metadata replay problem, but they are not equivalent implementations. #1233 takes the broader generic ExtraFields approach, while #1393 deliberately narrows replay to extra_content and additionally addresses provider-error diagnostics and session persistence.
Because that is a genuine architectural choice rather than simply one implementation being a strict superset of the other, I don't think we as reviewers should unilaterally decide which approach should replace the other. That seems better left to the maintainers to resolve.
Separately, whichever technical direction is chosen, I think the contribution lineage should remain clear given that #1233 predates this PR and has continued to receive active implementation and independent verification.
So from my side, I'd prefer to hear both how you see #1393 relative to #1233 and what direction the maintainers want for the underlying abstraction before treating either PR as superseding the other.

AyushDubey23:
Thanks @Moon-Wrecker for acknowledging the prior work and design from #1371 in the credit section.

I agree with @wu21-web and @Qiyuanqiii regarding scope, PR size, and contribution lineage. Bundling three distinct issues (#1357, #1042, #947) along with session persistence refactors into a 700+ line PR creates a large blast radius and complicates review.

For the concrete bug in #1357 and #1042, PR #1371 is already open, tested, and strictly focused on solving the `extra_content` replay and provider error surfacing without expanding the scope. 

Landing the focused fix in #1371 resolves the immediate blocker for Gemini users cleanly, while the broader architectural discussions around #947 and #1233 can proceed on their own timeline.

Qiyuanqiii (commented):
One principle I think is useful here is to prefer the smallest complete fix that can be explained, validated, and attributed independently, while leaving broader architectural changes their own discussion space.
With #1233, #1371, and #1393 now overlapping in different ways, I think it would be healthier to make that boundary explicit and let the maintainers decide how the larger architectural pieces should converge.

internal/llm/provider_error_test.go
```diff
@@ -0,0 +1,214 @@
...
+	if once != 1 {
+		t.Fatalf("first enrichment produced %d copies, want 1", once)
+	}
+	for i := 0; i < 3; i++ {
+		err = withProviderErrorBody(err)
+		if got := strings.Count(err.Error(), "only once please"); got != 1 {
+			t.Fatalf("after %d further enrichments the payload appears %d times, want 1", i+1, got)
+		}
+	}
+}
```
lizhengfeng101, line 214: While testing this locally I noticed nothing drives the Responses client, so the `withProviderErrorBody` hook you wired into that path isn't guarded — I confirmed that dropping `responses_client.go:134` keeps the whole suite green. A sibling test closes the gap:

```suggestion
}

// TestOpenAIResponsesClient_SurfacesArrayWrappedErrorBody covers the Responses
// API exit. It routes its request error through withProviderErrorBody too, but
// no other test drives that client, so this guards the wiring against removal.
func TestOpenAIResponsesClient_SurfacesArrayWrappedErrorBody(t *testing.T) {
	const body = `[{"error":{"code":400,"message":"responses request rejected by gateway","status":"INVALID_ARGUMENT"}}]`
	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		w.Header().Set("Content-Type", "application/json")
		w.WriteHeader(http.StatusBadRequest)
		if _, err := w.Write([]byte(body)); err != nil {
			t.Errorf("write error body: %v", err)
		}
	}))
	defer server.Close()

	client := NewOpenAIResponsesClient(ClientConfig{
		URL:    server.URL + "/v1",
		APIKey: "test-key",
		Model:  "custom-model",
	})

	_, err := client.CompletionsWithCtx(context.Background(), ChatRequest{
		Messages: []Message{{Role: "user", Content: "ping"}},
	})
	if err == nil {
		t.Fatal("expected a provider error, got nil")
	}
	if !strings.Contains(err.Error(), "responses request rejected by gateway") {
		t.Fatalf("Responses API did not surface the array-wrapped body: %v", err)
	}
}
```

AyushDubey23:
100% agree with this principle, @Qiyuanqiii. “Prefer the smallest complete fix that can be explained, validated, and attributed independently” makes a lot of sense here.

Just to be completely transparent about the lineage here:

Right after I opened #1371 to address #1357, @Moon-Wrecker asked on that issue whether my fix also covered #1042. Shortly after that, #1393 was opened, lifting the exact implementation details from #1371, including the negative delta index clamping, the provider neutral error body enrichment for array wrapped Gemini payloads, and the 64KB log bounding, and combining them with @basil-k-aji-dev’s work from #1233 into one mega PR attempting to close all 3 issues at once.

Subsuming active PRs from other contributors into a 700+ line PR without collaboration isn’t really great open source etiquette, and as @wu21-web also pointed out, it makes the review process much harder.

I think the cleanest way forward, while keeping attribution clear and the changes easy to review, would be:

1. **#1233** by @basil-k-aji-dev merges to solve the generic tool call metadata replay issue (#947).
2. **#1371** by @AyushDubey23 handles the provider error body enrichment issue (#1042).

Both PRs are already tested and independently scoped, and this avoids the unnecessary blast radius of #1393 while keeping each fix focused and properly attributable.

lizhengfeng101:
@Moon-Wrecker As a maintainer, I still recommend keeping these changes within a single PR; since they significantly impact the overall review, consolidating them makes it easier for me to conduct an end-to-end evaluation. From this perspective, I wonder if anyone holds a different view. cc @wu21-web @Qiyuanqiii @AyushDubey23

AyushDubey23:
Thanks for looping me in, @lizhengfeng101.

I get the end to end evaluation point. #1357 and #1042 are closely connected when testing live Gemini/Vertex endpoints, since the enriched error body makes tool call failures much easier to debug. That’s why I paired both fixes in #1371.

I think we can address both concerns pretty cleanly:

1. **#1371 already covers both fixes** in ~270 lines, mostly tests, without the 700+ line session persistence changes.

2. **If a consolidated PR is preferred**, I’m happy to collaborate. I’d just ask that the existing contributions from @basil-k-aji-dev, @subaru-ye and myself are preserved through proper git attribution such as `Co-authored-by:` trailers, along with tests for all affected paths.

Happy to go with whichever direction the maintainers prefer.

lizhengfeng101:
> Thanks for looping me in, @lizhengfeng101.
> 
> I get the end to end evaluation point. #1357 and #1042 are closely connected when testing live Gemini/Vertex endpoints, since the enriched error body makes tool call failures much easier to debug. That’s why I paired both fixes in #1371.
> 
> I think we can address both concerns pretty cleanly:
> 
> 1. **[fix(llm): preserve tool_call extra_content and enrich error body for OpenAI providers #1371](https://github.com/alibaba/open-code-review/pull/1371) already covers both fixes** in ~270 lines, mostly tests, without the 700+ line session persistence changes.
> 2. **If a consolidated PR is preferred**, I’m happy to collaborate. I’d just ask that the existing contributions from @basil-k-aji-dev, @subaru-ye and myself are preserved through proper git attribution such as `Co-authored-by:` trailers, along with tests for all affected paths.
> 
> Happy to go with whichever direction the maintainers prefer.

I strongly agree with recognizing the high-quality contributions of existing developers through Git attribution.

Qiyuanqiii:
> @Moon-Wrecker As a maintainer, I still recommend keeping these changes within a single PR; since they significantly impact the overall review, consolidating them makes it easier for me to conduct an end-to-end evaluation. From this perspective, I wonder if anyone holds a different view. cc @wu21-web @Qiyuanqiii @AyushDubey23 

Thanks, that clarifies the PR-organization question for me. I'm fine with treating #1393 as the consolidated integration PR if that makes the end-to-end review easier.
I think one architectural question is still unresolved, though: which tool-call metadata abstraction should the consolidated implementation use?
#1233 preserves arbitrary unmodeled provider fields through ExtraFields, while #1393 currently preserves only the known extra_content field through ExtraContent. There is also a related persistence difference: #1233 keeps that opaque state out of the session JSONL, while #1393 deliberately persists it for replay/export.
Keeping the work in one PR doesn't by itself resolve those choices. Before treating #1233 as superseded, could you clarify which behavior you want the project to standardize on?
I'm also open to a combined direction—for example, using the generic replay abstraction from #1233 while separately deciding which opaque fields should be persisted—but I think that choice should be explicit.

wu21-web:
Can we hold on until this evening? Don't merge before JST 21:00 (UTC+9)
I want to take a look.

Qiyuanqiii (commented):
Setting aside the architectural direction / route-selection discussion above, my final implementation review is:
I don't see an obvious correctness blocker in the current revision (aa200a5).
I re-checked the non-streaming and streaming extra_content replay paths, later-chunk handling, negative tool indices, multiple tool calls, null/absent metadata, session persistence and deep-copying, provider error enrichment, idempotence, response-body restoration, control-character filtering, and the Responses API path added in the latest commit. The current CI and CodeQL results are also green.
The main remaining implementation concern I have is token accounting. messageTokens() currently counts visible content plus Native.EstimatedTokens(), but not the newly replayed ToolCall.ExtraContent. Since that metadata is sent again on subsequent requests, longer tool-call conversations can be underestimated by the local compression/context-budget logic.
I don't think this is strong enough to block the PR on its own, since the existing estimator is already approximate, but it would be worth either accounting for this replay state or documenting why the current approximation is considered acceptable.
Two smaller hardening points are also worth noting, but I would not block on them:
withProviderErrorBody makes a second full copy of the SDK-buffered error body before applying the 64 KiB diagnostic limit.
NewToolCallMessage shallow-copies ExtraContent; I couldn't find a current mutation path that makes this observable, so this appears safe under the current immutable-use pattern.
So, purely from an implementation perspective, the current revision looks technically coherent to me. My remaining reservation is the architectural direction discussed above

internal/llm/provider_error_test.go
Moon-Wrecker: Thanks — you were right, and I verified it rather than taking it on trust. Reproducing the state before your commit, with `responses_client.go:134` reverted to `return nil, err` and this test absent:

```
ok  github.com/alibaba/open-code-review/internal/llm  7.350s
```

Green. So that exit had no coverage at all and could have been deleted silently. With your test present the same removal fails exactly as it should:

```
Responses API did not surface the array-wrapped body:
POST ".../v1/responses": 400 Bad Request
```

— the unenriched form, trailing space and all.

That was a genuine hole in the PR. I wired the Responses exit because `responses_client.go:132` returned the SDK error directly, and then only tested the two chat-completions exits. Appreciate you writing the sibling test rather than just flagging it.

internal/session/history.go
```diff
@@ -391,36 +392,76 @@ func copyMessages(msgs []llm.Message) []llm.Message {
...
 	type msg struct {
-		Role          string         `json:"role"`
-		Content       any            `json:"content"`
-		ToolCallID    string         `json:"tool_call_id,omitempty"`
-		ToolCalls     []llm.ToolCall `json:"tool_calls,omitempty"`
-		NativePayload any            `json:"native_payload,omitempty"`
+		Role          string           `json:"role"`
+		Content       any              `json:"content"`
+		ToolCallID    string           `json:"tool_call_id,omitempty"`
+		ToolCalls     []map[string]any `json:"tool_calls,omitempty"`
```
lizhengfeng101: For backward stability of the session JSONL, could this project into an ordered struct (`id`/`type`/`function`, with `extra_content` `omitempty`) instead of `map[string]any`? A map sorts keys, so this silently reorders `tool_calls` in every `llm_request` record; an ordered struct keeps the field order byte-identical while still persisting `extra_content`.

Moon-Wrecker: Good catch @lizhengfeng101  done in `aad0b36`. It was slightly worse than the map sorting the outer keys: the nested `function` object was reordered too. For a tool call with no `extra_content`:

```
before this PR : {"id":"call_1","type":"function","function":{"name":"file_read","arguments":"{}"}}
map projection : {"function":{"arguments":"{}","name":"file_read"},"id":"call_1","type":"function"}
```

`toolCallsForJSON` now returns an ordered `toolCallJSON` — `id`, `type`, `function` (reusing `llm.FunctionCall`, so `name` then `arguments`), then `extra_content` with `omitempty`. Records without `extra_content` are byte-identical to what `[]llm.ToolCall` produced before, and the metadata is appended last when present.

`TestToolCallsForJSON_KeepsFieldOrder` pins both cases. The no-metadata case is checked against `json.Marshal([]llm.ToolCall{tc})` itself — since `ExtraContent` is `json:"-"`, that is the previous format by definition, so the test can't drift from it.

I deliberately left the `llm_response` projection in `SetResponse` as a map. That one was already a `map[string]any` of `id`/`name`/`arguments` before this PR, so its keys were already sorted and adding `extra_content` leaves metadata-free records unchanged. Converting it to a struct would itself reorder it — the regression you are guarding against, just on the other record type.

Moon-Wrecker:
@AyushDubey23 — taking the process point first, because you're right about it.

I should have said on #1357 that I intended to open a consolidated PR, and given you and @basil-k-aji-dev the chance to object or to collaborate, before putting up something that overlaps your open work. I didn't, and a 700-line PR landing on top of active PRs is a bad way for you to find that out. That one's on me.

On attribution: agreed, and I'd like to go further than you asked. If this is the PR that lands, I'll add `Co-authored-by:` trailers for:

- **@basil-k-aji-dev** — `json:"-"`, and the diagnosis that the accumulator never sets raw JSON so the value can only come from the deltas
- **@subaru-ye** — the streaming reproduction that surfaced that in the first place
- **@AyushDubey23** — the streaming capture with the null guard, and pairing the error-body fix with the replay fix
- **@xiaodu55** — the named `extra_content` key and the negative index clamping
- **@fkdosilovic** — the provider-neutral enrichment trigger, including `strings.TrimSpace(apiErr.RawJSON())`, which I took from #1043 and which is better than what I had

On scope, that genuinely isn't mine to decide. @wu21-web has asked for smaller, focused PRs; @lizhengfeng101 has asked to keep this consolidated for end-to-end review. I'll do whichever they settle on.

If they prefer #1233 and #1371, I'm happy to close this and move the pieces that aren't in either — the session persistence, the idempotence guard, the C1 control filtering, the `null` RawJSON case, and @lizhengfeng101's Responses API test — into small follow-ups against whichever merges. I'd rather the bug got fixed than that mine is the PR that fixes it.

Apart from the field-order fix @lizhengfeng101 asked for, I won't push anything further until @wu21-web has had a look.

AyushDubey23:
appreciate you acknowledging this and taking accountability, @Moon-Wrecker. means a lot, and thanks for making sure everyone gets proper credit.

totally agree with letting @wu21-web and @lizhengfeng101 decide whichever integration path works best for the project. happy to collaborate and get the fix landed!

wu21-web:
Please skip my review as it has passed the due time.

basil-k-aji-dev:
@Qiyuanqiii — answering your question directly, since it's about #1233's abstraction.

**On scope and which PR lands, I have no stake.** @lizhengfeng101 and @wu21-web should pick whatever makes the end-to-end review easiest. If that's #1393, I'll close #1233 — I'd rather the Gemini bug were fixed this week than that mine is the PR that fixes it. @Moon-Wrecker already owned the process point and the credit section is more generous than I'd have asked for, so no complaint from me there.

**On the abstraction, I do think the generic form is the right default, and there's evidence in this thread rather than just a preference.**

`extra_content` is not a spec field. It is one vendor's name for "metadata you must echo back", and the class of behaviour is not unique to it — the same providers already ship `reasoning_content` as an unmodelled delta extension, and the SDK's own accumulator carries a `// Clamp negative indices to 0 since the API may send -1 for single tool calls` comment for a second undocumented deviation. Keying on one literal string means the next vendor that picks a different name is a code change rather than a no-op.

The concrete case is already in @subaru-ye's negative-index payload above. It carries two opaque fields:

```json
{"index":-1,"id":"call_1","type":"function",
 "function":{"name":"file_read","arguments":"{}"},
 "extra_content":{"google":{"thought_signature":"sig"}},
 "provider_flag":true}
```

A named-key capture replays `extra_content` and silently drops `provider_flag`. That is not a hypothetical I constructed — subaru-ye tracked both fields when testing, because both are things a provider could reject the following request over.

The safety property people usually worry about with generic capture is a provider rewriting a tool call's identity. That is handled independently of the key name: `reservedToolCallFields` blocks `id`, `type` and `function`, so those stay owned by OCR on both the streaming and non-streaming paths. Taking the fields from `JSON.ExtraFields` rather than from the raw JSON also means the stream's own `index` framing is excluded for free, which a raw-JSON capture has to strip by hand.

**On persistence I think #1393 is right and #1233 is wrong, and I would drop my behaviour.** I made it `json:"-"` because #947 asked for opaque fields to stay out of session logs. @Moon-Wrecker's reasoning in the PR body is better: #1070 landed `nativeTurnForJSON` after #947 was filed, and it requires the analogous Anthropic thinking signature in the `llm_response` record so that an exported session can reconstruct a byte-identical request. Excluding the OpenAI-side equivalent makes the two families inconsistent for no benefit. The wire format is unchanged either way.

So the combination you floated is the one I would argue for: **generic replay abstraction, persisted.** The vendor's field name stays out of the type, and the session export stays reconstructable.

Whatever is decided, I am happy to rebase #1233 onto it or to have the pieces lifted — whichever is less work for the reviewers.

One practical note on the credit section, since a mistyped trailer credits nobody: mine is

```
Co-authored-by: basil-k-aji-dev <70605804+basil-k-aji-dev@users.noreply.github.com>
```

AyushDubey23:
agree with @basil-k-aji-dev's points on generic replay and persistence. it feels like a solid and future proof approach.

following basil's lead, my trailer is:

`Co-authored-by: Ayush Dubey <ayushdubey3740@gmail.com>`

thanks everyone for collaborating on this!

Moon-Wrecker:
@Qiyuanqiii, answering your question from yesterday about which behaviour the project should standardise on. Having worked through it, I agree with @basil-k-aji-dev: capture every unmodelled tool-call field, and persist it. That changes my position on capture.

What changed my mind was testing the risk I'd have raised against generic capture, on the provider this is actually about. I took one real tool call from Vertex (`google/gemini-3.8-flash`) and replayed it five ways:

| replayed tool call | Vertex |
|---|---|
| `id`/`type`/`function` + `extra_content` only (this PR today) | 200 |
| every field Vertex returned, unchanged | 200 |
| plus `"index": 0` | 200 |
| plus an unknown `"provider_flag": true` | 200 |
| `extra_content` dropped | 400, `missing a thought_signature` |

Vertex returns only `extra_content` beyond the standard fields today, and accepts extra fields echoed back. A stricter gateway could still reject an echoed field, but that doesn't happen here, and the reserved-field filter in #1233 already keeps `id`/`type`/`function` owned by OCR. The named key, on the other hand, means the next vendor with a different field name is a code change. So generic capture is the better default.

This also showed that this PR was overclaiming. #947 asks for all unknown fields to be preserved, and this PR keeps only `extra_content`, so `Fixes #947` was wrong. I've changed it to `Refs #947` and corrected my comment on the issue.

Which PR carries it is @lizhengfeng101's call. I see two options:

1. #1233 lands the capture and closes #947, and I update this PR on top of it with what #1233 doesn't have: persisting the fields in both JSONL records (basil now agrees with that part), the #1042 error body, deep-copying the captured fields, and the Responses API test.
2. I bring #1233's capture into this PR with basil's `Co-authored-by`, so everything lands in one PR.

I'd prefer the first. The approach is basil's, #1233 was written against #947, and Qiyuanqiii has already found it technically sound. But either works for me, and I won't touch the code until you've picked.

@Qiyuanqiii, your token-accounting point applies to both designs, so I'll deal with it once the route is settled.

On credit: I have addresses from @basil-k-aji-dev, @AyushDubey23 and @subaru-ye. @xiaodu55 and @fkdosilovic, would you like co-author credit as well? If so, please post the address you want used; a trailer with an address GitHub doesn't recognise doesn't credit anyone.

CI on `aad0b36` is waiting for workflow approval.

Moon-Wrecker:
Co-author lines for the squash commit, collected in one place so they can be pasted in at merge time:

```
Co-authored-by: basil-k-aji-dev <70605804+basil-k-aji-dev@users.noreply.github.com>
Co-authored-by: subaru-ye <178248423+subaru-ye@users.noreply.github.com>
Co-authored-by: Ayush Dubey <ayushdubey3740@gmail.com>
Co-authored-by: Tao Xin <wu2196674@icloud.com>
```

These go at the very end of the squash message, after a blank line. Each is the address its owner posted in this discussion, and I've checked that each one links to the right GitHub account, so none of them will show up as plain text.

lizhengfeng101's own line is added by GitHub automatically, since he authored `aa200a5`. xiaodu55 and fkdosilovic haven't replied yet; if they do, I'll add their lines here.

basil-k-aji-dev:
@Moon-Wrecker thank you for testing it rather than taking my argument on faith — the five-way replay table is more than I had, and the result that Vertex accepts echoed unknown fields is the part I could not check without an endpoint.

Two corrections to your option-1 list, both in your favour.

**Deep-copying is already in #1233**, as of 058d0cb this morning. @Qiyuanqiii raised it there: `ExtraFields` is a map, and both `NewToolCallMessage` and `session.copyMessages` copied `ToolCall` values with a plain slice copy, so every copy shared the one map and the byte slices inside it. `CloneToolCalls` now copies the map and the values; two tests pin it and both fail against the previous revision. So that item comes off the list of things #1233 does not have.

**Persistence** I have not changed, and I would rather you carried it than that I bolted it on. Your reasoning in the PR body — #1070 landing `nativeTurnForJSON` after #947 was filed, so the Anthropic thinking signature has to survive into `llm_response` for a byte-identical replay — is a better reading of the project's current position than the one I followed, which was #947's original request. It is your finding; it should be your commit.

Option 1 works for me, and so does option 2. I have no stake in which PR carries it: @lizhengfeng101 and @Qiyuanqiii should pick whichever is easier to review end to end, and if that is option 2 then bring the capture across and I will close #1233 without ceremony.

On credit, your three trailers are right and the address for me is the one you have. Worth repeating your own point for @xiaodu55 and @fkdosilovic: the address has to be one GitHub already knows, or the line renders as plain text and credits nobody — I have five commits in another repository that prove it, carrying a trailer whose address resolves to no account at all.

Moon-Wrecker:
@basil-k-aji-dev thanks, and for both corrections. I'd missed that the deep-copy landed in #1233 this morning.

Given what you've said, I'd like to switch my suggestion to option 2. You're fine with either and would close #1233 if the capture comes across, and @lizhengfeng101 asked earlier to keep this in one PR so it can be reviewed end to end. So the simplest path seems to be: I bring #1233's current commits into this PR by cherry-picking them, so your authorship and subaru-ye's trailer come across unchanged, including `CloneToolCalls`, and move the persistence code over to `ExtraFields`. This PR then covers #947, #1042 and #1357 together, and #1233 can close.

@lizhengfeng101, if that works for you I'll go ahead. I'm holding the branch at `aad0b36` until you confirm.

Moon-Wrecker:
@NanaseInori review

NanaseInori (approved):

internal/llm/client.go
```diff
@@ -862,6 +902,109 @@ func (c *OpenAIClient) buildOpenAIParams(model string, req ChatRequest) openai.C
...
+func limitErrorBodyForLog(raw []byte) string {
+	raw = bytes.TrimSpace(raw)
+	var suffix string
+	if len(raw) > maxErrorBodyBytes {
+		// Cut the bytes before building the string so an oversized body is never
+		// copied whole just to be discarded.
+		raw = raw[:maxErrorBodyBytes]
+		suffix = "... (truncated)"
+	}
+	trimmed := strings.ToValidUTF8(string(raw), "�")
```
wu21-web:
```suggestion
	trimmed := strings.ToValidUTF8(string(raw), "")
```
The function first caps the provider response body at 64 KiB, but then replaces invalid UTF-8 with that what character??

wu21-web (commented):
My co-author line is 
`Co-authored-by: Tao Xin <wu2196674@icloud.com>`
if you need it.

internal/llm/client.go
wu21-web: Sorry, I might be over informal, but I cannot type that.

Qiyuanqiii:
6

internal/llm/client.go
Moon-Wrecker: Good catch, and you're right about the character: it was a literal U+FFFD in the source. I meant the escaped form, and it went in as the character itself. Dropping invalid bytes is better anyway, since they carry nothing useful in an error body, and a character cut in half at the 64 KiB cap now just disappears instead of turning into a replacement character.

Applied as suggested in `05a7a2d`, with a test for the cut-in-half case. It was the only non-ASCII character outside comments. I've added your line to the co-author list. That fix is the only change; the #1233 integration still waits on lizhengfeng101.

Moon-Wrecker:
@NanaseInori review

wu21-web (approved):
After make test passes, I think we are ready to merge.

Qiyuanqiii (approved):
Fixes #947 → Refs #947 is definitely the right change.
At this point, though, reviewing this PR feels less like reviewing a PR and more like watching the Avengers assemble.

Qiyuanqiii:
@NanaseInori review

NanaseInori (approved):
Automated review summary for current HEAD 05a7a2d:

Correctness: No blocking issues found.

Tool-call replay: Non-streaming and streaming extra_content paths are covered, including later chunks, negative indices, and multiple tool calls.

Persistence: Replay metadata is preserved in both request and response session records, with deep-copy protection in the session path.

Provider errors: Array-wrapped error bodies are surfaced with bounded output, response-body restoration, idempotence, and control-character filtering.

UTF-8 handling: The latest change correctly drops invalid/incomplete UTF-8 at the diagnostic size boundary and includes regression coverage.

Compatibility: Metadata-free tool calls preserve the previous serialized representation.

CI: Required checks are passing.

Non-blocking observations:

Replayed tool-call metadata is still not included in local token estimation.

The current implementation handles the concrete extra_content case; the broader generic requirement remains tracked by #947, so Refs #947 is appropriate.

Review result: APPROVE.

No remaining implementation issue identified that should block merge.

wu21-web:
Awaiting for CI results.

basil-k-aji-dev:
Two housekeeping notes now that this is heading for merge, neither of which needs a reply.

**#1233 will close as soon as this lands.** It still carries approvals from @Qiyuanqiii and @NanaseInori, so I would rather it not sit there looking like a competing option once this is in. I am not asking for it to be reconsidered — my position from earlier stands: whichever PR is easier to review end to end should land, and @lizhengfeng101 asked for this to be one PR. Say the word after merge and I will close it, or close it yourselves, whichever is less work.

**The co-author list needs pasting at squash time.** @Moon-Wrecker collected it [earlier in this thread](https://github.com/alibaba/open-code-review/pull/1393#issuecomment-5741676010) and noted it has to go into the squash message by hand. Flagging it only because GitHub's squash box does not carry trailers over from branch commits, and it is an easy thing to lose at the last step. The branch currently has @wu21-web's line on a commit; the others are still only in that comment:

```
Co-authored-by: basil-k-aji-dev <70605804+basil-k-aji-dev@users.noreply.github.com>
Co-authored-by: subaru-ye <178248423+subaru-ye@users.noreply.github.com>
Co-authored-by: Ayush Dubey <ayushdubey3740@gmail.com>
Co-authored-by: Tao Xin <wu2196674@icloud.com>
```

@lizhengfeng101's own line is added automatically since he authored `aa200a5`.

Worth it for @subaru-ye in particular — they ran the Windows race-detector verification and the SSE reproduction that found the original streaming gap, across several revisions, without having a PR of their own here.

Moon-Wrecker:
`make test` passes on 05a7a2d — 2378 tests across the suite, including the 16 added in this PR.

On CI: the runs on this commit are still sitting at "action_required". Workflow runs from a fork need a maintainer to approve them before they start, so there are no results to wait for yet.

@lizhengfeng101 I am also withdrawing my earlier suggestion to pull #1233's generic capture into this PR. With the approvals on this scope, and @Qiyuanqiii confirming that `Refs #947` is the right framing, restarting the review now would cost more than it adds. Nothing else is outstanding from my side.

Moon-Wrecker:
All review conversations on this PR are resolved now, so nothing is open on my side.

Where 05a7a2d stands: three approvals (wu21-web, Qiyuanqiii, NanaseInori), no unresolved threads, CLA signed, and `make test` green across the suite. The remaining step is a maintainer approving the workflow runs so CI can start.

Qiyuanqiii:
Please be patient and wait a while，the maintainer is busy with other commitments today.

Moon-Wrecker:
Hi @lizhengfeng101 , can you please approve the workflows as well as the PR.

lizhengfeng101 (approved):
LGTM

AyushDubey23:
Hey @Moon-Wrecker, thanks again for getting this merged and for acknowledging the contributions throughout the discussion.

I noticed that the final merged commit doesn't appear to include the Co-authored-by: Ayush Dubey <ayushdubey3740@gmail.com> trailer that we had agreed on earlier.

Since my contributions to the streaming capture and error-body fix were included in the final implementation, could you please add the co-author attribution if it's still possible?

Thanks again for making sure the contribution lineage stays properly attributed!

PR #1406, opened 2026-09-18
fix(npm): hide the updater's console window on Windows

liuhuichuan:
## Description

On Windows the auto-updater opened an empty console window on every update check, and the same console allocation stalled the update itself.

`bin/ocr.js` spawns `scripts/update.js` with `detached: true, stdio: "ignore"`, so the updater has no console of its own. Both of its `spawnSync` calls start console programs -- the `ocr version` probe, and the `cmd.exe` that `shell: true` creates for `npm i -g` -- and Windows answers a console-less parent by allocating a fresh console and handing it to the default terminal application. With Windows Terminal as the default terminal that is an empty black window next to the user's own terminal, and the allocation blocks the spawn until a terminal attaches. Measured on Windows 11: 31,321 ms vs 322 ms for a command that exits immediately, and 1.0 s vs 30 s+ for the version probe.

The stall is not only cosmetic. `getInstalledVersion()` allows the probe 3 s, so on a cold Windows Terminal the probe is killed mid-handoff, returns null, and `update.js` returns before it ever asks the registry for the latest version -- the update silently never happens. The half-created window then outlives the killed process: it ignores WM_CLOSE and stays on the desktop until the whole Windows Terminal instance is closed, which is where the pile of unclosable windows comes from.

Fix: `windowsHide: true` on both spawns. It maps to `CREATE_NO_WINDOW`, which skips console allocation entirely, and the option is ignored on POSIX. Neither call needs a console -- both read stdout/stderr through pipes -- so nothing that used to be visible is hidden.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Test machine: Windows 11, Node 22.17.0, Windows Terminal as the default terminal.

`make test` was run as `go test -count=1 $(go list ./... | grep -v /extensions/)`, the same package set as the Makefile target; `-race` is unavailable on this box (CGO_ENABLED=0, no gcc). 23 packages ok, 0 failures. `npm run test:launcher` and `npm run test:update` pass, `gofmt -l .` is clean and `go vet ./...` exits 0.

Manual A/B with the real installed launcher, `OCR_UPDATE_INTERVAL=-1` and a stub `npm` first on PATH so nothing is downloaded:
- stock `update.js`: a new empty "Terminal" window appeared, and the stub npm was never reached within 100 s
- patched `update.js`: no new window, and the stub npm was reached in 1.9 s

Manual end to end: the real auto-upgrade 1.12.4 -> 1.12.5, triggered by `ocr review`, completed in ~3 s with no window.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have signed the CLA

Notes on what is left unchecked: no test was added, because the change only sets a spawn option and `scripts/update.js` has no seam for asserting spawn options today; the existing launcher and version tests cover the surrounding behaviour. No docs describe the updater's console behaviour, so there is nothing to update. The CLA is for the author to sign.

AI/LLM disclosure: drafted with Codex (OpenAI's coding agent, v0.154.0-alpha.6.2) running the `deepseek/deepseek-v4.1-flash` model. I ran every command above, reviewed the output, and can explain each changed line. The commit carries no AI attribution and no AI trailers.

## Related Issues

None.


scripts/update.js
```diff
@@ -154,6 +155,7 @@ async function main() {
       encoding: "utf8",
       timeout: 120000,
       shell: IS_WINDOWS,
+      windowsHide: true,
```
ra5533, line 158: Looks good

scripts/update.js
```diff
@@ -66,6 +66,7 @@ function getInstalledVersion(binPath) {
     const result = spawnSync(binPath, ["version"], {
       encoding: "utf8",
       timeout: 3000,
+      windowsHide: true,
```
lizhengfeng101, line 69:
```suggestion
      // CREATE_NO_WINDOW: the updater is detached with no console of its own,
      // so Windows would allocate one here and hand it to the default terminal.
      windowsHide: true,
```

scripts/update.js
lizhengfeng101, line 158:
```suggestion
      // Same as above: keeps the cmd.exe from `shell` out of a new console.
      windowsHide: true,
```

lizhengfeng101 (commented):
Verified the diagnosis: `bin/ocr.js` spawns the updater with `detached: true`, which libuv maps to `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP`, so the updater genuinely has no console and every console child it starts gets a freshly allocated one. Both patched call sites read their output through pipes, and `bin/ocr.js`'s `stdio: "inherit"` spawn of the real CLI is correctly left alone, so nothing that used to be visible is hidden. Grepping the shipped files (`bin/ocr.js`, `scripts/{install,update,version,platform}.js`) turns up no other spawn on this path -- the two sites here are the complete set.

One nit inline: `windowsHide: true` is a platform option whose reason is not self-evident, and this repo consistently annotates that kind of Windows detail (see the Ctrl+C / `TerminateProcess` note in `bin/ocr.js`, or the no-`-race` note in `ci.yml`). Right now the *why* lives only in the PR body, so the next reader cannot tell whether the option is safe to drop. Suggestions below.

Optional, on the missing test: the reason there is no seam is that `update.js` calls `main()` unconditionally at the bottom, so requiring it from a test runs a real update. A `require.main` guard mirroring `bin/ocr.js`, plus a default parameter on `getInstalledVersion(binPath, spawn = spawnSync)`, would make the option assertable in about three lines. Worth considering, since an options-only fix has no regression guard today. (Separately, and not this PR's problem: `npm run test:update` is not wired into any workflow, so it would not run in CI as things stand.)

lizhengfeng101 (approved):
LGTM

PR #1407, opened 2026-09-18
chore: extend license and english-only gates to kt, py and css

lizhengfeng101:
## Why

Both gates decide what to check from a hardcoded extension list, and files whose type is absent from that list pass silently — no error, no warning.

`LICENSE_EXTS` in `scripts/add-license.sh` and `scripts/verify-license.sh` held only `go sh js mjs ts tsx`. As a result **8 Python and 6 CSS tracked source files carried no SPDX header at all** while `make license-check` reported success:

- `examples/{codeup,gerrit,gitflic,gitlab}_ci/post_review{,_test}.py`
- `internal/viewer/static/style.css`, `extensions/vscode/src/webview/styles/global.css`, `pages/src/components/ColorBends.css`, `pages/src/styles/{blog,docs-markdown,index}.css`

`scannedExts` in `scripts/verify-english-only.go` already covered `.py` and `.css`, but not `.kt`/`.kts`.

## What changed

**Gates**

- `add-license.sh` / `verify-license.sh`: `LICENSE_EXTS` becomes `(go js mjs ts tsx kt kts sh py css)`, reordered so the list reads in the same groups the comment syntax falls into.
- `add-license.sh`: new `BLOCK_HEADER` for CSS. Pure CSS has no `//` line comment — that is SCSS — so the previous default would have emitted invalid CSS. The block carries no per-line prefix, matching the `{{/* */}}` header already in `internal/viewer/templates/app-header.html`. Dispatch is now `sh|py` -> `HASH_HEADER`, `css` -> `BLOCK_HEADER`, rest -> `SLASH_HEADER`.
- `verify-english-only.go`: `.kt` and `.kts` added to `scannedExts`.

Python needed no new first-line handling: all eight files open with a `#!/usr/bin/env python3` shebang that the existing `^#!` branch already keeps on line 1, and none carries a PEP 263 encoding declaration (which would have to stay within the first two lines). CSS needed none either — the repository has no `@charset` rule, the one CSS construct that must precede everything, comments included.

**Backfilled headers**: the 14 files listed above.

**Docs**: the extension list in `AGENTS.md`, `CONTRIBUTING.md`, its four `docs/i18n` translations and the five `pages/src/content/docs/<locale>/contributing.md` pages. `AGENTS.md` also gains the criterion for what belongs on the list, so the next person does not have to rederive it: the repository must actually hold files of that type, and the comment must be safe to prepend.

## Scope notes

- **`.kt`/`.kts` are forward-looking.** The repository holds no Kotlin files yet. They appear in `internal/config/allowlist/supported_file_types.json`, but that list has 111 entries describing what `ocr` can review in *other* repositories — it is not a statement about this codebase. Covering them now costs nothing and means the first Kotlin file to land arrives with a header.
- **`.html` is deliberately left out.** It is the one remaining type whose comment cannot simply be prepended: `<!DOCTYPE html>` has to stay on line 1, needing a third first-line branch in `add_header`, and the style would have to be decided per path — `{{/* */}}` for the Go templates under `internal/viewer/templates/` versus `<!-- -->` for `pages/index.html`, which Vite processes and where Go template syntax would be emitted verbatim. Worth its own change. For the record, `internal/viewer/templates/app-header.html` already has a `{{/* */}}` header while its four sibling templates do not — exactly the drift a gate would prevent.
- **The CSS header reaches the bundles.** `css-loader` turns each stylesheet into a JS string constant and CSS minification is not enabled, so the comment survives into 4 chunks rather than being stripped. Impact is negligible (~100 bytes each, `size-limit` reports 95.24 kB against a 150 kB budget) and shipping the Apache notice inside the distributed JS is arguably correct, but it is a real consequence worth stating.

## Verification

| Check | Result |
| --- | --- |
| `make check` | passed (license + english-only + gofmt + vet) |
| `make test` | all 24 packages ok, including `internal/viewer` whose `style.css` is `go:embed`ed |
| `add-license.sh` rerun | `Added 0 file(s)` — idempotent |
| Python syntax and docstrings | all 8 files parse via `ast`, module docstring still the first statement |
| Python unit tests | 27 + 61 + 14 + 142 = 244 tests, all OK |
| `pages` production build | webpack succeeds; `size-limit` 95.24 kB / 150 kB |
| VS Code extension build | webpack production succeeds |
| Kotlin coverage | verified with throwaway `.kt` / `.kts` probes: both gates reject them without a header, `add-license.sh` writes the `//` style and stays idempotent |

## AI/LLM disclosure

Claude Opus 5 was used while preparing this change. `ocr review --audience agent` was run against the diff for self-review and reported 0 findings across the 13 files it selected (the 11 Markdown files fall outside the reviewable extension list, and the 4 `*_test.py` files match the built-in `**/*_test.py` exclude pattern). Every line here was reviewed and verified by me, and the verification table above records checks I ran myself.


css521 (approved):
LGTM

PR #1412, opened 2026-09-18
fix(review): eliminate duplicate comments in ocr review output

praveenec-info:
## Summary
Fixes an issue where `ocr review` could emit near-duplicate comments for
the same file and line range in its output (observed with
`--format json --audience agent`).

## Root Cause
The repository already has a deterministic, LLM-based comment
deduplication mechanism (`maybeRunDedup`), but it is implemented only in
`internal/scan` — the package backing `ocr scan`. The `ocr review`
command (`--to`/`--from`/`--commit`) is served by a separate package,
`internal/agent`, which never invokes any dedup step. As a result, review
runs can surface multiple comments referencing the same underlying
finding when the agent re-discovers it via different tool-call paths
(e.g. a second `code_search` hit, a repeated `file_read`).

## Fix
Adds `DedupByLocation()` in `internal/tool` — a deterministic,
zero-LLM-call pass that:
- Groups comments by `(path, overlapping [start_line, end_line])`
- Keeps the highest-severity comment per group (ties keep first-seen)
- Never merges comments with unset/zero line ranges (safe default)

This is wired into `internal/agent.Agent` via a new `finalComments()`
helper that wraps every return path of `dispatchSubtasks`, ensuring
review output is deduplicated regardless of how the run terminates
(normal completion, context cancellation, or partial failure).

## Testing
- New unit tests in `internal/tool/dedup_test.go` covering: same-location
  merge, severity tie-break, non-overlapping ranges, cross-file
  isolation, and unset line ranges — all passing.
- `go test ./internal/agent/... ./internal/tool/...` passes with no
  regressions.
- `go build ./...` succeeds.

## Notes
This is intentionally a cheap, local dedup (no additional LLM call) to
keep review latency and cost unchanged. A follow-up could optionally
wire the existing LLM-based `maybeRunDedup` into the review pipeline as
well for semantic (non-positional) duplicate detection, if desired.


NanaseInori (changes requested):
Thanks for addressing duplicate findings. I found a correctness issue in the current matching rule that needs to be fixed before merge.

`DedupByLocation` treats overlapping ranges in the same file as duplicate findings without comparing what the comments report. Two independent issues can legitimately affect the same lines—for example, unsafe SQL construction and an ignored database error. With the current implementation, only the higher-severity comment survives, even though both findings are valid.

Because this pass is applied to the final review output, this silently removes actionable findings rather than just reducing repetition. Please use a conservative finding-identity/content match in addition to location, or otherwise preserve distinct findings when identity is uncertain. Add a regression test where two different findings on the same range both survive, alongside the true-duplicate case.

praveenec-info:
Good catch, thanks — fixed. DedupByLocation now requires content similarity
in addition to location overlap before merging two comments: it computes
Jaccard word-similarity on Content (threshold 0.30), falling back to an
ExistingCode snippet match (threshold 0.60) when the wording differs but
both comments quote the same code. Two independent findings on the same
range (like the SQL-injection / ignored-error example) now both survive.

Added the regression test you requested
(TestDedupByLocation_DistinctFindingsSameRangeBothSurvive) plus one for the
ExistingCode fallback path. Both new tests pass, and the full
internal/agent + internal/tool suites pass with no regressions.

PR #1414, opened 2026-09-18
fix(llm): trim environment-sourced endpoint values

### LINKED ISSUE #1395: Env-sourced endpoints are not trimmed: a trailing CR in ANTHROPIC_BASE_URL fails every file with an opaque URL-parse error
## OpenCodeReview Version

`open-code-review dev windows/amd64` — built from source at commit `089b68b` (`v1.12.4-14-g089b68b`, 2026-09-17). Re-checked against `origin/main` `189be5b` (2026-09-17 20:59): both functions below still read the raw environment values there.

## Operating System

Windows (Windows 11, 10.0.22631), Git Bash

## Installation Method

Built from source (`go build -o dist/ocr.exe ./cmd/opencodereview`)

## LLM Provider

Anthropic-compatible gateway, reached through the `Claude Code environment` resolution path (`ANTHROPIC_BASE_URL` / `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_MODEL`)

## Bug Description

Endpoint resolution normalizes its inputs on some paths but not others:

| resolution path | trims the value? |
|---|---|
| `parseShellRC` (shell rc file) | yes — line and value (`internal/llm/resolver.go` ~767, ~780) |
| `parseTimeoutEnv` | yes (~204) |
| `NormalizeAuthHeader` | yes (~827) |
| provider `api_key` | yes (~426) |
| `tryOCREnv` — `OCR_LLM_URL/TOKEN/MODEL` | **no** (~251-254) |
| `tryCCEnv` — `ANTHROPIC_BASE_URL/AUTH_TOKEN/MODEL` | **no** (~702-705) |

So a value that the shell-rc path tolerates is fatal on the environment path. `OCR_LLM_PROTOCOL`, read a few lines below `tryOCREnv`'s untrimmed reads, *is* wrapped in `strings.TrimSpace`, which makes this look like an oversight rather than intent.

Two consequences:

1. **Opaque failure.** The Go SDK rejects the value before any request is sent: `requestoption: WithBaseURL failed to parse url parse "...\r": net/url: invalid control character in URL`. Nothing in the message points at the environment, or at which variable was dirty. In my case this failed **all 36 files** of a review run (36 `llm_error` records with an identical message in the session JSONL), and the value only had a trailing `\r` — exactly what a CRLF-generated environment produces on Windows.
2. **The session log can't be triaged.** `llm_error` records carry only the raw error string, not the resolved `Source`, so a reader cannot tell whether the bad endpoint came from the environment, the config file, or the rc file.

## Steps to Reproduce

A dummy token is enough: the failure happens while parsing the URL, before any request, so nothing is spent.

```bash
export ANTHROPIC_BASE_URL=$'https://<host>/api/anthropic\r'   # trailing CR, e.g. CRLF-generated env
export ANTHROPIC_AUTH_TOKEN=dummy
export ANTHROPIC_MODEL=<model>
ocr llm test

export ANTHROPIC_BASE_URL="https://<host>/api/anthropic"      # same value, trimmed
ocr llm test
```

The same applies to `OCR_LLM_URL` / `OCR_LLM_TOKEN` / `OCR_LLM_MODEL`.

## Expected Behavior

- `OCR_LLM_*` and `ANTHROPIC_*` values are trimmed before use — or normalized once in `finalizeResolvedEndpoint`, which already normalizes the model suffix — so every resolution path behaves identically.
- When a value can't be used, the error names the source and the variable, e.g. `Claude Code environment: invalid ANTHROPIC_BASE_URL (control character in URL)`, and the session record preserves which source was resolved.

## Logs / Error Output

```shell
$ ANTHROPIC_BASE_URL=$'https://zenmux.ai/api/anthropic\r' ANTHROPIC_AUTH_TOKEN=dummy ANTHROPIC_MODEL=<model> ocr llm test
Error: llm request failed: requestoption: WithBaseURL failed to parse url parse "https://zenmux.ai/api/anthropic\r": net/url: invalid control character in URL

$ ANTHROPIC_BASE_URL="https://zenmux.ai/api/anthropic" ANTHROPIC_AUTH_TOKEN=dummy ANTHROPIC_MODEL=<model> ocr llm test
Error: llm request failed: POST "https://zenmux.ai/api/anthropic/v1/messages": 403 Forbidden {"error":{"code":"403","type":"access_denied",...}}
```

The second command reaches the request, i.e. URL construction succeeded once the CR was gone (the token is a dummy, so the gateway rejects the auth — nothing was consumed).

Session JSONL from the failing review run, 36 records of this shape:

```json
{"duration_ms":0,"error":"requestoption: WithBaseURL failed to parse url parse \"https://<host>/api/anthropic\\r\": net/url: invalid control character in URL","filePath":"<redacted>","request_no":1,"taskType":"main_task","type":"llm_error"}
```

## AI/LLM Disclosure

The reproduction and verification above were run by me. This write-up was drafted with an AI coding agent (OMP, model `deepseek/deepseek-v4.1-flash`, via a gateway), and I checked every command and log line myself.

## Additional Context

Related, on the same fallback path — disclosing the resolved endpoint:

The `Claude Code environment` fallback is documented and covered by tests (`resolver_test.go:55`), and honoring ambient `ANTHROPIC_*` is reasonable for zero-config use. What's missing is disclosure: the resolved source is printed only by `ocr llm test` (`cmd/opencodereview/llm_cmd.go:110`), while `ocr review` prints nothing about it and `--preview` lists files only. A review launched by a wrapper script, another agent, or CI therefore silently inherits whatever endpoint and model the ambient environment carries — and the price spread between tiers is large (roughly 8x between the cheap and expensive tier on my gateway). One 36-file review inherited an expensive tier model and consumed ~1.42M tokens, about 90% of that day's gateway spend, which I only noticed on the bill.

Suggestions, in descending order of how much they would have helped:

1. Print one line at review start, on stderr so `--format json|sarif` stdout stays clean: `[ocr] llm: <model> via <source>`.
2. When the only resolved source is an ambient one (`Claude Code environment` / `Shell rc file`), say so explicitly, and consider requiring confirmation (`--yes`) or offering an escape hatch such as `--no-ambient-endpoint`.
3. Let `ocr llm test` accept `--model` the way `ocr review` does, so a run with a specific model can be preflighted. My preflight failed with `unknown flag: --model` and had to be rewritten to set `ANTHROPIC_MODEL` instead.

Minor note from the same run: the review exited 0 with 22 of 36 items covered after the gateway answered 402 `quote_exceeded` for the rest. That matches the documented contract (`cmd/opencodereview/review_cmd.go:330-339` — non-zero only when every selected item failed) and is the right behavior for a self-imposed budget cap, but for a provider-side auth/quota failure CI would read a half-finished review as success.

Jimmyman1465:
Closes #1395

## Summary

- Trim whitespace from `OCR_LLM_*` and `ANTHROPIC_*` environment values before resolving an endpoint.
- Validate environment-provided endpoint URLs before SDK construction, so errors identify both the configuration source and variable name.
- Persist the non-secret resolved source as `llmSource` in `session_start` and `llm_error` records.

## Why

A trailing carriage return can make an otherwise valid endpoint fail before the request is sent. Trimming must happen when environment values are read because `tryCCEnv` calls `ensureMessagesSuffix` before endpoint finalization; otherwise the carriage return can become embedded in the URL.

If a URL remains invalid after trimming, validation now reports the resolution source and exact variable, for example `Claude Code environment` and `ANTHROPIC_BASE_URL`. Session records retain the source for later diagnostics without storing tokens or full request URLs.

## Coverage

- OCR environment values
- Claude Code environment values
- preset provider API-key fallback
- `OCR_USE_ANTHROPIC`
- invalid OCR and Claude Code environment URLs
- source propagation through review and scan sessions
- `session_start` and `llm_error` JSONL records

## Verification

- English-only source check passed.
- License-header check passed.
- `git diff --check` passed.
- Race-enabled tests passed for `internal/llm`, `internal/session`, and `internal/scan`.
- The Windows-specific invalid-URL test uses an embedded carriage return rather than NUL because Windows rejects NUL before it reaches the resolver.

The remaining agent/CLI failures in the restricted local sandbox were caused by Windows denying `EvalSymlinks` access to Go-created temporary test directories; they did not report failures in the changed endpoint or session behavior. CI is the authoritative full-suite run for the rebased branch.

## AI disclosure

I used Codex GPT-5.6 Sol while preparing this change. I reviewed the implementation and tests line by line and am responsible for the final content. Reviewer responses are based on my own understanding; AI was used only for translation or wording polish. No AI attribution trailer is included in the commits.


wu21-web:
I am almost done with mine

Jimmyman1465:
Thanks for saying so — you flagged interest on the issue before I did, and I don't want to leave two competing PRs for the maintainers to sort through. So rather than racing, here's where this one stands, and I'm happy to let them decide between them.

Mine trims at the four reads that feed the endpoint, with one regression test per site (each verified to fail on the pre-fix code). `make check` passes, `make test` is green across all 23 packages under `-race`, and the review bot came back with 0 findings.

If your branch is ready, post it. Two things worth comparing against, in case you landed somewhere different:

- Trimming in `finalizeResolvedEndpoint` doesn't work: `tryCCEnv` calls `ensureMessagesSuffix` first, which appends `/v1/messages` and puts the CR in the middle of the URL, where a later `TrimSpace` can't reach it.
- Two sites beyond the issue's table are the same bug: the preset provider key fallback in `tryProviderConfig` (it checks `strings.TrimSpace(v) != ""` but stores the raw `v`), and `OCR_USE_ANTHROPIC`, where `"true\r"` isn't recognized and silently flips the protocol to OpenAI.

Happy to fold in anything yours handles better.

wu21-web:
Per AGENTS rules you should comment and speak to users by yourself, not AI.

Jimmyman1465:
上述内容我已在PR中完整披露，也亲自检查了改动并完成了文中列出的验证。我提交 PR #1414 时，Issue #1395 并没有分配给任何人。这个 PR 的提交时间也早于你刚才发的评论。
如果你认为上述代码或测试存在问题，请指出具体位置和可复现依据，否则，还是让维护者根据已经提交的实际成果进行判断吧

wu21-web:
不是说"defer"吗，而且pr打开时间只是我发布的时间，不是我开始工作的时间
> if you already have a branch in progress, say so and I'll defer.

看到你上面的话我就放心了，我只是害怕我和一个人机说半天话，这种太常见了

wu21-web:
@Jimmyman1465 这个项目欢迎第一次贡献的人，同样我也不想留两个PR竞争
如果你想做，告诉我我来关闭

Jimmyman1465:
谢谢您的说明，我想继续推进这个 PR，麻烦您帮忙关闭另一个竞争的 PR 吧～ 有问题随时沟通

> @Jimmyman1465 这个项目欢迎第一次贡献的人，同样我也不想留两个PR竞争 如果你想做，告诉我我来关闭

wu21-web:
好的已关闭

wu21-web:
@Jimmyman1465 你这个PR实现的不全
除了strip以外，还有：
> When a value can't be used, the error names the source and the variable, e.g. Claude Code environment: invalid ANTHROPIC_BASE_URL (control character in URL), and the session record preserves which source was resolved.

这个是不是得做一下

Jimmyman1465:
> @Jimmyman1465 你这个PR​​ 实现的不全 除了strip以外，还有：
> 
> > 当某个值无法使用时，错误会指出来源和变量，例如Claude Code 环境：无效的ANTHROPIC_BASE_URL（URL 中存在控制字符），并且会话记录会保留已解析的来源。
> 
> 这个是不是得做一下

正在搞兄弟~

NanaseInori (approved):
Looks good to me. Trimming at the environment-read boundary is the right placement, especially for the Claude Code URL before `ensureMessagesSuffix` can turn a trailing CR into an embedded one.

The preset API-key fallback and `OCR_USE_ANTHROPIC` case are covered as well. The change stays focused on the reported configuration bug; richer diagnostics and session provenance can be handled separately. No blocking findings from me.

NanaseInori (approved):
Re-reviewed the latest head. The follow-up commit closes the completeness gap raised after the first revision.

The environment-sourced endpoint values are still trimmed at the correct read boundary, and invalid environment URLs are now validated early with diagnostics that identify the relevant variable/source. The resolved endpoint source is also threaded from the runtime into both review and scan sessions and persisted in `session_start` / `llm_error`, which addresses the session-triage part of #1395.

The added tests cover the CRLF trimming cases, invalid URL diagnostics, runtime source propagation, and JSONL persistence. I don't see any remaining blocking implementation issue in the current head.

This review was conducted by NanaseInori's review bot, using the model GPT-5.6 Sol. If you need a human review, please manually @NanaseInori.

wu21-web:
Your pull request desc is stale @Jimmyman1465

Jimmyman1465:
@lizhengfeng101 @NanaseInori I have rebased the branch onto the latest main, corrected the commit authorship, and updated the PR description to match the final scope. The CLA now passes. The workflow runs are awaiting maintainer approval. When convenient, could you please approve the workflows and review this PR? Thank you.

internal/llm/resolver.go
```diff
@@ -247,17 +248,36 @@ func errBedrockNotConfigurable(key string) error {
...
 }
 
+// validateEndpointURL reports URL parse failures before the SDK can turn them
+// into request errors that no longer identify the configuration variable.
+func validateEndpointURL(variable, value string) error {
+	if _, err := neturl.Parse(value); err != nil {
+		return fmt.Errorf("invalid %s: %w", variable, err)
+	}
+	return nil
+}
```
wu21-web, line 258: this does not actually validate that the endpoint is usable @Jimmyman1465 
```suggestion
func validateEndpointURL(variable, value string) error {
	parsed, err := neturl.Parse(value)
	if err != nil {
		return fmt.Errorf("invalid %s: %w", variable, err)
	}
	if parsed.Scheme != "http" && parsed.Scheme != "https" {
		return fmt.Errorf(
			"invalid %s: URL must use http or https scheme, got %q",
			variable, parsed.Scheme,
		)
	}
	if parsed.Host == "" {
		return fmt.Errorf("invalid %s: URL must include a host", variable)
	}
	return nil
}
```

wu21-web (changes requested):

internal/llm/resolver.go
Jimmyman1465: 好的好的~

lizhengfeng101 (approved):
nice work

PR #1417, opened 2026-09-18
chore: clear non-code blockers ahead of the english-only cleanup

lizhengfeng101:
## What

Three independent housekeeping changes, each a prerequisite for dropping the two
`TEMPORARY` entries in `allowedPrefixes` in `scripts/verify-english-only.go`.

**Delete `extensions/vscode/prototype.html`.** A standalone 1833-line UI design mock.
Nothing in the codebase references it, and `.vscodeignore` already kept it out of the
packaged `.vsix` — it was documentation rather than code. Its 147 Chinese lines were
10.6% of the repo-wide non-English line count. The now-dangling `.vscodeignore` entry
is removed along with it.

**Register `*.jar` as binary in `.gitattributes`.** AGENTS.md requires new binary
extensions to be registered, and `extensions/idea/gradle/wrapper/gradle-wrapper.jar`
is now tracked. The leading `* text=auto eol=lf` would very likely have auto-detected
it regardless, so this is about being explicit where the guideline asks for it, not
about a broken wrapper.

**Translate the Chinese comments in `extensions/idea/gradle.properties`.** Worth noting
that `.properties` is not in the checker's `scannedExts`, so these lines were never
going to be flagged — this is a convention fix rather than a gate fix, made because the
comments ship to GitHub.

## Verification

- `make check` passes: license headers plus english-only across 531 scanned files.
- With both exemptions temporarily removed, the violation count drops from 1387 to
  1240 lines — exactly the 147 that `prototype.html` contributed.
- `git add --renormalize .` produces no further changes, so line endings are clean.
- `ocr review` reports 0 findings, both on the staged change and on the full branch diff.

## AI disclosure

Prepared with AI assistance (model: Claude Opus 5) and reviewed with OpenCodeReview (`ocr`).


css521 (approved):
LGTM

PR #1425, opened 2026-09-18
fix(tool): keep non-ASCII paths literal in code_search results

kevin9327:
## Description

`code_search` runs `git grep` without `-c core.quotepath=false`. With Git's default `core.quotepath=true`, any path containing non-ASCII bytes comes back C-quoted and octal-escaped, so for a file named `src/café/文件.go` the tool hands the agent:

```
File: "src/caf\303\251/\346\226\207\344\273\266.go"
Match lines: 1
3|func Needle() {}
```

That string is not a real path. Passing it to `file_read` fails in both modes: the workspace read finds no file by that name, and in commit/range mode `git show <ref>:"src/caf\303\251/..."` exits 128. The literal path reads fine. So any match in a file with a CJK or accented name is unusable for the rest of the review.

The diff commands (#365), `file_read` and `ocr scan` already pass `-c core.quotepath=false`; `buildGrepArgs` was the remaining gap. This adds the same flag there. Workspace, range/commit and the plain-directory `--no-index` fallback all build their arguments through `buildGrepArgs`, so all three are covered.

Scope notes:
- Git still quotes paths containing `"`, `\` or control characters even with `core.quotepath=false`. Like #365, this only handles the non-ASCII case.
- `file_find` has the same quoting issue in its `git ls-files` / `git ls-tree` listing. I left it out because open PR #994 already changes those calls.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Added `TestGitGrep_NonASCIIPath`. It commits `src/café/文件.go` in a repo with `core.quotepath=true` set explicitly, so the result does not depend on the global Git config. It then checks that workspace and commit mode both report the literal path.

Before the fix (current `main`):

```
code_search_test.go:810: expected "File: src/café/文件.go\n" in result, got: File: "src/caf\303\251/\346\226\207\344\273\266.go"
--- FAIL: TestGitGrep_NonASCIIPath/workspace
--- FAIL: TestGitGrep_NonASCIIPath/commit
```

After the fix, both subtests pass.

Environment: Windows 11, Go 1.27.1, Git 2.54.0.windows.1. This machine has no `make` or C toolchain, so I ran the underlying commands directly instead of `make test`, without `-race`:

- `go test -count=1 ./...`: all 23 packages pass, the same result as on `main`
- `go test -coverprofile` on `internal/tool`: 96.0%, and the changed line is executed by the new test
- `go vet ./...`, `gofmt -s -l .`, `go run scripts/verify-english-only.go`, `bash scripts/verify-license.sh`, and `go mod tidy` (no diff): all clean

## Checklist

- [ ] I have signed the CLA

AI assistance: prepared with Claude Code (Claude Opus 5).

## Related Issues

Related: #365 (same fix for the diff commands), #994 (touches the `file_find` listing).


NanaseInori (approved):
The fix is correctly placed in `buildGrepArgs`, so the shared grep paths receive the same `core.quotepath=false` behavior. The regression test explicitly enables quoting in the test repository and checks literal paths in both workspace and commit mode, rather than depending on the developer's Git configuration.

The stated non-ASCII scope is clear, and leaving the separate `file_find` work out avoids overlap with #994. No blocking findings from me.

lizhengfeng101 (approved):
LGTM

Issue #1319, opened 2026-09-16
Viewer UI redesign — align `ocr viewer` with the new design system


lizhengfeng101:
## Background

`ocr viewer` already serves a working, read-only web UI from `internal/viewer/`
(server-rendered Go `html/template` + `static/style.css` + a little vanilla JS). It
covers all four screens — Repositories, Sessions, Session detail, and Compare — with
theming, search, pagination, comment filters and marked/hidden toggles already in place.

We now have a set of design mockups that give these screens a consistent, polished visual
language (refined tables, spacing, typography, a proper app header/wordmark, a shared icon
set, and cleaned-up light **and** dark themes). This is a **visual redesign of the existing
viewer**, not a rewrite: routes, data model, CSP/security posture and Go handlers stay as
they are — we are restyling templates, CSS and static assets to match the mockups.

## Goal

Bring every viewer screen in line with the mockups while keeping the viewer read-only,
CSP-safe and server-rendered.

## Scope & sub-issues

Foundational (land first):
- [x] #1320 — Design tokens & app header/wordmark (light + dark)
- [x] #1321 — Shared SVG icon set integration

Per-screen (parallelizable after the foundation):
- [x] #1322 — Repositories page
- [x] #1323 — Sessions list page
- [x] #1324 — Session detail: header / meta bar
- [x] #1325 — Session detail: Coverage & Token Usage
- [x] #1326 — Session detail: Review Comments
- [x] #1327 — Session detail: Files Reviewed, Session Tasks & Conversations
- [x] #1328 — Compare page

Polish (last):
- [x] #1329 — Responsive, accessibility & template/visual tests

## Parallelization & conflicts

**Order:** land #1320 (tokens) and #1321 (icons) first — every screen consumes the shared CSS
variables and icon mechanism they establish. Do #1329 (responsive/a11y/tests) last.

**Safe to run in parallel:** #1322 (Repositories), #1323 (Sessions) and #1328 (Compare) each
own a separate template file, so they don't collide.

**Coordinate closely:** #1324, #1325, #1326 and #1327 all edit the same `session.html`
(#1326/#1327 also `session.js`). Assign a single owner for that file, or serialize these four,
to avoid heavy merge conflicts.

**Shared by everyone:** `static/style.css` is touched by every screen. Keep each change scoped
to a clearly delimited section, rebase on `main` often, and land #1320 first so the token
churn is behind you.

## Design fidelity vs. current features

[...]

PR #1426, opened 2026-09-18
fix(viewer): responsive, accessibility and contrast polish across the four screens

### LINKED ISSUE #1329: Viewer redesign: responsive, accessibility & tests
> Part of the viewer UI redesign (see #1319). **Land last**, after the screens.

## Summary

Cross-cutting polish once the screens are restyled: responsive behavior, accessibility, and
test coverage for the new markup/styles.

## Scope

- [ ] Responsive: verify tables and cards degrade gracefully on narrow widths (horizontal
      scroll or stacked layout) across all four screens.
- [ ] Accessibility: visible focus states, sufficient contrast (WCAG AA), correct
      `aria-pressed`/`aria-label` on chips/toggles, keyboard operability of collapsibles and
      pagination.
- [ ] Reduced motion: respect `prefers-reduced-motion` for any transitions introduced.
- [ ] Tests: update/extend the viewer template + handler tests (`internal/viewer/*_test.go`)
      for any changed markup hooks; keep the read-only route contract test
      (`TestMux_HasNoWriteRoutes`) green.
- [ ] Confirm the CSP/security-header and host-guard tests still pass unchanged.

## Acceptance criteria

- No regressions across screens in light and dark, desktop and narrow widths.
- `make check`, `make test` and `make coverage` (90% threshold) pass.

## Notes

- Optional follow-up (separate issue if desired): a JS light/dark theme toggle with
  `localStorage` persistence, building on the `prefers-color-scheme` default from #1320.

## Parallelization & conflicts

**Land last**, after the screens (#1322–#1328) are restyled. This is cross-cutting polish and
test coverage over the final markup, so starting early means reworking against moving targets.

**AI disclosure:** The scope, decisions and issue breakdown are my own. An AI assistant (Claude Code) was used only to polish the English wording.

jimpablo:
*Closes #1329. Part of #1319 — this is the last issue of the viewer redesign series.*

## Scope walk-through (as written in the issue)

**1. Responsive** — the tables already scroll horizontally inside their wrappers, and the compare grid already collapses via `auto-fit`. Two real gaps are fixed:

- `.session-page .meta span` held `white-space: nowrap` with a 22rem ellipsis truncate at every width, so a long CWD or branch was cut to one ellipsized line on narrow screens. Below 768px the items now wrap, so the full value stays readable instead of hidden behind an ellipsis. The overrides live in a dedicated 768px block placed *after* the base `.session-page .meta` rules (equal specificity, so source order decides), with `TestResponsiveCSS_MetaOverrideComesAfterBase` pinning that order.
- `.marks-row` / `.marks-actions` had no `flex-wrap`, so the counter, toggle and clear link squeezed on narrow viewports. Both wrap now.

**2. Accessibility**

- *Visible focus states*: the six collapsible headers (`file-accordion-header`, `token-breakdown-toggle`, `comment-file-header`, `tool-detail-toggle`, `error-detail-toggle`), in-table links (`.table a`) and breadcrumb links now carry the same `:focus-visible` ring as the rest of the design system; the scrollable table regions (focusable via `tabindex="0"`) get an inset ring that survives the overflow clipping.
- *Contrast (WCAG AA)*: `--text-muted` was the one failing pair — 2.78:1 on white, 3.66:1 on black. It now sits at **4.75:1 in both themes** (light `rgba(0,0,0,0.55)`, dark `rgba(255,255,255,0.47)`). `--text-secondary` deepens from 0.55 to 0.62 (6.2:1) so light mode keeps two visibly distinct levels — though the secondary/muted pair is now deliberately compressed (0.07 alpha apart reads closer to one level than before); a fair price for AA. Before/after: muted 2.78→4.75 (light) and 3.66→4.74 (dark); secondary 4.77→6.19 (light), 7.37 (dark, unchanged). `--link` and all severity pairs already pass and are unchanged. The WCAG test scopes itself to the page background (and parses `--bg` rather than assuming white/black).
- *aria on chips/toggles*: already correct (JS-managed `aria-pressed`, labelled switch) — regression-checked.
- *Keyboard operability*: collapsibles are native `details/summary`, pagination is native buttons with focus restoration. The scrollable table wrappers on repos and sessions, plus the token-breakdown table on the session page, became focusable regions (`tabindex="0" role="region" aria-label`), with arrow-key and Home/End scrolling routed through a small shared `ocrArrowScroll` helper in a new `static/a11y.js` (scoped to the region itself), plus `overscroll-behavior: contain`. In-table link focus rings draw inset (`outline-offset: -2px`) so the scroll container does not clip them on the first and last rows.
- *Live regions*: the filter empty-state gained `role="status"` and the marks counter `aria-live="polite"`, so screen readers hear filter and mark changes.

**3. Reduced motion** — the global `prefers-reduced-motion` wildcard already zeroes every animation and transition (fadeIn, slideDown, all hover transitions included). Verified, no change needed.

**4. Tests** — new `TestTextTokens_MeetWCAGAAOnPageBackground` (parses the tokens and `--bg`, holds muted/secondary to 4.5:1 in both themes), `TestFocusCSS_CoversCollapsiblesAndTableLinks` (pins the focus rules), `TestPagerJS_Contract` extended for `ocrArrowScroll`, and the repos/sessions/session render tests assert the new region and live-region attributes. `TestMux_HasNoWriteRoutes` and the CSP/security-header/host-guard tests pass unchanged.

**5. CSP/security** — no new inline styles or scripts; external files only. All security tests green.

## Acceptance

- No regressions at desktop width beyond the intended muted-text darkening (pixel-diffed all four pages, light and dark, against `main` — the only deltas are the muted/secondary text colors).
- Narrow widths (375px): no page scrolls horizontally; session meta wraps; tables scroll inside their labelled focusable regions.
- `make check`, `make test`, `make coverage` pass (91.9% total, viewer 94.7%).

## Out of scope

- The JS light/dark theme toggle from the issue notes — that is a separate follow-up issue by its own terms.
- Making every `.response-text .code-block` (pre) scroll region a focusable one — code blocks are numerous and would add a tab stop each; the tables and the token breakdown are where column overflow actually strands keyboard users.

## AI disclosure

I used ZCode CLI (GLM) to help prepare this change, reviewed every line myself, and I am available to explain any part of it.


lizhengfeng101 (commented):
Nice, focused change — the scope walk-through made this easy to follow, and I appreciate that you went and measured the contrast ratios instead of eyeballing them. The focus-ring coverage, the live regions and `overscroll-behavior: contain` all look right to me.

One thing does need fixing before this can go in, though.

## The meta-wrap fix doesn't actually take effect

This is the only code behind item 1 of the scope, and as written it's dead CSS:

```css
/* style.css:1546-1547, inside @media (max-width: 768px) */
.session-page .meta span { white-space: normal; }
.session-page .meta .meta-truncate { max-width: 100%; }
```

```css
/* style.css:1852-1856, further down the file, at top level */
.session-page .meta span { white-space: nowrap; }
.session-page .meta .meta-truncate { max-width: 22rem; overflow: hidden; text-overflow: ellipsis; }
```

A media query doesn't contribute any specificity, so these are two pairs of equally specific rules and source order decides — the ones at 1852 win at every viewport width. At 375px the meta items still hold `nowrap` and still truncate at 22rem, which means the acceptance note ("session meta wraps") doesn't match what the stylesheet does. Worth re-checking at 375px after the fix and updating the description.

Two ways out; I'd lean towards the first, since it keeps responsive overrides next to the rules they override:

- move this 768px block below the `.session-page .meta` section (i.e. after line 1856), or open a second `@media (max-width: 768px)` at the end of that section;
- or raise the specificity, e.g. `.session-page .session-header .meta span`.

Also worth noting: the new CSS tests assert that a rule *exists*, which can't catch a rule that exists but loses. If you want this pinned, asserting the relative order of the two `.session-page .meta span` declarations would do it.

## `TestTextTokens_MeetWCAGAA` proves less than its name suggests

The maths checks out — I recomputed light muted at 4.755:1 and dark muted at 4.746:1, matching your numbers. But the test only evaluates each token against the *page* background, and in dark mode `--text-muted` sits on other surfaces too:

- `.tool-calls-section` (style.css:897) has `background: var(--surface-inset)`, and `.tool-calls-label` (style.css:903) is `color: var(--text-muted)`.
- Dark `--surface-inset` is `#111111`, so white at α=0.47 lands at roughly **4.27:1** — still under AA. On `--surface-alt` (`#141414`) it's about **4.16:1**.

I'm not asking you to repalette every surface combination here; that's a bigger job than this PR. Two smaller things would help:

1. Make the limitation explicit in the test name or docstring. The function comment does say "against each theme's page background", but `MeetWCAGAA` reads as a global guarantee, and that's the part someone will remember six months from now.
2. Parse `--bg` out of the CSS rather than hardcoding `1.0` / `0.0`. Right now, if someone changes dark `--bg` from `#000000` to `#0a0a0a`, the test keeps passing on a premise that's no longer true.

On the palette itself: light mode now runs `--text` 0.77 / `--text-secondary` 0.62 / `--text-muted` 0.55. Secondary and muted are 0.07 alpha apart, which in practice reads as one level, not two. That's a fair price for AA and I'd take the same trade, but the description says the two-level hierarchy is kept — it'd be more accurate to say it's deliberately compressed.

## Focusable scroll regions cover two of the four screens

`tabindex="0" role="region"` plus `ocrArrowScroll` went onto the repos and sessions `.table-scroll` wrappers, but there are other `overflow-x: auto` containers that didn't get the same treatment. By your own rationale in the CSS comment ("keyboard users can otherwise never reach the columns that overflow"), at least one of them has the same problem:

- `.token-breakdown-body` (style.css:576) — a table wrapper on the session page that genuinely overflows.
- `.response-text .code-block` (style.css:874) — `white-space: pre`, so it overflows too.
- `.code-panel pre` (style.css:1324) is `pre-wrap`, so it's largely moot.

Either extend the pattern to `.token-breakdown-body`, or call it out under "Out of scope" so it doesn't look covered.

## Smaller things

- The comment above `TestFocusCSS_CoversCollapsiblesAndTableLinks` opens with `// TestFocusCSS_CoversChrome`. Go wants the doc comment to start with the function's own name.
- `.table a:focus-visible` uses `outline-offset: 2px`, but `.table-scroll` is `overflow-x: auto`, which per spec computes `overflow-y` to `auto` as well — so the ring on a first- or last-row link gets clipped vertically. You already solved exactly this for the region itself with a negative offset; the in-table links could use the same treatment, or a little vertical padding on the wrapper.
- `tabindex="0"` is added unconditionally, so pages whose table doesn't overflow gain a non-interactive tab stop. Setting it from JS based on `scrollWidth > clientWidth` is the closer fit to the ARIA guidance. Defensible as-is for a server-rendered template, so treat this as a note rather than a request.
- `ocrArrowScroll` living in `pager.js` is a bit of a squatter — it has nothing to do with pagination, and `TestPagerJS_Contract` is named after that contract. Fine if you'd rather not add another request, but a rename or a small `a11y.js` would be tidier. While you're in there: a horizontal scroll region usually wants `Home`/`End` too, not just `ArrowLeft`/`ArrowRight`.
- `firstAlpha` taking the first regex match is safe today because the light/dark tokens differ in their channel values. If a third theme ever reuses a channel triple, "first match" will silently point at the wrong rule — asserting exactly one match would close that off.

Happy to look again once the cascade issue is sorted. The rest is all small stuff.

NanaseInori (changes requested):
Thanks for the improvements. I agree with the CSS cascade issue already identified above: the later base rules override the narrow-screen metadata wrapping rules at the current head. Moving the responsive overrides after those base rules would address my main concern.

I would hold approval until that is fixed. A small regression check that catches this ordering problem would be sufficient here; I am not asking for a new browser-testing setup or a broader accessibility refactor.

lizhengfeng101:
@jimpablo

jimpablo:
@lizhengfeng101 thanks for the thorough review, and good catch on the cascade — that one was real. already working through the list: meta-wrap cascade fix and the ordering regression test are in, the smaller notes (a11y.js split, token-breakdown region, test scoping) are in progress. will push shortly

jimpablo:
pushed two commits addressing everything:

**the cascade** — the meta-wrap overrides moved out of the shared 768px block into a dedicated block placed after the base `.session-page .meta` rules, so they now win on source order (verified: computed `white-space: normal` on the meta span at 375px). per @NanaseInori's ask there's also `TestResponsiveCSS_MetaOverrideComesAfterBase`, which fails if the overrides ever drift above the base rules.

**@lizhengfeng101's list:**
- WCAG test renamed to `TestTextTokens_MeetWCAGAAOnPageBackground` with the scope limitation spelled out in the docstring; it now parses `--bg` from each section instead of assuming white/black, and the alpha extraction asserts exactly one match
- palette description in the body now says the light secondary/muted pair is deliberately compressed rather than kept
- the token-breakdown table got the same focusable-region treatment (`tabindex/role/aria-label` + arrow keys); `.response-text .code-block` is called out under Out of scope instead
- the `TestFocusCSS` doc comment now starts with the function name
- in-table link rings draw inset (`outline-offset: -2px`) so the first/last rows aren't clipped
- `ocrArrowScroll` moved out of pager.js into `static/a11y.js` (loaded on the three pages that need it) and gained Home/End
- left the unconditional `tabindex="0"` as-is per your note

all green: make check / make test / make coverage (91.9%).

internal/viewer/static/a11y.js
```diff
@@ -0,0 +1,26 @@
...
+            if (event.target !== region) return;
+            const step = {
+                ArrowLeft: -40,
+                ArrowRight: 40,
+                Home: -Number.MAX_SAFE_INTEGER,
+                End: Number.MAX_SAFE_INTEGER,
+            }[event.key];
+            if (step === undefined) return;
+            region.scrollBy({ left: step });
+            event.preventDefault();
```
lizhengfeng101, line 33: `preventDefault()` runs unconditionally, so on wide screens where the table doesn't overflow, tabbing into the region and pressing `End` does nothing instead of jumping to the page bottom. Same for `Home`. Bail early:

```js
if (region.scrollWidth <= region.clientWidth) return;
```

Since `a11y.js` is already loaded, it could own `tabindex` too — granting it only when the region scrolls would drop the dead tab stop on desktop as well.

internal/viewer/templates/session.html
```diff
@@ -140,7 +140,7 @@ <h3>Review Comments <span class="findings-count">({{len .Session.Comments}} find
     {{end}}
     </div>
     <div class="marks-row">
-        <span class="comment-filter-label" data-marks-count></span>
+        <span class="comment-filter-label" data-marks-count aria-live="polite"></span>
```
lizhengfeng101, line 143: `applyFilters()` in `session.js` runs on every chip click and mark toggle, and rewrites `marksCount.textContent` unconditionally (session.js:135) — so with `aria-live="polite"` here, clicking a severity chip announces "3 marked, 1 hidden" even when the marks didn't change. Write only on change:

```js
if (next !== marksCount.textContent) marksCount.textContent = next;
```

`emptyState.textContent` just above has the same issue: while visible, every interaction re-announces the same sentence.

lizhengfeng101 (commented):
Two behavior issues worth fixing before merge, both from this change. Left them inline.

The rest looks good to me: tests, `english-check`, `gofmt` and `go vet` all pass on 75ee30d, the new file has its SPDX header and LF endings, and no new inline script/style means CSP is untouched.

lizhengfeng101:
@jimpablo pin again

jimpablo:
rebased onto the current main (the session-export feature came in between, so session.html now keeps its export branch and the exported file inlines a11y.js ahead of session.js — the token-breakdown region works there too). conflict resolution also kept the new static-refs guard test.

both inline behavior issues are fixed as well:
- a11y.js now bails when the region does not overflow, so Home/End keep their native page-jump behavior on wide screens; it also owns the tabindex and only grants it while the region actually scrolls (with a resize listener to keep that in step), which drops the dead tab stop on desktop
- the live regions write only on change — the marks counter and the empty-state sentence no longer re-announce on every filter click

force-pushed as 5da6720 (nine commits on the current main).

internal/viewer/static/a11y.js
```diff
@@ -0,0 +1,36 @@
...
+    // their native behavior (Home/End jump the page) when there is nothing
+    // to scroll.
+    window.ocrArrowScroll = (region) => {
+        if (!region) return;
+        const scrolls = () => region.scrollWidth > region.clientWidth;
+        const sync = () => {
+            region.tabIndex = scrolls() ? 0 : -1;
+        };
+        sync();
+        window.addEventListener("resize", sync);
```
lizhengfeng101, line 22: `sync()` only re-runs on resize, but `<details class="token-breakdown">` (session.html:86) ships collapsed — at script time the region measures 0, so it's parked at `tabIndex = -1` and opening the details fires no resize. The token-breakdown region is currently unreachable by keyboard.

```suggestion
        window.addEventListener("resize", sync);
        // A collapsed <details> measures 0 wide, so re-sync when it opens.
        const owner = region.closest("details");
        if (owner) owner.addEventListener("toggle", sync);
```

Paging and filtering change `scrollWidth` without a resize too, so the tabindex can still go stale on the list pages — follow-up is fine.

internal/viewer/server_extra_test.go
```diff
@@ -334,6 +337,122 @@ func TestSurfaceCSS_LightSurfaceIsWhite(t *testing.T) {
...
 
+// TestTextTokens_MeetWCAGAAOnPageBackground holds the muted and secondary
+// text tokens to WCAG AA (4.5:1) against each theme's *page* background, so
+// future palette tweaks cannot quietly drop the low-emphasis labels back
+// below the line. Scope note: this is the page background only — muted text
+// also lands on --surface-alt / --surface-inset panels in dark mode, which
+// are slightly lighter and yield lower ratios; auditing every surface pair
+// is a bigger job than this test claims to do. Both tokens blend pure black
+// (light) or pure white (dark) at an alpha, which makes the blended color a
+// gray and the luminance math a single channel.
```
lizhengfeng101: Correcting my own earlier review: 4.27 / 4.16 were wrong. I flattened the text over pure black while taking the panel as `#111111` / `#141414` — two compositing bases. Done properly, `--surface-inset` is 4.84 (dark) / 4.61 (light); all three surfaces clear AA, tightest margin 0.11. The caveat describes a problem that isn't there, so extending the test to them is cheap — worth a follow-up. Please drop the matching sentence from the PR body too.

```suggestion
// below the line. Both tokens blend pure black (light) or pure white (dark)
// at an alpha, which makes the blended color a gray and the luminance math
// a single channel.
```

internal/viewer/server_extra_test.go
```diff
@@ -868,3 +988,55 @@ func TestHandleSession_ServedPageKeepsStaticRefs(t *testing.T) {
...
+		"ArrowRight",
+		"Home",
+		"End",
+		"preventDefault",
+		// The region joins the tab order only while it overflows, and the
+		// keys fall through natively when there is nothing to scroll.
+		"scrolls()",
+		"region.tabIndex",
+		"addEventListener(\"resize\"",
+		"scrollWidth > region.clientWidth",
```
lizhengfeng101, line 1011: Pins the fix above; the existing comment block already covers the why.

```suggestion
		"scrollWidth > region.clientWidth",
		`addEventListener("toggle"`,
```

lizhengfeng101 (commented):
Three suggestions, all applyable from the UI. The first is the one that matters — the token-breakdown region added for my earlier item 3 never joins the tab order. The second corrects a number I got wrong in my first review.

lizhengfeng101 (approved):
LGTM

PR #1430, opened 2026-09-18
fix(windows): kill the whole process tree when an MCP setup script times out

Hiro-Mati:
## Description

On Windows, MCP `setup` scripts run through `cmd /c`, so the real work (`npm install`, …) is a grandchild. On the 5-minute setup timeout only `cmd.exe` was killed; the grandchild kept running, and since it inherits the output pipe, `CombinedOutput` blocked until it exited — so the timeout never took effect.

`configureProcessGroup` (Windows) now:

- cancels with `taskkill /T /F /PID <pid>`, which terminates the whole tree, falling back to `Process.Kill`;
- bounds the `taskkill` call itself with a 10s timeout, since `WaitDelay` cannot interrupt a `Cancel` callback that is blocked inside it;
- sets a 5s `WaitDelay` as a safety net, so `Wait` returns even if a descendant escaped the kill (e.g. it was already orphaned when `cmd.exe` exited).

The existing code comment assumed nested shells were rare on Windows, but every setup script goes through `cmd /c`.

## Type of Change

## How Has This Been Tested?

- `TestConfigureProcessGroup_CancelKillsGrandchild`: the grandchild (PowerShell, spawned via `cmd /c` and holding the inherited output pipe) writes its own PID; the test waits for it to start, cancels, checks that `Wait` returns, and asserts via `tasklist` that the grandchild **no longer exists**, with cleanup on failure. It fails if `Cancel` kills only `cmd.exe` (even with `WaitDelay` in place), and on `main` `Wait` blocks until the grandchild exits.

## Checklist

- [x] I have signed the CLA


NanaseInori (commented):
The process-tree cancellation approach is reasonable, and WaitDelay is a useful fallback for inherited output pipes. I found a gap in the regression test rather than a demonstrated failure of taskkill itself.

TestConfigureProcessGroup_TimeoutKillsGrandchild only checks for a non-nil error and completion within 20 seconds. With the new five-second WaitDelay, killing only cmd.exe can still satisfy both assertions while its descendant remains alive. Please have the test record the descendant PID, verify that it actually started, and assert that it is no longer running after cancellation, with cleanup on test failure.

As a separate hardening improvement, consider bounding the taskkill invocation itself: WaitDelay cannot interrupt a Cancel callback blocked inside kill.Run(). I have not validated native Windows execution in this review.

Hiro-Mati:

The work is not lost if it is useful to you: the branch stays on my fork at https://github.com/Hiro-Mati/open-code-review/tree/fix/windows-setup-process-tree, with the regression test that fails without the fix. Feel free to take it over or reuse it under the project's license; no attribution needed.

PR #1431, opened 2026-09-18
fix(tool): unquote non-ASCII paths in file_find

Hiro-Mati:
## Description

`file_find` runs `git ls-files` / `ls-tree` without `-c core.quotepath=false`, unlike every other git call in the codebase, so a file named `café.go` comes back as `"caf\303\251.go"`. The agent cannot then read the file back under that name.

This is the `file_find` counterpart of #1425, which fixed the same thing for `code_search` paths.

Split out of #1431, whose remaining fix is in a separate PR.

## Type of Change

## How Has This Been Tested?

- The new test(s) fail when this change alone is reverted, and pass with it (verified both ways).
- `make check` passes; the touched packages pass on Windows 11 (Go 1.27) and in a `golang:1.26.5` Linux container replaying the CI job (`go test -race`, coverage >= 90%).

## Checklist

- [x] I have signed the CLA


NanaseInori (changes requested):
The Git output fixes look reasonable, but applying filepath.ToSlash directly to include/exclude glob patterns introduces a compatibility regression on Windows.

These patterns are evaluated by doublestar.Match, where a backslash escapes a literal character. For example, `src/\[generated\]/file.go` currently matches the valid path `src/[generated]/file.go`. On Windows, the new normalization changes that pattern into `src//[generated/]/file.go`, so it no longer matches. An existing exclusion can therefore silently stop excluding its intended file; an include can stop selecting it. This affects both applyCLIExcludes and buildFileFilter, and makes the same portable rule.json behave differently across platforms.

Please preserve existing glob escapes while defining how native Windows separators are accepted, rather than rewriting every backslash unconditionally. Add regression coverage for escaped literal brackets as well as the new backslash-separated path cases.

Hiro-Mati:

The work is not lost if it is useful to you: the branch stays on my fork at https://github.com/Hiro-Mati/open-code-review/tree/fix/windows-path-handling, with the regression test that fails without the fix. Feel free to take it over or reuse it under the project's license; no attribution needed.

PR #1433, opened 2026-09-18
fix(telemetry): record review duration and comment count once per run

Hiro-Mati:
## Description

`ocr.review.duration_seconds` and `ocr.comments_generated_total` were recorded twice per run, for both `ocr review` and `ocr scan`: once in the agent (`dispatchSubtasks` / `Run`) and again in `emitRunResult`.

Each metric is now recorded exactly once, in the agent's `Run`:

- **Duration**: a `defer` at the top of `Agent.Run` (review and scan), so it is recorded on every return path — successful, partial and failed runs alike. `emitRunResult` is skipped when `ag.Run` fails (e.g. `scan_cmd.go` returns early), so it cannot be the single recording site.
- **Comment count**: recorded once after dispatch, as before.
- `emitRunResult` no longer records either metric.

## Type of Change

## Tests

- `internal/telemetry/testing.go` adds `EnableMetricsForTest`, which routes metrics to a caller-supplied `MeterProvider` so other packages can assert what a code path records.
- `TestRun_RecordsOneDurationSamplePerRun` (scan): successful, partial (one file fails) and all-failed runs each record **exactly one** duration sample. It fails on the previous revision of this PR (0 samples).
- `TestEmitRunResult_RecordsNoRunMetrics`: `emitRunResult` records neither metric. It fails on `main` (the original double count).

## How Has This Been Tested?

## Checklist

- [x] I have signed the CLA


NanaseInori (changes requested):
Removing the duplicate recording is the right goal, but emitRunResult is not reached on every completed execution path.

In scan_cmd.go, an error from ag.Run(ctx) returns before emitRunResult. For example, when all dispatched file scans fail, dispatchSubtasks returns an error. Previously its deferred RecordReviewDuration still recorded that run; after this change the remaining recording site is skipped, so the failed scan contributes zero duration samples rather than one. This also biases the duration histogram by dropping failing runs.

Please put the single duration recording in a run-finalization path that is reached on both success and failure, while keeping result emission free of duplicate recording. Add regression coverage showing that successful, partial, and all-failed scans each contribute exactly one duration sample, with comment counts recorded only once where applicable.

Hiro-Mati:

The work is not lost if it is useful to you: the branch stays on my fork at https://github.com/Hiro-Mati/open-code-review/tree/fix/telemetry-double-count, with the regression test that fails without the fix. Feel free to take it over or reuse it under the project's license; no attribution needed.

PR #1434, opened 2026-09-18
fix(config): write config.json atomically

Hiro-Mati:
## Description

`saveConfig` wrote `~/.opencodereview/config.json` — which holds API keys — with `os.WriteFile`, which truncates first. An interruption mid-write could leave a truncated or empty config.

It now writes to a same-directory temp file (`os.CreateTemp`, created 0600), syncs, then renames over the target. A symlinked `config.json` is still written through to its target, and the temp file is removed on every failure path. Symlinks are resolved by `resolveSymlinkTarget`, which follows the link chain without requiring the final target to exist (unlike `filepath.EvalSymlinks`), so a link to a not-yet-created config keeps working as it did with `os.WriteFile`. Loops are bounded (40 hops).

## Type of Change

## How Has This Been Tested?

- New tests cover replacement without leftover temp files, temp cleanup when the final rename fails, writing through a symlink to an existing target and to a not-yet-created one (this last test fails on the previous revision), and symlink-loop detection. They guard the new code path; the crash-mid-write scenario itself is not simulated.

## Checklist

- [x] I have signed the CLA


NanaseInori (changes requested):
The temporary-file replacement is an improvement over truncating the active configuration, but there is a symlink compatibility regression in the reused path resolver.

When `config.json` is a symlink to `real.json`, the target's parent directory exists, and `real.json` has not been created yet, the previous os.WriteFile call follows the symlink and creates the target. resolveOutputCommitPath now calls filepath.EvalSymlinks, which fails because the final target does not exist, so saveConfig can no longer initialize that configuration. I reproduced this before/after behavior with the extracted file-writing helpers on Linux.

Please preserve write-through behavior for this case: resolve the link target without requiring the final file to exist, create the temporary file in the target directory, and leave the symlink itself intact. Add a regression test with an initially absent target; the current symlink test only covers an existing target.

lizhengfeng101:
Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/alibaba/open-code-review?pullRequest=1434) before we can accept your contribution.

Hiro-Mati:

The work is not lost if it is useful to you: the branch stays on my fork at https://github.com/Hiro-Mati/open-code-review/tree/fix/atomic-config-write, with the regression test that fails without the fix. Feel free to take it over or reuse it under the project's license; no attribution needed.

PR #1439, opened 2026-09-18
fix(allowlist): exclude pytest-style test_*.py from review

arif-js:
## Summary

Expanded Python's built-in exclude list to cover pytest's default test-file pattern: `**/test_*.py`

This prevents ordinary pytest test files from being selected for review and sent to the model. `conftest.py` is deliberately **not** excluded: pytest does not treat it as a test module (`python_files`), it is auto-loaded as a local
plugin that can implement collection/runtime hooks and shared fixture or environment setup, so it stays reviewable by default.

## Behavior

Before/after using `ocr review --preview` on the same changeset:

- Reviewed files: `6` -> `4`
- Excluded:
  - `app/test_handler.py`
  - `tests/test_api.py`
- Still reviewed:
  - `app/handler.py`
  - `app/contest_utils.py`
  - `conftest.py`
  - `src/test_helpers.go`

Users who want test files reviewed can add them to `include`, which overrides the default-path gate.

## Tests

Added coverage for the new matches and non-matches:

- Matches: `test_*.py`
- Non-matches: `conftest.py`, `contest_utils.py`, `attestation.py`, `latest.py`

Updated the assertions in `allowed_ext_test.go` that pinned the previous behavior.

## Docs

Added `**/test_*.py` to the enumerated default-exclusion list in `review-rules.md` for all maintained locales: `en`, `ja`, `ko`, `ru`, `zh`.

Verified:

- `make test` with race detection
- `make check`
- `make coverage` at 92.5%

## AI Disclosure

AI/LLM assistance: implemented with Claude Code; I reviewed every line.


internal/config/allowlist/default_exclude_patterns.json
```diff
@@ -8,6 +8,8 @@
   "**/test/**/*_test.py",
   "**/tests/**/*_test.py",
   "**/*_test.py",
+  "**/test_*.py",
+  "**/conftest.py",
```
wu21-web: I do not really recommend you exclude `**/confest.py`, it is true executable code in Python ...
They should be reviewed all the time.


```suggestion
  "**/conftest.py",
```

internal/config/allowlist/allowed_ext_test.go
```diff
@@ -174,12 +174,25 @@ func TestIsExcludedPath(t *testing.T) {
...
-		// Python test files
-		{"python test file", "tests/test_utils.py", false}, // pattern is *_test.py, not test_*.py
+		// Python test files. pytest collects both `test_*.py` and `*_test.py`
+		// (python_files), and conftest.py is test fixture code by definition.
+		{"python pytest prefix at root", "test_utils.py", true},
+		{"python pytest prefix nested", "app/test_handler.py", true},
+		{"python pytest prefix in tests dir", "tests/test_utils.py", true},
+		{"python pytest prefix deeply nested", "packages/core/app/test_api.py", true},
+		{"python conftest at root", "conftest.py", true},
+		{"python conftest nested", "tests/unit/conftest.py", true},
```
wu21-web: as above
```suggestion
		{"python pytest prefix at root", "test_utils.py", true},
		{"python pytest prefix nested", "app/test_handler.py", true},
		{"python pytest prefix in tests dir", "tests/test_utils.py", true},
		{"python pytest prefix deeply nested", "packages/core/app/test_api.py", true},
```

NanaseInori (changes requested):
The `**/test_*.py` addition looks correct to me. Pytest's default `python_files` patterns are `test_*.py` and `*_test.py`, and OCR already excludes the latter, so covering the prefix form closes a real gap. The positive and negative regression cases around that pattern also look good.

I do share the concern raised above about `**/conftest.py`, with one distinction: the issue is not merely that it is executable Python. Pytest does not treat `conftest.py` as one of its `python_files` test-module patterns; it auto-loads it as a local plugin. A conftest can implement collection/runtime hooks and shared fixture or environment setup, so excluding it by default removes pytest control/infrastructure code from review rather than simply completing the missing test-file naming convention.

For this focused fix, I would keep `**/test_*.py` and leave `conftest.py` reviewable by default unless the project explicitly decides that pytest plugin/fixture infrastructure should also be part of the built-in exclusion policy. That broader policy choice can be discussed separately if desired.

There is also a documentation update needed here. `review-rules.md` explicitly enumerates the built-in default path exclusions and explains their include-override behavior. After adding `test_*.py` (and `conftest.py` if it is ultimately retained), that list becomes stale. Please update the maintained documentation locales accordingly.

With those points addressed, I do not see another blocking issue in this patch.

This review was conducted by Qiyuanqiii's review bot, using the model GPT-5.6 Sol. If you need a human review, please manually @.

arif-js:
> The `**/test_*.py` addition looks correct to me. Pytest's default `python_files` patterns are `test_*.py` and `*_test.py`, and OCR already excludes the latter, so covering the prefix form closes a real gap. The positive and negative regression cases around that pattern also look good.
> 
> I do share the concern raised above about `**/conftest.py`, with one distinction: the issue is not merely that it is executable Python. Pytest does not treat `conftest.py` as one of its `python_files` test-module patterns; it auto-loads it as a local plugin. A conftest can implement collection/runtime hooks and shared fixture or environment setup, so excluding it by default removes pytest control/infrastructure code from review rather than simply completing the missing test-file naming convention.
> 
> For this focused fix, I would keep `**/test_*.py` and leave `conftest.py` reviewable by default unless the project explicitly decides that pytest plugin/fixture infrastructure should also be part of the built-in exclusion policy. That broader policy choice can be discussed separately if desired.
> 
> There is also a documentation update needed here. `review-rules.md` explicitly enumerates the built-in default path exclusions and explains their include-override behavior. After adding `test_*.py` (and `conftest.py` if it is ultimately retained), that list becomes stale. Please update the maintained documentation locales accordingly.
> 
> With those points addressed, I do not see another blocking issue in this patch.
> 
> This review was conducted by Qiyuanqiii's review bot, using the model GPT-5.6 Sol. If you need a human review, please manually @.

docs point fixed in [880e0f2](https://github.com/alibaba/open-code-review/pull/1439/commits/880e0f2bd0747cb711a0f277478d8d4fddbbbe13): `**/test_*.py` is added to the enumerated default-exclusion list in review-rules.md for all maintained locales (en, ja, ko, ru, zh).

internal/config/allowlist/default_exclude_patterns.json
arif-js: Agreed. Removed in [880e0f2](https://github.com/alibaba/open-code-review/pull/1439/commits/880e0f2bd0747cb711a0f277478d8d4fddbbbe13), so conftest.py is reviewable by default again.

internal/config/allowlist/allowed_ext_test.go
arif-js: Addressed in [880e0f2](https://github.com/alibaba/open-code-review/pull/1439/commits/880e0f2bd0747cb711a0f277478d8d4fddbbbe13).

wu21-web (approved):

NanaseInori (approved):

lizhengfeng101 (approved):
LGTM

PR #1444, opened 2026-09-18
feat(llm): support Claude on Google Vertex AI

### LINKED ISSUE #56: Support Claude on Google Vertex AI

stivenallan:
## Summary

Closes #56. Adds the `anthropic-vertex` protocol, mirroring the existing `anthropic-bedrock` support end to end: requests are authorized from [Application Default Credentials](https://cloud.google.com/docs/authentication/application-default-credentials) instead of an API key, and the region + project decide the host instead of a configured URL.

```bash
gcloud auth application-default login

ocr config set provider                     vertex
ocr config set model                        claude-sonnet-5
ocr config set providers.vertex.gcp_region  us-east5
ocr config set providers.vertex.gcp_project my-project
```

- `internal/llm`: new `ProtocolAnthropicVertex` canonical protocol, a `vertex` built-in preset (ambient auth, no `api_key`), and `NewAnthropicVertexClient` built on the official SDK's `vertex` package. Region and project are validated *before* the SDK is touched — `vertex.WithGoogleAuth` panics on an empty region and on unresolvable credentials, so both failure modes are turned into a deferred `initErr` instead of a stack trace, the same way `NewAnthropicBedrockClient` avoids `bedrock.WithLoadDefaultConfig`'s panic.
- `cmd/opencodereview`: `ocr config set` (`gcp_region`/`gcp_project` on both built-in and custom providers), the interactive `ocr config provider` wizard, and `ocr llm test` (reports region/project instead of a URL) all recognize the new protocol, following the exact pattern already used for bedrock.
- Fixes `providerAcceptsAWSSettings`, which gated on the preset's `AmbientAuth` flag alone. That was a correct discriminator only while bedrock was the sole ambient-auth preset — with vertex also `AmbientAuth: true`, the old check would have let `aws_region` silently apply (and do nothing) on a `vertex` provider entry. Added the mirror-image `providerAcceptsGCPSettings` and a two-directional regression test.
- Docs: adds a "Google Vertex AI" section to the English configuration guide, matching the existing "AWS Bedrock" section's structure. Left `ja`/`zh`/`ru`/`ko` translations untouched rather than machine-translate technical docs — happy to follow up if a maintainer/translator wants that split out.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .` (clean)
- [x] `go test ./...` — full suite green, including:
  - 16 new tests in `internal/llm/vertex_test.go` (protocol registration, resolver behavior, ambient-auth gating, client-construction failure modes for missing region/project/credentials, `GOOGLE_CLOUD_PROJECT` fallback)
  - 11 new tests in `cmd/opencodereview/vertex_config_test.go` (config round-trip, `ocr config set` field validation, the `providerAcceptsAWSSettings`/`providerAcceptsGCPSettings` cross-protocol regression, TUI ambient-auth flow)
  - 5 pre-existing tests updated where they asserted `anthropic-vertex` was rejected (they were placeholders anticipating this feature)


stivenallan:
Update: tested this live against a real Vertex AI project (billing enabled, API enabled, ADC login) and found a real bug in the first commit, now fixed in the latest push.

**The bug:** `NewAnthropicVertexClient` built the auth context with `context.WithTimeout(...)` + `defer cancel()`, copying the pattern `NewAnthropicBedrockClient` uses for `awsconfig.LoadDefaultConfig`. That pattern is safe for the AWS SDK because the returned `aws.Config` carries no reference to the context used to load it. It is **not** safe for `google.FindDefaultCredentials`: the context passed in gets embedded in the returned `Credentials.TokenSource`, which reuses it for every future token refresh, not just the one during construction. The deferred `cancel()` fired the instant the constructor returned, so the very first real request failed with:

```
Post "https://oauth2.googleapis.com/token": context canceled
```

Fixed by using `context.Background()` for that call instead of a canceled-after-construction context.

**What I verified live, end to end**, once that was fixed:
- `gcloud auth application-default login` → ADC picked up correctly
- `ocr config set provider vertex` / `gcp_region` / `gcp_project` → resolves and authorizes correctly
- Request reaches Google, model is recognized, and the request is correctly authorized/routed — it currently stops at a `429 RESOURCE_EXHAUSTED` from Google's own default (zero) quota for `anthropic-claude-sonnet` on a fresh project, which is a Google Cloud quota policy on their side (unrelated to this PR — same thing would happen calling the raw API directly), not something this code can work around.
- Also confirmed via Google's docs that `claude-sonnet-5` requires the `global` or multi-region (`us`/`eu`) endpoint rather than a specific region like `us-east5` (regional endpoints only support Sonnet 4.6 and earlier) — worth double-checking the built-in model list / docs example region against that if a maintainer wants the example to use a model that supports regional endpoints, or keep `global` as the recommended default (which is what I ended up documenting).

NanaseInori (changes requested):
Thanks for adding Vertex support and for explaining the credential-context fix. I reviewed `2a3b0893`; keeping the credential context alive after construction addresses the earlier first-request failure.

I have two remaining points to address before approval.

**Bound token refresh independently of the Messages request.**

The long-lived background context is appropriate for the token source, but the token exchange also needs a finite timeout. For user ADC credentials, the source retains that context, and the OAuth transport calls `Token()` synchronously before sending the model request. A stalled token endpoint can therefore keep the call blocked even after the Messages request deadline expires or its context is canceled.

Please keep the credential source usable after construction while separately bounding token-refresh HTTP requests. A mocked token-endpoint regression should cover both successful use after construction and a stalled refresh. This should not restore the constructor-level deferred cancellation that caused the previous bug. My local reproduction covered this context/transport boundary, not the complete SDK or a live Vertex request.

**Use supported model/endpoint combinations in the examples.**

The configuration guide still pairs `claude-sonnet-5` with `us-east5`, and the custom-provider example pairs it with `europe-west1`. The current official documentation places newer models on global or multi-region endpoints. Please use a supported pairing—for example, `global` for the general example and `eu` for the European multi-region example—or choose a model supported by the specific region. The sample output and PR description should be updated consistently.

Please also carry the provider-table addition into the locale docs required by the repository's provider rule, as already noted in the automated review.

The protocol registration, ambient-auth routing, and separation of AWS/GCP settings otherwise look reasonable. These requests are focused on making authentication failures bounded and the documented setup usable.

This review was conducted by Qiyuanqiii's review bot, using the model GPT-6 Astra Pro. If you need a human review, please manually @.

Qiyuanqiii (changes requested):
One process issue needs to be resolved before this PR can move forward.

The repository's AI-assisted development policy explicitly requires that:

1. AI/LLM use is disclosed in the initial issue or pull request, including the tools and models used.
2. AI/LLMs are **not attributed as commit authors or contributors**, including through `Assisted-by`, `Co-developed-by`, `Co-Authored-By`, or similar trailers.

Both commits in this PR currently contain:

`Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>`

Please **don't attribute an AI model as a co-author or contributor in this repository**. This directly conflicts with the project's contribution rules and needs to be corrected before the PR can be approved.

Please rewrite the commit history to remove these AI co-author trailers. The PR description also needs an explicit AI disclosure naming the actual tool(s) and model(s) used during development. Since that disclosure was not present when the PR was opened, please add it now and make sure future AI-assisted contributions disclose this upfront.

Please also confirm that you have personally reviewed and understand all AI-generated code and text in the final patch, as required by the contribution guidelines.

wu21-web:
I no longer think this is nessicary. That is a stale issue. Plus Google Vertex + Claude filters out almost every user. It is not worth the cost.

PR #1449, opened 2026-09-19
docs(assurance): correct API key storage claims

### LINKED ISSUE #1440: docs(assurance): ASSURANCE_CASE claims API keys are never stored in config files, but CLI persists api_key to config.json
## Summary
`ASSURANCE_CASE.md` says API keys are read exclusively from environment variables and are never embedded in config files. The CLI actually persists `api_key` / `auth_token` into `~/.opencodereview/config.json`.

This is a documentation / assurance-case accuracy issue (not an exploitable vulnerability by itself). Config writes use mode `0600`, which is good — but the assurance claim as written is incorrect and can mislead security reviewers / consumers of the assurance case.

### ASSURANCE_CASE.md claims
- Threat T2: Keys read from environment variables only; never logged, written to output files
- Fail-safe defaults: API keys must be explicitly provided via environment variables
- OWASP A07 row: API keys are read exclusively from environment variables, never embedded in code or config files, never logged

Source: https://github.com/alibaba/open-code-review/blob/main/ASSURANCE_CASE.md

### Product behavior
- CLI example: `ocr config set providers.anthropic.api_key "$ANTHROPIC_API_KEY"`
- `ProviderEntry` includes `api_key` JSON field
- `LlmConfig` includes `auth_token` JSON field
- Config path defaults to `~/.opencodereview/config.json` and is saved via `saveConfig`

Source: https://github.com/alibaba/open-code-review/blob/main/cmd/opencodereview/config_cmd.go

## Proposed fix
Update `ASSURANCE_CASE.md` (and any mirrored claims) to document the real model:

1. Keys may come from env vars, `api_key_cmd` / `auth_token_cmd`, or the user config file
2. When stored in config, files are written with restrictive permissions (`0600`)
3. Keys must not be logged / echoed in cleartext (masking already exists for `config set` display)

Optional: a one-line README security note pointing to the same precedence.

## AI disclosure
Assisted by passive static analysis tooling; all claims verified against public source on GitHub. I reviewed the evidence myself.

## Request
Please **assign this issue to me (@Erebuzzz) before I implement** so we avoid duplicate work. Happy to open a focused `docs/` PR after assignment.

Frank-zhu0404:
## Description

`ASSURANCE_CASE.md` claimed API keys are env-only and never stored in config files. The CLI actually persists `api_key` / `auth_token` to `~/.opencodereview/config.json` with mode `0600` (`saveConfig` in `cmd/opencodereview/provider_cmd.go`).

This PR aligns the assurance case with real credential sources:

1. Environment variables (`OCR_LLM_TOKEN`, provider-specific vars)
2. Credential commands (`api_key_cmd` / `auth_token_cmd`)
3. Optional plaintext persistence in the user config file (mode `0600`)

Also updates Threat T2, Fail-safe defaults, Separation of privilege, and OWASP A07 rows to match.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

Docs-only change. Verified claims against `saveConfig` / config path behavior and grepped that no other mirrored "env-only / never in config" assurance statements remain in `ASSURANCE_CASE.md`. Ran `go run scripts/verify-english-only.go` (pass).

Cross-checked threat table + Saltzer/Schroeder + OWASP A07 wording against CLI config persistence (`0600`) and masking on `ocr config set` display.

## Checklist

- [x] I have signed the CLA

AI disclosure: assisted by Grokbuild (Grok 4.6) for drafting the ASSURANCE_CASE wording; human-reviewed every line against CLI config persistence behavior. Codex (gpt-5.6-sol) used for read-only review only (VERDICT=PASS). No AI commit trailers.

## Related Issues

closes #1440


NanaseInori (changes requested):
The overall correction is needed, but the new `0600` assurance claim is still stronger than the current implementation guarantees.

`saveConfig` currently writes with `os.WriteFile(path, data, 0o600)`. In Go, that permission is used when the file is created; if the file already exists, `WriteFile` truncates and rewrites it without changing its existing permissions. So an existing `config.json` with broader permissions (for example `0644`) can remain broadly readable after `ocr config set` persists an API key.

Because this document is an assurance case, please either narrow the wording to say that newly created config files are created with mode `0600` (without implying existing permissions are tightened), or land an implementation that actually enforces `0600` on existing files before making the unconditional claim. The rest of the credential-source correction looks consistent with the documented behavior.

This review was conducted by NanaseInori's review bot, using the model GPT-5.6 Sol. If you need a human review, please manually @NanaseInori.

Frank-zhu0404:
Thanks @NanaseInori — good catch on `WriteFile` only applying mode on create.

Addressed in c992b9c:
- `saveConfig` already followed `os.WriteFile(..., 0o600)` with `os.Chmod(path, 0o600)`; added an explicit comment documenting why the chmod is required for existing files.
- Added `TestSaveConfigEnforcesModeOnExistingFile` so a seeded `0644` `config.json` becomes `0600` after save.
- Updated `ASSURANCE_CASE.md` to state that `saveConfig` enforces `0600` on create and on subsequent writes (including tightening a broader existing mode).

`go test ./cmd/opencodereview/ -run TestSaveConfig` passes.

Frank-zhu0404:
Follow-up to @NanaseInori's CHANGES_REQUESTED on the 0600 claim:

`saveConfig` now calls `os.Chmod(path, 0o600)` after every `WriteFile`, so existing broader modes (e.g. 0644) are tightened on subsequent persists — not only newly created files. Added `TestSaveConfigEnforcesModeOnExistingFile` and updated the assurance wording to match.

Commit: c992b9c. Ready for re-review.

NanaseInori (changes requested):
The new revision identifies the previous issue correctly, but the implementation still does not enforce the behavior that the documentation and the new regression test now claim.

`provider_cmd.go` adds a comment explaining that `WriteFile(..., 0o600)` does not tighten an existing file, but the actual `saveConfig` implementation still only calls `os.WriteFile(path, data, 0o600)`; there is no `os.Chmod(path, 0o600)` or equivalent permission-tightening step in the current diff.

At the same time, `TestSaveConfigEnforcesModeOnExistingFile` now seeds the config at `0644` and expects `0600` after `saveConfig`, so on Unix-like systems the new test should fail with the current implementation. `ASSURANCE_CASE.md` also now states that `saveConfig` always chmods existing files to `0600`, which is not yet true in code.

Please add the actual permission-enforcement step (preferably ensuring the file is restricted before sensitive data is exposed), then keep the regression test and assurance wording aligned with that implementation. Once that is in place, the original blocker should be resolved.

This review was conducted by NanaseInori's review bot, using the model GPT-5.6 Sol. If you need a human review, please manually @NanaseInori.

Frank-zhu0404:
Thanks for the re-review — I double-checked the PR head (`c992b9c`) and `alibaba/open-code-review@main`.

`saveConfig` already calls `os.Chmod(path, 0o600)` immediately after `os.WriteFile(...)` on both `main` and this branch (see `cmd/opencodereview/provider_cmd.go` around the `saveConfig` body). Commit `c992b9c` only added the clarifying comment + `TestSaveConfigEnforcesModeOnExistingFile`; it did not need to introduce a new Chmod call because that call was already present upstream.

I just re-ran locally:
```
go test ./cmd/opencodereview/ -run TestSaveConfigEnforcesModeOnExistingFile -count=1
--- PASS
```

Could you take another look at the full `saveConfig` function (not only the `c992b9c` hunk)? Happy to also harden further by chmod'ing an existing file *before* `WriteFile` so the mode is restricted before sensitive bytes land — say if you'd like that ordering change.

cmd/opencodereview/provider_cmd.go
```diff
@@ -425,6 +425,9 @@ func saveConfig(path string, cfg *Config) error {
 	if err != nil {
 		return fmt.Errorf("marshal config: %w", err)
 	}
+	// WriteFile applies 0o600 only when creating the file; existing files keep
+	// prior permissions. Chmod enforces 0600 on every persist (e.g. after
+	// ocr config set writes an API key into a previously world-readable config).
 	if err := os.WriteFile(path, data, 0o600); err != nil {
 		return fmt.Errorf("write config: %w", err)
 	}
```
NanaseInori, line 436: This keeps the fix minimal while ensuring an existing broadly-readable config is tightened *before* credentials are written. A missing file is still created as `0600` by `WriteFile`.

```suggestion
	// WriteFile applies 0o600 only when creating the file; existing files keep
	// their prior permissions. Tighten an existing config before writing any
	// credential material; a missing file will be created as 0600 below.
	if err := os.Chmod(path, 0o600); err != nil && !os.IsNotExist(err) {
		return fmt.Errorf("chmod config: %w", err)
	}
	if err := os.WriteFile(path, data, 0o600); err != nil {
		return fmt.Errorf("write config: %w", err)
	}
```

ASSURANCE_CASE.md
```diff
@@ -59,25 +59,35 @@ OCR is a CLI tool that:
...
 | T6 | Malicious LLM response | 2 | JSON schema validation on response structure; line number bounds checking against actual diff ranges |
 | T7 | Dependency vulnerabilities | All | `govulncheck` runs in CI; Dependabot monitors for updates; `go.sum` provides integrity verification |
 
+### Credential sources
+
+API keys are never hardcoded in the binary. They may be supplied at runtime from:
+
+1. Environment variables (`OCR_LLM_TOKEN`, provider-specific variables such as `ANTHROPIC_API_KEY`). Env-only use is supported: a complete `OCR_LLM_*` or Claude Code `ANTHROPIC_*` environment needs no config file.
+2. A credential command (`api_key_cmd` / `auth_token_cmd`) whose stdout is the secret. The command string may be stored in config; the secret itself is not.
+3. Optional persistence in the user config file (`~/.opencodereview/config.json`) as `api_key` or `auth_token`. `ocr config set` and interactive provider setup write this file in plaintext; `saveConfig` (`cmd/opencodereview/provider_cmd.go`) creates it with mode `0600` and always `chmod`s to `0600` afterward so an existing broader mode (for example `0644`) is tightened on every persist.
```
NanaseInori: If the implementation is changed to tighten permissions before writing, this wording should describe that ordering rather than claiming the `chmod` happens afterward.

```suggestion
3. Optional persistence in the user config file (`~/.opencodereview/config.json`) as `api_key` or `auth_token`. `ocr config set` and interactive provider setup write this file in plaintext; `saveConfig` (`cmd/opencodereview/provider_cmd.go`) creates new config files with mode `0600` and tightens existing config files to `0600` before persisting credentials.
```

NanaseInori (commented):
Concrete inline suggestions for the remaining permission-enforcement blocker:

This review was conducted by NanaseInori's review bot, using the model GPT-5.6 Sol. If you need a human review, please manually @NanaseInori.

Frank-zhu0404:
Pre-write `Chmod` landed at `9b3e9fc`: existing `config.json` is tightened to `0600` before `WriteFile` so sensitive bytes are not written while the file remains broadly readable; post-write `Chmod` is retained as defense in depth. `TestSaveConfigEnforcesModeOnExistingFile` still passes. Please re-review.

Qiyuanqiii:
@NanaseInori finish

cmd/opencodereview/provider_cmd.go
Frank-zhu0404: Applied your suggested pre-write `os.Chmod` + ASSURANCE wording at `6de3a94`.

Dropped post-write chmod to match the minimal suggestion.

`TestSaveConfigEnforcesModeOnExistingFile` still PASS — please re-review.

ASSURANCE_CASE.md
Frank-zhu0404: Applied your suggested pre-write `os.Chmod` + ASSURANCE wording at `6de3a94`.

Dropped post-write chmod to match the minimal suggestion.

`TestSaveConfigEnforcesModeOnExistingFile` still PASS — please re-review.

Frank-zhu0404:
@NanaseInori your suggested pre-write `os.Chmod` + ASSURANCE wording are landed at `6de3a94` (applied on top of `9b3e9fc`).

- `saveConfig`: tighten existing config to `0600` with `os.Chmod` before `WriteFile`; missing file still created as `0600`
- ASSURANCE_CASE updated to describe that ordering
- `TestSaveConfigEnforcesModeOnExistingFile` PASS

Please re-review this head. Once the bot blocker is clear, happy to request a human look from you as noted in the review footer.

NanaseInori:
> @NanaseInori finish

Acknowledged. Bot review will be temporarily skipped for this PR.

Qiyuanqiii:
My bot review hasn't been set up yet. Otherwise, it should end with @Qiyuanqiii.

cmd/opencodereview/provider_cmd_test.go
```diff
@@ -102,6 +102,44 @@ func TestSaveConfig(t *testing.T) {
...
+func TestSaveConfigEnforcesModeOnExistingFile(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Windows does not honor Unix file modes the same way")
+	}
+
+	dir := t.TempDir()
+	path := filepath.Join(dir, "config.json")
+	if err := os.WriteFile(path, []byte(`{"provider":"old"}`), 0o644); err != nil {
+		t.Fatalf("seed config: %v", err)
+	}
```
Qiyuanqiii, line 114: Could we explicitly call `os.Chmod(path, 0o644)` after creating the test file? With `umask 077`, `WriteFile` creates it as `0600`, so the permission check below fails before `saveConfig` runs. Setting the mode explicitly would keep the test independent of the local umask.

```suggestion
	if err := os.WriteFile(path, []byte(`{"provider":"old"}`), 0o644); err != nil {
		t.Fatalf("seed config: %v", err)
	}
	if err := os.Chmod(path, 0o644); err != nil {
		t.Fatalf("set seed permissions: %v", err)
	}
```

Qiyuanqiii (commented):
Thanks for pointing this out. You were right: our bot's earlier automated reviews missed the existing post-write `Chmod`. The claims that there was no permission-enforcement step and that the new test would necessarily fail were incorrect. Sorry for the confusion and the extra work this caused. Those claims should not block this PR.

At `6de3a94`, the full function now tightens permissions before writing, and the assurance wording matches that ordering. There are three separate follow-ups:

- **Test setup:** with `umask 077`, the new test creates its seed file as `0600`, then expects `0644`. A Linux probe reproduced that setup mismatch. The inline comment suggests explicitly setting the seed mode; this is the concrete test fix I'd recommend.
- **Directory handling (non-blocking):** calling `saveConfig` directly with a directory does change its mode to `0600` before the write fails. However, the CLI reads the config first and rejects a pre-existing directory before reaching `saveConfig`. A Linux CLI check confirmed the directory stayed at `0755`. A regular-file guard could improve the helper, but the pre-existing-directory scenario does not justify a blocking review.
- **Test coverage (non-blocking):** the test checks the final mode, so it passes with either pre-write or post-write chmod. It doesn't verify the ordering, but that is a coverage gap, not evidence that the current implementation is wrong.

The Linux checks above were focused probes, not a full Linux test-suite run.

Frank-zhu0404:
Thanks @Qiyuanqiii — appreciated the correction on the earlier Chmod miss, and the umask catch.

Applied your suggested seed fix at `2ffa3fe`: after `WriteFile(..., 0o644)` the test now calls `os.Chmod(path, 0o644)` so `TestSaveConfigEnforcesModeOnExistingFile` is independent of local umask. Local re-run still PASS.

Noted the two non-blocking follow-ups (directory-mode guard; ordering-aware coverage) — happy to take those in a follow-up if maintainers want them.

@NanaseInori per your review footer, requesting a **human review** on this head (`2ffa3fe`) now that the permission-enforcement blocker is cleared and the umask-stable seed is in.

cmd/opencodereview/provider_cmd_test.go
Frank-zhu0404: Done in `2ffa3fe` — `os.Chmod(path, 0o644)` after seed WriteFile. Test PASS under default umask.

Qiyuanqiii:
@NanaseInori Resume Review

Qiyuanqiii (approved):
LGTM

Qiyuanqiii (approved):
GPT once again mistook “function-level side effects” for a blocker that real commands can actually trigger, while missing the already-verified call chain. So I manually intercepted the message send. That said, apart from the non-blocking issues mentioned above, I have no further concerns.

NanaseInori (approved):

lizhengfeng101 (approved):
nice work

PR #1456, opened 2026-09-19
🏗️ feat: add architecture consistency review rules

cyberspace-cs:
## 背景

现有规则主要覆盖代码质量和安全，但缺少对**架构一致性**的检查——比如依赖方向是否正确、边界是否泄漏、循环依赖等。这对于大型项目尤其重要。

## 新增规则集

新增 `architecture_consistency.md` 规则文件，覆盖五大类架构检查：

| 类别 | 检查点 |
| --- | --- |
| **依赖方向** | 高层模块依赖低层实现细节、业务逻辑依赖框架类 |
| **边界泄漏** | 内部实现通过公共接口暴露、配置跨模块泄漏 |
| **循环依赖** | 模块互相 import、功能模块互相依赖共享工具 |
| **单一职责** | 一个模块处理多个不相关关注点、横切关注点分散 |
| **抽象层次不匹配** | 高抽象代码混低层次实现、调用方依赖具体实现而非接口 |

## 注册方式

在 `system_rules.json` 的 `path_rule_map` 中注册，匹配 `**/*.{py,go,java,ts,js,tsx,jsx}` 所有主要代码文件。

## 设计考虑

- **精准优先**：只在确认违反是真实且有影响时才标记
- **不重复现有规则**：现有的 dead code、安全检查等不重复
- **可扩展**：后续可以加更多架构检查点

## 价值

这个规则集让 OCR 不仅能检查代码质量，还能检查架构健康度，对大型项目更有价值。


internal/config/rules/system_rules.json
```diff
@@ -1,6 +1,10 @@
 {
   "default_rule": "default.md",
   "path_rule_map": {
+    "**/*.{py,go,java,ts,js,tsx,jsx}": "architecture_consistency.md",
```
chaojixinren: I think this changes the behavior more than intended. `path_rule_map` uses first-match-wins, so this broad pattern will cause most Go/Python/Java/TS/JS files to resolve only to `architecture_consistency.md`, instead of their existing language-specific rules.

For example, a `.go` file would no longer use `go.md`.

If the goal is to add architecture checks on top of the existing rules, I think this probably needs rule composition rather than another first-match path rule.

internal/config/rules/system_rules.json
```diff
@@ -1,6 +1,10 @@
...
+    "**/*llm*.{py,ts,js,go,java}": "llm_security.md",
+    "**/*agent*.{py,ts,js,go,java}": "llm_security.md",
+    "**/*prompt*.{py,ts,js,go,java}": "llm_security.md",
```
chaojixinren: Where does `llm_security.md` come from? I don't see this rule file in the repository or in this PR, and the PR description only mentions architecture consistency.

Was this part of another unfinished change and accidentally included here?

internal/config/rules/rule_docs/architecture_consistency.md
```diff
@@ -0,0 +1,30 @@
+> Architecture consistency review: check that code respects the project's architectural boundaries, dependency directions, and separation of concerns. Only flag when you are confident the violation is real and material; do not nitpick minor style differences that don't cross architectural boundaries.
```
chaojixinren: I think we should be more explicit that the reviewer must first establish the project's existing architectural conventions before reporting these issues.

Some examples below (domain/infra separation, service DTOs vs ORM objects, depending on interfaces) assume a particular layered/clean architecture style, which may not be intended by every project. Otherwise this could encourage the model to impose a new architecture rather than check consistency with the existing one.

chaojixinren:
I think this is an important change because it directly affects OCR's review rules and therefore changes what kinds of issues OCR will detect by default.

The idea of adding architecture consistency checks is valuable, but changes at this layer should be handled carefully since they can affect review behavior across many repositories and languages.

I've left inline comments for the concrete issues I found. Once those are addressed, I think this is worth continuing to improve.

PR #1471, opened 2026-09-20
fix(diff): let git decide which changed files are ignored

Hiro-Mati:
## Description

The built-in `.gitignore` matcher disagreed with git:

- `filepath.Match` uses `\` as separator on Windows, so `*` crossed `/`: `/*.go` dropped every `.go` change in subdirectories;
- a pattern with a slash in the middle (`config/local.json`) matched at any depth instead of being anchored;
- only the root `.gitignore` was read, so nested negations (`!*.gen.go`) were ignored and files git lists as not ignored were dropped;
- untracked names were `TrimSpace`d (`" lead.go"` became `"lead.go"`); the read then failed and the file vanished silently.

Changed files are now checked with one batched `git check-ignore --no-index -z --stdin` call; `--no-index` keeps the existing intent that ignore rules apply to tracked changes too. If git cannot answer, the old root-`.gitignore` matcher runs with a warning (now slash-correct on Windows). Untracked files are listed with `-z`, and an unreadable one produces a warning.

Behavior note: since git decides, `.git/info/exclude` and `core.excludesFile` now count as well, so a tracked change matching a user's global excludes is no longer reviewed. This is consistent with how the root `.gitignore` is already applied to tracked changes. Commit and range modes still use the working tree's ignore rules, as before.

## Type of Change

## How Has This Been Tested?

- New regression test(s) fail on `main` and pass with this change (verified both ways).
- `make check` passes; full unit suite passes on Windows 11 (Go 1.27) and in a `golang:1.26.5` Linux container replaying the CI job (`go test -race`, coverage >= 90%).

## Checklist

- [x] I have signed the CLA


Hiro-Mati:

The work is not lost if it is useful to you: the branch stays on my fork at https://github.com/Hiro-Mati/open-code-review/tree/fix/gitignore-matching, with the regression test that fails without the fix. Feel free to take it over or reuse it under the project's license; no attribution needed.

PR #1472, opened 2026-09-20
fix(diff): parse quoted and ambiguous diff --git headers

Hiro-Mati:
## Description

Two header shapes broke `diffHeaderRe` (`^diff --git a/(.+?) b/(.+)$`):

- git C-quotes paths containing `"`, `\` or control characters even with `core.quotepath=false` (`diff --git "a/b\"q.go" "b/b\"q.go"`). The regex never matched, so the file's headers and hunks were folded into the **previous** file's diff.
- an unquoted path containing ` b/` (e.g. `x b/y.go`) was split at the wrong place for non-renames; the file was reviewed with no content under a wrong path.

The new `parseDiffHeader` reads each side as a quoted or bare token, decodes git's C-style escapes, and for bare headers prefers the split where both halves are equal (`a/P b/P`). `rename from`/`rename to` values are unquoted too. A rename whose two different paths both contain ` b/` stays ambiguous in the header line, but `rename from/to` still give the right paths, as before.

## Type of Change

## How Has This Been Tested?

- New regression test(s) fail on `main` and pass with this change (verified both ways).
- `make check` passes; full unit suite passes on Windows 11 (Go 1.27) and in a `golang:1.26.5` Linux container replaying the CI job (`go test -race`, coverage >= 90%).

## Checklist

- [x] I have signed the CLA


Hiro-Mati:

The work is not lost if it is useful to you: the branch stays on my fork at https://github.com/Hiro-Mati/open-code-review/tree/fix/diff-header-quoted-paths, with the regression test that fails without the fix. Feel free to take it over or reuse it under the project's license; no attribution needed.

PR #1488, opened 2026-09-20
fix(viewer): shorten the session id cell on the session list

### LINKED ISSUE #1469: Viewer: don't display text `Session`, also unnecessary to display full session id on the session list page
### OpenCodeReview Version

v1.12.7 (85cecfe)

### Operating System

macOS (Apple Silicon)

### Installation Method

npm (global)

### LLM Provider

### Bug Description

On session list page, column `SESSION ID`, better not prefix text `Session: `, and better not display the full session id. Instead, just display the first part of session id, e.g. `8a3f01bc-...`.
Image

### Steps to Reproduce

Just open viewer and visit a session list page.

### Expected Behavior

Clearer and shorten text.

### Logs / Error Output

```shell

```

### AI/LLM Disclosure

### Additional Context

_No response_

basil-k-aji-dev:
Fixes #1469

The session list rendered `Session: ` followed by 30 characters of a 36-character UUID:

```
Session: b029c726-7b6b-46aa-b923-9fea9f…
```

Both halves are redundant. The column header already says `Session ID`, so the word is repeated on every row, and the tail of a UUID is not what a reader uses to tell two sessions apart — the first segment does that on its own.

It now renders:

```
b029c726…
```

## What does not change

- The full identifier stays in the `title` attribute, so hovering still shows it.
- It stays in the `href`, so the link target is unaffected.
- The column header keeps its name.

This is a one-line template change; no Go code is touched.

## Tests

`TestRenderTemplate_SessionsTableMockup` pinned the old string and now pins the new one. Reverting the template fails it with the rendered cell in the message:

```
--- FAIL: TestRenderTemplate_SessionsTableMockup
    rendered sessions page missing "<td class=\"col-session\">...>b029c726…</a></td>"
```

`go test ./internal/viewer/` passes, and `gofmt -l internal/viewer/` is clean.

## Not verified

I did not open the viewer in a browser on macOS, which is where @iredmail reported it. The change is in the template text only, so the rendered markup is what the test asserts, but I have not looked at the column width on a real screen.


lizhengfeng101 (approved):
LGTM

PR #1496, opened 2026-09-20
fix(scan): keep whitespace in filenames from git ls-files -z

### LINKED ISSUE #1493: bug(scan): TrimSpace corrupts valid filenames from git ls-files -z
### OpenCodeReview Version

main @ a003b9341a65130b024829101ea35494b56569e1

### Operating System

macOS (Apple Silicon)

### Installation Method

Built from source

### LLM Provider

OpenAI

### Bug Description

`gitLs` correctly requests NUL-delimited paths with `git ls-files -z`, but then applies `strings.TrimSpace(f)` to every pathname. Leading and trailing whitespace are valid filename data, not record formatting.

For example, a tracked file named ` normal.go` becomes `normal.go`. Later `Lstat`/read operations target the wrong path and the real tracked file is skipped, so a full scan can silently lose coverage.

### Steps to Reproduce

1. Create a Git repository containing a tracked Go file with a leading space:

   ```bash
   git init
   printf 'package p\n' > ' normal.go'
   git add . && git commit -m initial
   ```

2. Confirm Git preserves the leading byte in its NUL-delimited output:

   ```bash
   git -c core.quotepath=false ls-files -z | od -An -tx1c
   ```

3. Run the scan enumeration path (for example, `ocr scan`).
4. Observe that `gitLs` returns `normal.go` instead of ` normal.go`; the subsequent filesystem lookup targets a nonexistent path and the tracked file is skipped.

### Expected Behavior

NUL-delimited entries should be preserved exactly. After splitting on `\x00`, the implementation should discard only empty records and must not trim any pathname bytes. The scan should include files with valid leading or trailing whitespace.

### Logs / Error Output

```shell
The problematic transformation is:


for _, f := range files {
    f = strings.TrimSpace(f)
    if f != "" {
        result = append(result, f)
    }
}
```

### AI/LLM Disclosure

### Additional Context

Relevant source:
- https://github.com/alibaba/open-code-review/blob/a003b9341a65130b024829101ea35494b56569e1/internal/scan/provider.go#L263-L270
- https://github.com/alibaba/open-code-review/blob/a003b9341a65130b024829101ea35494b56569e1/internal/scan/provider.go#L100-L113

AI/LLM disclosure: This report was prepared with OpenAI Codex (GPT-5) for source inspection and issue drafting. The Git behavior was reproduced locally by the tool; no project code was modified.

shreyanshj10:
## Description

`gitLs` asks git for NUL-delimited paths with `ls-files -z`, which is exactly the format that makes odd filenames survive intact. It then ran `strings.TrimSpace` on every record, so any leading or trailing space in a pathname was thrown away before the path was used.

A tracked file named ` spaced.go` came back from `gitLs` as `spaced.go`. That path does not exist on disk, so the `os.Lstat` in `Enumerate` failed, the file was skipped with a "cannot stat" warning, and the scan quietly lost coverage of it.

The fix is to stop trimming. After splitting on `\x00` the only record worth discarding is the empty one left by the trailing NUL, and that filter was already there. Nothing else in the function changes.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

I added `TestProvider_Enumerate_KeepsWhitespaceInFilenames` to `internal/scan/provider_test.go`. It commits a file named ` spaced.go` alongside a plain one and asserts both come back from `Enumerate`. On `main` it fails with:

```
[ocr] WARNING: cannot stat spaced.go: ... The system cannot find the file specified.
--- FAIL: TestProvider_Enumerate_KeepsWhitespaceInFilenames
    provider_test.go:270: paths = ["plain.go"], want [" spaced.go" "plain.go"]
```

and it passes with this change. I also ran `go test -count=1 ./internal/...`, `go vet`, `gofmt -s`, and the license and english-only check scripts.

The test uses a leading space only. A trailing space is equally valid data to fix here, but Windows strips it from a path component, so a trailing-space fixture would not survive the Windows CI job.

## Checklist

- [ ] I have signed the CLA

AI/LLM disclosure: I used Claude (Opus) as an assistant while working on this change. I read and reviewed every line myself, and I will answer review questions in my own words.

## Related Issues

Closes #1493


lizhengfeng101:
Thanks for this — you correctly identified that `strings.TrimSpace` on each `-z` record silently drops filenames with leading/trailing spaces, defeating the whole point of the NUL-delimited format.

We've merged #1497, which removes that `TrimSpace` call and also fixes a second defect in the same `gitLs` function (the runner path merging stderr into stdout via `CombinedOutput`). As #1497 covers both issues, I'm closing this PR as superseded.

Thanks again for the detailed explanation and the test case. 🙏

PR #1497, opened 2026-09-20
fix(scan): keep NUL-delimited pathnames from git ls-files intact

### LINKED ISSUE #1493: bug(scan): TrimSpace corrupts valid filenames from git ls-files -z

basil-k-aji-dev:
## Description

`gitLs` asks git for NUL-delimited paths and then runs `strings.TrimSpace` over every record. Leading and trailing spaces are filename bytes, not record framing — `-z` exists precisely so that they need no escaping. A tracked ` normal.go` became `normal.go`, the later `Lstat` found nothing, and the file left the scan on a warning. A full scan reports success having reviewed one file fewer than the repository holds.

While fixing that I found a second instance of the same defect in the same function. `gitLs` takes stdout only on its `p.runner == nil` branch, with a comment explaining why it must:

```go
// Use Output (stdout only), not CombinedOutput: with -z, git emits
// NUL-delimited paths on stdout, and merging stderr in would corrupt
// the filename parsing below.
```

The other branch — the one every real `ocr scan` takes, since `agent.go:319` passes `a.args.GitRunner` — called `Runner.Run`, which is `CombinedOutput`. `Runner.Output` already existed and is now used instead.

That one is worse than it sounds. With `-z` there is no line structure to resynchronise on, so a warning git writes while still exiting 0 does not arrive as a stray record: it is glued to the front of the first real pathname. In the test fixture, one unreadable directory costs the scan *every* file, not one.

## Type of Change

## How Has This Been Tested?

Two new tests in `internal/scan/provider_paths_test.go`, both verified to fail without the patch:

| test | without the patch |
|---|---|
| `TestProvider_Enumerate_PreservesWhitespaceInTrackedPaths` | `enumerated ["plain.go"], want [" leading.go" "plain.go" "trailing.go "]` |
| `TestProvider_Enumerate_StderrWarningIsNotParsedAsAPath` | `enumerated [], want exactly ["kept.go"]` |

The first runs against a real repository through both the `nil` and injected-`Runner` paths. The second makes `git ls-files --others` warn and still exit 0 by way of a mode-0000 directory, which is the shape that matters — on a non-zero exit `gitLs` returns the error and parses nothing.

Both skip on Windows, which normalises away the filenames the fixtures depend on; the second also skips as root, who reads a 0000 directory regardless of its mode. The parsing under test is platform-independent.

One caveat on my local run: `make test` uses `-race`, and this machine's kernel has a 47-bit VMA where ThreadSanitizer needs 48, so every package aborts before running. I ran the suite as `go test -count=1 ./...`, which is green, plus `make check` and `make english-check`. Race coverage on these changes is therefore untested on my side and rests on CI.

## Checklist

- [ ] I have updated the documentation accordingly (if applicable) — no user-facing surface changes
- [x] I have signed the CLA

AI/LLM disclosure: written with AI assistance — Claude Code, model Claude Opus 5. I reviewed every line, and each of the two defects is backed by a test I ran against unpatched `main` to confirm it fails there first. The reasoning above and in any review thread is my own.

## Related Issues

Closes #1493


lizhengfeng101 (approved):
LGTM

PR #1491, opened 2026-09-20
docs(review): synchronize exclude & secret patterns

wu21-web:
## Description
The documentation for review rules are currently stale, discovered while reviewing #1439 

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring (no functional changes)
- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] `make test` passes locally
- [x] Manual testing (describe below)
Ran `prettier`, which didn't produce any diff.
## Checklist

- [x] My code follows the project's coding style (`go fmt`, `go vet`)
- [x] I have performed a self-review of my code
- [ ] I have added tests that prove my fix is effective or my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have updated the documentation accordingly (if applicable)
- [ x I have signed the CLA
- [x] I did not use AI/LLM to create this PR, or I disclosed the tool/model below and reviewed its output; I did not attribute commits to AI and will answer maintainer questions and review comments myself without AI/LLM.
AI used as a translator.

## Related Issues

n/a

PR #1490, opened 2026-09-20
feat(session): add 'ocr session rm' to delete a saved session

basil-k-aji-dev:
Closes #1468

Adds `ocr session rm` to delete one persisted session from `~/.opencodereview/sessions/`.

```bash
ocr session rm 9f2c1b4a-7e35-4d61-b2f0-6c8a41d9e72b
ocr session rm 9f2c1b4a-7e35-4d61-b2f0-6c8a41d9e72b --yes
ocr session rm 9f2c1b4a-7e35-4d61-b2f0-6c8a41d9e72b --repo ~/work/my-project
```

@iredmail asked for a delete button on the Viewer's session list. @Qiyuanqiii [declined that](https://github.com/alibaba/open-code-review/issues/1468#issuecomment-5747553000) on the grounds that the Viewer's read-only property is a tested invariant, and @iredmail has since [accepted the shell command](https://github.com/alibaba/open-code-review/pull/1490#issuecomment-5750271985) as the resolution.

## Behaviour

**The id is enough on its own.** The command searches every repository, so it runs from any directory. `--repo` narrows the search rather than enabling it. This came from @wu21-web and @Qiyuanqiii on this PR; the first version required `--repo` or the right working directory, which pushed an implementation detail of the storage layout onto the caller.

**An ambiguous id is listed, not guessed at.** `encodeRepoPath` replaces separators with `-`, so two repository paths can share one sessions directory. Where an id resolves to more than one file, the candidates are printed and nothing is deleted.

**The summary is printed before the prompt**, in the same format `ocr session list` uses, so the answer is given against the branch, start time and file count rather than an opaque id.

**A non-interactive stdin answers no.** A pipeline or CI job sees EOF immediately, so those callers pass `--yes` deliberately.

**A corrupt session is still deletable**, since that is the record most likely to want removing. One that cannot be *read* is reported instead.

## Where review attention is worth spending

**The destructive primitive enforces its own boundary.** `DeleteSessionAt` rebuilds the target from validated parts rather than trusting the `Location` it is handed, and refuses a `Location` whose `Path` disagrees with its own fields. Found by @NanaseInori: the earlier version checked only the basename, and a `Location` with `../..` in `EncodedDir` deleted a file two directories above the sessions root and returned `nil`.

[...]

PR #1499, opened 2026-09-20
fix(config): exclude dependency and build-output directories by default

### LINKED ISSUE #1494: fix(config): cover dependency and build-output directories in default_exclude_patterns.json
## Description

`default_exclude_patterns.json` currently covers test files, snapshots and generated code, but not **dependency directories or build output**. The most direct evidence is line 17: `**/oh_modules/**` (the HarmonyOS dependency directory) is there, while `**/node_modules/**` is not.

Two reasons `.gitignore` does not cover this:

1. **Once a directory is committed, `git diff` stops looking at `.gitignore`.** Reviewing that commit — or any range spanning it — pulls in every file. This repository has a real case: the root `.gitignore` had `/node_modules` (anchored to the root), so `extensions/frontend/node_modules` was never ignored and got committed in c640c109. A single review picked up 389 files before it was removed in 27c43fa8.
2. **The extension allowlist does not help.** Files inside these directories are `.js`, `.ts`, `.json`, `.css` — all in `supported_file_types.json`, so they pass the extension gate in `whyExcluded` (`internal/agent/selection.go:92`) and reach the default-path gate on the next line.

Worth knowing before adding entries: `.map`, `.lock`, `.sum`, `.class` and `.jar` are **not** in the extension allowlist, so source maps, `yarn.lock`, `Cargo.lock`, `go.sum` and compiled artifacts are already excluded upstream — no pattern needed. But `package-lock.json` and `pnpm-lock.yaml` resolve to `.json` / `.yaml`, which *are* allowed, so those need explicit path patterns.

## Scope

- File: `internal/config/allowlist/default_exclude_patterns.json`
- Tests: `internal/config/allowlist/allowed_ext_test.go`

Add at least the following. Matching is case-insensitive and uses doublestar globs, so `**` crosses directory boundaries and handles nesting at any depth:

```
JS/TS deps and output
  **/node_modules/**
  **/bower_components/**
  **/.pnpm-store/**
  **/.yarn/{cache,unplugged,releases,sdks,patches}/**
  **/package-lock.json
  **/pnpm-lock.yaml
  **/npm-shrinkwrap.json
  **/*.min.{js,css}
  **/dist/**
  **/.next/**
  **/.nuxt/**
  **/.svelte-kit/**
  **/.astro/**
  **/.turbo/**

Vendored deps
  **/vendor/**
  **/.bundle/**

JVM
  **/target/**
  **/.gradle/**

Python
  **/__pycache__/**
  **/.venv/**
  **/venv/**
  **/site-packages/**
  **/*.egg-info/**
  **/.tox/**
  **/.mypy_cache/**
  **/.pytest_cache/**
  **/.ruff_cache/**

Swift/iOS
  **/Pods/**
  **/Carthage/**
  **/.build/**

Other
  **/obj/**
  **/.dart_tool/**
  **/.terraform/**
  **/coverage/**
```

`**/vendor/**` supersedes the existing `**/vendor/**/*.{jsonnet,libsonnet}` on line 40 — drop that line.

**Additions beyond this list are welcome.** If you know a dependency or build-output directory from an ecosystem we missed, include it and say which toolchain produces it. Two names that are deliberately left out: `**/build/**` and `**/bin/**` — plenty of projects keep hand-written sources there (this repository's own `bin/ocr.js`, for one), so the false-exclusion risk outweighs the benefit.

## Acceptance Criteria

- [ ] Patterns above are present in `default_exclude_patterns.json`, with the redundant `vendor` jsonnet line removed
- [ ] A nested path such as `extensions/frontend/node_modules/@babel/core/lib/index.js` is reported as excluded by `IsExcludedPath`
- [ ] Test cases cover nesting at depth, plus at least one non-excluded lookalike (e.g. `src/distribution/index.js` must NOT be excluded by `**/dist/**`)

## Context

Found while reviewing a session that had 389 `extensions/frontend/node_modules/**` files in it.

`internal/diff/git.go` has a separate blocklist (`providerDirIgnoreDirs`) that drops these directories earlier, before diffs are even parsed. It only matches at the repository root, which is its own bug — out of scope here. This issue is about the configurable default list, which already has the right glob semantics for the job.

basil-k-aji-dev:
## Description

Adds the dependency and build-output patterns from #1494 to `default_exclude_patterns.json`, and drops the now-redundant `**/vendor/**/*.{jsonnet,libsonnet}` line.

I checked the two premises in the issue before writing anything, because both are what make this worth doing rather than a `.gitignore` problem:

- **`.gitignore` does not cover it.** Once a directory has been committed, `git diff` stops consulting `.gitignore` for it. `c640c109` committed `extensions/frontend/node_modules` because the root `.gitignore` had `/node_modules` anchored to the root; `27c43fa8` removed it, but reviewing either commit, or any range spanning them, still pulls in all 389 files.
- **The extension allowlist does not cover it.** I confirmed against `supported_file_types.json`: `.map`, `.lock`, `.sum`, `.class` and `.jar` are absent, so source maps, `yarn.lock`, `Cargo.lock`, `go.sum` and compiled artifacts are already stopped at the extension gate and need no pattern. But `.json`, `.yaml`, `.js`, `.ts`, `.css`, `.cjs`, `.mjs` and `.hcl` are all present, so `package-lock.json`, `pnpm-lock.yaml` and everything inside these directories reaches the path gate unfiltered.

### One behaviour change that needs a decision, not just a review

`**/vendor/**` is broader than what it replaces, and the narrow pattern was narrow on purpose. `allowed_ext_test.go` said so:

```go
// The pattern is extension-scoped: IsExcludedPath applies every pattern to every
// path, so a bare **/vendor/** would also drop vendored Go and PHP sources.
{"go under vendor still reviewed", "vendor/github.com/pkg/errors/errors.go", false},
{"php under vendor still reviewed", "vendor/monolog/monolog/src/Logger.php", false},
```

The issue asks for `**/vendor/**` and for that line to go, so those two cases now assert `true` and carry a comment saying #1494 reversed the earlier decision deliberately. Flagging it here rather than burying it: this is the one change in the PR that stops something being reviewed that used to be, and if the intent was only to widen it for jsonnet, this is the line to push back on.

### Additions beyond the issue's list

Six, each from a toolchain that writes the directory and no hand-written source:

| pattern | written by |
|---|---|
| `**/.angular/**` | Angular CLI build cache |
| `**/.parcel-cache/**` | Parcel bundler cache |
| `**/.docusaurus/**` | Docusaurus build cache |
| `**/.stack-work/**` | Haskell Stack build output |
| `**/.pnp.cjs`, `**/.pnp.loader.mjs` | Yarn PnP, generated — `.cjs`/`.mjs` are both in the allowlist |
| `**/.terraform.lock.hcl` | Terraform, and it sits *outside* `.terraform/` so the directory pattern misses it — `.hcl` is in the allowlist |

`**/build/**` and `**/bin/**` stay out, per the issue. This repository's own `bin/ocr.js` is the counter-example, and both are pinned as *not* excluded so a later well-meaning addition fails the test rather than silently dropping files.

## Type of Change

## How Has This Been Tested?

Against the acceptance criteria:

- `extensions/frontend/node_modules/@babel/core/lib/index.js` is excluded, as are `node_modules/left-pad/index.js` and `a/b/c/d/node_modules/pkg/dist/index.js` — root, workspace and depth.
- Ten lookalikes assert the opposite, because a directory pattern matching a prefix of a segment instead of the whole segment is how this kind of change quietly deletes coverage: `src/distribution/index.js`, `src/objects/model.ts`, `app/targeting/rules.go`, `src/vendors/stripe.ts`, `ios/podsmith/Helper.swift`, `tools/venvironment/setup.py`, `app/coverages/report.ts`, `src/node_modules_helper.ts`, `src/minified.ts`, `src/package-lock-utils.ts`.
- 45 positive cases in total, one per ecosystem, so a later edit to any single pattern fails a named test rather than the whole block.

One existing test outside this package needed updating and I want it visible rather than folded into the diff: `TestWhyExcluded_UserIncludePattern` used `vendor/baz.go` purely as "a path outside the include patterns that nothing else excludes", to demonstrate that include is additive. `**/vendor/**` invalidates the fixture, not the assertion, so the path is now `cmd/baz.go` with a comment saying why it changed.

Local run caveat: `make test` uses `-race`, and this machine's kernel has a 47-bit VMA where ThreadSanitizer needs 48, so every package aborts before running. I ran `go test -count=1 ./...`, which is green, plus `make check` and `make english-check`.

## Checklist

- [ ] I have updated the documentation accordingly (if applicable) — the pattern-syntax docs in `allowed_ext.go` are unchanged and still accurate
- [x] I have signed the CLA

AI/LLM disclosure: written with AI assistance — Claude Code, model Claude Opus 5. I reviewed every pattern individually, checked each one against `supported_file_types.json` and the existing tests rather than assuming, and found the `vendor` reversal and the `TestWhyExcluded_UserIncludePattern` fixture by running the suite. The reasoning above and in any review thread is my own.

## Related Issues

Closes #1494


Qiyuanqiii (commented):
The implementation itself looks solid to me. The new patterns use the existing path-matching semantics cleanly, the nesting/lookalike coverage is good, and CI/CodeQL are green.

I do think the documentation needs to move with this change before merge, though.

The current Viewer/rule documentation does not merely omit these new patterns; it describes a different filtering model. In particular, the existing `review-rules.md` / `architecture.md` text says that `default_path` is primarily the built-in test-file exclusion gate, while noisy directories such as `vendor/`, `node_modules/`, and `target/` are filtered earlier by the diff provider.

After this PR, that is no longer generally true. Nested dependency/build directories such as:

- `extensions/frontend/node_modules/...`
- `foo/vendor/...`
- `foo/target/...`

can now reach the per-file filter and be excluded by `default_path`. That distinction is user-visible because a user `include` rule can override a default-path exclusion, whereas the provider-level root-directory blocklist cannot be overridden.

There is also a sequencing issue with #1491: that PR is currently synchronizing the five localized `review-rules.md` files to the pre-#1499 exclude list. If #1491 lands first and this PR lands unchanged afterwards, the documentation becomes stale again immediately.

Could we update the relevant documentation after rebasing on #1491 (or otherwise coordinate the two PRs), including the five localized `review-rules.md` files and the corresponding `architecture.md` explanation of the filtering layers?

I don't have a code-level blocker beyond that. The broader `**/vendor/**` behavior is a real coverage change, but #1494 explicitly asks for that pattern, so I consider that a maintainer policy decision rather than an implementation defect in this PR.

basil-k-aji-dev:
@Qiyuanqiii you were right that this was not an omission but a contradiction, and the docs are updated. Pushed.

**What was wrong.** Both documents said `default_path` matches built-in *test-file* patterns, and that noisy directories are handled earlier at the diff level. I checked `ProviderDirPrefix` before writing anything, and your reading is exact:

```go
if relPath == dirPart || strings.HasPrefix(relPath, prefix) {
```

`prefix` is `"vendor/"`, so it matches only at the repository root. `vendor/pkg/x.go` never reaches the per-file filter; `api/vendor/pkg/x.go` always did reach it, and after this PR it is excluded there by `default_path`.

**The user-visible half is the part I would have under-documented without your review.** The two outcomes are not interchangeable: `provider_directory` cannot be brought back by an `include` rule and `default_path` can. Both documents now say which path reports which, with that exact pair of examples, rather than leaving a reader to infer it.

Changed in five locales each: `review-rules.md` gains the dependency and build-output group beside the test group, and corrects the "happens earlier" sentence to say root-only. `architecture.md` gets the same correction in its filtering-layers section, and gate 5 no longer says **test-file**.

**On sequencing with #1491.** You are right that whichever of us lands second makes the other stale. I have not rebased onto it, and I would rather say why than do it silently: #1491 is @wu21-web's and is still open, so rebasing onto an unmerged branch would put their commits inside my PR and make this one unreviewable for the thing it is actually about.

What I have done instead is make my change additive to the same sections, so the conflict is small and mechanical. My preference, if it helps you sequence them:

- **#1491 first**, then I rebase and take their list as the base, keeping only my two group additions and the root-only correction. I am happy to do that rebase the day it lands.
- **This one first** is also fine, but then #1491 needs to pick up the new patterns before it merges, which is more work for @wu21-web than it is for me.

Either way I will follow whichever order you pick rather than race it.

@wu21-web, since this touches the five files you are already syncing: if you would rather carry the pattern-list half in #1491 yourself and leave me only the root-only correction, say so and I will cut mine back. You have also just told me on #1490 that my translations need work, which I accept, so I would rather you had the option than have me push four more locales past you.

**On `**/vendor/**`.** Agreed that it is a policy call rather than a defect, and I flagged it in the PR description for the same reason: it stops vendored Go and PHP sources being reviewed, which `allowed_ext_test.go` previously asserted on purpose. #1494 asks for the pattern explicitly, so I implemented it and made the reversal visible rather than quiet, but it is @lizhengfeng101's to confirm.

wu21-web:
You are so thoughtful. Yes indeed, you have to pick up the new docs session after my PR gets merged.

basil-k-aji-dev:
Agreed, #1491 first. Once it lands I will rebase onto your list and keep only the two additions this PR needs: the dependency and build-output group, and the correction that the diff-level blocklist matches at the repository root only.

Ping me when it merges, or I will pick it up when I see it.

lizhengfeng101:
@basil-k-aji-dev please rebase main

Qiyuanqiii:
ci failing

Qiyuanqiii (approved):
LGTM

pages/src/content/docs/en/review-rules.md
```diff
@@ -189,10 +191,27 @@ generated code, and vendored dependencies:
...
+…and dependency or build-output directories:
+
+- `**/node_modules/**`
+- `**/bower_components/**`
+- `**/vendor/**`
+- `**/target/**`
+- `**/dist/**`
+- `**/__pycache__/**`, `**/.venv/**`, `**/site-packages/**`
+- `**/Pods/**`, `**/Carthage/**`
+- `**/.next/**`, `**/.nuxt/**`, `**/.gradle/**`, `**/.terraform/**`, …
```
wu21-web, line 203: I suggest adding these to the list already there instead of creating a new list per translation, because they are all in the same json file.

Qiyuanqiii:
@NanaseInori review

internal/config/allowlist/default_exclude_patterns.json
```diff
@@ -55,5 +54,52 @@
...
+  "**/Pods/**",
+  "**/Carthage/**",
+  "**/.build/**",
+
+  "**/obj/**",
+  "**/.dart_tool/**",
+  "**/.terraform/**",
+  "**/.terraform.lock.hcl",
+  "**/.stack-work/**",
+  "**/coverage/**"
```
lizhengfeng101: `**/coverage/**` also swallows a hand-written domain directory named `coverage` (for example an insurance/health module at `src/coverage/plan.ts`). Matching is case-insensitive and segment-exact, so the `coverages` lookalike test does not guard this case — the plural is `false` under both patterns. Narrowing to istanbul/nyc's report subdir keeps the artifact excluded while leaving business code reviewable. The existing positive case `coverage/lcov-report/index.js` still matches.

```suggestion
  "**/coverage/lcov-report/**"
```

internal/config/allowlist/allowed_ext_test.go
```diff
@@ -391,6 +393,75 @@ func TestIsExcludedPath(t *testing.T) {
...
+
+		// Lookalikes: a directory pattern must match a whole path segment, not
+		// a prefix of one, or these ordinary sources disappear from review.
+		{"distribution is not dist", "src/distribution/index.js", false},
+		{"objects is not obj", "src/objects/model.ts", false},
+		{"targeting is not target", "app/targeting/rules.go", false},
+		{"vendors is not vendor", "src/vendors/stripe.ts", false},
+		{"podsmith is not Pods", "ios/podsmith/Helper.swift", false},
+		{"venvironment is not venv", "tools/venvironment/setup.py", false},
+		{"coverages is not coverage", "app/coverages/report.ts", false},
```
lizhengfeng101, line 456: Pin the singular case too, so a future revert to `**/coverage/**` fails a named test rather than silently dropping the file. The existing `coverages` case cannot catch this — the plural is `false` under both patterns.

```suggestion
		{"coverages is not coverage", "app/coverages/report.ts", false},
		{"coverage as a domain dir is reviewed", "src/coverage/plan.ts", false},
```

NanaseInori (commented):
Re-reviewed the current head (`4410d29`).

The implementation and regression coverage still look good to me, and the earlier documentation issue about the provider-level versus `default_path` filtering semantics has been addressed correctly.

I agree with @wu21-web's documentation comment, and I think there is a concrete consistency issue behind it rather than only a formatting preference.

After the rebase, the localized `review-rules.md` files currently document both:

`**/vendor/**/*.{jsonnet,libsonnet}`

and:

`**/vendor/**`

However, `default_exclude_patterns.json` removes the former and replaces it with the latter. So the documentation now describes an exclusion pattern that no longer exists in the source of truth.

This appears to have been left behind while resolving the documentation conflicts with #1491: the existing synchronized list was retained, and the new dependency/build-output patterns were added as a separate list.

I suggest applying this PR's pattern changes directly to the existing list, as @wu21-web proposed, across the maintained locales. In particular, the old `**/vendor/**/*.{jsonnet,libsonnet}` entry should be removed when `**/vendor/**` is added, so the documented list continues to match `default_exclude_patterns.json`.

I don't see another implementation blocker at the current head. The filtering behavior, nested-path coverage, lookalike regressions, and provider/default-path distinction look coherent, and the current CI, CodeQL, Pages, translation-sync, and plugin-contract checks are green.

Requesting changes only for the documentation/source-of-truth mismatch above.

lizhengfeng101 (approved):
LGTM

basil-k-aji-dev:
@NanaseInori @wu21-web your two comments landed a few minutes before this merged, so I have taken both in #1541.

@NanaseInori, you were right: the localized `review-rules.md` files still documented `**/vendor/**/*.{jsonnet,libsonnet}` after the rebase removed it from the json, so the page described an exclusion that no longer exists.

@wu21-web, also right that one json file should not appear as two lists.

Checking both turned up something neither of you had reason to look for. The page listed **74** of the **97** patterns in the file. That gap predates this PR; the list is hand-maintained and has fallen behind more than once.

So #1541 generates the single list from the json, and adds a test that fails when the two disagree in either direction. I checked it against both before relying on it, and the stale-entry direction reproduces @NanaseInori's finding exactly:

```
en/review-rules.md lists "**/vendor/**/*.{jsonnet,libsonnet}",
which is not in default_exclude_patterns.json
```

That test would have caught this before merge, which is the part worth having.

@lizhengfeng101, your `**/coverage/**` narrowing and the `src/coverage/plan.ts` pin are both on main and unchanged here.

PR #1500, opened 2026-09-20
fix(config): cover dependency and build-output directories in default_exclude_patterns.json

### LINKED ISSUE #1494: fix(config): cover dependency and build-output directories in default_exclude_patterns.json

jimpablo:
## Description

`default_exclude_patterns.json` covered test files, snapshots and generated code, but not dependency directories or build output: `**/oh_modules/**` was excluded while `**/node_modules/**` was not. `.gitignore` does not backstop this — once such a directory is committed, any review spanning that commit pulls in every file inside it (this repository's own `extensions/frontend/node_modules` incident put 389 files into one session). The extension allowlist does not help either, because these directories are full of `.js`/`.ts`/`.json`/`.css` files.

This adds the dependency and build-output patterns from the issue:

- JS/TS: `**/node_modules/**`, `**/bower_components/**`, `**/.pnpm-store/**`, `**/.yarn/{cache,unplugged,releases,sdks,patches}/**`, `**/package-lock.json`, `**/pnpm-lock.yaml`, `**/npm-shrinkwrap.json`, `**/*.min.{js,css}`, `**/dist/**`, `**/.next/**`, `**/.nuxt/**`, `**/.svelte-kit/**`, `**/.astro/**`, `**/.turbo/**`
- Vendored: `**/vendor/**` (wholesale now — supersedes the jsonnet-scoped line), `**/.bundle/**`
- JVM: `**/target/**`, `**/.gradle/**`
- Python: `**/__pycache__/**`, `**/.venv/**`, `**/venv/**`, `**/site-packages/**`, `**/*.egg-info/**`, `**/.tox/**`, `**/.mypy_cache/**`, `**/.pytest_cache/**`, `**/.ruff_cache/**`
- Swift/iOS: `**/Pods/**`, `**/Carthage/**`, `**/.build/**`
- Other: `**/obj/**`, `**/.dart_tool/**`, `**/.terraform/**`, `**/coverage/**`

`**/build/**` and `**/bin/**` are deliberately left out, per the issue.

Test changes in `internal/config/allowlist/allowed_ext_test.go`: nesting-at-depth positives per group — including `extensions/frontend/node_modules/@babel/core/lib/index.js` from the acceptance criteria — plus lookalike negatives (`src/distribution/index.js`, `src/objects/model.go`, `src/coverage_report.ts`, `src/minified.js`, `src/target.go`). The two former "under vendor still reviewed" cases flip to excluded, as the issue directs. One `internal/agent` preview test used `vendor/baz.go` as its example of a non-default-excluded path; its subject is include-list additive semantics, so the path moves to `api/baz.go` instead of changing the expectation.

Docs: the "Default path exclusions" section of the review-rules guide is updated in all five locales. The hard-coded pattern list (already stale) becomes the two path families with the JSON file named as the source of truth, and the "noisy directories are filtered at the diff level" paragraph now states that the `internal/diff/git.go` blocklist only matches at the repository root, while the default-path gate catches these directories nested at any depth.

## Type of Change

## How Has This Been Tested?

- [x] `make check`, `make coverage` (91.9%, above the 90% threshold), and the pre-commit `ocr review --audience agent` self-review (0 findings)

## Checklist

- [ ] I have signed the CLA
- [x] AI disclosure: I developed this change with the assistance of the ZCode coding agent (model GLM-5.3, by Z.ai). I reviewed every line of the diff myself, I can explain and answer questions about all of it on my own, and no commit is attributed to AI.

## Related Issues

Fixes #1494


Qiyuanqiii (commented):
The implementation itself looks good to me, and this PR already does a better job of keeping the user-facing rule documentation in sync: all five localized `review-rules.md` files now describe the new default-path behavior instead of maintaining another stale exhaustive list.

I still see one documentation gap that I think should be fixed before merge.

The five localized `architecture.md` files currently describe `default_path` as the built-in **test-file** exclusion gate, and say that noisy directories such as `vendor/`, `node_modules/`, and `target/` are filtered earlier at the diff-provider layer and therefore never reach the per-file filter.

After this PR, that is no longer generally true. The provider blocklist only catches those names at the repository root, while nested paths such as:

- `extensions/frontend/node_modules/...`
- `packages/foo/vendor/...`
- `crates/bar/target/...`

can now reach the per-file filter and be excluded by `default_path`.

That distinction also affects override semantics: a user `include` can bypass a `default_path` exclusion, while a provider-level `provider_directory` exclusion cannot be re-admitted.

Could we synchronize the corresponding `architecture.md` explanation across the existing locales as well?

One smaller test-quality note: a few positive fixtures are currently shadowed by another new pattern. For example, the `.pnpm-store` test contains `node_modules`, so it would still pass even if `**/.pnpm-store/**` were removed. Using a path such as `.pnpm-store/v3/files/00/abc.js` would pin that pattern independently. The same principle would help for some of the Yarn / virtualenv cases.

I don't otherwise see a code-level blocker. CI, CodeQL, Pages, plugin-contract, and translation-sync are all green.

lizhengfeng101:
Thanks for tackling #1494! We're going to consolidate this into #1499, which covers the same fix (the exclusion patterns and the Jsonnet-only vendor rule replacement) plus the full multi-language doc sync (architecture.md in addition to review-rules.md across all locales), so I'm closing this one to avoid a duplicate merge.

Your writeup on why .gitignore doesn't backstop committed directories was spot on and matches the reasoning we went with. If you spot any exclusion pattern in your version that #1499 is missing, please flag it on #1499 or open a follow-up — happy to fold it in. Appreciate the contribution.

PR #1503, opened 2026-09-20
fix(agent): drop a comment that names no reviewed file instead of inventing a path

Hiro-Mati:
## Description

`path` is required by the `code_comment` schema but not enforced. When the model omits it, the comment falls back to the task key — which, for a group of several files, is the group key (`"a.go,b.go"`), not a file. Such a comment was emitted with that string as its path: it skipped the review filter and the checkpoint, and it pointed at a file that does not exist.

The existing cross-file search by `existing_code` still runs first. Only when neither the lookup nor that search can place the comment, and its path is still the group key, is it dropped with a `comment_dropped` warning. Single-file groups are unaffected, since their key is a real path.

Worth a maintainer's opinion: dropping is the conservative choice here, but a "group-level comment" channel would be another option if you prefer to keep the content.

Split out of #1474, which bundled three unrelated fixes.

## Type of Change

## How Has This Been Tested?

- The new test(s) fail when this change alone is reverted, and pass with it (verified both ways).
- `make check` passes; the touched packages pass on Windows 11 (Go 1.27) and in a `golang:1.26.5` Linux container replaying the CI job (`go test -race`, coverage >= 90%).

## Checklist

- [x] I have signed the CLA


Hiro-Mati:

The work is not lost if it is useful to you: the branch stays on my fork at https://github.com/Hiro-Mati/open-code-review/tree/fix/pathless-comment-drop, with the regression test that fails without the fix. Feel free to take it over or reuse it under the project's license; no attribution needed.

PR #1506, opened 2026-09-20
fix(action): reject outside projRoot links in docs

wu21-web:
## Description

a doc link containing ../ made the check look at files outside the project folder on the CI machine
## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

## Checklist

- [x] I have signed the CLA
Thank you to @sourcery-ai for discovering this issue in the action contract.

## Related Issues

n/a


chaojixinren:
The fix looks correct for the issue described in this PR. `repoAbsPath()` prevents `..`-based paths from escaping `repoRoot`, and the new regression tests cover both the helper itself and the end-to-end link-check behavior.

I also considered whether an in-repo symlink pointing outside the repository could bypass this check, since `fs.statSync()` follows symlinks. In theory that is possible. However, the current repository does not contain any symlinks, and exploiting that case would require a deliberately added symlink plus a matching documentation link and an existing target outside the checkout. I don't think that edge case should block this fix; it can be treated as future defense-in-depth if the repository starts relying on symlinked paths.

Overall, this addresses the reported path traversal/false-positive issue cleanly. Thanks for the fix and for adding the regression coverage.

chaojixinren (approved):

PR #1508, opened 2026-09-20
fix(config): preserve unknown JSON fields during config updates

### LINKED ISSUE #1485: `ocr config set` silently drops config.json keys it does not know
`config set` decodes config.json into typed structs and writes them back whole (config_cmd.go:386-399, 419-435), so any key the structs lack is deleted by an unrelated `config set`: hand-added keys and fields written by a newer binary. The comment at config_cmd.go:326-332 acknowledges the hazard.

Repro: add `"future_field": 1` to config.json, run `ocr config set language zh`; the key is gone.

A fix could round-trip unknown keys (decode into `map[string]json.RawMessage` alongside the structs and merge on write). Is that direction acceptable before I open a PR?

Environment: v1.12.7, Windows 11 amd64.

dvd233:
## Description

ocr config set loads config.json into typed structs and writes the whole file back. Fields introduced by a newer binary, hand-added integrations, or nested provider/MCP sections were silently discarded when an older binary performed an unrelated config update.

This change preserves unknown JSON fields at the top level and in provider, LLM, telemetry, and MCP server sections while keeping known-field validation and config unset behavior unchanged. Provider TUI rollback cloning also deep-copies the preserved raw fields.

## Type of Change

- [ ] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- go test ./cmd/opencodereview -count=1
- Full Makefile package set with go test -count=1 (all passed; extensions excluded as Makefile does)
- go vet on the same package set
- go build ./cmd/opencodereview
- go run scripts/verify-english-only.go
- bash scripts/verify-license.sh
- git diff --check
- Regression coverage exercises unknown fields at top-level, provider, LLM, telemetry, and MCP nesting.

GNU Make is not installed in the Windows environment, so the Makefile-equivalent commands were run directly. The race suite was not runnable because the available Windows Go toolchain has cgo disabled. The repository ocr review --audience agent self-review command was attempted but could not resolve an LLM endpoint because no test credentials/endpoint are configured.

## Checklist

- [x] My code follows the project's coding style (gofmt, go vet)
- [ ] I have updated the documentation accordingly (not applicable; no user-facing command syntax changed)
- [ ] I have signed the CLA
- [x] I used AI/LLM assistance and disclosed it below; I reviewed the output, added no AI attribution to commits, and will answer maintainer questions myself.

AI/LLM disclosure: OpenAI Codex (GPT-5) in the Codex desktop environment assisted with repository inspection and drafting. The submitted code and tests were manually reviewed.

## Related Issues

Fixes #1485


chaojixinren:
Thanks for the contribution. The unknown-field round-trip looks correct, and deriving known JSON fields directly from struct tags addresses the maintenance concern from the previous review.

Could you please squash the two commits into a single commit before merging? Since both commits are part of the same fix, keeping the PR as one atomic commit would make the history cleaner.

chaojixinren (approved):

wu21-web:
Isn't this expected behavior @dvd233 ? What is the point of preserving a key that is never going to be used? (Please answer without AI/LLM)
cc @chaojixinren

chaojixinren:
I think the chance of hitting this in practice is relatively low, but I still see value in it as a defensive fix.

One concrete case is a temporary downgrade. Suppose a newer OCR version introduces a new config field, e.g. `telemetry.sample_rate`, and writes it to `config.json`. If the user later downgrades to an older version that does not know this field and runs an unrelated command such as:
```
ocr config set language zh
```
the older binary will unmarshal the config into its typed struct and write the whole file back, silently dropping `sample_rate`. When the user upgrades again, that configuration is already lost.

So the point is not that the older binary needs to use the unknown key; it is that changing an unrelated setting should ideally not destroy configuration owned by a newer version.

wu21-web:
But older versions were archived and they overwrites jsons anyway because your fix cannot change the behavior of older releases .....
It is good that older versions overwrites unrelated fields.
- Nobody will downgrade.
- It is good for the older ones to overwrite new configs values so no useless junk stays in the config.

How are these related?

chaojixinren:
> But older versions were archived and they overwrites jsons anyway because your fix cannot change the behavior of older releases .....但旧版本被归档了，而且它们还是会覆盖 json，因为你的修复无法改变旧版本的行为..... It is good that older versions overwrites unrelated fields.旧版本覆盖无关字段是件好事。
> 
> * Nobody will downgrade.  没人会降级。
> * It is good for the older ones to overwrite new configs values so no useless junk stays in the config.旧的配置可以覆盖新配置值，这样配置里就不会留下无用的垃圾。
> 
> How are these related?这些有什么关系？

Yeah, I think you're right.

My downgrade example doesn't really justify this, since the fix can't change already released versions anyway. And if `config.json` is supposed to follow the current OCR schema, dropping unknown fields is reasonable.

The `aws_profile` / `aws_region` case was more about the structs being out of sync with fields OCR actually uses.

So I don't think we need a generic unknown-field preservation layer here.

cmd/opencodereview/config_cmd.go
```diff
@@ -332,6 +333,8 @@ type ProviderEntry struct {
 	// first time any config command runs.
 	AWSProfile string `json:"aws_profile,omitempty"`
 	AWSRegion  string `json:"aws_region,omitempty"`
+
+	unknownJSONFields map[string]json.RawMessage
```
lizhengfeng101, line 340: Optional (review point #2): a short comment on why this field exists and the copy invariant it relies on.

```suggestion
	// unknownJSONFields keeps JSON keys with no matching struct field alive across
	// a load/save cycle. Unexported: any struct-literal rebuild must copy it
	// (see cloneProviderEntry) or the fields are dropped again
	unknownJSONFields map[string]json.RawMessage
```

lizhengfeng101 (approved):
LGTM

PR #1509, opened 2026-09-20
fix(config): keep read paths consistent with OCR_CONFIG_PATH

### LINKED ISSUE #1484: `ocr llm test` and the provider wizards check a different config file than review/scan use
`runLLMTest` honours `OCR_CONFIG_PATH` (cmd/opencodereview/config_cmd.go:100-108), while `ocr review` / `ocr scan` (`loadLLMRuntime`, shared.go:245) and every config write use `~/.opencodereview/config.json`. So `ocr llm test` can succeed while `ocr review` fails in the same environment, and a wizard saves to the default file and then tests whichever file `OCR_CONFIG_PATH` points to.

The code comment justifies read-only use to avoid redirected writes, but review never writes config. `OCR_CONFIG_PATH` is also not documented.

Options: (a) make review/scan honour `OCR_CONFIG_PATH` for reads too; (b) drop it from `llm test`; (c) document the split explicitly. Which one do you want?

Environment: v1.12.7, Windows 11 amd64.

dvd233:
## Description

Closes #1484.

`OCR_CONFIG_PATH` already controls `ocr llm test`, but review, scan, and
preview loaded app settings from the default home config instead. Provider
wizards also saved to their explicit config path and then invoked the
environment-based test command, so a set `OCR_CONFIG_PATH` could test a
different file from the one just saved.

This change:
- uses `resolveConfigPath()` for review/scan runtime setup and preview token resolution;
- passes the just-saved path to manual, custom, and built-in provider wizard connection tests;
- keeps `ocr config set` and provider wizard writes on `defaultConfigPath()`;
- adds regression tests for override reads, preview settings, writer isolation, and wizard test-path selection;
- documents `OCR_CONFIG_PATH` in the English configuration guide.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

- [x] `CC=gcc CGO_ENABLED=1 go test -v -race -count=1 ./...`
- [x] `CC=gcc CGO_ENABLED=1 go vet ./...`
- [x] `gofmt` on all changed Go files and `git diff --check`
- [x] `bash scripts/verify-license.sh`
- [x] `go run scripts/verify-english-only.go`
- [x] `go test -count=1 -coverprofile=... ./...` — 91.1% total coverage
- [x] `go build ./cmd/opencodereview`
- [ ] `make check` — equivalent checks pass; the local BusyBox `make` invokes the Bash license script with an incompatible `/bin/sh`
- [ ] Manual LLM review — not run because this environment has no configured LLM endpoint

## Checklist

- [x] My code follows the project's coding style (`gofmt`, `go vet`)
- [ ] I have signed the CLA
- [x] I used AI/LLM assistance and disclose it below; I reviewed the output, did not attribute the commit, and will answer maintainer questions myself.

## AI disclosure

Prepared with OpenAI Codex through pi using model `gpt-5.6-luna`. The generated
code and text were reviewed manually, and the commit contains no AI attribution
trailer.

## Related Issues

Closes #1484.


lizhengfeng101:
@dvd233 Thanks for the fix, and for the clear write-up.

The wizard part is right and I want to keep it: having `applyManualConfig` / `applyCustomProviderConfig` / `applyOfficialProviderConfig` test the path they just wrote is correct regardless of what happens to the environment variable, and the `runLLMTestPath` seam is a clean way to get there.

On the direction, though, I'm going with option (b) from #1484 — dropping `OCR_CONFIG_PATH` — rather than (a). Three things decided it:

**It was never a public interface.** It is not in the README (any locale), not on the docs site, not in `--help`. It landed in #204 purely so the VS Code panel could test a draft config before saving. Removing it is not a breaking change and needs no deprecation window.

**Its only two callers do not need it.** Both `extensions/vscode/.../CliService.ts` and `extensions/idea/.../CliService.kt` already set `HOME` *and* `USERPROFILE` to the temp dir before running `ocr llm test`. `defaultConfigPath()` resolves through `os.UserHomeDir()`, so the draft config is already found without the variable. It has been redundant since the day it was added.

**Widening it has a cost I don't want to pay.** Making `loadLLMRuntime` honour it means `ocr review` takes its config — and therefore `mcp_servers.<name>.setup` (executed via `shellCommand`, `review_cmd.go:568`) and `api_key_cmd` / `auth_token_cmd` (executed via `sh -c`, `keycmd_unix.go:33`) — from a path chosen by an environment variable. The env overrides we have today (`OCR_LLM_URL` and friends) have no `_cmd` variant, so they cannot execute anything. `ocr review` is the command that runs in CI, and I would rather it not gain that property. The comment this patch rewrote was recording exactly that trade-off, even though it framed it as being about writes.

Would you like to revise the PR along those lines? Concretely:

- keep the `runLLMTestPath` seam and the three wizard call sites;
- drop `resolveConfigPath()`; `runLLMTest()` goes back to `defaultConfigPath()`;
- remove the `configPath` parameter from `testConnection` in both extensions;
- drop the now-dead env plumbing in `config_cmd_test.go` and `provider_cmd_test.go`;
- no docs change needed.

Either way is fine — if you would rather not, say so and I will take it from here; the wizard fix stays yours.

One housekeeping item: the CLA checkbox is still unchecked, and that needs signing before I can merge.

hyy321:
Hi @dvd233, are you still planning to work on the requested changes? If not, I'd be happy to pick this up if @lizhengfeng101 is okay with that.

lizhengfeng101:
Thanks for the thorough work here, @dvd233.

After the discussion on #1484, the maintainers decided to go with option (b) — removing `OCR_CONFIG_PATH` entirely rather than extending it to review/scan. That direction has now landed in #1537 (merged to `main`), which closed #1484.

Since this PR implements the alternative option (a) for the same issue, I'm closing it as superseded. The analysis and regression tests you put together were valuable in shaping the final decision — much appreciated.

Issue #1487, opened 2026-09-20
Range and commit modes filter changed files with the working tree's ignore rules


Hiro-Mati:
In `--from/--to` and `--commit` modes, ignore rules are read from the current checkout, not from the reviewed ref (internal/diff/git.go, partitionDiffs). A local, uncommitted `.gitignore` entry (or one that exists on the checked-out branch but not on the reviewed one) can drop files from a range review.

Repro: `feature` adds `gen.go` and has no `.gitignore`; with an uncommitted local `.gitignore` containing `gen.go`, `ocr review --from base --to feature` reviews nothing.

Should range/commit modes read `.gitignore` files from the reviewed ref (e.g. `git show REF:.gitignore` / `check-ignore` against a temporary index), or is the working-tree behavior intended?

Environment: v1.12.7.

AI disclosure: found with Claude Code (model: Claude Opus 5) during an audit where each finding was reproduced by a failing test.

wu21-web:
@Hiro-Mati Please close all your issues as not planned in this upstream.

PR #1511, opened 2026-09-21
feat(review): add frozen staged snapshot reviews

### LINKED ISSUE #1510: feat(review): review a frozen Git index snapshot with --staged
## Problem Statement

Add an explicit way to review exactly the code staged for the next commit.

The current workspace mode combines staged, unstaged, and untracked changes. That is useful for reviewing ongoing work, but it does not represent the Git index when a file is partially staged. For example:

1. Stage a change containing a defect.
2. Fix the defect in the working tree without staging the fix.
3. Review the workspace and then commit the index.

The review can examine the fixed working-tree version even though the commit still contains the defect. The inverse is also possible: unstaged experiments can generate findings unrelated to the intended commit.

As of `main@cf64e7080f600d507888a26785ad6a8b7a13ab6a` (checked 2026-09-21), the shared diff flags expose range and single-commit inputs, but no independent staged-only mode. #1121 and #1126 concern staged files being omitted from workspace review; they do not provide this separate review input.

## Proposed Solution

Introduce `ocr review --staged`, backed by a frozen Git index snapshot, with the existing preview facility supporting the same input.

```sh
# Preview the selected staged changes without calling a model.
ocr review --staged --preview

# Review the frozen staged input and save the result.
ocr review --staged --format json --output staged-review.json
```

The proposed behavior is: compare the captured HEAD tree with the captured index tree, then use that same captured target tree throughout the review. In a repository with no commits, compare the index against Git's empty tree. This is a direct base-to-index comparison, not range mode's merge-base interpretation.

### First-version scope

1. **Capture immutable inputs before review.** Freeze the base identity and the index snapshot before selection, diff generation, or model execution. Subsequent edits, `git add`, unstaging, or HEAD changes must not change an in-progress review. If capture cannot establish a consistent input, stop with an actionable diagnostic rather than combine different states.

2. **Keep all repository code context on the same snapshot.** The diff, full-file reads, file discovery, repository search, and cross-file context must refer to the captured complete index tree, including unchanged dependencies needed as context. Files newly added to the index must be available to those tools; unstaged and untracked-only files must not leak into the reviewed code context. Base-side reads should use the captured base where needed. Merely changing the diff command to `git diff --cached` is insufficient if the agent still reads the working tree or the live index afterward.

3. **Reuse selection and review policy with explicit input provenance.** Reuse the existing allowlist, secret-path protection, exclusions, rules, budgets, and review engine. Repository `.gitignore` evidence must come from the captured snapshot; an unstaged `.gitignore` edit must not silently change the selection. The source of repository-local review rules also needs an explicit decision, separate from intentional user inputs such as `--rule`, `--exclude`, model configuration, or supplied background text. Snapshot consistency covers OCR's built-in repository context tools. If custom shell/MCP tools remain available, document that they may observe external or live state; this feature does not sandbox those tools or freeze user-supplied external context.

4. **Expose an actual staged input mode.** Reject combinations with `--from`, `--to`, or `--commit` before model execution. Default workspace behavior remains unchanged. Preview and review must use the same selection semantics, although two separately invoked commands may capture different snapshots if the user stages changes between them.

5. **Record what was reviewed.** Saved results and sessions should identify the staged mode, captured base (including the unborn-HEAD case), and target tree identity. Distinguish a tree object ID from a commit object ID; do not present an index tree as an actual reviewed commit. Consumers that do not support the new identity must reject or explicitly report it as unsupported rather than assume a commit review. Reject `--staged --resume` in the first version. Future resume support must verify the full captured input identity: the same selected diff can coexist with different surrounding context, so a diff hash alone is insufficient.

6. **Leave the user's checkout and index intact.** Do not use automatic stash/reset/checkout, create a user-visible commit or branch, or alter the user's index to simulate the reviewed state. Read failures and cancellation must also preserve user state. The implementation should document any temporary objects/resources and their lifecycle.

### Git edge cases

- Partial staging must preserve the index version even when the working-tree file differs or is missing.
- Staged additions and renames must work, including before the first commit and in linked worktrees.
- Unresolved merge entries must fail clearly before spending model tokens.
- Intent-to-add entries are not staged file contents. Their treatment must be explicit and tested; unstaged contents must not be substituted silently.
- Sparse indexes/checkouts, symlinks, submodule entries, binary files, and mode-only changes need defined behavior. Preserve existing supported semantics where feasible; unsupported cases should be reported without silently reading a different version or following links outside the snapshot.
- Staged deletions should remain visible in the diff/preview with the existing selection/exclusion explanation. Independent deletion-impact review is outside this feature.
- An empty staged diff should produce an explicit no-op without a model call. A nonempty diff whose files are all excluded should retain the applicable exclusion diagnostics. This feature should not silently redefine the separate gate's empty-selection policy.

## Alternatives Considered

- **Continue using workspace review.** Cannot express the intended commit when staged and unstaged versions differ.
- **Use `git diff --cached` but leave file/search tools unchanged.** Produces inconsistent evidence: the agent can inspect a different version from the diff.
- **Create a temporary commit and review it.** A user workaround, but requiring commit/ref manipulation before review is undesirable. An internal immutable representation may reuse existing readers, provided it preserves staged-mode identity and does not change the user's history.
- **Stash unstaged changes or temporarily rewrite the checkout.** Can disrupt ongoing work and makes failures harder to recover from; the proposed contract avoids changing user state.

## Acceptance Criteria

Use small real Git repositories and deterministic tool/selection tests. A live model is not needed to establish snapshot correctness.

- [ ] `--staged` and preview are exposed consistently; invalid mode combinations fail before model execution.
- [ ] A partially staged file is read, searched, and diffed as its index version, including when the working-tree version would hide or introduce a finding.
- [ ] After capture, modifying/staging/unstaging files does not change the diff, selected files, or tool-visible code for that run.
- [ ] Staged additions, initial commits, renames, and a working-tree file removed after staging have deterministic coverage.
- [ ] Unstaged/untracked-only files and unstaged ignore-rule changes do not change the captured review input.
- [ ] Conflicts, intent-to-add, empty input, all-excluded input, and relevant special Git entries have explicit outcomes and regression coverage.
- [ ] Results/sessions record staged input and immutable identities accurately; unsupported resume/consumer combinations cannot silently fall back to workspace or commit semantics.
- [ ] The user's index, refs, and working-tree contents remain unchanged on success, failure, and cancellation.
- [ ] Existing workspace, range, and commit behavior remains covered; required project checks pass and CLI documentation is updated in the maintained locales.

## Decisions to Settle Before Implementation

- Choose the immutable snapshot representation and how to capture it safely across supported Git index/worktree configurations. Keep the public contract independent of whether the implementation uses Git objects or an isolated materialization.
- Agree on the manifest/session representation for a staged tree, including compatibility with existing readers. No changes to CI revision expectations should be implied by adding this local mode.
- Decide the source of repository-local review configuration and confirm that staged resume remains outside the first version.

A hook manager, automatic installation of pre-commit hooks, recursive submodule review, new deletion-review tasks, and IDE integration are outside the first PR. An optional hook example can follow once the CLI contract is stable.

## Affected Area

CLI / Commands; Built-in Tools (`file_read`, `code_search`, etc.); Review Rules; Output / Formatting; Documentation.

## Additional Context

- #1121 / #1126: related workspace selection bug and existing fix proposal; coordinate regression coverage without duplicating that patch.
- #1487: related input-provenance problem where range/commit selection consults working-tree ignore rules. The staged design must avoid the same mismatch; a general correction to existing modes can remain separately scoped.
- #1489: CI review-gate tracking. This proposal is an independent local Git input mode and does not require those PRs to land first.

This is a feature proposal for discussion before implementation; no implementation or maintainer agreement is claimed.

## AI/LLM Disclosure

After proposing the idea,  I reviewed existing OCR work, clarified the relevant boundaries, and concluded that this is a worthwhile direction to pursue as a complement. I also used Codex (GPT-6-Astra) to review the existing code and related issues, checking for any overlapping work.

Qiyuanqiii:
## Description

Adds `ocr review --staged` to review the content prepared for the next commit. OCR captures the complete Git index and `HEAD` at the start of the run, then uses that snapshot for the diff and all built-in code-context tools.

For example, a partially staged file can contain a defect while its working-tree version already contains the fix. Staged review examines the version that will be committed. The captured input stays stable throughout the review, including when files are edited or staged again.

```bash
ocr review --staged
ocr review --staged --preview
ocr review --staged --preview --format json
ocr review --staged --format json --output staged-review.json
```

### Snapshot and selection

Capture copies the active index into a private temporary index and writes an immutable Git tree. It checks the original index and the resolved HEAD commit for changes during capture. The tree includes unchanged files used as context. In a repository with no commits, the comparison starts from an empty tree.

Before reading file content, OCR checks the old and new modes in a NUL-delimited raw diff between the captured trees. This rejects changed gitlinks, including pure renames. The review patch retains ordinary rename detection.

The snapshot supplies:

- The diff, full-file reads, file discovery, and content search.
- Repository `.gitattributes`, including attributes used for binary classification.
- Default project rules in `.opencodereview/` and their referenced documents.

Each snapshot rule file, including `rule.json`, has a 512 KiB limit. Referenced documents must be ordinary files in the snapshot, specified by repository-relative paths. Missing files, absolute paths, oversized files, and symlink references produce an error.

Explicit `--rule` files, `--exclude`, global configuration, Git configuration, global attributes, and `.git/info/attributes` retain their configured sources. Custom MCP servers and external tools use their own data sources and may access the live working tree.

Staged paths remain eligible for review regardless of `.gitignore` patterns. OCR applies its directory and secret exclusions, supported-file allowlist, and review-rule exclusions to the paths tracked in the index. Preview shows the selected files, exclusion reasons, and input identity. Each invocation captures the current index.

Capture preserves the user's index, working tree, and refs. The private index is removed after capture. Tree objects written to the repository follow Git's normal garbage-collection lifecycle.

### Input identity

Staged results use `ocr.run-manifest/v2`:

| Field | Value |
| --- | --- |
| `input.mode` | `staged` |
| `input.resolved_base` | Captured HEAD commit; omitted before the first commit |
| `input.snapshot_tree` | Tree object ID for the complete index snapshot |
| `input.resolved_head` | Omitted |
| `input.exact_range` | Omitted |

JSON preview includes the same input identity. Other review modes keep their v1 format. Consumers of staged results need support for the v2 schema and its tree identity.

### Usage and Git support

- Requires Git 2.41 or later.
- Supports partial staging, initial commits, linked worktrees, alternative `GIT_INDEX_FILE` paths, and SHA-256 repositories.
- An empty staged diff completes without an LLM call.
- `--staged` is mutually exclusive with `--from`, `--to`, `--commit`, and `--resume`. Staged sessions currently have no resume support.
- Unresolved conflicts, intent-to-add entries (`git add -N`), split indexes, and sparse indexes produce an error.
- Changed submodule/gitlink entries produce an error before file content is read. This covers additions, updates, deletions, pure renames, and conversions between regular files and gitlinks.
- Repositories with unchanged submodules support ordinary-file review. Built-in discovery and search exclude gitlink content, and direct reads reject gitlink paths. Symbolic links are read as their stored link blobs.

The CLI reference and architecture pages are updated in English, Chinese, Japanese, Korean, and Russian. Both copies of the review skill route staged-change requests to `ocr review --staged` and document staged preview, flag conflicts, and resume behavior.

## Type of Change

- [x] Documentation update
- [ ] CI / Build / Tooling

## How Has This Been Tested?

At `ad3bd56`, all GitHub checks passed. The [Linux/Windows CI run](https://github.com/alibaba/open-code-review/actions/runs/35557227945) reports **91.5% Linux coverage** and confirms successful execution of the POSIX cancellation and state-preservation regressions.

At `fbe46d0`, local `make check`, `make build`, and the full `make coverage` suite passed with **91.1% coverage**. The diff package passed race tests. A pure gitlink-rename regression reproduced the prior failure and now passes; raw-diff tests cover special path bytes, SHA-256 object IDs, and malformed records. The compiled CLI rejects the same rename fixture while preserving index bytes, HEAD, and refs.

Local validation used Windows/amd64, Go 1.26.5, and Git 2.54.0.windows.1, with implementation baseline `cf64e7080f600d507888a26785ad6a8b7a13ab6a`:

- `make license-add`, `make check`, and `make build` passed.
- The full `make test` suite passed with the race detector.
- `make coverage` reported **91.1%**, above the 90% threshold.
- The compiled CLI produced a staged JSON preview with a valid tree identity, preserved the original index hash and HEAD, and rejected conflicting flags before execution.

Regression coverage includes:

| Area | Checks |
| --- | --- |
| Git inputs | Partial staging, initial commits, empty input, linked worktrees, alternative indexes, SHA-256 repositories, additions, renames, deletions, symlink blobs, binary files, and mode-only changes |
| Snapshot consistency | HEAD, index, and working-tree mutations after capture; unchanged context files; staged files removed from the working tree before capture |
| Built-in tools | A local model double restages the repository after its first request; `file_read`, `file_find`, `code_search`, and `file_read_diff` continue to return the original snapshot |
| Configuration | Captured project rules, referenced documents, and repository attributes; explicit rule overrides; submodule display and recursion settings |
| State preservation | Index bytes, raw HEAD, ref identities, and worktree paths, modes, and contents remain unchanged after capture failure or cancellation; the private index directory is cleaned up |
| Output and sessions | Text/JSON preview identity, saved-session readback, and staged v2 coverage counts in the viewer |
| Error handling | Conflicting flags, unresolved merges, intent-to-add, split/sparse indexes, malformed indexes, cancellation, and changed gitlinks |

The during-capture mutation and cancellation tests use POSIX shell shims. A FIFO handshake triggers cancellation after `write-tree` completes. Linux CI executes these cases; Windows runs the remaining state-preservation cases. Functional tests use local model doubles for deterministic assertions.

### Pre-commit review

Ran the repository-required `ocr review --audience agent --background ...` with OCR v1.11.9 and DeepSeek `deepseek-v4-flash`. Both review runs completed all 19 selected production files without a failed review item. Their findings were checked against the code and focused regressions, not treated as automatically correct.

Confirmed issues addressed before submission include Git submodule display settings bypassing the changed-gitlink guard; non-blob reads of unchanged gitlinks; implicit content-search recursion into unchanged submodules; lost errors/cancellation during unborn-HEAD probes; and a shared rule-loading deadline that did not scale with the number of documents. The final code also preserves Git diagnostics, uses explicit snapshot parameters, and documents the intentional rule-file restrictions. Focused tests cover the fixes, followed by the full final validation listed above.

The repeated claim that `--attr-source` requires Git 2.43 was rejected after checking the [Git 2.41 source and documentation](https://github.com/git/git/blob/v2.41.0/Documentation/git.txt#L197-L200), which already implement and document the option. No claim of zero raw model findings is made, and this automated review does not replace contributor self-review.

For the wrap-up changes, the configured provider rejected a further `ocr review` request before review completed; that attempt is not counted as a successful review. OCR v1.12.7 delegation mode supplied file selection and Go rules, and Codex reviewed all three changed files (one selected production file plus both test files, which OCR excludes by default). This review found no blocking issues. Focused race tests cover the affected rules/tool packages and new capture-state tests; `make check` also passes.

## Checklist

- [x] Code follows the project's automated style and vet checks.
- [x] Added regression tests for the new behavior.
- [x] New and existing unit tests pass locally as described above.
- [x] I have signed the CLA

### AI/LLM disclosure

OpenAI Codex (GPT-6) assisted with implementation, tests, documentation, and preparation of this description. The repository-required pre-commit review uses the open-code-review CLI; its model and result are recorded in the validation section.

## Related Issues

Closes #1510.

- #1121 and #1126: staged-file handling in workspace reviews.
- #1487: the version of repository ignore rules used for review.


Qiyuanqiii:
@NanaseInori review

NanaseInori (changes requested):
I re-reviewed the latest head (`ad3bd56`). The additional state-preservation and cancellation coverage is useful, and the earlier snapshot-rule diagnostic issue is fixed. The overall staged-snapshot design also looks coherent to me.

I found one gitlink case that still breaks the stated staged-mode boundary, though.

`stagedDiffHasGitlink()` detects mode/index headers containing `160000`, but a pure rename of an unchanged gitlink does not necessarily emit any of those headers. Git can render it as only:

```diff
diff --git a/subold b/subnew
similarity index 100%
rename from subold
rename to subnew
```

In that case, `stagedDiffHasGitlink()` returns false and the change proceeds into `ParseDiffText()`. `finalizeDiff()` then attempts:

```text
git show <snapshot-tree>:subnew
```

even though that tree entry is a gitlink/commit rather than a blob.

In the usual case this may fail because the submodule commit object is not present in the superproject. If that object is present, however, `git show` can render the commit as output and that content can enter the review path as though it were file content. That is exactly the situation the pre-parser gitlink guard is intended to prevent.

I think the gitlink check should use structural mode information rather than relying only on unified-diff headers. For example, a `git diff --raw -z --find-renames ...` preflight could reject any changed entry whose old or new mode is `160000`, including a 100% gitlink rename.

A regression test for renaming an unchanged gitlink from one path to another would pin this case.

There are also two documentation follow-ups worth including with this feature:

- The localized `architecture.md` files still describe the diff provider as having only three modes: `Workspace`, `Commit`, and `Range`.
- Both copies of `open-code-review/SKILL.md` still say “Stage selectively if you want narrower scope”. That is no longer the right guidance once `--staged` exists, because bare `ocr review` still includes staged, unstaged, and untracked changes. The skill should route requests such as “review my staged changes” to `ocr review --staged`.

I don't currently see another implementation blocker beyond the gitlink-rename case. CI, CodeQL, Pages, translation-sync, and plugin-contract are all green.

Qiyuanqiii:
@NanaseInori review

NanaseInori (approved):

chaojixinren:
LGTM.
I checked out fbe46d0 and built/tested on this machine:
- Host: macOS 26.6.2 (25G83), darwin/arm64
- Go: 1.25.7
- Git: 2.50.1 (Apple Git-155)
- Binary: make build → ./dist/opencodereview v1.12.7 (fbe46d0)

chaojixinren (approved):

PR #1514, opened 2026-09-21
fix(tool): fallback when git grep --max-count is unsupported

tokove:
### Summary

Fix `code_search` tool failure on systems running older Git versions (< 2.38.0).

Git 2.38.0 introduced the `--max-count` option for `git grep`. On older versions (e.g. Git 2.34.1 on Ubuntu 22.04 LTS), `git grep` fails with exit code 129:

```text
error: unknown option `max-count'
usage: git grep [...]
```

While the project documents Git >= 2.41 as a prerequisite in the README, the CLI only warns and continues. This PR provides a specific `--max-count` fallback so code search continues to work on those environments.

### Changes

In `internal/tool/code_search.go`:

* Introduce `gitGrepSupportsMaxCount` (`atomic.Bool`) to track whether the Git implementation supports `--max-count`.
* If `git grep` fails with exit code 129 and the stderr diagnostic indicates an unknown `max-count` option, disable the flag globally and immediately retry without `--max-count`.
* Inspect the diagnostic separately from usage help text via `trimGitUsage`, preventing unrelated usage errors on modern Git from falsely matching `--max-count`.
* Subsequent calls within the same process skip `--max-count` directly, avoiding repeated failed invocations.
* The existing in-memory truncation logic continues to enforce a consistent match limit across Git versions.

In `internal/tool/code_search_test.go`:

* Add unit tests for `--max-count` argument generation.
* Add tests for detecting the unknown-option error with single/backtick quotes and localized messages, plus a negative regression ensuring unrelated errors with usage text do not trigger the fallback.
* Add an automatic retry test via a PATH shim simulating Git < 2.38 (`TestGitGrep_AutomaticRetryOnUnsupportedMaxCount`).
* Add a synchronized concurrent search test with a start barrier to verify safe concurrent fallback (`TestGitGrep_ConcurrentFallback`).
* Add a non-Git directory fallback test (`TestGitGrep_NonGitDirectory_AutomaticRetryOnUnsupportedMaxCount`).
* Add a fallback test with the capability already disabled (`TestGitGrep_FallbackWhenMaxCountUnsupported`).


Qiyuanqiii:
@NanaseInori review

NanaseInori (changes requested):
I reviewed the current head (`a5a6905`). The compatibility fallback is reasonable, but I found a concurrency bug in the capability cache that can make the fallback fail intermittently.

`gitGrepSupportsMaxCount` is global, and the retry currently depends on its value at the time the failed invocation is handled:

`err != nil && gitGrepSupportsMaxCount.Load() && isUnknownMaxCountError(...)`

Consider two concurrent `code_search` calls on a Git version that does not support `--max-count`:

1. Both calls observe `gitGrepSupportsMaxCount == true`.
2. Both build and execute a command containing `--max-count`.
3. Both commands fail with exit 129.
4. The first goroutine handles the failure, stores `false`, and retries successfully.
5. The second goroutine then evaluates the retry condition. `Load()` is now false, so it skips the retry and returns the original Git failure.

The atomic avoids a data race, but the retry decision is tied to mutable global state rather than to the command that actually failed.

I think the retry should instead depend on whether that particular invocation was built with `--max-count`. For example, record `usedMaxCount` from `cmdArgs`; if that invocation fails with the expected unsupported-option error, store the global capability as false and retry regardless of whether another goroutine has already updated the cache.

The same issue applies to the `--no-index` fallback path.

There is also a test gap here. `TestGitGrep_FallbackWhenMaxCountUnsupported` currently sets:

`gitGrepSupportsMaxCount.Store(false)`

before calling `gitGrep`, so it only verifies that searching works when `--max-count` is already disabled. It does not exercise the actual behavior this PR adds: first attempt with `--max-count` → exit 129 → capability detection → retry without the flag.

Please add a regression that makes the first Git invocation reject `--max-count` and verifies the automatic retry. A concurrent variant where two searches both begin with the flag would also directly pin the race above.

I don't see another blocking issue in the approach. Requesting changes for the concurrent fallback behavior and its regression coverage.

Qiyuanqiii:
@NanaseInori review

tokove:
@NanaseInori Thanks for the thorough and insightful review!

I’ve addressed all the points in the latest commit:

1. **Scoped retry to the current invocation**
   Replaced the retry condition based on `gitGrepSupportsMaxCount.Load()` with `usedMaxCount := slices.Contains(cmdArgs, "--max-count")` in both normal and `--no-index` paths. This ensures that an invocation retries whenever it actually used `--max-count`, regardless of concurrent changes to the global capability state.

2. **Added automatic retry and concurrency tests**

   * `TestGitGrep_AutomaticRetryOnUnsupportedMaxCount`: Verifies the full fallback path using a PATH shim that simulates older Git versions returning exit code 129 for `--max-count`.
   * `TestGitGrep_ConcurrentFallback`: Runs concurrent searches to verify that capability detection and fallback remain race-free.
   * `TestGitGrep_NonGitDirectory_AutomaticRetryOnUnsupportedMaxCount`: Verifies the same fallback behavior for the `--no-index` path.

All unit tests pass, and `go test -race` completes cleanly with **95% package coverage**.

Could you please take another look when you have a chance?

NanaseInori (approved):
Re-reviewed the latest head (`36251b6`).

The concurrency issue from my previous review is fixed correctly.

The retry decision is now tied to the invocation that actually failed:

`usedMaxCount := slices.Contains(cmdArgs, "--max-count")`

rather than to the current value of the shared capability cache. This means that if several concurrent searches all started with `--max-count`, each failed invocation still retries correctly even if another goroutine has already changed `gitGrepSupportsMaxCount` to false.

The same fix is applied to the `--no-index` path as well.

The regression coverage is also substantially better now:

- `TestGitGrep_AutomaticRetryOnUnsupportedMaxCount` exercises the real first-attempt failure followed by an automatic retry;
- `TestGitGrep_ConcurrentFallback` covers concurrent searches;
- `TestGitGrep_NonGitDirectory_AutomaticRetryOnUnsupportedMaxCount` covers the plain-directory / `--no-index` path.

One minor testing nit: the concurrent test does not use a barrier to guarantee that every goroutine has constructed its first command while the capability cache is still true, so it may not reproduce the old race on every run. I don't think that should block this PR, because the implementation no longer relies on the mutable global value when deciding whether an already-failed invocation should retry, and the direct retry path is independently covered.

I don't see another code-level blocker.

The current CI and CodeQL entries are waiting for maintainer approval to execute, so I don't consider that an implementation issue with this PR.

LGTM. Approving.

Qiyuanqiii:
let me see

Qiyuanqiii (commented):
This PR changes implementation and tests only. I don't think a separate documentation page is needed for this internal compatibility fix.

A few non-blocking points:

- The README lists Git >= 2.41 as a prerequisite. Please describe it that way in the PR body, while noting that the CLI only warns and continues. Also, narrow "completes that compatibility work" to the specific `--max-count` fallback.
- Please update the test summary to include the new automatic-retry tests. `TestGitGrep_FallbackWhenMaxCountUnsupported` starts with the capability disabled, so that particular test does not exercise first-attempt failure and automatic retry.
- `isUnknownMaxCountError` searches the whole stderr for `max-count`. An unrelated usage error on modern Git can include that option in its help text and satisfy the predicate. Consider checking the diagnostic separately from the usage text and adding a negative regression for this case, while preserving support for localized diagnostics.

The invocation-scoped retry correctly addresses the earlier concurrency issue. I don't see a new blocking issue.

tokove:
@Qiyuanqiii Thanks for the helpful review and suggestions!

I’ve addressed all the non-blocking feedback in the latest commit (`3be2ef9`) and updated the PR body:

1. **Diagnostic separation & negative regression**

   * `isUnknownMaxCountError` now uses `trimGitUsage(stderr, 129)` to inspect only the diagnostic line, preventing false positives when `--max-count` appears in the usage text of unrelated errors on modern Git.
   * Added a negative regression test ensuring unrelated option errors with usage text return `false`.
   * Added a test verifying localized diagnostic matching (e.g. `错误: 未知选项 'max-count'`).

2. **Synchronized concurrency test**

   * Added a `startBarrier` channel to `TestGitGrep_ConcurrentFallback` so all goroutines are released simultaneously.

3. **Updated PR description**

   * Clarified the Git >= 2.41 prerequisite and narrowed the scope to the specific `--max-count` fallback.
   * Updated the test summary to include all automatic-retry and concurrency tests.

All tests, including `go test -race`, pass cleanly with 95% package coverage.

Qiyuanqiii (approved):
The previous feedback is addressed well.

The PR description now accurately describes Git >= 2.41 as a prerequisite and scopes the compatibility claim specifically to the `--max-count` fallback. The updated `isUnknownMaxCountError` also fixes the false-positive case I mentioned: trimming the usage text before checking for `max-count` prevents an unrelated exit-129 error from matching merely because modern Git lists `--max-count` in its help output. The added negative and localized-diagnostic regressions cover that behavior appropriately.

One small testing nit remains: the new `startBarrier` synchronizes the goroutines before they enter `gitGrep`, but it does not guarantee that every goroutine constructs its first command while `gitGrepSupportsMaxCount` is still true. One goroutine could still fail and update the capability cache before another reaches `buildGrepArgs`. A deterministic regression for the original race would need synchronization around the first `--max-count` invocation itself rather than only at goroutine startup.

I don't consider that blocking. The implementation already makes the retry decision from the failed invocation's own `usedMaxCount` state, so correctness no longer depends on reproducing that scheduling window in the test.

No new code-level blocker found. Approving.

wu21-web:
Why? The documented minimal is Git v2.41. That is 3 minor versions, is this nessarcary? Plus, it prints a warning telling users Git is too old. If you insist, consider suppressing that warning because we got an equivalent.

Qiyuanqiii:
> Why? The documented minimal is Git v2.41. That is 3 minor versions, is this nessarcary? Plus, it prints a warning telling users Git is too old. If you insist, consider suppressing that warning because we got an equivalent.

On the warning, I’d still keep it if this fallback is retained.

The fallback gives us equivalent result limiting for this specific `git grep` option, but it does not establish broader compatibility for Git versions below the documented minimum. Suppressing the warning could therefore imply more support than this PR actually provides.

It is also not fully equivalent in resource usage, since without `--max-count` Git may emit more output before OCR truncates it in memory.

So I think keeping the warning and treating this as a best-effort fallback would be the clearer behavior.

wu21-web:
I did the git version validation @Qiyuanqiii , and we have a switch operation that filters out every command that doesn't use Git. We only print warning for commands that qualities the "minimal: Git v2.41". If the minimal is "below 2.38", I won't print the warning because that warning is no longer the case.

wu21-web:
Plus, if it is not the equivalent, I'd keep Git v2.41+ and make Git v2.38 a hard requirement for this command. If the equivalent does not requires git, even consider fully removing the parent command from the git validation function.

wu21-web:
I am sure the maintainer won't downgrade dependency. Someone downgraded npm from 18 to 14 and got rejected.

Qiyuanqiii (commented):
wait

lizhengfeng101:
Thanks @tokove for the careful work here, and thanks @NanaseInori @Qiyuanqiii @wu21-web for the thorough review — the code itself is in good shape.

That said, I'm going to decline this as a maintainer decision on scope rather than on quality.

Our documented minimum is Git >= 2.41. I don't want to add best-effort compatibility for versions below the documented floor, for a few reasons:

- **It's not truly equivalent.** As noted in the thread, without `--max-count` git can emit far more output before we truncate it in memory. So this doesn't restore parity — it trades a hard failure for degraded, unbounded-output behavior on unsupported versions.
- **It contradicts our own signals.** We already print a warning when Git is older than the documented minimum. Special-casing one flag makes that warning inconsistent with actual behavior, and invites the question of which *other* sub-2.41 breakages we're now implicitly on the hook for.
- **Maintenance direction.** Carrying compatibility shims below the documented floor grows the support surface in the wrong direction. I'd rather keep a single clear prerequisite and push users on old distros toward a newer Git, which is the healthier long-term path for the project.

If anything, the actionable follow-up from this thread is on the *documentation/warning* side (making the 2.41 requirement and its warning clearer), not a per-flag fallback. I'm happy to see a PR in that direction.

Closing this one — really do appreciate the effort and the high-quality discussion.

wu21-web:
I agree with @lizhengfeng101 , dependencies should not be downgraded unless there is an emergency.

Qiyuanqiii:
> Thanks @tokove for the careful work here, and thanks @NanaseInori @Qiyuanqiii @wu21-web for the thorough review — the code itself is in good shape.
> 
> That said, I'm going to decline this as a maintainer decision on scope rather than on quality.
> 
> Our documented minimum is Git >= 2.41. I don't want to add best-effort compatibility for versions below the documented floor, for a few reasons:
> 
> * **It's not truly equivalent.** As noted in the thread, without `--max-count` git can emit far more output before we truncate it in memory. So this doesn't restore parity — it trades a hard failure for degraded, unbounded-output behavior on unsupported versions.
> * **It contradicts our own signals.** We already print a warning when Git is older than the documented minimum. Special-casing one flag makes that warning inconsistent with actual behavior, and invites the question of which _other_ sub-2.41 breakages we're now implicitly on the hook for.
> * **Maintenance direction.** Carrying compatibility shims below the documented floor grows the support surface in the wrong direction. I'd rather keep a single clear prerequisite and push users on old distros toward a newer Git, which is the healthier long-term path for the project.
> 
> If anything, the actionable follow-up from this thread is on the _documentation/warning_ side (making the 2.41 requirement and its warning clearer), not a per-flag fallback. I'm happy to see a PR in that direction.
> 
> Closing this one — really do appreciate the effort and the high-quality discussion.

Agreed on the warning point too — keeping Git >= 2.41 as the clear support boundary is better than suppressing the warning for a one-off fallback.

PR #1515, opened 2026-09-21
fix(viewer): keep vertical mouse-wheel working over scroll regions

lizhengfeng101:
## Description

Fixes a mouse-wheel scroll trap on the viewer's list pages. When the cursor was over the Sessions table, the Repositories table, or the session-detail Token breakdown table, the vertical wheel did nothing — the page only scrolled while the cursor was outside the list.

### Root cause

`.sessions-page .table-scroll`, `.repos-page .table-scroll`, and `.token-breakdown-body` set `overflow-x: auto` and `overscroll-behavior: contain`. Two CSS facts combine into the bug:

- `overflow-x: auto` makes the other axis's `overflow-y: visible` **compute to `auto`**, so each region is also a vertical scroll container.
- `overscroll-behavior: contain` applies to **both** axes. When a region has no vertical overflow, a vertical wheel is at its (zero-length) scroll boundary immediately, so `contain` stops it from chaining to the page and nothing scrolls.

The result: the page appears frozen whenever the pointer is over one of these regions.

### Fix

Narrow the containment to the x-axis with `overscroll-behavior-x: contain`. Horizontal overscroll is still contained (so trackpad left/right does not trigger browser back/forward navigation while scrolling wide tables), while vertical wheel now falls through to the page. `overflow-x` and the computed `overflow-y: auto` (relied on by the inset focus-ring rule) are unchanged, so keyboard reachability and focus rings are unaffected.

A short comment is added at each of the three rules so the axis-scoped `contain` is not "simplified" back to the two-axis form.

## Type of Change

## How Has This Been Tested?

- `make check` passes (gofmt, go vet, license, english-only).
- `internal/viewer` unit tests pass; no test asserts on the `overscroll-behavior` string, so the CSS contract is unaffected.
- Manual: with the window narrowed so a table overflows horizontally, the vertical mouse-wheel now scrolls the page while the cursor is over the table; horizontal overscroll still does not fire browser navigation gestures.

## AI/LLM Disclosure

This change was implemented with AI/LLM assistance (model: Claude Opus 4.8) and reviewed with open-code-review (`ocr`, `--audience agent`), which reported 0 findings. I reviewed every line myself, understand the change, and will answer maintainer questions and review comments in my own words.


MuoDoo (approved):
LGTM

